PR #1, opened 2017-02-12
RFC to describe the RFC process

zimbatm:
The goal is to define a process to unblock issues and PRs where there is clearly a need or good contribution but no clear outcome. The hope here is that by putting down all the context it will help the community reach a consensus. I also want to keep the process lite-weight to avoid hindering the contribution process.

The proposal is mainly lifted from the [Rust community](https://github.com/rust-lang/rfcs/blob/master/text/0002-rfc-process.md). This is mainly a starting point for the discussion, we should probably adapt the RFC process to our community.

One of the big difference with our community is that we don't have paid members that can implement the RFCs. That makes it hard to "accept" a RFC and have it implemented without having a "champion" that first agrees to be working on this. So I propose that instead we should have the RFC submitter try to gain at least traction of one other community member.

There are also some open questions regarding the feedback process. What is the acceptable response period for a feedback?

Actions after merge:
* [ ] Add cleaned-up RFC to README.md (will be handled in #2)
* [ ] Move repo to the NixOS org
* [ ] Redirect relevant issues and PRs to this repo
* [ ] Go through the process with another RFC to improve the process

### [Rendered](https://github.com/zimbatm/rfcs/blob/rfc-process/rfcs/0000-rfc-process.md)


rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,127 @@
...
+* Fill in the RFC
+* Submit a pull request. The pull request is the time to get review of
+the design from the larger community.
+* Build consensus and integrate feedback. RFCs that have broad support
+are much more likely to make progress than those that don't receive any
+comments.
+
+At this point, the person submitting the RFC should find at least one "buddy"
+that will help him bring the RFC to reality. The goal is to improve the
+chances that the RFC is both desired and lilely to be implemented.
```
nbp: nit: lilely -> likely

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,127 @@
...
+* Copy `0000-template.md` to `rfcs/0000-my-feature.md` (where
+'my-feature' is descriptive. don't assign an RFC number yet).
+* Fill in the RFC
+* Submit a pull request. The pull request is the time to get review of
+the design from the larger community.
+* Build consensus and integrate feedback. RFCs that have broad support
+are much more likely to make progress than those that don't receive any
+comments.
+
+At this point, the person submitting the RFC should find at least one "buddy"
```
teh: I'd go even further and find a "shepherd" who has the implied right to pull in people who are currently blocking progress. I don't mean "blocking" in a bad way. AFAICT all PRs die due to indifference or lack of time, not because anyone is maliciously blocking.

My hope is that "ownership" will  help avoiding no-one being responsible for moving things forward to either accepted or rejected.

rfcs/0000-rfc-process.md
zimbatm: Fixed in 254918a364420f60ced755ad5ee2f69646665acf

rfcs/0000-rfc-process.md
zimbatm: Yes that's what the Rust community is also using: https://github.com/rust-lang/rfcs#the-role-of-the-shepherd . I think it implies some form of structure like sub-teams and such which is why I didn't go with this, the nix community structure is much more amorphous. In my mind the "buddy" plays a similar role but depends on it's gravity, kudos, community credits, to make things happen.

zimbatm:
Thanks all for your reviews. I am still looking for a "buddy" for this RFC to adhere to it's own saying, anyone interested?

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,127 @@
...
+  - Removing language features
+  - Big restructuring of nixpkgs
+  - Introduction of new interfaces or functions
+  - A controversial change
+
+Some changes do not require an RFC:
+
+  - Adding, updating and removing packages in nixpkgs
+  - Additions only likely to be _noticed by_ other developers-of-nix,
+    invisible to users-of-nix.
```
edolstra: What are "users" and "developers" in this context?

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,127 @@
...
+* Fill in the remaining metadata in the RFC header, including links for
+  the original pull request(s) and the newly created issue.
+* Commit everything.
+
+Once an RFC becomes active then authors may implement it and submit the
+feature as a pull request to the nix or nixpkgs repo. An 'active' is not a
+rubber stamp, and in particular still does not mean the feature will
+ultimately be merged; it does mean that in principle all the major
+stakeholders have agreed to the feature and are amenable to merging it.
+
```
edolstra: This needs some more info about the lifecycle. See e.g. https://www.python.org/dev/peps/pep-0001/. For example, what's the status of a finalized RFC? Once PEPs reach "final" state, they're no longer modified and are mostly of historic interest - they're not primary documentation because any relevant doc changes should be made to the manual. This is in contrast to IETF Standards Track RFCs which *are* primary documentation.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,127 @@
...
+proposed here. If desired, we should likely look to Python's [PEP] process for
+inspiration.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+1. Does this RFC strike a favorable balance between formality and agility?
+2. Does this RFC successfully address the aforementioned issues with the current
+   informal RFC process?
+3. Should we retain rejected RFCs in the archive?
```
edolstra: I would say "yes", they can just be marked as "rejected" like PEPs.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,127 @@
...
+
+You need to follow this process if you intend to make "substantial" changes to
+the Nix ecosystem. What constitutes a "substantial" change is evolving based
+on community norms, but may include the following.
+
+  - Any semantic or syntactic change to the language that is not a bugfix.
+  - Removing language features
+  - Big restructuring of nixpkgs
+  - Introduction of new interfaces or functions
+  - A controversial change
```
edolstra: Also, anything that expands the scope of the project in a significant way. This includes adding support for new platforms or adding major subprojects (think NixUP, nix-make, ...).

Really, anything that has the potential of inflicting a lot of crosscutting technical debt or maintenance cost.

edolstra: Also, adding new package abstractions or idioms (think various auto-updaters, `callPackages`, `composableDerivation`, `builderDefs`, language-specific functions, ...). Deviations from the standard style make it harder for other people to understand / modify a package, so there should be a good reason for introducing them.

Of course, we should probably also better document what the standard style is :-)

teh:
I think this is a good test case. Let's say I want to be your RFC-buddy, what can I do? The text says "that will help him bring the RFC to reality" - what kind of actions are expected of me?

(unrelated, maybe use the gender-neutral them, i.e. "will help them bring")

zimbatm:
@teh I'm proposing 08b081a as the definition of the role, what do you think? If you still agree of being the buddy for the RFC it would make sense to catch up using a higher throughput protocol like voice, video or IRL :)

rfcs/0000-rfc-process.md
```diff
@@ -94,6 +94,15 @@ rubber stamp, and in particular still does not mean the feature will
 ultimately be merged; it does mean that in principle all the major
 stakeholders have agreed to the feature and are amenable to merging it.
 
+### Role of the "buddy"
+
+To goal for assigning a "buddy" to the RFC is multifold. The main
+responsability is to make himself available for to the author to move the RFC
```
teh: "make themselves" :) & "with them"

rfcs/0000-rfc-process.md
zimbatm: added 1a37875. I'm trying to avoid over-specifying the rules. Potentially we could add a BDFL line like: if @edolstra says so :-)

rfcs/0000-rfc-process.md
zimbatm: Okay so it means adding a status metadata

rfcs/0000-rfc-process.md
zimbatm: I don't really know what's the best solution. I agree that RFCs shouldn't have to be updated but things might come up during the implementation that might trigger another round of discussion.

rfcs/0000-rfc-process.md
zimbatm: That was imported from the Rust RFC where developers and users are more clearly defined I think. I will remove that for now.

zimbatm: 6563750

rfcs/0000-rfc-process.md
zimbatm: Fixed in 65d32e1, keep them coming :)

rfcs/0000-rfc-process.md
zimbatm: Another line of reasoning would be that closed PRs are still available on github and the repo only includes things we intend to implement, which are probably more relevant to users.

zimbatm: It also means that closed PRs can get resurrected more easily in case our opinion changes over time.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,136 @@
...
+submit an RFC first.
+
+## What the process is
+
+In short, to get a major feature added to the Nix ecosystem, one should first
+go throught the RFC process in order to improve the likelyhood of inclusion.
+Here are roughly the steps that one would take:
+
+* Fork the RFC repo https://github.com/NixOS/rfcs
+* Copy `0000-template.md` to `rfcs/0000-my-feature.md` (where
```
danbst: Where can I find this template?

rfcs/0000-rfc-process.md
moretea: So is it OK to create a PR first, and then find a buddy (by e.g. sending an email to the list)?

rfcs/0000-rfc-process.md
moretea: What about having the following state machine:
```
new -> proposed ->  accepted -> implemented
                    |     ^
                    v     |
                    revising
       
and: new -> closed, proposed -> closed, revising -> closed, accepted -> closed
(but not implemented -> closed)
```

If a RFC has been accepted, but not implemented yet, we could start revising it if there are issues that need RFC level discussions.

zimbatm:
See PR #2

zimbatm:
Yes I think so. The RFC doesn't have to be in a finished state before
submitting the PR. I expect it to be easier to find a buddy with a first
draft available to express your idea.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,136 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+Many changes, including bug fixes and documentation improvements can be
+implemented and reviewed via the normal GitHub pull request workflow.
+
+Some changes though are "substantial", and we ask that these be put through a
+bit of a design process and produce a consensus among the Nix community and
+the core team.
```
regnat: Does the distinction between the community and a "core team" makes sense for nix ?

rfcs/0000-rfc-process.md
regnat: I don't really understand the exact role of the buddy in our case. For the rust community it makes sense because of the structure of the project (the shepherd belongs to Rust's development team which is a well-defined subset of the community), but as nix community is far less structured, it seems to me that the person submitting the RFC will *de facto* be his own buddy

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,136 @@
...
+* Assign an id, using the PR number of the RFC pull request. (If the RFC
+  has multiple pull requests associated with it, choose one PR number,
+  preferably the minimal one.)
+* Add the file in the `rfcs/` directory.
+* Create a corresponding issue on the appropriate repo (NixOS/nix, NixOS/nixpkgs, ...).
+* Fill in the remaining metadata in the RFC header, including links for
+  the original pull request(s) and the newly created issue.
+* Commit everything.
+
+Once an RFC becomes active then authors may implement it and submit the
```
regnat: Many changes won't probably drain an unconditional approval from everyone in the community − see the everlasting discussion about the wiki for example ;).

It should be useful to have a well-defined notion of when a RFC is accepted for cases where it is not obvious ( Sufficient consensus between a few selected members of the community ? Vote ? Call for @edolstra's decision ?)

rfcs/0000-rfc-process.md
regnat: I think, rejected RFCs should be kept in the git history − at least the ones that raised some discussion from the community. The history of closed PRs isn't as convenient to browse as the actual git repo and is less reliable (the code of a PR can easily disappear or change if the submitter deletes or overwrites his branch iirc).

rfcs/0000-rfc-process.md
zimbatm: It's in #2 for now. I thought it would make sense to have the implementation separate.

rfcs/0000-rfc-process.md
zimbatm: Maybe "co-author" is more accurate?

The goal is multifold. I believe that having someone with whom the author can work more closely with would help to advance the RFC. If the author is not embedded in community, the co-author can lend some credence to the proposal. He can also help respond to other people's question as he hopefully shares enough context with the author. And also help out weed the usually typos and just provide a quicker iteration speed in general.

rfcs/0000-rfc-process.md
zimbatm: Most of the value of the RFC is derived from having developers being forced to serialise their thoughts to human language instead of code. It's mainly a communication tool. Unless the RFC gets largely discarded when faced with implementation realities it doesn't need to be amended after merge.

rfcs/0000-rfc-process.md
zimbatm: Probably not. We have a "core team" but the notion is a bit fuzzy and probably changes from person to person. Removed in 48e3cef

rfcs/0000-rfc-process.md
zimbatm: Totally agreed. One of the reason for going though the RFC process is to avoid everlasting PRs that don't get consensus. The idea is that the RFC will allow to expose the underlying problem and thinking without getting drowned in implementation details. But we could have a similar issue on the RFC submission for sure.

As usual @edolstra is the BDFL. Because of the concurrency limit, hopefully RFCs will be quicker for him to evaluate and take an executive decision. I don't have a perfect answer for that as it's a human problem. Some people want perfect solutions while others are willing to accept a compromise as long as the proposed approach improves on what we have already.

rfcs/0000-rfc-process.md
ttuegel: I think revision after implementation is more important than before. Everybody always has a favorite bikeshed color, so it's a real pain when you reach a consensus after weeks of discussion only to find that what you need to build isn't a bikeshed at all!

To put it less colorfully, there needs to be space to experiment, especially when it comes to changes that affect a wide section of our project. There are always unknowns, and there are always unknown-unknowns. It's important to get feedback from people who are actually using an implementation before setting that implementation in stone. Projects that insist on one-way innovation with no room for experimentation end up strangling themselves with bad decisions that seemed ingenious at the time. The core Haskell projects are a great example of this. If you can't fix a severe security vulnerability without a multi-year deprecation period, something has gone horribly wrong!

I'm not suggesting that this proposal is rushing down that path, but I think there should explicitly be room for experimentation in the proposal lifecycle. Leaving room explicitly is why [C4](https://github.com/zeromq/rfc/tree/master/42) has a "draft" status for public contracts. I'm thinking along these lines:
```
new -> proposed -> accepted -> implemented (draft) -> final
                       ^                |
                       |                |
                       +--  revision  --+
```
The `draft` stage should be a "handshake agreement": the user understands that the feature may change in response to data gathered from use, but the implementer understands this should only be done with adequate warning and for a really good reason. The `final` stage would be regarded as "set in stone": this feature will only change following a new RFC and an extended deprecation period.

Well, I wanted to give my two cents, but it looks like I dropped a buck-fifty on the table; sorry about that!

zimbatm: No worries :)

How I see it, the RFC process is more about doing a project proposal. Right now our de-facto "project manager", the person who can make wide decisions is Eelco, and he doesn't have time to extract meaning from reading code on large pull-requests. The goal here is to push some of that decision-making upfront, before all the work is done. It's more costly to the developer but should help Eelco. It doesn't mean that *everything* has to be decided upfront though. I agree that we want to avoid doing waterfall design.

If that's not clear from the existing wording then it should be revised. Maybe "RFC" is a bad term because it's too close to IETF RFCs. How about I rename them to NIP (Nix Improvement Proposals)? I think that was @moretea's idea. (danger: bikeshed ahead :)

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,135 @@
...
+This is the bulk of the RFC. Explain the design in enough detail for somebody
+familiar with the ecosystem to understand, and implement.  This should get
+into specifics and corner-cases, and include examples of how the feature is
+used.
+
+## When you need to follow this process
+
+You need to follow this process if you intend to make "substantial" changes to
+the Nix ecosystem. What constitutes a "substantial" change is evolving based
+on community norms, but may include the following.
```
nbp: nit: This should as well include making a RFC to describe the processes by which the community interact around the projects.

rfcs/0000-rfc-process.md
nbp: Ok, I thought that a graphviz document, with comment would be easier to follow.  There is one thing which remains to be determine in the following, which is the set of person responsible for accepting or rejecting proposals.

```dot
digraph rfc-state {
  /* Draft stage, any newly create RFC is in a draft stage. Provide draft implementation for the project, such as
   *  Nix, Nixpkgs without adding pull request to other repository, but refer to the draft-ed implementation in forked
   *  version of the projects. */
  draft;

  /* Submitted stage, a comity of persons (to be defined) evaluates the feasibility and compare benefits against the
   * drawbacks.  Ideally the peers (buddy, mentors, ...) should have help clear-up the ground for any questions that
   * might be asked by the comity. */
  submitted;

  /* Accepted stage, a comity of persons, agreed with the feasibility and that benefit outweigh the drawback and that a
   * complete implementation would be implemented in the projects. */
  accepted;

  /* Rejected stage, a comity of persons, disagree with the feasibility and/or that the drawbacks outweigh the benefit.
   * The comity should provide a comments explaining the choice. Any reject draft should not be submitted until all 
   * comments are addressed, and not within a period of 3 months (in order to prevent recurrent submissions) */
  rejected;

  /* Abandoned stage, nor the authors nor the peers (buddy, mentor, ...) can think of an alternative to compensate for
   * the rejection comments. */
  abandoned;

  /* Implemented stage, the proposal is available main-stream, and should be the default way of doing. */
  implemented;

  /* Replaced stage, another proposal is implemented and the content of this proposal should be ignored, unless
   * explicitly mentioned by an implemented proposal. */
  replaced;

  draft -> submitted [label="submitted by the author"];
  submitted -> accepted [label="validated by comity"];
  submitted -> rejected [label="rejected by comity"];
  rejected -> draft;

  draft -> abandoned [label="no alternative, or person to pursue"];

  accepted -> implemented [label="all components merged, and the feature is usable"];
  accepted -> replaced [label="better alternative found before the implementation"];
  implemented -> replaced [label="better alternative found after the implementation"];
}
```

rfcs/0000-rfc-process.md
nbp: If so, maybe we should make a repository for draft proposal, in which people work on, and accepting or rejecting would imply moving these files to the proper directory.

nbp:
Just a note, people seems to have started annotating their pull request / issues with the RFC tag:
https://github.com/NixOS/nixpkgs/issues?utf8=%E2%9C%93&q=RFC%20in%3Atitle

I took the current list and sorted the issues / pull requests in 3 categories: Submittable, Draft, and Rejected.  I evaluated these based on my personal opinion and the following criteria:
 * the quality of the first comment. (detailed with motivation & others)
 * the impact of the changes. (number of files modified or number of person impacted is high)

Submittable as RFC:
 * https://github.com/NixOS/nixpkgs/issues/21662: RFC: Graphical installer for NixOS
 * https://github.com/NixOS/nixpkgs/pull/10851: Shipping Security Updates
 * https://github.com/NixOS/nixpkgs/issues/15932: RFC: Generating AppStream Metadata
 * https://github.com/NixOS/nixpkgs/issues/7220: RFC: Harden(ed) NixOS
 * https://github.com/NixOS/nixpkgs/pull/1875: nixos: add grsecurity module (RFC)

Draft (not actionable, or not described enough) RFC:
 * https://github.com/NixOS/nixpkgs/issues/23181: [RFC] Modularize the firewall and nixify the rules
 * https://github.com/NixOS/nixpkgs/pull/22096: [RFC] Add lifecycle stages to nixpkgs
 * https://github.com/NixOS/nixpkgs/pull/17886: [WIP/RFC] Make `cmakeFlags` more ergonomic
 * https://github.com/NixOS/nixpkgs/issues/17044: RFC: Unified PAM settings for different display managers
 * https://github.com/NixOS/nixpkgs/pull/15799: WIP/RFC: stdenv: deprecate passing a string for configureFlags
 * https://github.com/NixOS/nixpkgs/pull/14766: [RFC/WIP] multiple-outputs: Change the order of outputs to have binaries in the primary output
 * https://github.com/NixOS/nixpkgs/issues/14532: RFC: remove node packages
 * https://github.com/NixOS/nixpkgs/pull/13542: RFC: Style and shellchecks
 * https://github.com/NixOS/nixpkgs/issues/13132: RFC: poor man's nix-option via package argument names
 * https://github.com/NixOS/nixpkgs/pull/10468: [RFC] libssl abstraction, default to libressl instead of openssl

Rejected as RFC (should be an ordinary PR):
 * https://github.com/NixOS/nixpkgs/pull/23190: [RFC] version: Extend /etc/os-release
 * https://github.com/NixOS/nixpkgs/pull/23175: [RFC] nixos-version: Add lsb_release compatibility
 * https://github.com/NixOS/nixpkgs/pull/21766: RFC: More common updater script implementations 
 * https://github.com/NixOS/nixpkgs/pull/21513: [WIP][RFC] signing-party: 2.2 -> 2.5
 * https://github.com/NixOS/nixpkgs/pull/19966: RFC: drop nodejs-4_x and packages
 * https://github.com/NixOS/nixpkgs/pull/19622: kdeconnect: [WIP/RFC] let dbus activate via systemd 
 * https://github.com/NixOS/nixpkgs/pull/19502: [RFC][WIP] vim module: support pkg and vimrc specification
 * https://github.com/NixOS/nixpkgs/pull/18818: Rfc htpdate
 * https://github.com/NixOS/nixpkgs/issues/16513: [RFC] fetchFromGitHub for release archives 
 * https://github.com/NixOS/nixpkgs/issues/14538: RFC: Live environment with nix store included in initrd
 * https://github.com/NixOS/nixpkgs/pull/14000: Use fix and extend function for all-packages.nix
 * https://github.com/NixOS/nixpkgs/pull/13181: [WIP, RFC] networking: add veth pairs via nix config
 * https://github.com/NixOS/nixpkgs/pull/12736: [RFC] hostbridge and IPv6 for containers
 * https://github.com/NixOS/nixpkgs/pull/11649: [RFC] claws-mail: enable PGP plugin by default
 * https://github.com/NixOS/nixpkgs/pull/10799: [RFC][WIP] uboot refactoring and new board support
 * https://github.com/NixOS/nixpkgs/pull/10582: [RFC] go: add packages rogpeppe/godef and 9fans/go
 * https://github.com/NixOS/nixpkgs/pull/6282: [RFC] Stop writing to obsolete /proc/sys/kernel/hotplug
 * https://github.com/NixOS/nixpkgs/pull/4657: RFC: Add buildExternalNixExpression
 * https://github.com/NixOS/nixpkgs/pull/3915: RFC: Agda builder
 * https://github.com/NixOS/nixpkgs/pull/2951: RFC: collectd: enable a bunch of plugins (by adding more deps)

I also added https://github.com/NixOS/nixpkgs/pull/10851 (submittable) and https://github.com/NixOS/nixpkgs/pull/14000 (rejected) as additional examples.  I classified https://github.com/NixOS/nixpkgs/pull/14000 as rejected, even if the change might have been controversial, it did not had impact out-side the limited set of developers/reviewers of this code.  It would have been a good RFC, but only as part of a larger plan to introduce overlays, or to add a better cross-compilation support, but not as a single set of commits.

I think this list is a good starting point to qualify precisely what belongs as an RFC or not, and help remove these RFC tags from existing pull requests / issues.

aneeshusa:
:+1: from me for this effort! Can we please add a LICENSE (say Apache 2) up front? It's a lot harder to do down the line (see https://github.com/rust-lang/rfcs/issues/1259 for example).

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,135 @@
...
+
+  - Any semantic or syntactic change to the language that is not a bugfix.
```
aneeshusa: style nit: remove trailing period for consistency

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,135 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+There is a danger that the additional process will hinder contribution more
+than it would help. We should stay alert that the process is only a way to
+help contribution, not an end in itself.
+
+# Alternatives
+[alternatives]: #alternatives
```
zimbatm: The idea is to list prior-art. This mainly shows that the author has done research. I agree that it would be nice to also have a justification of why those aren't adequate but that's not a strict requirement. Some of them might be obvious (eg: already discarded).

rfcs/0000-rfc-process.md
zimbatm: Renamed buddy to co-author in 2d315c6. I agree that it would be nice to refine the co-author's responsibilities but it's another thing to reach consensus on. Would it be alright to leave it ambiguous for now?

rfcs/0000-rfc-process.md
zimbatm: The interesting bit about rejected PRs is the discussion that's happening around it, which is currently tied to github. Presuming the desire to merge is to keep the history of it, I'm not convinced that a closed PR would be much worse than a merge. It also means that it's one less thing to manage as a maintainer.

rfcs/0000-rfc-process.md
zimbatm: One thing at a time :)

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,135 @@
...
+
+At this point, the person submitting the RFC should find at least one "buddy"
+that will help them bring the RFC to reality. The goal is to improve the
+chances that the RFC is both desired and likely to be implemented.
+
+Whomever merges the RFC should do the following:
+
+* Assign an id, using the PR number of the RFC pull request. (If the RFC
+  has multiple pull requests associated with it, choose one PR number,
+  preferably the minimal one.)
```
zimbatm: Good point. Fixed in b3db396

rfcs/0000-rfc-process.md
zimbatm: thanks, fixed as part of 983dca3

rfcs/0000-rfc-process.md
zimbatm: Ok that's the reason why the RFC needs to be either renamed at the very start or very end: otherwise all the existing comments are being hidden by github. This will be part of the CONTRIBUTING.md or PULL_REQUEST.md so users do the right thing.

zimbatm:
Please vote: is [AsciiDoctor](http://asciidoctor.org/) an acceptable alternative text format on top of Markdown? In some cases like here it would be nice to have a bit more markup to express things like diagrams and list definitions.

aneeshusa:
I'm not sure about how one would layer AsciiDoctor on top of Markdown (i.e. in `.md` files), but I'd definitely like it as an alternative (i.e. `.adoc` files).

zimbatm:
Yes that would be as an alternative. The downside would be to maintain two template files.

nbp:
I would prefer a single format, in order to avoid having to define one reference, and potentially have different versions.  Do we expect much RFC to have content which is defined graphically?  Could we just provided an image as part of the RFC?

teh:
To me this seems "ready enough", how about we merge unless anyone objects? I would add two more comments:

* Add a note that this is a "working document" and open to well-reasoned changes
* Move the RFC repo to NixOS to institutionalize ownership.

What do you think?

zimbatm:
@aneeshusa regarding licensing, I added a CC BY-SA 4.0 license to the repo. Apache 2 is fine but targeted at licensing software.

zimbatm:
@teh I would like to address the state machine discussion before calling it ready. I think the current text doesn't do enough to explain what are the different steps. I would like to propose a simplified state though

PR -> find co-author -> submit for community approval -> iterate -> merge, abandon or reject

The RFC wouldn't need to be updated during implementation as it's goal is to gain community approval. All the useful documentation should be copied over to the implementation instead.

zimbatm:
Alright, I think this RFC is ready now. The state machine is very simplistic but should be enough to bootstrap the process.

Unless there are some major objections I would like to move forward with this. We can always create further RFCs or PRs to refine the process.

Vote, yay, nay?

@edolstra can you help me move this to the nixos org?

Ekleog:
No major objections here, just a few comments to compare this to IETF's RFCs, as they are the predecessors of all kinds of RFC systems that I know of. (Sorry for waking up to this so late.)

1. IETF's RFCs are numbered only after being approved, not before. Before, they are called Internet Drafts, and are not numbered. This could be mirrored by not having the RFCs numbered at the time of the PR but at the time of the merge. Advantages, in my opinion: any RFC number less than the last one is a valid RFC, there is no way to refer to RFCXXXX while RFCXXXX has not been approved. Drawbacks: more "work" merging such a RFC. (well, actually this can be considered as a major issue, as RFCs can't reasonably be re-numbered and thus the scheme, whichever is eventually chosen, will not be possible to change)
2. "community vote" as not a well-defined term, as "community" can not be expressed in terms of a number of persons, "vote" is not a possible way to do (much as for the IETF). A way to fix this is to just write "rough consensus" (like the IETF does), and that's actually what likely would happen anyway. (Yes, I'm nitpicking, sorry)
3. As much as I love all the work @edolstra has done (thanks again!), I'm not sure it's a great idea to encode this BDFL-iness in the RFC. People will naturally turn towards him if he makes a point, but in the current state I feel like nix / NixOS is already heavily centralized in the hands of a handful of persons, him being the foremost. I think writing this down in "law" is not necessarily a good thing for the future (including for him, as that creates a lot of pressure I guess). Then, that's just my opinion, others  may of course disagree.

Hoping that may help!

leenaars:
I would suggest to use restructuredText as an alternative to its poor cousin Markdown (which is cumbersome beyond the simplest use). It is a mature and large ecosystem, has a rich semantic and is supported by many tools - and yet is simple to use. For The Commons Conservancy DRACC's we made this choice, as you can add metadata like numbering, copyright etc inside the format (and thus ready for machine processing) without the need to add a non-standardised header etc.

We have a description of our publication process here, taking into account some things

       https://dracc.commonsconservancy.org/0000  (webserver is running on NixOS, btw)

The metadata at the top looks like this in the source files:

:Title: Introduction to the DRACC series
:DRACC: 0000
:Category: Regulatory
:Scope: Global
:Authors: Leenaars, M.A.G.J.; Šuklje, M.
:Date: January 2017
:Copyright: [The Commons Conservancy]

But that allows us to produce overviews like:

     https://dracc.commonsconservancy.org

(which makes you happy when you have hundreds of them)

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,132 @@
...
+
+The "RFC" (request for comments) process is intended to provide a consistent
+and controlled path for new features to enter the nix language, packages and
+OS, so that all stakeholders can be confident about the direction the
+ecosystem is evolving in.
+
+# Motivation
+[motivation]: #motivation
+
+There are a number of GitHub issues, PRs and mailing-list discussions where
```
leenaars: Are you sure you want to encode a specific hosting tool inside these documents? Why not say 'issues from the tracker'?

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+You need to follow this process if you intend to make "substantial" changes to
+the Nix ecosystem. What constitutes a "substantial" change is evolving based
+on community norms, but may include the following.
+
+* Any semantic or syntactic change to the language that is not a bugfix
+* Removing language features
+* Big restructuring of nixpkgs
+* Expansions to the scope of nixpkgs (new arch, major subprojects, ...)
+* Introduction of new interfaces or functions
+* A controversial change
```
spinus: I'm not sure this should be here, it's very subjective. It can even start a meta discussion is something is controversial or not :-)

zimbatm: I think what you're saying is controversial :)

shlevy:
What should I look at to get a high level overview of the status here? There's a lot of commentary :grinning:

fpletz:
Also, I don't think we have enough opinions from other active community members here.

Specifically I would love to see comments from: @domenkozar @globin @grahamc @joachifm @garbas @peti @rbvermaa @FRidh @abbradar @dezgeg @copumpkin @vcunat @aszlig @Ericson2314

rfcs/0000-rfc-process.md
copumpkin: For nixpkgs, probably true, but for Nix the program there is certainly a core team, that consists mostly of @edolstra, with a bit of @shlevy 😄 it'd be nice to expand that group, but we might as well be explicit about it until then

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+
+Some changes do not require an RFC:
+
+* Adding, updating and removing packages in nixpkgs
+* Fixing security updates and bugs that don't break interfaces
```
copumpkin: Does this include refactoring existing stuff? I often find a lot of duplication and try to factor it out. I'd hate to have to make a whole RFC just to say "quit copying and pasting this stuff around". Even if that ends up changing a substantial amount of code.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+  understand why new features are being merged
+
+As an alternative, we could adopt an even stricter RFC process than the one
+proposed here. If desired, we should likely look to Python's [PEP] process for
+inspiration.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+To be solved in the future:
```
copumpkin: Is there an RFC process for updating the RFC process?

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+* Fill in the RFC
+* Submit a pull request. Rename the rfcs with the PR number. (eg: PR #123 would
+  be `rfcs/0123-my-feature.md`)
+* Get initial feedback on the PR.
+
+At this point, the person submitting the RFC should find at least one "co-author"
+that will help them bring the RFC to completion. The goal is to improve the
+chances that the RFC is both desired and likely to be implemented.
+
+Once the author is happy with the state of the RFC, he tells it's ready for
```
grahamc: This sentence is a bit awkward "he tells it's ready..." (1) gender, (2) who do they tell? Their cat?

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+wider community review. And the review process happens.
+
+After the review process has settled a decision is taken by community vote,
+with @edolstra as a BDFL that can override any decision.
+
+Whomever merges the RFC should do the following:
```
grahamc: Since this person is the subject of the verb, I think it is _Whoever_.

rfcs/0000-rfc-process.md
teh: IMO no, refactoring implies no substantial change in functionality. It'd be bad if we discouraged your amazing work :(

Ericson2314:
A bit busy so I just skimmed, but I've really enjoyed https://github.com/rust-lang/rfcs/ and https://github.com/ghc-proposals/ghc-proposals, so yay thanks @zimbatm!

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+This RFC is used to bootstrap the process. Further RFCs can be used to refine
+the process.
+
+# Motivation
+[motivation]: #motivation
+
+There are a number of changes that are significant enough that they could
+benefit from wider community consensus before being implemented. Either
+because they introduce new concepts, big changes or are controversial enough
+that not everybody will agree on the direction to take.
```
FRidh: Therefore, the purpose of this RFC is ...

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+
+Some changes though are "substantial", and we ask that these be put through a
+bit of a design process and produce a consensus among the Nix community.
+
+This is the bulk of the RFC. Explain the design in enough detail for somebody
+familiar with the ecosystem to understand, and implement.  This should get
+into specifics and corner-cases, and include examples of how the feature is
+used.
+
+## When you need to follow this process
```
FRidh:
```
## When this process is followed

This process is followed when one intends to make ...
```

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+on community norms, but may include the following.
+
+* Any semantic or syntactic change to the language that is not a bugfix
+* Removing language features
+* Big restructuring of nixpkgs
+* Expansions to the scope of nixpkgs (new arch, major subprojects, ...)
+* Introduction of new interfaces or functions
+* A controversial change
+
+Some changes do not require an RFC:
```
FRidh: Certain changes

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+
+* Adding, updating and removing packages in nixpkgs
+* Fixing security updates and bugs that don't break interfaces
+
+If you submit a pull request to implement a new feature without going
```
FRidh: Pull requests that contain any of the afore mentioned 'substantial' changes may be closed if there is no RFC connected to the proposed changes.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+through the RFC process, it may be closed with a polite request to
+submit an RFC first.
+
+## What the process is
```
FRidh:
```
## Description of the process
```

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+
+In short, to get a major feature added to the Nix ecosystem, one should first
+go throught the RFC process in order to improve the likelyhood of inclusion.
```
FRidh: Is the purpose just to improve the likelihood of inclusion? I think there are other reasons. In any case, this is not the place to mention the reasons again (although I do understand why you pick something here).

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+Here are roughly the steps that one would take:
+
+* Fork the RFC repo https://github.com/NixOS/rfcs
+* Copy `0000-template.md` to `rfcs/0000-my-feature.md` (where 'my-feature' is
+  descriptive. don't assign an RFC number yet).
+* Fill in the RFC
+* Submit a pull request. Rename the rfcs with the PR number. (eg: PR #123 would
+  be `rfcs/0123-my-feature.md`)
```
FRidh: I don't think we should use PR numbers because I can imagine that at some point we may want to make minor changes and those changes could go through PR's. If you want to link to a discussion, I think it is better to be explicit about it in the RFC.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+* Get initial feedback on the PR.
```
FRidh: Get or Request? Furthermore, one wants feedback on the RFC, not the PR.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+
+At this point, the person submitting the RFC should find at least one "co-author"
```
FRidh: > The person submitting the RFC should find at least one "co-author"

Does it matter whether they search for one now ("At this point") or already had one?

rfcs/0000-rfc-process.md
FRidh: When a RFC is rejected, it would be stored in the repo, and a section would be added explaining why it was rejected. In this section we could refer to discussions, but obviously those references may break.

rfcs/0000-rfc-process.md
FRidh: And how is this announced? Mailing list?

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+that will help them bring the RFC to completion. The goal is to improve the
+chances that the RFC is both desired and likely to be implemented.
+
+Once the author is happy with the state of the RFC, he tells it's ready for
+wider community review. And the review process happens.
```
FRidh: Are there any more details on how that should happen? Is there a certain fixed time window, say 4 weeks, for the review process? Or do we keep that open? The author initiates the review process, but who takes care of settling on a decision? Is it the task of the author?

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+
+After the review process has settled a decision is taken by community vote,
```
FRidh: What does community vote mean? Majority vote? Consensus? And who is consider part of the community? I'm asking because we've had quite a lot of controversial issues in the past.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+with @edolstra as a BDFL that can override any decision.
+
+Whomever merges the RFC should do the following:
+
+* Create a corresponding issue on the appropriate repo (NixOS/nix,
```
FRidh: Why do we create an issue on any of those repo's as well? Is that to simply show it is being discussed? If that's the case, shouldn't we also inform via the mailing list?

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+  NixOS/nixpkgs, ...).
+* Fill in the remaining metadata in the RFC header, including links for the
+  original pull request(s) and the newly created issue.
+* Commit everything.
+
+Once an RFC becomes active then authors may implement it and submit the
```
FRidh: Should we include this sentence? One could always make an implementation and open a PR. I suppose we might want to write here what our preference is regarding where the discussion takes place.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+feature as a pull request to the nix or nixpkgs repo. An 'active' RFC is not a
```
FRidh: > An 'active' RFC

Is that an *accepted* RFC?

Also, what are the possible status of an RFC? I don't see them mentioned here.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+  on design issues
+
+The co-author doesn't necessarily have to agree with all the points of the RFC
+but should generally be satisfied that the proposed additions are a good thing
+for the community.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+There is a danger that the additional process will hinder contribution more
```
FRidh: danger -> risk

rfcs/0000-rfc-process.md
FRidh: If its for prior-art then I'm of the opinion it should be mentioned *before* the actual proposal, and should be discussed at that point. 

@nbp 
If I would write a report where I investigate possible solutions, then I would do also as you say, have a section per alternative, and then a discussion where one (or none) is chosen. I can imagine that in a NEP (that's how we're going to call them, right? Nix Enhancement Proposal) that choice has already been made by the authors of the NEP and therefore they're only discussed briefly either before or after the actual proposal.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+than it would help. We should stay alert that the process is only a way to
+help contribution, not an end in itself.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Retain the current informal RFC process. The newly proposed RFC process is
```
FRidh: ah, so it was intended to be informal, I didn't get that :)

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,148 @@
...
+author: zimbatm
+co-authors: teh, MoreTea
+related-issues: https://github.com/zimbatm/rfcs/pull/2
+---
+
+# Summary
+[summary]: #summary
+
+The "RFC" (request for comments) process is intended to provide a consistent
+and controlled path for new features to enter the nix language, packages and
```
abbradar: Nitpick: capitalize Nix.

rfcs/0000-rfc-process.md
abbradar: > throught

through

rfcs/0000-rfc-process.md
abbradar: I agree: we should define this process in more detail. For example let's say "period of initial discussion is two weeks, and may be extended (by whom?) if it's felt that more discussion is needed".

rfcs/0000-rfc-process.md
abbradar: It talks about merging RFC here IIUC, so they would be "issue: implement RFC #0000" -- no discussion needed at this point.

abbradar:
Ugh, instead of adding it to @FRidh's comment threads GitHub separated mine -- they lost context that way.

FRidh:
@abbradar not if you go to the tab "Files changed" https://github.com/zimbatm/rfcs/pull/1/files

rfcs/0000-rfc-process.md
zimbatm: fixed in 9322c2e

rfcs/0000-rfc-process.md
zimbatm: 4464a7e

rfcs/0000-rfc-process.md
zimbatm: 92cb660

rfcs/0000-rfc-process.md
zimbatm: 0f4a464

rfcs/0000-rfc-process.md
vcunat: I don't think voting approaches would be good here.  Admittedly, I haven't really seen any "community vote" in action, but I think the notion tends to push the discussion itself to a bad direction, and achieving fair voting conditions seems to take lots of work.

As mentioned before, I'd personally also prefer striving for some kind of consensus on each RFC.

rfcs/0000-rfc-process.md
zimbatm: b1d580b

rfcs/0000-rfc-process.md
vcunat: We might be keeping the number of the PR that first introduced the RFC.  There are some advantages to having a stable identifier already when the discussion starts, and keeping it forever.  (I sometimes slightly struggle with this "renaming" of IETF RFCs.)

rfcs/0000-rfc-process.md
zimbatm: The RFC process is mostly about bringing up discussions before implementing something that nobody wants. These are more like guidelines for when it's desirable to bring up the discussion.

If code is changed that doesn't break public interfaces I think it's not necessary to raise a RFC. Even for rarely used public interfaces, for a certain measure of "rare", I don't think that a RFC is necessary.

To goal is more that if for example you think that the mkDerivation "buildInputs" should be renamed to "inputs", then other people should be notified before it happens.

rfcs/0000-rfc-process.md
zimbatm: e6de427 thanks

rfcs/0000-rfc-process.md
zimbatm: ad8bb6c

rfcs/0000-rfc-process.md
FRidh: Maintainers of this rfc repo can assign a RFC number when a PR is opened so it can be used directly. I agree it makes sense to always include the PR number in the RFC.

rfcs/0000-rfc-process.md
zimbatm: @FRidh yes there are other reasons. I see the RFCs mainly as a communication tool, it forces developers to serialize their thinking instead of presenting a bunch of code on which the reviewer has to reverse the context.

@abbradar fixed in dd962df

rfcs/0000-rfc-process.md
FRidh: Perhaps we should consider having a Technical Committee like e.g. Debian.

rfcs/0000-rfc-process.md
zimbatm: What alternative numbering system do you propose? Numbering by PR makes it easy to avoid having two RFCs with the same number. It's not a big deal if RFCs are not numbered continuously.

rfcs/0000-rfc-process.md
zimbatm: Removed the line as I think it's described below. 9faf879

rfcs/0000-rfc-process.md
zimbatm: The idea is that it's possible to design a RFC alone, but at this point they should try to rally at least on co-author to help them with the process. If they already found one before then it's fine as well.

Any idea on how to change the sentence to convey that properly?

rfcs/0000-rfc-process.md
zimbatm: @grahamc Agreed, any idea on how to improve the sentence?

@FRidh Mailing list seems reasonable. The process is a bit vague but it's intentional, I would prefer to update the process based on current usage rather than come up with something too artificial. Rust has multiple sub-teams each responsible for their RFC process but we don't have that. Potentially we could start forming sub-teams in a further RFC if we feel like it would be useful.

rfcs/0000-rfc-process.md
FRidh: Continuous numbering with RFC repo maintainers assigning a RFC number after a draft of the RFC has been approved and merged.

I think PEP 1 may be a good source of inspiration, although I am not familiar with RFC's in other projects so I can't really compare.

rfcs/0000-rfc-process.md
zimbatm: The process can always be refined later. Right now we don't have enough data points to know what's a good review period. We also lack a "core team" that can take executive decisions of what is "accepted". I think we'll have to rely on a general consensus on all the reviewer participants.

I agree that these are good questions but I'm also wary that answering them will delay this RFC for another month. Month in which other RFCs might also happen successfully.

rfcs/0000-rfc-process.md
zimbatm: It's a good question and I don't know how to answer it. The best thing I can say is that we probably need a "core team" that can make executive decisions. I don't think it should block this RFC though, because not all RFCs are controversial. Making this controversial would affect even more RFCs.

rfcs/0000-rfc-process.md
zimbatm: Thanks, I'll trust you on that 1eb8a17

rfcs/0000-rfc-process.md
zimbatm: Yeah feel free to suggest better wording for this sentence.

rfcs/0000-rfc-process.md
zimbatm: fb723ea

rfcs/0000-rfc-process.md
zimbatm: And I aswered to all your other comments :)

rfcs/0000-rfc-process.md
zimbatm: Anther RFC why ? :D 
To be real, in most cases I think we can just submit a PR with the proposed changes to the README.md

rfcs/0000-rfc-process.md
zimbatm: I don't really see the advantage. Numbering by PR allows the author to obtain his number without having to coordinate with other people. It's an implementation detail but github tends to lose comments on renaming of a file so if the author can submit his PR and rename straight away then it makes this easier. On this very PR I couldn't rename it because comments had already started on the file.

rfcs/0000-rfc-process.md
zimbatm: Let's aim for consensus instead of vote for now. That should work for most RFCs unless they are controversial.

zimbatm:
Please review c28db71 which removes the voting condition. For now let's aim for a general consensus instead.

rfcs/0000-rfc-process.md
vcunat: I'd go without a committee, at least for now (too big jump).  And it would just make the problem indirect and introduce more complications, i.e. we would have to decide how to choose committee members, etc.

rfcs/0000-rfc-process.md
zimbatm: I replaced 'active' with 'accepted' which should make things a bit clearer

rfcs/0000-rfc-process.md
zimbatm: Submitting a RFC is already a pretty big barrier for community work, I'm a bit concerned of putting a too formal process in place.

rfcs/0000-rfc-process.md
zimbatm: Alright, all rejected RFCs will be moved to the ./rejected folder for now 9830d2d

At some point we'll be able to use the metadata to generate a website to list all the RFCs, at that point we can merge the folder back and add a "status" header.

zimbatm:
Thanks @vcunat @FRidh and @abbradar for the new comments, it definitely made the RFC better.

Unless some major objections come up I hope that next week will see the RFC officially merged.

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,154 @@
...
+  descriptive. don't assign an RFC number yet).
+* Fill in the RFC
+* Submit a pull request. Rename the rfcs with the PR number. (eg: PR #123 would
+  be `rfcs/0123-my-feature.md`)
+
+At this point, the person submitting the RFC should find at least one "co-author"
+that will help them bring the RFC to completion. The goal is to improve the
+chances that the RFC is both desired and likely to be implemented.
+
+Once the author is happy with the state of the RFC, she/he should seek for
```
FRidh: she/he -> they

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,154 @@
...
+wider community review by stating the readyness of the work. Advertisement on
+the mailing-list and IRC is an acceptable way of doing that.
+
+After a number of rounds of review the discussion should settle and a general
+consensus should emerge. This bit is left intentionnaly vague and should be
```
FRidh: intentionally

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,154 @@
...
+to the feature and are amenable to merging it.
+
+Whoever merges the RFC should do the following:
+
+* Fill in the remaining metadata in the RFC header, including links for the
+  original pull request(s) and the newly created issue.
+* Commit everything.
+
+If a RFC is rejected, whoever merges the RFC should do the following:
+* Move the rfc to the rejected folder
```
FRidh: If the status is `rejected` then what's the point of putting it in a separate folder? And how about other statuses? Will they also be put in separate folders?

rfcs/0000-rfc-process.md
```diff
@@ -0,0 +1,154 @@
...
+* Fill in the remaining metadata in the RFC header, including links for the
+  original pull request(s) and the newly created issue.
+* Include a summary reason for the rejection
+* Commit everything
+
+## Role of the "co-author"
+
+To goal for assigning a "co-author" is to help move the RFC along.
```
FRidh: The goal

rfcs/0000-rfc-process.md
vcunat: What other statuses?  In the beginning they're unmerged and they should get either merged (among accepted) or "merged" into the rejected folder.

vcunat: From previous discussion it seemed useful that we somehow "permanently record" the rejected proposal with a summary of arguments, including *why* it was rejected.  Keeping all the information just in PR discussion threads has some disadvantages.

edolstra:
Looks good to me. The main remaining issue is that it leaves the decision procedure unspecified, but I agree we can leave that for now. The risk is that it creates the same problem as the current PR system (i.e. that PRs stay in limbo with no clear decision), but at least it's an improvement to require a general consensus, instead of the current situation where any of a few dozen committer can merge controversial changes.

Who should have write access to the rfcs repo?

vcunat:
> any of a few dozen committer can merge controversial changes

This will still be able to, *technically*, but I think everyone has been implicitly following the rule that controversial changes aren't merged without wider approval.

layus:
> [...], but I think everyone has been following an implicit rule that controversial changes aren't merged without wider approval.

Yes and not. For two different PR introducing similar changes, I have seen one stalling under controversial discussions, and the second one getting merged by another committer unaware of the discussion.

vcunat:
I suppose it was a case not considered controversial by the merging side, but it's still good to get things a bit more standardized.  (I personally tend to be more careful with merges than what's usual, I think, even if the changes aren't "controversial".)

rfcs/0000-rfc-process.md
zimbatm: Just to restate the goal: the goal is to keep a record of any unsuccessful PRs because on github it's possible for the submitter of the PR to remove the original code and then the context is lost.

In that regards, the `rejected` folder should contain any RFC that is not accepted. It would also include abandoned RFCs. If you have a better name for it I'm happy to change it.

rfcs/0000-rfc-process.md
zimbatm: thanks, fixed in 0b188c5

rfcs/0000-rfc-process.md
zimbatm: 0b188c5

rfcs/0000-rfc-process.md
zimbatm: 0b188c5

zimbatm:
Alright, I think all the remaining issues have been fixed now. Time for merging!!!

rfcs/0000-rfc-process.md
FRidh: I'm of the opinion we should merge RFC's sooner and assign them a draft status.

zimbatm:
@edolstra I sent you an invite for the repo, this will let you move it to the NixOS org.

In regards to access, since RFCs are mainly a communication tool I would suggest just giving the same access as nixpkgs. I think that having a core team would be helpful to make executive decisions when necessary but that can be defined in another RFC along with their responsibilities and process to join/leave the core team.

fpletz:
Ping. What's the problem in migrating this to the NixOS organization? We have reached a rough consensus in the community for the RFC process. This should be a good start.

cc @rbvermaa

domenkozar:
I can migrate if @edolstra agrees :)

edolstra:
@domenkozar Yes please do.

domenkozar:
@zimbatm can you give me admin access to this repo?

domenkozar:
Done.

vcunat:
GitHub now nicely redirects from the old links :-)

PR #6, opened 2017-04-01
[RFC 0006] Runtime references interface

shlevy:
Rendered: https://github.com/shlevy/rfcs/blob/runtime-references-interface/rfcs/0006-runtime-references-interface.md

### DIFF
--- a/rfcs/0006-runtime-references-interface.md
+++ b/rfcs/0006-runtime-references-interface.md
@@ -1,0 +1,64 @@
+---
+feature: runtime-references-interface
+start-date: 2017-04-01
+author: Shea Levy
+co-authors: (find a buddy later to help our with the RFC)
+related-issues: https://github.com/NixOS/rfcs/pull/6
+---
+
+# Summary
+[summary]: #summary
+
+Currently, nix itself scans derivation outputs for the hashes of the
+(transitive closure of) the inputs of the build and registers those it
+finds as references. This feature would provide an optional interface
+for builds to specify their own runtime references, and possibly for
+nix to tell the build what its build time inputs are, bypassing nix's
+own hash scanning.
+
+# Motivation
+[motivation]: #motivation
+
+The existing hash scanning works very well for most use cases. There
+are times, however, where we have to hack around the constraints it
+imposes (e.g. `nukeReferences`, not storing the configure flags used
+to build gcc, storing the path of a file in a comment to ensure the
+build depends on it, etc.). Moreover, every time we need to tweak the
+logic we need a new nix release. Finally, recursive nix will involve a
+more complex reference checking story. Having an interface to
+explicitly specify runtime references would improve all of these
+issues.
+
+# Detailed design
+[design]: #detailed-design
+
+Nix will put all the build time references (or requisites?) into
+`$NIX_BUILD_TOP/build-time-references`. After the build completes, nix
+will check `$NIX_BUILD_TOP/runtime-references` and, if it exists, skip
+its builtin hash scanning and just use the file. Both files are
+newline-separated lists of store paths. Parse errors in the
+`runtime-references` file, or non-existent store paths, fail the build.
+
+Alternatively, since recursive nix will require a more complex
+interface with the host nix process anyway, we can just expose a unix
+domain socket (possibly just a socketpair) to the build to query the
+build time references and set the run time references.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Arguably this will be a slight increase in complexity, but IMO it's
+more moving complexity to a more appropriate place.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The main alternative is just sticking with the status quo and hacks,
+which mostly works and can be accomodated (though not super cleanly)
+in the recursive nix regime
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+The exact specifics of the interface (see "Alternatively" in the
+[detailed design](#detailed-design)).
--- a/rfcs/0006-runtime-references-interface.md
+++ b/rfcs/0006-runtime-references-interface.md
@@ -23,7 +23,8 @@ The existing hash scanning works very well for most use cases. There
 are times, however, where we have to hack around the constraints it
 imposes (e.g. `nukeReferences`, not storing the configure flags used
 to build gcc, storing the path of a file in a comment to ensure the
-build depends on it, etc.). Moreover, every time we need to tweak the
+build depends on it, building disk images that don't retain references
+to their contents, etc.). Moreover, every time we need to tweak the
 logic we need a new nix release. Finally, recursive nix will involve a
 more complex reference checking story. Having an interface to
 explicitly specify runtime references would improve all of these
@@ -44,6 +45,11 @@ its builtin hash scanning and just use the file. Both files are
 newline-separated lists of store paths. Parse errors in the
 `runtime-references` file, or non-existent store paths, fail the build.
 
+To avoid duplicating the default reference scanning behavior in Nix,
+we could also make a tiny tool part of Nix itself that wraps the
+`scanForReferences` API, and then expose that in the `<nix>` namespace
+so tooling can get the full path to that tool.
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
@@ -55,7 +61,11 @@ more moving complexity to a more appropriate place.
 
 The main alternative is just sticking with the status quo and hacks,
 which mostly works and can be accomodated (though not super cleanly)
-in the recursive nix regime
+in the recursive nix regime. Tools like `nukeReferences` and similar
+can approximate the mechanism we propose in this RFC in many cases,
+but with disk images in particular they fall short, because the paths
+are valid _inside_ the image, but we don't want to retain references
+in the store that built it.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/rfcs/0006-runtime-references-interface.md
+++ b/rfcs/0006-runtime-references-interface.md
@@ -12,9 +12,9 @@ related-issues: https://github.com/NixOS/rfcs/pull/6
 Currently, nix itself scans derivation outputs for the hashes of the
 (transitive closure of) the inputs of the build and registers those it
 finds as references. This feature would provide an optional interface
-for builds to specify their own runtime references, and possibly for
-nix to tell the build what its build time inputs are, bypassing nix's
-own hash scanning.
+for builds to specify their own runtime references and for nix to tell
+the build what its build time inputs are, bypassing nix's own hash
+scanning.
 
 # Motivation
 [motivation]: #motivation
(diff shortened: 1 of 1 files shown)


shlevy:
@edolstra

bennofs:
What's the motiviation for providing access to the build time references? Those are just the references of the .drv, correct?

shlevy:
@bennofs That will be needed if we want to implement hash scanning in `stdenv.mkDerivation` (which would eventually enable us to remove it from the nix codebase all together).

shlevy:
@bennofs Or if we want to do hash scanning and then modify the results somehow, or whatever.

shlevy:
I'll update the motivation.

moretea:
I like the file based approach, which could in theory be extended to talk to the Nix daemon by making the `$NIX_BUILD_TOP/runtime-references`  a named pipe.

An additional benefit of this approach is that it enables outputting runtime dependencies from packagemanager tools like `bundix` automatically.

rfcs/0006-runtime-references-interface.md
```diff
@@ -0,0 +1,77 @@
...
+Nix will put all the build time references (or requisites?) into
+`$NIX_BUILD_TOP/build-time-references`. After the build completes, nix
+will check `$NIX_BUILD_TOP/runtime-references` and, if it exists, skip
+its builtin hash scanning and just use the file. Both files are
+newline-separated lists of store paths. Parse errors in the
+`runtime-references` file, or non-existent store paths, fail the build.
+
+Alternatively, since recursive nix will require a more complex
+interface with the host nix process anyway, we can just expose a unix
+domain socket (possibly just a socketpair) to the build to query the
```
teh: AFAICT right now when using nix as a package manager there is no nix-daemon interaction. Would this still be the case after this change?

shlevy: I don't understand the question. There is plenty of nix daemon interaction when using nix as a package manager, except in single user mode.

teh: Apologies, single-user mode is what I meant.

shlevy: This wouldn't affect the usage of a daemon outside of a nix build, no.

shlevy:
@moretea Yes, after further thought I'm pretty convinced the file based approach is the right way. Will update the RFC.

edolstra:
How will you prevent dangling references, i.e. packages that work on the system where they were built but suddenly don't work when copied to another system because some reference is not present?

I think `nukeReferences` / `removeReferencesTo` is a more sound approach because it physically obliterates all references to a path. So if some reference to the path actually *is* necessary to the functioning of the package, the package will break deterministically - you don't get a danging reference that works sometimes.

Also, this approach moves the entire responsibility for detecting references to the builder. The builder cannot say "ignore references to path P". Instead, it has to specify *all* desired references, which basically means it needs to duplicate Nix's dependency scanning.

I also predict that this feature will rapidly be (ab)used to implement optional runtime dependencies.

shlevy:
> How will you prevent dangling references, i.e. packages that work on the system where they were built but suddenly don't work when copied to another system because some reference is not present?

Packages should not touch the default scanning unless they know what they're doing. We can also currently store a path inside e.g. a compressed file to get around reference detection.

> So if some reference to the path actually is necessary to the functioning of the package, the package will break deterministically - you don't get a danging reference that works sometimes.

Again, this should not be used casually. Currently we can use global references/PATH/etc. if we choose in nixpkgs, and the answer to this concern there is: don't do that unless you're sure it's the right thing

> it needs to duplicate Nix's dependency scanning.

Well, part of the point is to eventually *move* nix's dependency scanning to stdenv. But yes, this is true.

> I also predict that this feature will rapidly be (ab)used to implement optional runtime dependencies.

I'm not convinced this is abuse, though of course it needs to be used with care.

copumpkin:
I like this and it gets around some of my concerns about putting more of the java ecosystem into Nix. Currently, we'll lose any runtime references inside jar files unless we turn off all jar compression, which gets awkward. Another example are wide character strings used for store references.

@edolstra much of this boils down to being able to move more stuff to nixpkgs and experiment with features without being constrained by the Nix release cycle. There's lots of ways for us to shoot ourselves in the foot in nixpkgs today, so I guess it all boils down to how much Nix wants to protect us from ourselves. I don't think this introduces too much opportunity for screwups but it'll likely be obvious when it does cause weirdness, right?

Ericson2314:
Somewhat relevant: https://github.com/NixOS/nix/issues/1080

shlevy:
@edolstra I think this is in a state for a yes-or-no decision, unless you have further feedback. For stuff like this, does it make sense to open it to the community or just have you decide?

copumpkin:
I want to propose another important application for this: disk images. In a disk image, I don't want to nuke the references, because that'll break my image. I also don't want to leave the references untouched, because then my image will retain references to store paths that built it. In some cases, I can compress or otherwise mangle the image, but that feels like a nasty hack to work around the absence of this functionality.

shlevy:
@copumpkin Can you open a PR adding that to this RFC?

copumpkin:
A PR against your PR source branch?

shlevy:
Yep!

copumpkin:
@shlevy and I have updated the RFC a bit to mention images and propose a way to share reference scanning logic between Nix and nixpkgs. @edolstra any new thoughts?

edolstra:
I'm not really in favor of this. Nix's conservative, automatic runtime dependency detection is one of its most important features. Moving responsibility for detecting dependencies to builders is a really major change. If we do this, then Nix's runtime dependency detection would not be fundamentally different from (say) RPM, which (IIRC) comes with a whole library of functions for detecting various kinds of dependencies, but you have to trust that they're doing a good job.
So I strongly disagree that dependency scanning should be moved to stdenv. Stdenv should not take over the job of the package manager.

In almost all cases, a reference either does not matter, in which case you can just get rid of it using `nukeReferences`, or it *does* matter, in which case it should not be ignored via the proposed interface since that will produce a dangling reference and therefore the danger of missing runtime dependencies. There are a few exceptions like disk images, but I'm not convinced that they're common enough to compromise on one of Nix's core properties.

It would be nice to move `nukeReferences` functionality into Nix, e.g. by having builders do something like `echo $path > $TMPDIR/references-to-nuke`. Nix can then get rid of the references via regular hash rewriting on NARs. This would be more efficient than letting builders do it (since Nix is going to scan the entire output anyway), and more correct (since hash rewriting operates on the entire NAR, while `sed` hacks don't).

A somewhat theoretical objection: currently it's possible in principle to reconstruct references from a NAR, given a Nix store. So if the Nix database is lost, you can reconstruct it by scanning every path in the store for references to every other path. This property would be lost if builders are responsible for determining references.

shlevy:
> Nix's runtime dependency detection would not be fundamentally different from (say) RPM, which (IIRC) comes with a whole library of functions for detecting various kinds of dependencies, but you have to trust that they're doing a good job.

I'm going to close this now, but I have to say this objection is ridiculous. If 95% of builds have their dependencies detected using the exact same logic as before, except it's managed by a script in stdenv rather than a C++ program in nix, and the rest have carved out carefully chosen exceptions, and that would make dependency detection not fundamentally different than RPM, then we're already not fundamentally different from RPM.

copumpkin:
@edolstra while I'm disappointed by this, can we treat it as a shared goal? If you don't like this approach, is there an approach you would like (or that you've thought of in the past) that achieves the same goals without the concerns you raised? 

I feel like we could probably get much of the benefit that @shlevy and I are seeking with a couple of files similar to what you're proposing: `$NIX_BUILD_TOP/references-to-ignore` (could be a magic value of `*` to say ignore all references, when e.g., making bootstrap tools or images), or `$NIX_BUILD_TOP/references-to-add`, which would provide us a non-output way to tell Nix that e.g., our compressed java jar files contain store paths that should be considered. Yes, this would lose your theoretical benefit above, but while nice, I'd prefer to get practical benefits out of all this, and the jar dependency issue is something I actually care about and makes me hesitate to add stuff to nixpkgs. I realize that we can already today create a dummy file adjacent to the jar in the output folder called something like `build-support/pile-o-references` for the jar case, but:

1. That means we can't do this with single-file outputs
2. We again run into your original objection that we have to duplicate some of Nix's scanning behavior to produce that file, or increase the space requirements by leaving an uncompressed version of the file lying around

When I think about it, it seems actually pretty rare to want to **nuke** references (in the binary rewriting sense, which is error-prone and can miss things), and the times we do it are just to hide references from Nix's automatic dependency scanning. We could still have a `$NIX_BUILD_TOP/references-to-nuke` to tell Nix to do the nuking for us, but I care less about that). There are also probably shenanigans we can do today with `unsafeDiscardStringContext` that might work similarly to the nuke/ignore behavior I want, but it feels much scarier and easier to screw up.

bennofs:
@copumpkin adding a file to `$out/build-support` would have the advantage of preserving the property that the nix database can be rebuild from the contents. Would it really be such a bad thing if you can't have single-file outputs with forced references?

@edolstra how do you feel about a solution where we specify explicitly which references to ignore? Then, while someone *could* come along and just scan the output and put all references there, that is much more complicated and is probably not going to happen by accident.

This whole issue seems to be about Nix's inability to "understand" different file types. @edolstra , are you opposed to any kind of smarter hash scanning (since that kinda feels like "the RPM way" to me), or just this extreme solution of moving everything to the stdenv side?

copumpkin:
@bennofs it'd still be duplicating Nix scan logic, if I for example scanned a jar for Nix references and dumped them to `$out/build-support`. Furthermore, I don't really have the information to do a proper Nix-style scanner, because I don't have any way (or do I?) to access the transitive requisites that Nix would scan for. I could search for `/nix/store/<somelonghash>` but that feels messier.

zimbatm:
How I see it this would move the nix hash scanning from runtime to buildtime. It would still be the same algorithm with some possible extensions.

I wonder how much time it would save on operations like GC to have all the hashes readily available in a single file per derivation output.

copumpkin:
@zimbatm it already happens at "build time", except not anywhere we can control from a build. The current build mechanism is basically as follows:

1. Nix when evaluating keeps track of all derivations your derivation might require. It uses these both when constructing a sandbox and when scanning for references.
2. Nix runs your build script
3. Nix scans your full outputs for all the hash references it tracked in step 1 (so if you include a random literal `/nix/store/abjkslfjdsklfj-foo` in your build script, I don't think that would count)
4. Nix records the outputs it found in its database for easy retrieval later

We're basically proposing a way of hooking into step 3 so that we can make it write something different to its database in step 4.

bennofs:
@copumpkin hmm. Perhaps you could copy the jar (or some other "plaintext" that you generated) (or symlink it) to some special location, and Nix will then scan it and place the result in `$out/somewhere` ?

copumpkin:
@bennofs: yeah, that's I guess mostly what I was hoping for.

How's this for a super easy proposal: if `$NIX_BUILD_TOP/override-references` (with optional per-separate-output suffix) exists, Nix will scan that file/folder instead of the equivalent output (using the exact same scanning logic we have today). That lets us suppress references (for images and bootstrap tools), add new references (for jars), and otherwise customize references _if a build explicitly opts in_. It would also be easy to `grep` in nixpkgs.

PR #7, opened 2017-04-01
Add future work section to the template

shlevy:
@zimbatm @edolstra

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -39,3 +39,9 @@ What other designs have been considered? What is the impact of not doing this?
 [unresolved]: #unresolved-questions
 
 What parts of the design are still TBD or unknowns?
+
+# Future work
+[future]: #future-work
+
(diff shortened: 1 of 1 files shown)


zimbatm:
@shlevy can you outline the motivation for this change?

shlevy:
Lots of changes imply or enable future work if they happen to take full advantage (e.g. a new nix feature allows us to improve something in nixpkgs), and having some thought up front as to what those changes entail makes the scope of the change clearer, even if those future work changes aren't strictly necessary to finish the feature itself.

zimbatm:
Would it be enough to add a related-to section where one could link and comment various items?

shlevy:
Sure, that works fine as well.

shlevy:
(FYI I don't have write access here)

Profpatsch:
Sounds like a good addition; most research papers have a “Future Work” section as well.

PR #8, opened 2017-04-02
[RFC 0008] Readonly recursive Nix

shlevy:
Rendered: https://github.com/shlevy/rfcs/blob/readonly-recursive-nix/rfcs/0008-readonly-recursive-nix.md

### DIFF
--- a/rfcs/0008-readonly-recursive-nix.md
+++ b/rfcs/0008-readonly-recursive-nix.md
@@ -1,0 +1,96 @@
+---
+feature: readonly-recursive-nix
+start-date: 2017-04-02
+author: Shea Levy
+co-authors: (find a buddy later to help our with the RFC)
+related-issues: https://github.com/NixOS/nix/issues/13
+---
+
+# Summary
+[summary]: #summary
+
+Allow nix builds to perform readonly operations on the subset of the
+nix store exposed to them.
+
+# Motivation
+[motivation]: #motivation
+
+The primary motivation for this feature is as an incremental step
+toward full recursive nix, described in [NixOS/nix#13][nix-issue-13]
+and to be fleshed out in its own RFC. However, even without the full
+solution this feature would replace the `exportReferencesGraph` nix
+feature and make it more general.
+
+# Detailed design
+[design]: #detailed-design
+
+There are a number of components to this work:
+
+## StoreViewStore
+[store-view-store]: #StoreViewStore
+
+A `nix::Store` implementation that forwards readonly requests to an
+underlying `Store` implementation, filtering both requests and
+responses to ensure that only paths that should be accessible to the
+build are returned (so e.g. `queryAllValidPaths` will only return the
+build time requisites of the build, not the full host store).
+
+Alternatively, this can just be a special mode of the nix daemon
+rather than a proper store api implementation.
+
+## FdDaemonStore
+[fd-daemon-store]: #FdDaemonStore
+
+A `nix::RemoteStore` implementation that opens new connections by
+making a request to a datagram-oriented unix domain socket passed into
+it as a parameter.
+
+## Build setup
+[build-setup]: #build-setup
+
+### Daemon availability
+
+The nix build code must ensure there is some daemon running. If the
+current build is a child of a nix-daemon process, it can just reuse
+the parent, otherwise it must spawn its own private nix-daemon process.
+This can be shared by all builds.
+
+### Connection setup
+
+For each build, the build code should create a socketpair, passing one
+socket to the build via `NIX_REMOTE` as appropriate for the
+[FdDaemonStore][fd-daemon-store], and reading packets from the other.
+When connection requests are received, the build loop should create a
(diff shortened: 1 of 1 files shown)


shlevy:
@edolstra

rfcs/0008-readonly-recursive-nix.md
```diff
@@ -0,0 +1,96 @@
...
+[FdDaemonStore][fd-daemon-store], and reading packets from the other.
+When connection requests are received, the build loop should create a
+new connection with the daemon, tell the daemon what it needs to know
+to set up a [StoreViewStore][store-view-store], and then pass the
+client socket through to the build.
+
+Multiple processes in the build may be trying to open connections in
+parallel, but as long as each only consumes one response it doesn't
+matter if the response is the exact one corresponding to the specific
+request made.
```
edolstra: I don't understand what this means. It seems to say there is some sort of race with multiple connections but I don't see how.

edolstra:
The idea seems mostly fine. A few comments:

* The main objection is that it significantly increases the attack surface from Nix builds, since they can now talk to a root process outside of the sandbox.

* Regarding daemon availability: A simpler approach would be to move all the daemon code from src/nix-daemon to libstore as a Daemon class that can be instantiated by DerivationGoal. So we just fire off a thread to handle incoming connections from the build. This also makes it easy to pass the StoreViewStore to the daemon (Daemon would just take a Store as a constructor argument).

* I don't see why FdDaemonStore is needed. Wouldn't the regular RemoteStore work? The only thing needed is a way to set the socket path (at least for non-sandbox builds where we can't create `/nix/var/nix/daemon-socket/socket`).

* I don't understand the name "StoreViewStore". Maybe RestrictedStore?

rfcs/0008-readonly-recursive-nix.md
shlevy: All processes in the build share the same socket FD they use to request a new connection, those requests could beinterleaved.

shlevy:
@edolstra thanks for taking a look!

> The main objection is that it significantly increases the attack surface from Nix builds, since they can now talk to a root process outside of the sandbox.

True, I'll include that with the next draft (and the counterpoint that the daemon is already meant to be accessible to everyone in a normal setup)

> Regarding daemon availability: A simpler approach would be to move all the daemon code from src/nix-daemon to libstore as a Daemon class that can be instantiated by DerivationGoal. So we just fire off a thread to handle incoming connections from the build. This also makes it easy to pass the StoreViewStore to the daemon (Daemon would just take a Store as a constructor argument).

Sure, makes sense.

> I don't see why FdDaemonStore is needed. Wouldn't the regular RemoteStore work? The only thing needed is a way to set the socket path (at least for non-sandbox builds where we can't create /nix/var/nix/daemon-socket/socket).

Inheriting a file descriptor is just an alternative to specifying the socket path.I slightly prefer the former, but not strongly, I'll change this to do it that way instead if you prefer.

> I don't understand the name "StoreViewStore". Maybe RestrictedStore?

Yeah, that's btter. Though the more I think about it the more it seems like this is more a parameter of the daemon than an actual standalone store implementation anyway.

joachifm:
> and the counterpoint that the daemon is already meant to be accessible to everyone in a normal setup

If I understand correctly, the concern is not necessarily other local users but that the proposal increases the feasibility of attacking the host by injecting malicious code into the build process.

edolstra:
Right. Currently sandboxed builds don't have access to the daemon socket.

shlevy:
Right, it's not a complete counterpoint, just that the daemon already should be pretty secure.

shlevy:
We could put this behind an off-by-default option.

Ericson2314:
I'd like to see some of the comparison with import-from-derivation fleshed out. For example, I'm missing why this is more expressive.

shlevy:
@edolstra @Ericson2314 Updated in response to your comments.

Ericson2314:
@slevy I'll respond in detail later but I think all those IFD drawbacks are very surmountable.

shlevy:
@edolstra can we expect a response either way from you here?

edolstra:
This looks good to me, thanks!

shlevy:
@edolstra Awesome! :confetti_ball: 

@zimbatm So what's the next step from the perspective of the RFC process?

Ericson2314:
@edolstra, would you also be OK in an RFC for improving import-from-derivation? They are in many cases competing ways to achieve the same thing, but implementation-wise there is no conflict.

0xABAB:
I think the RFC document (the one in the first message) is so vague that I have no idea what problems it intends to solve. Just based on that, I am already against.

zimbatm:
@shlevy the RFC is very dry right now. @edolstra and you can understand it because you have a lot of shared context on the nix implementation.

To help the user's perspective, would you mind adding an example of nix code that would use the feature and then describe the various steps that the runtime would take when invoking the code?

shlevy:
@zimbatm good call, will do when I get a chance

PR #9, opened 2017-04-04
[RFC 0009] Nix rapid release

shlevy:
Rendered: https://github.com/shlevy/rfcs/blob/nix-rapid-release/rfcs/0009-nix-rapid-release.md

### DIFF
--- a/rfcs/0009-nix-rapid-release.md
+++ b/rfcs/0009-nix-rapid-release.md
@@ -25,9 +25,9 @@ more considerate of what we consider merge-ready.
 # Detailed design
 [design]: #detailed-design
 
-1. Get nix 1.12 out ASAP. Identify the bugs that need fixing and
+1. Get a new nix out ASAP. Identify the bugs that need fixing and
    optionally features that absolutely need finishing (or possibly
-   temporary reverting), get people to test, and release it.
+   temporary reverting), get people to test, and release it as `2.0.0`.
 2. Do new non-trivial work in feature branches
 3. If a feature branch may introduce regressions, including
    performance regressions, ensure they are tested by relevant parties
--- a/rfcs/0009-nix-rapid-release.md
+++ b/rfcs/0009-nix-rapid-release.md
@@ -14,37 +14,33 @@ Nix master is always release-ready and new releases are cut often.
 # Motivation
 [motivation]: #motivation
 
-Currently, nix releases are unpredictable and new features (and, less
-often, bug fixes) sit on master for a long time before people can rely
-on them. nix 1.12 in particular has a large number of changes that are
-not generally available yet. A rapid release policy would allow more
-users to benefit from the latest greatness while also forcing us to
-make our improvements to nix more incremental and unitary and to be
-more considerate of what we consider merge-ready.
+Currently, nix releases are unpredictable and new features (and, less often, bug fixes) sit on master for a long time before people can rely on them.
+nix 1.12 in particular has a large number of changes that are not generally available yet.
+A rapid release policy would allow more users to benefit from the latest greatness while also forcing us to make our improvements to nix more incremental and unitary and to be more considerate of what we consider merge-ready.
 
 # Detailed design
 [design]: #detailed-design
 
-1. Do new non-trivial work in feature branches
-2. If a feature branch may introduce regressions, including
-   performance regressions, ensure they are tested by relevant parties
-   before merge
-3. Release master frequently. Community members can make a request for
-   a new release and it's almost always granted.
+1. Do new non-trivial work in feature branches.
+
+2. If a feature branch may introduce regressions, including performance regressions, ensure they are tested by relevant parties before merge.
+
+3. Release master frequently.
+   Community members can make a request for a new release and it's almost always granted.
+
 4. If need arises, start maintenance branches that *only* do bug fixes.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-This will change developer workflows. It also may make some large-scale
-features harder to develop, though in practice when following this
-policy on other projects it hasn't seemed much of an issue.
+This will change developer workflows.
+It also may make some large-scale features harder to develop, though in practice when following this policy on other projects it hasn't seemed much of an issue.
 
 # Alternatives
 [alternatives]: #alternatives
 
-Status quo, mostly. We could also switch to a more up-front planned
-releases, or merge windows, or time-based releases.
+Status quo, mostly.
+We could also switch to a more up-front planned releases, or merge windows, or time-based releases.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
@@ -54,7 +50,6 @@ The steps needed to get 1.12 out.
 # Future work
 [future]: #future-work
 
-1. Get a new nix out ASAP. Identify the bugs that need fixing and
-   optionally features that absolutely need finishing (or possibly
-   temporary reverting), get people to test, and release it as `2.0.0`.
-2. Switch to semver
+1. Get a new nix out ASAP.
+   Identify the bugs that need fixing and optionally features that absolutely need finishing (or possibly temporary reverting), get people to test, and release it as `2.0.0`.
+2. Switch to semver.
--- a/rfcs/0009-nix-rapid-release.md
+++ b/rfcs/0009-nix-rapid-release.md
@@ -2,7 +2,7 @@
 feature: nix-rapid-release
 start-date: 2017-04-04
 author: Shea Levy
-co-authors: (find a buddy later to help our with the RFC)
+co-authors: John Ericson (@Ericson2314)
 related-issues: (will contain links to implementation PRs)
 ---
 
--- a/rfcs/0009-nix-rapid-release.md
+++ b/rfcs/0009-nix-rapid-release.md
@@ -14,9 +14,9 @@ Nix master is always release-ready and new releases are cut often.
 # Motivation
 [motivation]: #motivation
 
-Currently, nix releases are unpredictable and new features (and, less often, bug fixes) sit on master for a long time before people can rely on them.
-nix 1.12 in particular has a large number of changes that are not generally available yet.
-A rapid release policy would allow more users to benefit from the latest greatness while also forcing us to make our improvements to nix more incremental and unitary and to be more considerate of what we consider merge-ready.
+Currently, Nix releases are unpredictable and new features (and, less often, bug fixes) sit on master for a long time before people can rely on them.
+Nix 1.12 in particular has a large number of changes that are not generally available yet.
+A rapid release policy would allow more users to benefit from the latest greatness while also forcing us to make our improvements to Nix more incremental and unitary and to be more considerate of what we consider merge-ready.
 
 # Detailed design
 [design]: #detailed-design
@@ -50,6 +50,6 @@ The steps needed to get 1.12 out.
 # Future work
 [future]: #future-work
 
-1. Get a new nix out ASAP.
+1. Get a new Nix out ASAP.
    Identify the bugs that need fixing and optionally features that absolutely need finishing (or possibly temporary reverting), get people to test, and release it as `2.0.0`.
 2. Switch to semver.
--- a/rfcs/0009-nix-rapid-release.md
+++ b/rfcs/0009-nix-rapid-release.md
@@ -33,11 +33,8 @@ more considerate of what we consider merge-ready.
    performance regressions, ensure they are tested by relevant parties
    before merge
 4. Switch to semver
-5. Release master frequently. Personally I'm in favor of every new
-   successful hydra build being a release (and, once we get proper PR
-   integration, we shouldn't have merges before hydra is green), but
-   probably a more realisitic step would be that people can make a
-   request for a new release and it's almost always granted.
+5. Release master frequently. Community members can make a request for
+   a new release and it's almost always granted.
 6. If need arises, start maintenance branches that *only* do bug fixes.
 
 # Drawbacks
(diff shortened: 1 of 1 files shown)


shlevy:
@edolstra

shlevy:
(I really wish I had commit access here so I could request reviewers etc....)

vcunat:
👍 for the principle at least (I haven't read details yet).

Profpatsch:
+1, one addition: Release nix `1.12` as nix `2.0.0` and do semver from there.

shlevy:
@profpatsch good idea, pushed.

vcunat:
The release number policy seems a mostly independent issue to me, and I wouldn't like to get blocked on that.  For example, version 2.0.0 might suggest backward-incompatible changes in the nix _language_, but that's not really planned AFAIK.

shlevy:
It's not completely independent, but I'm not averse to changing it if that's the only concern.

domenkozar:
I'm very much in favor of this :)

Profpatsch:
> For example, version 2.0.0 might suggest backward-incompatible changes in the nix language, but that's not really planned AFAIK.

Completely revamping the CLI interface (which is a major interface change) and rewrite of major parts in C++ (trimming 40MB from the package manager size) should be enough for a 2.0 I think.
Also there hasn’t been a release in over a year, so there’s probably lots of breaking changes apart from that.

shlevy:
The old interface still exists, FWIW.

shlevy:
@domenkozar Co-author?

shlevy:
@zimbatm This RFC clearly has community support, but no one has stepped up to be a coauthor. Frankly I think that requirement just creates more work than needed and requires RFC authors to be pests :frowning: . Any chance we can remove it?

Ekleog:
I don't really think 3 days' notice is enough to gather all counter-arguments that could potentially arise, but I guess that's orthogonal to the coauthor requirement (that said, if noone is stepping up to coauthor, given the really basic requirements for it, maybe it's not that much wanted by the community? -- anyway this discussion about removing coauthor should go somewhere, like in comments in a RFC to remove the requirement)

Then, here are two issues with this, to me. First, I think words like "Personally" should not appear in a RFC. The final text should be the One True Source of Wisdom once accepted, so personal opinion should not matter. But the real issue is that this RFC is too extreme in my opinion.

Stability of a program as defined by "ability to make a release" greatly varies depending on the project. Having master always building and passing the test suite is not an issue but clearly not enough, as test suites always let things through. I think that having a commit on master always meaning a release is just too much, for it means *every* PR has to be completely bug-free before being released.

That said, I think having a faster release cycle would indeed be useful. But there is a need to actually do something consistent, that lets bugs introduced by new PRs be discovered by the testers running off master branch (or even by manual functional testing, just before a release). An intermediate solution may be to simply tag all issues with a milestone, and as soon as all the issues tagged with the next milestone are solved it can be released.

This, assuming not too many features are tagged for the next milestone, means a fast release cycle (hopefully one new feature per release at most, serious issues can even get a bugfix release by just setting them as the only issue for the next milestone), better tracking of what still needs to be done for next release, yet still allows bugs to be discovered on master before the release, and also to be fixed before it by making the release depend on them.

What do you think about it?

shlevy:
@Ekleog
> I don't really think 3 days' notice is enough to gather all counter-arguments that could potentially arise, but I guess that's orthogonal to the coauthor requirement

Yeah, I wasn't expecting everyone to chime in in 3 days! Just that a bunch of people have expressed lots of support for this one. I don't think it's really a reflection of how much interest there really is, but maybe I'm wrong there.

> anyway this discussion about removing coauthor should go somewhere, like in comments in a RFC to remove the requirement

Yeah, I wanted to open up an issue but this repo has them disabled :open_mouth: Not sure this really merits an RFC... And who would coauthor it? :grin: 

> First, I think words like "Personally" should not appear in a RFC. The final text should be the One True Source of Wisdom once accepted, so personal opinion should not matter.

Fair point, I'll remove that bit.

> Having master always building and passing the test suite is not an issue but clearly not enough, as test suites always let things through.

This is always true and why we have bug fixes and PR authors should be careful to manuall test and think of corner cases. I don't think we get significant extra test coverage by having long release cycles.

> I think that having a commit on master always meaning a release is just too much, for it means every PR has to be completely bug-free before being released.

That's removed now, if there has been a big feature merge recently we can hold off until testing can happen if need be.

> That said, I think having a faster release cycle would indeed be useful. But there is a need to actually do something consistent, that lets bugs introduced by new PRs be discovered by the testers running off master branch (or even by manual functional testing, just before a release).

I'm not sure what this means/how it differs from my idea

> An intermediate solution may be to simply tag all issues with a milestone, and as soon as all the issues tagged with the next milestone are solved it can be released.

In my experience these just grow, and aren't maintained well.

> What do you think about it?

Personally I think just a delay/testing window manually imposed when a change is worrisome is a better approach.

Ekleog:
On 04/07/2017 09:23 PM, Shea Levy wrote:
> This is always true and why we have bug fixes and PR authors should be
> careful to manuall test and think of corner cases. I don't think we get
> significant extra test coverage by having long release cycles.

Well, I've had experience with uncovering bugs introduced after the last
release while testing a new feature I was implementing on another big
community-based project, so I guess it's actually useful.

>     That said, I think having a faster release cycle would indeed be
>     useful. But there is a need to actually do something consistent,
>     that lets bugs introduced by new PRs be discovered by the testers
>     running off master branch (or even by manual functional testing,
>     just before a release).
> 
> I'm not sure what this means/how it differs from my idea

Well, that was just stating that people who test corner cases of a new
commit on master can find a bug in a previous commit of master, and hope
it's not released yet.

>     An intermediate solution may be to simply tag all issues with a
>     milestone, and as soon as all the issues tagged with the next
>     milestone are solved it can be released.
> 
> In my experience these just grow, and aren't maintained well.

If it's strictly defined as in "when a release is made, a number of
features for the next release is picked from the open issues list, then
nothing else than *important* bugfixes can be added to the milestone",
and not adding any feature before the release, then it wouldn't grow,
would it? And it doesn't require any maintenance over just closing the
issues and picking which ones to work on ahead of time, which anyway is
necessary.

It would be a kind of "feature freeze": at release time, the features
for the next release are already selected.

>     What do you think about it?
> 
> Personally I think just a delay/testing window manually imposed when a
> change is worrisome is a better approach.

Unfortunately worrisome changes are maybe not the only important ones.
Like, I've sometimes unwillingly written code that looked innocuous, did
little, but as the project was huge (like nixpkgs is), it had a
side-effect on a completely unrelated part of the code. This kind of bug
introduction is what can hopefully be found by someone working on
another feature that would touch this part of the code.

Then of course, the number of features added between two releases would
have to be small, in order not to have something too slow to move
forward, but it would give at least a view of why the next release is
not out yet, which is what I think we maybe need most (I'm pretty sure
there is a valid reason), and a way to know what to fix in order to get
it out.

It would be a kind of improved TODO-list, in my mind.

That said, I'll let someone else chime in on this topic, as obviously
I'll be biased towards my idea, and I think I explained it clearly
enough for it to be understood :)

zimbatm:
@shlevy the co-author is more interesting for long RFCs, it's to avoid having 1000 comments on typo and such when the important bit is the content. At the end of day it's just a process to help us.

I feel like this RFC is a reaction to 1.12 taking ages to be released. This is probably better solved by defining what's left to do to get 1.12 out of the door no?

Who is the release manager for nix, @edolstra? What are the steps required to get it released?

shlevy:
@zimbatm It's not just about 1.12, it's been the norm for all previous nix releases and makes it difficult to solve problems that involve upstreaming features.

Ericson2314:
@shlevy I really ought not to be coauthor as I've never contributed to Nix itself and will be traveling in May, but if no one steps up I'll do it. It's very important that this happen.

zimbatm:
@shlevy since this is a process-type RFC it should probably also identify who is in charge of release management. Who has the power to do it and who is volunteering to do it, what are the responsibilities, ... Otherwise the RFC will be discussed and there will be nobody to implement it.

shlevy:
@zimbatm I'm happy to do it, and I would not be surprised if there are others if not me, but only @edolstra can decide that.

shlevy:
@edolstra Can you chime in here? This one seems to need your goahead

shlevy:
@edolstra can we expect a response either way from you here?

aneeshusa:
Big +1 from me on this RFC, especially having master more "release-ready" at all times - I work on other projects where this makes it much nicer to develop. I also think the RFC process can help with larger changes to Nix itself and help avoid regressions.
For Nix 1.12 (or 2.0) specifically, I would consider fixing https://github.com/NixOS/nix/issues/1356 necessary before release.

Ericson2314:
I will say, if https://github.com/NixOS/rfcs/pull/14 is passed I will have a much more enjoyable time implementing it were this RFC policy.

Ericson2314:
One compromise position is to have a `staging` branch for DB schema transitions.

Ekleog:
OK, so re-reading this thread, everyone seems to agree this RFC is needed, except me saying sometimes a bit of time between development of a new feature and a release may help.

Having re-read the RFC, I still think it's much better than the status quo, and would like to withdraw my negative comments (the thing I think I was opposed to was having a release on each master commit, the current state of the RFC looks great to me).

With this move made there seem to be no longer any negative comment.

Could we move forward, then? (I'm thinking having long-lived RFCs where everyone agrees is not a good thing for future RFCs to come up: it reduces the motivation of people to actually make them)

PS: I'm assuming @edolstra's not answering this thread is an implicit way to tell the community “you made the RFC process, now use it and stop relying on me so much”. Which would mean we should go forward and land this RFC if we actually want it implemented. If I'm wrong, please, by all means say it.

shlevy:
@Ekleog I don't think that's a fair way to interpret @edolstra's lack of response here. I've been explicitly chastised for adding release tags to the nix repo without his say up until now and he is the primary active developer on the project. I don't see a way to proceed with this issue in particular without his say-so. And it would take 2 minutes for him to leave a comment suggesting that we should go ahead based on community consensus here.

domenkozar:
Imho, the right way to go is to have a release manager, the same way we did for NixOS.

shlevy:
Happy to do that or find someone to do it, but that process itself will require at least initial signoff from @edolstra

Ekleog:
@shlevy You are right about the 2 minutes required to leave a comment suggesting we should go ahead based on community consensus, but what stuns me is that he actually changed the PR title on Apr 10 (meaning he noticed and read it), yet left no comment... Obviously, I may be wrongly interpreting it, and you have much more experience than me :)

Ericson2314:
I offered a few changes, mainly making the first two bullets stricter. What do you all think?

vcunat:
I'm afraid the efforts are doomed unless/until Eelco shows some support.  (I don't know if that might be helped by someone pledging some effort to make the releases happen.)

shlevy:
I'm willing to commit to reasonable efforts (including ensuring test coverage, making a timeline, tracking down bugs, etc.) to get the next release out and set us up to move on to a rapid release model if approved.

shlevy:
@edolstra Can you please let us know what further info/changes/commitments would be necessary to enable to you make a decision on this?

0xABAB:
I am not in favor of solving a technical problem by changing the process. I also don't see the need, because it is possible to control exactly the version of every single component in a running system as an end user (which is a fairly unique capability of NixOS).

Ekleog:
Given `nixpkgs` has to be compatible with stable nix (per policy), not having nix 1.12 blocks a number of potential changes (like the use of encryption as per RFC #5 when it's accepted, or `nixos-prepare-root` that would make for a cleaner implementation of #12, for the two that I can think of).

So having nix 1.12 released and then trying to release a bit more often would most likely help, even though obviously it doesn't mean to let bugs sink in a release, and this process of releasing often while not letting partially written modifications in is (I think) described here.

0xABAB:
@edolstra does vastly more work (apparently 9 times the number of commits of the second committer) on Nix. As such, while it would be cool if he would like to modify his way of working, we can't actually make him do this; without @edolstra there defacto is no Nix (but nixpkgs would still be viable).

I think if @edolstra doesn't ever respond, the only option is to fork.

I am not saying we should fork, because overall `nix` works fairly well and seems to improve over time. 

This RFC is a bit like Australia and the UK voting about whether the US should invade Syria. 

The way this should have been handled was sending an e-mail to @edolstra (assuming he doesn't mind those from @shlevy) asking whether he thinks it would be a good idea and if the most important stakeholder was on board (that's what the RFC process prescribes even) then it would have been a good time to ask what the rest of the community thinks about it (if at all), since defacto @edolstra's opinion is the only one which matters, due to the fact of him being the owner of the repository. 

The only reason for asking the community would be to decrease the risk of alienating the community, but the risk of that is so incredibly low in that case that it doesn't seem useful to even ask.

zimbatm:
I still think this RFC is a reaction to 1.12 taking ages to be released. And @edolstra is not communicating which makes it hard to know what to do next. I can only guess but I'm assuming there are two major reasons not to do the release:

* Not enough testing. The tool is not stable enough.
* The new UX is not ready. It's a big change which has deep implications for the future so understandably one would want to make sure it's in the right shape.

Based on those assumptions I would say that the best path forward is to (1) find and fix all remaining stability issues. (2) make a nix 1.12 release where the `nix` dispatcher is disabled. That would let us have the release while still maturing the `nix` dispatcher.

vcunat:
- (UX not ready): I don't think that's a reason to stop the release.  The old `nix-foo` interface still works OK on `1.12pre*` and we can explicitly say that the new `nix` command is experimental and possible to change completely.

vcunat:
- (not enough testing): look at [Hydra](http://hydra.nixos.org):
  > Hydra 0.1.1234.abcdef (using nix-1.12pre5413_b4b1f452).

  I think that's quite some testing by itself, at least for some parts of nix.  It has one specific configuration, but some potential problems, e.g. recent setuid/setgid "failures" in builders, are detected that way.  I've been running 1.12 on my machine where I do most of nix* stuff, as another small point.

PR #10, opened 2017-04-04
[RFC 0010] Nixpkgs development support

shlevy:
Rendered:  https://github.com/shlevy/rfcs/blob/nixpkgs-development-support/rfcs/0010-nixpkgs-development-support.md

### DIFF
--- a/rfcs/0010-nixpkgs-development-support.md
+++ b/rfcs/0010-nixpkgs-development-support.md
@@ -1,0 +1,54 @@
+---
+feature: nixpkgs-development-support
+start-date: 2017-04-04
+author: Shea Levy
+co-authors: (find a buddy later to help our with the RFC)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Add a `development-support` section to `nixpkgs` with functions useful
+for those using nix in their software development.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, nixpkgs has some features, like `callCabal2nix` and the
+`env` attribute of haskell packages, that don't really have anything
+to do with package management per se but rather provide support for
+developers who use nix to manage their projects. With the merging of
+the `builtins.exec` nix feature, there are a number of developer tools
+I'd like to make widely available, such as something to translate a
+.gitignore into a filterSource call or do no-sha256 fetchgit with
+caching. It would be good to have a single blessed place to put these
+kinds of tools.
+
+# Detailed design
+[design]: #detailed-design
+
+Add `pkgs/development-support`, add `pkgs.development-support` calling
+the functions defined in `pkgs/development-support`.
+
+Any tool relying on `builtins.exec` will take that function as an
+argument, rather than using the builtin directly, so that all of
+nixpkgs evaluates properly without
+`allow-unsafe-native-code-during-evaluation`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Arguably nixpkgs is about package management and shouldn't necessarily
+be the one source for all things shared nix.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Separate repository with these tools defined, but then we lose tight
+integration with the packages they depend on.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+None I think.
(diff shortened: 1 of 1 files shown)


shlevy:
@edolstra @peti @domenkozar @globin

edolstra:
This sounds like it could easily become a dumping ground of random crap. What would be the inclusion criteria? The alternative - using separate repositories - seems fine for development since people can use `builtins.fetchgit` to get them. If some tool becomes popular, it can always be merged into the Nixpkgs repo.

Also, I'm afraid `builtins.exec` may become a Pandora's box turning Nix into an imperative language.

shlevy:
I'm reluctant to assume the availability of `builtins.fetchgit` until 1.12 is released.

We don't have explicit inclusion criteria elsewhere in nixpkgs, do we? It should be something generally usefuil and maintainable and up to basic standards of code cleanliness.

Alternate repositories can't be kept in tight lockstep with nixpkgs. This is why we have nixos in nixpkgs, right?

`builtins.exec` is only turned on if you ask for it, and it solves some significant real problems. Do you have an alternative?

edolstra:
Another issue: How does this differ from `pkgs/build-support`?

rfcs/0010-nixpkgs-development-support.md
```diff
@@ -0,0 +1,54 @@
...
+# Detailed design
+[design]: #detailed-design
+
+Add `pkgs/development-support`, add `pkgs.development-support` calling
+the functions defined in `pkgs/development-support`.
+
+Any tool relying on `builtins.exec` will take that function as an
+argument, rather than using the builtin directly, so that all of
+nixpkgs evaluates properly without
+`allow-unsafe-native-code-during-evaluation`.
```
edolstra: I don't see why this is necessary.

rfcs/0010-nixpkgs-development-support.md
```diff
@@ -0,0 +1,54 @@
...
+the `builtins.exec` nix feature, there are a number of developer tools
+I'd like to make widely available, such as something to translate a
+.gitignore into a filterSource call or do no-sha256 fetchgit with
+caching. It would be good to have a single blessed place to put these
+kinds of tools.
+
+# Detailed design
+[design]: #detailed-design
+
+Add `pkgs/development-support`, add `pkgs.development-support` calling
```
edolstra: -> `pkgs.developmentSupport`

shlevy:
@edolstra `build-support` is useful things that help nix builds. development-support is mostly evaluation time stuff.

rfcs/0010-nixpkgs-development-support.md
shlevy: It makes it possible with a grep (possibly even enforced before merging) that nothing you build in nixpkgs can cause a problem if you have `allow-unsafe-native-code-during-evaluation` on.

rfcs/0010-nixpkgs-development-support.md
shlevy: Right, fixing

copumpkin:
I think it's partly about discoverability and partly about co-versioned-ness. In many ways the closer analog to what you're talking about feels like the maintainers folder and possibly @aszlig's md5 collision abomination. These both strongly suggest to me that we want tooling for working on the nixpkgs repo, and want many of the goodies that Nix provides when working on Nixpkgs itself

shlevy:
@copumpkin Want to coauthor?

shlevy:
Created https://github.com/shlevy/nixpkgs-development-support for now

rfcs/0010-nixpkgs-development-support.md
```diff
@@ -0,0 +1,54 @@
...
+# Motivation
+[motivation]: #motivation
+
+Currently, nixpkgs has some features, like `callCabal2nix` and the
+`env` attribute of haskell packages, that don't really have anything
+to do with package management per se but rather provide support for
+developers who use nix to manage their projects. With the merging of
+the `builtins.exec` nix feature, there are a number of developer tools
+I'd like to make widely available, such as something to translate a
+.gitignore into a filterSource call or do no-sha256 fetchgit with
```
michalrus: :heart_eyes:

0xABAB:
Another feature I would not want to see. Others have given enough arguments already.

PR #13, opened 2017-05-04
[RFC 0013] Ergonomic `cmakeFlags`

aneeshusa:
Rendered: https://github.com/aneeshusa/nixos-rfcs/blob/ergonomic-cmakeFlags/rfcs/0013-ergonomic-cmakeFlags.md

There are still a few TODOs, but most of the RFC is available.

### DIFF
--- a/rfcs/0013-ergonomic-cmakeFlags.md
+++ b/rfcs/0013-ergonomic-cmakeFlags.md
@@ -1,0 +1,142 @@
+---
+feature: ergonomic_cmakeFlags
+start-date: 2016-08-20
+author: Aneesh Agrawal
+co-authors: (find a buddy later to help out with the RFC)
+related-issues: https://github.com/NixOS/nixpkgs/pull/17886
+---
+
+# Summary
+[summary]: #summary
+
+Make usage of `cmakeFlags` in Nixpkgs more ergonomic
+by representing them as structured data in the form of attribute sets,
+and passing them more intelligently to the builder.
+
+# Motivation
+[motivation]: #motivation
+
+Improvements:
+- Enable first-class Nix handling of `cmakeFlags`,
+  including easier extension and overriding
+- Handle spaces, newlines and other whitespace correctly
+- Make it possible to reference build-time environment variables directly
+
+# Detailed design
+[design]: #detailed-design
+
+There are two main parts to this change:
+- Using attribute sets for `cmakeFlags` in Nix
+- Passing `cmakeFlags` to the builder more intelligently
+
+## Using Attribute Sets for `cmakeFlags`
+
+Almost all flags passed to `cmakeFlags` are setting CMake cache options
+with `"-D<name>=<value>"` or `"-D<name>:<type>=<value>"`.
+Because the CMake cache is essentially a set of keys and values,
+model `cmakeFlags` as an attribute set in Nix as well.
+
+`mkDerivation` will take an incoming `cmakeFlags` set and convert each
+key/value pair to an appropriate `"-D"` option for CMake. String values
+are used as is, while integers are converted to strings and booleans are
+converted to `"ON"` and `"OFF"`. This presents the following benefits:
+- Removal of `"-D"` visual noise everywhere for ease of reading.
+- The `=` operater is literal, allowing extra whitespace for more ease
+  of reading.
+- CMake options are now first class Nix values that can be operated on
+  directly with the full strength of the Nix language. For example,
+  boolean values can be used directly because the serialization logic is
+  consolidated into `mkDerivation`, getting rid of helpers like `edf`
+  and making it easier to re-use feature flags like `pythonSupport`.
+- Sets automatically sort options, improving cache re-use.
+- An override which updates a CMake flag option will only incur
+  (amortized) O(1) cost with a set, as opposed to O(n) cost with an
+  (unsorted) list, which would require a full linear search. It's also
+  shorter to write: just use `//`.
+
+A few other CMake flag types are supported:
+- Setting a value to null will cause its key to be unset via `"-U<name>"`.
+- The `generator` key is mapped to the `"-G<value>"` option.
+- The `extraArgs` key is an escape hatch; any flags in this list of
+  strings are not pre-processed, but concatenated to the generated list.
+
+Because `cmakeFlags` will now be processed by `mkDerivation`,
+`overrideAttrs` should be used instead of `overrideDerivation`
+when overriding `cmakeFlags`.
+
+TODO: Add example(s)
+
+## Passing `cmakeFlags` to the builder more intelligently
+
+Nix's regular conversion of lists to an environment variable
+simply concatenates the contents with spaces,
+which breaks if any of the flags themselves contain spaces.
+`cmakeFlagsArray` was added as a way to get around this limitation,
+but is unsatisfactory to use because items must be appended via bash
+in a preConfigure hook.
+
+Instead, pass the list to the builder in a smarter way:
+instead of relying on Nix's conversion to a string,
+perform our own conversion by escaping double quotes,
+double quoting each item, and passing the flags as a Bash array,
+which is hydrated via `eval`.
+
+This handles flags with spaces, newlines and other whitespace,
+as well as double quotes, faithfully.
+We already use `eval` heavily in Nixpkgs builders,
+so this does not pose any additional security risks.
+
+Additionally, the use of eval makes it possible to reference build-time
+environment variables such as `$out` directly when setting `cmakeFlags`,
+instead of needing to set up these flags in `preConfigure`.
+
+Make the list available during preConfigure as a bash array, so any
+dynamic modifications to the CMake flags can be done there.
+
+These changes make also `cmakeFlagsArray` redundant,
+so remove and replace it in all cases with `cmakeFlags`.
+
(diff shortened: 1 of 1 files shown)


Ericson2314:
This sounds great—both the design proper and the follow-up proposals, which mirrored exactly my thoughts as I read the detailed design.

The modularity problem is interesting, and I'd like to come back to it, but the core benefit here is enough I'm happy to punt for now.

Also fwiw I learned the other day there is some extra stuff to pass for all cmake derivations with `cmakeFlags` for cross compilation. This would also make doing that more elegant.

aneeshusa:
@Ericson2314 the cross-compilation use case sounds like it could make a good example to add to the RFC, do you mind expanding on that?

I do have some vague ideas on the modularity aspect, but agreed on wanting to punt that for now.

Ericson2314:
Sure https://github.com/NixOS/nixpkgs/pull/25218/files#r113298669

Ericson2314:
BTW https://github.com/NixOS/nixpkgs/blob/master/lib/composable-derivation.nix is some ancient thing for `configureFlags`.

0xABAB:
If the implementation of this is rock solid, I am in favor. If it is implemented OKish, I am not on board. I have concerns about `eval` in particular. Merely suggesting `eval` seems to suggest a lack of general shell scripting experience. 

So, the general idea of having special support for CMake is reasonable, although not strictly needed, but the devil will be in the details. I would consider all existing instances of `eval` to be candidates for being a bug. I.e., just because someone used eval, doesn't mean we should continue to do it.

globin:
@aneeshusa: In order to move this forward the @NixOS/rfc-steering-committee would like to open this up for shepherd nominations in case you are still interested in completing this and think it generally is ready for further discussion? (see the new RFC workflow defined in #36 and in the updated README)

aneeshusa:
@globin I'm still interested in this, yes! I was just looking at the associated nixpkgs PR yesterday and [left a comment](https://github.com/NixOS/nixpkgs/pull/17886#issuecomment-452584555) with a number of improvements to the approach, many based on new features of Nix 2 and which fill in some of the TODOs here.

I'd like to get a chance to update this RFC (likely this weekend) before opening it up for a ton of debate. Not 100% sure of the process but seems reasonable to open this up for shepherd nomiations in the meantime.

Mic92:
@aneeshusa we will meet next Thursday for the shepherd nominations.
It would be great if you can update your status until then.

globin:
I think it would be fine to open this up for nominations then right away, since the general idea is clear, fleshing it out with the shepherds is the idea anyway!

=> **Open for nominations**

I'd also hereby nominate @Ericson2314, @matthewbauer as shepherds.

shlevy:
I nominate @Ericson2314

shlevy:
Oops guess @globin beat me to it :joy:

Ericson2314:
@cleverca22 nix-level setup hooks was how @aneeshusa originally did it. There is also a third way which is a `buildCMakePackage`. I think I learn towards that as getting nix-level setup hooks to compose sanely is some work.

layus:
I agree with you. Something in the lines of `mkCMakeDerivation` makes more sense than adding support for very generic names like `generator` in the already featureful `mkDerivation`. `pkgs.cmake.mkDerivation` is also a good option.

Ericson2314:
We should probably standardize `buildFooPackage` vs `mkFooDerivation` too. Odd we mainly have the former.

edolstra:
I'm strongly in favor of doing this as part of a migration to structured attributes. That way, the implementation doesn't require any special handling at evaluation time (like a `mkCMakeDerivation`) and can be handled in the cmake setup hook just as it is today.

Ericson2314:
@edolstra that's fine, but in general getting setup hooks to commute / play nice is already somewhat non trivial. I don't think setup hooks a sustainable model with or without structured attrs.

globin:
Per meeting of the @NixOS/rfc-steering-committee: Shepherding Team is @edolstra as leader, @Ericson2314 and @matthewbauer as long as they both agree otherwise we will follow up with further candidates.

Ericson2314:
I accept!

aneeshusa:
I agree with @edolstra that structured attributes would be the best way to handle this. A couple questions as I haven't gotten the chance to play with them:
- What are the downsides to turning them on, other than requiring Nix 2? @cleverca22 you mentioned `it can break things` - can you expand?
- Based on the answers to that question, what would the timeline to turn structured attributes on globally look like? Have we as a community started planning a path forward for that?

One idea is to enable structured attrs conditionally. One option would be checking for `cmakeArgs` directly in `mkDerivation`, which is tight coupling as @cleverca22 points out. A better idea I had was to enable structured attrs if there are any fields which are non-derivation attrsets - this would fail evaluation without structured attrs so it's safe to do (can't break existing code), isn't cmake specific, and presumably lets us start using structured attributes today.

That being said, if there's good reasons to wait on structured attrs entirely, I would love to use one of the other mechanisms as a fallback so we can take advantage of the better interface in the meantime, and clean up the implementation later.

About the other ideas:
1. adding directly to mkDerivation: My PR used to do this. I'm not a fan of this, hence why I changed my PR over to nix-level setup hooks.
1. nix-level setup hooks: I currently have these implemented (prototype stage) in the accompanying PR. These are a bit icky (I don't think there would be a need for these once we have structured attrs), but they seem like the least-bad alternative. Still on the fence about if these are worth doing with the prospect of a structured attrs future. One question: @Ericson2314, why would we need composability of these? I would expect any of these to only touch attributes they own/consume themselves and wasn't planning on supporting composability.
1. buildCMakePackage - I don't want to make cmake special here, I really like our existing setup hooks user experience where you simply have to add `cmake` as a native build input and I don't want to degrade that. That also breaks composability if every tool that needs special code to run has a separate function.

Apologies for the slow reply - once I get some feedback on these points I'll try to add these alternatives to the RFC, hopefully this weekend.

aneeshusa:
Also, I think the `status: open for nominations` label is out of date - can that please be updated?

Ericson2314:
> this would fail evaluation without structured attrs so it's safe to do (can't break existing code), isn't cmake specific, and presumably lets us start using structured attributes today.

Actually today richer data is often just stripped away. Also arrays in particular become strings, which the setup script bash much work around.

> why would we need composability of these? I would expect any of these to only touch attributes they own/consume themselves and wasn't planning on supporting composability.

What happens if `cmake` and say `meson` or `build2` are in the same derivation? Which one gets to do the build or configure phases? Building relies on one such build tool being primary and delegating to the others, which setup hooks aren't really qualified to figure out.

matthewbauer:
Yeah I can help shepherding this RFC.

edolstra:
@aneeshusa @Ericson2314 @matthewbauer Are you interested in having a chat (e.g. on Hangouts or IRC) to discuss how to move forward with this IRC? Maybe some time next week?

matthewbauer:
@edolstra Yes, that would be okay with me.

aneeshusa:
Thanks for the ping. This week is a bit busy for me, but definitely happy to chat sometime next week - IRC sounds good.

edolstra:
@aneeshusa @Ericson2314 @matthewbauer  How about next week (April 15-19)? I'm relatively unconstrained wrt time.

Ericson2314:
https://www.when2meet.com/?7756595-quZ4l OK scheduling take 2!

Ericson2314:
@edolstra @matthewbauer @aneeshusa Let's....try again? Would something be better than www.when2meet.com to coordinate a time?

aneeshusa:
Unfortunately, looks like I don't have as much time to contribute to Nix/Nixpkgs as I thought :( I'm going to step down from this RFC and close this PR. However, I'm still very interested in this feature (I still run an updated version of original branch locally) and would love for a new champion with more time to pick this up! Process-wise, I imagine you would make a new PR from your own fork but with the same RFC number? Not sure.

My most recent thoughts here: Per [my previous comment](https://github.com/NixOS/rfcs/pull/13#issuecomment-459456166) I looked into conditionally turning on structured attrs, but it ended up causing infinite recursion errors. I took another look at the Nix manual and learned that structured attrs mode also writes out a `.attrs.sh` file that the existing bash builder should be able to use with no or hopefully a minimum of modifcation. So, what I imagine the next steps here would be:
- turn on structured attrs unconditionally in nixpkgs, using the `.attrs.sh` to avoid having to rewrite all of `stdenv`
- teach the cmake setup-hook to take advantage of the `.attrs.json` data, either with `jq` or shell-ing out to e.g. Python

edolstra:
Thanks @aneeshusa! Yeah, I think switching to structured attrs globally is the way to go, though obviously a rather big operation. But then we get ergonomic `cmakeFlags` almost for free.

Ericson2314:
@edolstra there's a number of changes that would be nice for a "stdenv 2.0", and I'm curious what you think as to how they should land (piecemeal?, one big leap?)

aneeshusa:
Due to Nix writing out`.attrs.sh`, I would hope that switching to structured attrs would be a small change that shouldn't require large changes to stdenv or be visible to consumers of stdenv - definitely something that I'd rather land incrementally than wait to bundle into a "stdenv 2.0". If this isn't possible for some reason, I'd want to know more about why not and figure out ways to make it possible to land incrementally.

PR #14, opened 2017-06-03
[RFC 0014] Improve import from derivation

Ericson2314:
Fix some issues with importing from derivations so it is nicer to use both manually and with CI. If this RFC is implemented, hydra.nixos.org should allow importing from derivations.

[Rendered](https://github.com/Ericson2314/nix-rfcs/blob/import-from-derivation/rfcs/0000-import-from-derivation.md)

### DIFF
--- a/rfcs/0000-import-from-derivation.md
+++ b/rfcs/0000-import-from-derivation.md
@@ -9,7 +9,8 @@ related-issues: (will contain links to implementation PRs)
 # Summary
 [summary]: #summary
 
-Fix some issues with import-from-derivation so it is nice to use and suitable for use in nixpkgs.
+Fix some issues with importing from derivations so it is nicer to use both manually and with CI.
+Once this RFC is implemented, hydra.nixos.org should allow importing from derivations.
 
 # Motivation
 [motivation]: #motivation
--- a/rfcs/0000-import-from-derivation.md
+++ b/rfcs/0000-import-from-derivation.md
@@ -2,7 +2,7 @@
 feature: import-from-derivation
 start-date: 2017-05-19
 author: John Ericson (@Ericson2314)
-co-authors: (find a buddy later to help our with the RFC)
+co-authors: Joe Hermaszewski (@expipiplus1)
 related-issues: (will contain links to implementation PRs)
 ---
 
--- a/rfcs/0000-import-from-derivation.md
+++ b/rfcs/0000-import-from-derivation.md
@@ -19,7 +19,7 @@ Sometimes the entire dependent graph cannot be known statically (before any buil
 Examples of this would be a package with its own nix-based build system—the dependency must be downloaded first,
 or leveraging a language specific build tool (Cabal, Cargo, etc) to generate a build plan.
 
-Import-from-derivation and recursive Nix are two ways to achieve this dynamism—the interleaving of planning the graph and building it.
+Import-from-derivation and recursive Nix are two ways to achieve this dynamism—the interleaving of planning the build graph and executing it.
 Implementation-wise, however, they are completely orthogonal so there's no reason not work on both.
 
 Currently there are some issues with import-from-derivation, however. Per @shlevy's read-only recursive nix RFC:
--- a/rfcs/0000-import-from-derivation.md
+++ b/rfcs/0000-import-from-derivation.md
@@ -1,0 +1,47 @@
+---
+feature: (fill me in with a unique ident, my_awesome_feature)
+start-date: (fill me in with today's date, YYYY-MM-DD)
+author: (name of the main author)
+co-authors: (find a buddy later to help our with the RFC)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+One paragraph explanation of the feature.
+
+# Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected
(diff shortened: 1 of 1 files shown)


zimbatm:
@Ericson2314 did you get my review by email? Let me know what I can do to help.

rfcs/0000-import-from-derivation.md
```diff
@@ -0,0 +1,119 @@
...
+
+The fundamental change is rather than evaluation unconditionally triggering execution, evaluation will suspend yielding the execution needed to be done so it can make further progress on resumption.
+This is probably easier to explain in terms of the implementation steps needed.
+
+## Black-hole unwinding
+
+There's currently a bug with `nix-repl` where if computation is interrupted (say with `SIGINT`), and the same computation is begun again, one will get an infinite recursion error despite there being no infinite recursion.
+This is because the thunks stay black-holed—the indication that they are currently evaluated—despite the computation being aborted.
+To fix this, we want aborted computations to instead unwind their stack, un-black-holing any thunks back the way they were before.
+Mathematically, black-hole unwinding ensures that evaluation is idempotent.
```
zimbatm: Just a thought: this seems useful on it's own. Maybe split it in it's own RFC, I suspect smaller RFCs are more likely to make progress. Or just implement it?

Ericson2314: Seeing that that's just a bug fix, I might just open an issue. That issue can be linked from any tracking the RFC as a whole if it is accepted.

Ericson2314: Opened https://github.com/NixOS/nix/issues/1407

rfcs/0000-import-from-derivation.md
```diff
@@ -0,0 +1,119 @@
...
+`nix-build --dry-run` is thus a synonym of `nix-build --dry-run=0`.
+`nix-build --dry-run=infinity` is the semantics of `--dry-run` today—perform as many rounds as needed to get the final static build plan.
+
+Because we pause evaluation each round, we can build the imported derivations "normally".
+That includes taking advantage of build remotes, which avoids the problem of import from derivation DOSing the hydra evaluator and or imported derivations requiring a different platform than the evaluation machine.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Overlap with recursive Nix.
```
zimbatm: Do you have a link to what recursive Nix is? Just so the reader can compare.

0xABAB:
The technical writing could be a lot better, but I get the gist of it. 

It appears that this is just an instance of "recursive make considered harmful", though. My longterm concern is that this will lead to a new workflow where people will start making use of the feature, which means that the system as a whole becomes dependent on more dynamic features while they could also be implemented with more static features (as it is now). 

In short, I would say this is a controversial feature. It's not like I can stop you, but I probably wouldn't add this feature if I had complete control over the project.

Ericson2314:
Yes, dynamism should be the escape the escape hatch of last resort. I hope the burden of writing concrete syntax with an external program is enough to dissuade anyone from using it more.

Ericson2314:
I should edit this RFC to also mention readFile. ReadFile of a derivation needs analogous improvements.

domenkozar:
So if I understand correctly, to summarize: 

1) Hydra would evaluate an expression until it hits IFD and halt there
2) it would build the IFD paths
3) start with 1) using the new built paths to continue the eval

Ericson2314:
@domenkozar do you mean " start 1) again..."? If so yes. It could also just let `nix-instantiate` do that automatically, but avoid keeping the eval process around it's probably best to do that by hand with `--dry-run`.

edolstra:
@Ericson2314 We're doing triage on the RFC backlog, are you still interested in pursuing this RFC?

The issue I see with this RFC is that it doesn't really work given Hydra's current design, specifically, the strict split between the evaluator and the queue runner. The evaluator is supposed to add rows to the `Builds` table, so it can't delay building IFDs - it has to produce a top-level derivation. Likewise the queue runner doesn't have access to the jobset inputs, so it can't resume evaluation.

This could be solved by having an integrated evaluator/queue runner, but then we don't need any partial evaluation tricks: we can just build IFDs as normal build steps during evaluation. (The only issue is that during evaluation there is no Build yet to account the BuildStep to, but that can be hacked around by having a temporary Build or whatever.)

The downside of integrating the evaluator and queue runner is that it reduces security a bit, since (for example) the jobset Nix expression could try to read SSH credentials of the queue runner during evaluation using `readFile`.

Ericson2314:
@edolstra Also, thank you for that description of the hydra status quo. I'll try to better separate the Nix itself and hydra concerns from this PR, but on the hydra side I'm thinking that top-level derivations must not be auto-generated, but their dependencies may be. My new idea already supported that on the Nix side, and from your description, that would be allow Hydra to keep its existing division of labor. [If we had to refactor hydra that much, I'd declare it "totalled", i.e. it would be better to just use a rewrite that would also fix a slew of other issues.]

edolstra:
In Nix, you can't have a .drv file without knowing *all* its dependencies. This is independent from any evaluation issues.

Ericson2314:
@edolstra yes this would be a recursive nix, thing, where drv1 would build drv2 (and drv2's deps). drv1 would be delayed evaluation of nix expressions, which we may or may not wish to special-case purely as an optimization.

PR #19, opened 2017-10-29
RFC for a maintainers file

moretea:


Profpatsch:
We introduced the [GitHub CODEOWNERS file](https://github.com/NixOS/nixpkgs/blob/master/.github/CODEOWNERS) a while back.

zimbatm:
I didn't know about the Github CODEOWNERS file. There is probably a documentation problem, what is the best place to put it?

Could the RFC be extended to include the places where the CODEOWNERS would be used?

I also found https://github.com/softprops/git-codeowners which looks really nice, let's get that packaged in.

zimbatm:
`nix-env -iA gitAndTools.git-codeowners` on master

zimbatm:
So apparently, CODEOWNERS only works if the people have push access to the repo. This is something we want to avoid in the long run. https://github.com/isaacs/github/issues/989#issuecomment-320475904

Ekleog:
Hmm... Just to bounce back on “only works if the people have push access to the repo. This is something we want to avoid in the long run.”

In my opinion having github able to add a backdoor into nixpkgs is an issue with the current setup (likely noone would notice an additional commit added by github).

In order to solve this issue, I see no other solution than signing all git commits (I wanted to write and submit an RFC somewhere down these lines, but will wait until https://github.com/NixOS/rfcs/pull/18 is solved, given a single RFC stuck in limbo is enough for me). This way each git commit, so each change of the nixpkgs repository, would have a single person asserting “this is not a backdoor that I know of,” and access could be revoked if it was maliciously used.

However, this scheme is pretty much incompatible with not giving people push access to the repo. Or do I misunderstand?

(BTW, a scheme where all commits are pgp-signed could also enforce that only certain OpenPGP keys are allowed to change some directories, once the scripts are setup to check for it on each pull and on hydra, so I guess that it could interoperate nicely with a MAINTAINERS or CODEOWNERS file?)

FRidh:
@Ekleog signing of commits is a different issue, please discuss that elsewhere.

@zimbatm why would we want to avoid that? The following is written in the header of our `CODEOWNERS` file:
```
# This file is used to describe who owns what in this repository. This file does not
# replace `meta.maintainers` but is instead used for other things than derivations
# and modules, like documentation, package sets, and other assets.
#
# For documentation on this file, see https://help.github.com/articles/about-codeowners/
# Mentioned users will get code review requests.
```
We have code-owners, and maintainers. The code-owners have merge-rights and own "the bigger picture" and maintainers (who may not have merge-rights) maintain individual expressions.

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,95 @@
...
+the packages themselves. Introducing a maintainers file enables us to mark
+maintainers for the remaining parts.
+
+# Motivation
+[motivation]: #motivation
+
+<!--  Why are we doing this? -->
+There is no explicit (machine-readable) description of who maintains what.
+
+Currently, a bot is used to ping people based on the heuristic of who authored
```
FRidh: That bot was replaced a while ago by the `CODEOWNERS` file.

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,95 @@
...
+
+Currently, nixpkgs does not have explicit maintainers for anything besides
+the packages themselves. Introducing a maintainers file enables us to mark
+maintainers for the remaining parts.
+
+# Motivation
+[motivation]: #motivation
+
+<!--  Why are we doing this? -->
+There is no explicit (machine-readable) description of who maintains what.
```
FRidh: `CODEOWNERS` file describes who owns portions.
The `maintainers` fields describe who maintain certain expressions.

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,95 @@
...
+For example, if someone took over maintenance of the NixOS module system, nbp 
+will still get pinged all the time, because he has created most of the files
+in there.
+
+Furthermore, there is only the rather black-and-white distinction between
+people who have commit access, and those that do not. A maintainers file
+might open further automation to delegate merge access.
+
+<!-- What use cases does it support? -->
+With a maintainers file, we can ping the correct maintainer(s) on a PR.
```
FRidh: We can already. By evaluating Nixpkgs we see which derivations are affected and can thus ping those maintainers. We just need  a bot for it.

shlevy:
Erm, why do we want to limit push access to the repo exactly? Liberal granting of the commit bit is a great policy IME.

zimbatm:
It's a bit early for the feedback, @moretea and me are still working on the RFC and the rationale will be made clearer when it will be submitted for wider review. I hope that all these questions will be answered by then.

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,126 @@
...
+reviewers in place for all files.
+This RFC is part of a larger body of work that is necessary to archieve this goal.
+To enable this, we must be able to map all nixpkgs files to maintainers.
+
+
+# Motivation
+[motivation]: #motivation
+<!--  Why are we doing this? -->
+Currently we have two mechanism in place that explictly describes who the owner is.
+It either is defined in the `metadata.maintainers` in a `stdenv.mkDerivation`, or when it matches a
```
globin: `meta.maintainers`

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,126 @@
...
+pattern in `CODEOWNERS` file.
+
+The metadata.maintainers only covers the packages, and does not used to determine reviewers
```
globin: see above

globin: does not used -> is not used

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,126 @@
...
+for packages on GitHub.
+
+The [`CODEOWNERS`](https://help.github.com/articles/about-codeowners/)
+is only used to help GitHub select a reviewer and is centrally managed in a single file.
+One of the problems is that this loses the data locality, unlike the `.gitignore` files, on which
+the format is heavily inspired.
+This makes further delegation of maintainer responsibilities harder.
+
+These two systems do not integrate together.
+The meta data in` pokgs/*` is used by Hydra packages break, and the CODEOWNERS is only used by
```
globin: pkgs -> pkgs

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Requirements
+The input to the maintainers script is a list of files the output will be a machine readable list
+of maintainers. It will support a `--why` option to print out which decisions points were
+encountered, in order to debug why someone is a maintainer of a file.
+
+This script will read in a MAINTAINERS files, with support for nested delegation like how the
+`.gitignore` files work.
+
+For files in `pkg/`, the script will try to use the metadata in the packages to find the maintainer.
```
globin: pkgs

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,126 @@
...
+
+TBD.
+
+There are two high level options to take:
+- Data format, such as toml, yaml, or the
+  [Linux kernel's maintainer file format](https://github.com/torvalds/linux/blob/master/MAINTAINERS)
+- Use Nix file + script to invoke this.
+
+## Implementation
+In order to facilitate integration in the GitHub PR + Hydra work by @globin and @gchristensen,
```
globin: This is the issue wrt the comment: https://github.com/mayflower/nixbot/issues/9

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,126 @@
...
+GitHub for reviews.
+
+The `CODEOWNERS` format is not extensible at all; there is no way to specify e.g. that some parts
+require one positive review of the maintainers and others might require more than one.
+
+Furthermore, GitHub is not suitable to handle the permission that the Nix community needs;
+currently people either have full commit access, or none at all.
+
+Therefore, we should implement a format and automation to handle both the data from the
+`metadata.maintainers` and some maintainers file.
```
globin: see above

moretea:
@globin yes, the intention is that pkgs/* is special cased in the script, as is described in the RFC.

edolstra:
Previous related discussion: https://github.com/NixOS/nixpkgs/issues/13602

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,129 @@
...
+## Requirements
+The input to the maintainers script is a list of files the output will be a machine readable list
+of maintainers. It will support a `--why` option to print out which decisions points were
+encountered, in order to debug why someone is a maintainer of a file.
+
+This script will read in a MAINTAINERS files, with support for nested delegation like how the
+`.gitignore` files work.
+
+For files in `pkgs/`, the script will try to use the metadata in the packages to find the maintainer.
+If this information cannot be retrieved, it falls back to the rules in the MAINTAINERS file.
```
edolstra: It should also use the meta.maintainers field of NixOS modules.

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,129 @@
...
+
+
+## Maintainers file format
+
+TBD.
+
+There are two high level options to take:
+- Data format, such as toml, yaml, or the
+  [Linux kernel's maintainer file format](https://github.com/torvalds/linux/blob/master/MAINTAINERS)
+- Use Nix file + script to invoke this.
```
edolstra: This seems like a good option, since then we can include lib/maintainers.nix from that file. And it can easily be converted to JSON for processing by other tools (`nix eval --json`). The format could be something like:
```
with import lib/maintainers.nix;

[
  { description = "Stdenv";
    paths = [ "pkgs/stdenv/*" ... ];
    maintainers = [ eelco ... ];
  }

  { description = "Haskell Infrastructure";
    paths = [ "pkgs/top-level/haskell-packages.nix"
              "pkgs/development/compilers/ghc/*"
              "pkgs/development/haskell-modules/lib.nix"
            ];
    maintainers = [ peti ... ];
  } 
]
```

rfcs/0019-maintainers.md
globin: (and tests)

edolstra:
@shley Whether we want to *require* approval from maintainers is a policy issue separate from having a maintainers file. The latter is useful in any case, because it enables the bot to automatically ping the maintainers who are responsible for the files changed by the PR.

I do think requiring approval for core subsystems or cross-cutting concerns is a good idea, since the situation where any of >80 committers can change (say) stdenv or lib is pretty scary.

moretea:
Relevant issue: https://github.com/NixOS/nixpkgs/issues/13602

edolstra:
I would like to go ahead with this, perhaps with some minor changes:

* Not sure if the `delegate` attribute is needed at the moment. We can also just import and concatenate other maintainers files (e.g. `// (import pkgs/development/ruby-modules/maintainers.nix)`).

* It's a bit confusing to have a `maintainers.nix` and a `lib/maintainers.nix`. Maybe the latter can be merged into the former? (It was always a bit strange to have data like maintainer names in `lib`...)

@moretea Are you still interested in implementing a bot to auto-assign PRs based on the maintainers file?

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,197 @@
...
+This script will read in a maintainers files, with support for nested delegation like what is
+possible with `.gitignore` files.
+
+For packages in `pkgs/`, modules in `nixos/modules` and tests in `nixos/tests/`,
+the script will try to use the metadata in the packages to find the maintainer.
+If this information cannot be retrieved, it falls back to the rules in the MAINTAINERS file.
+
+## Maintainers file format
+The maintainers file format will be a Nix expression.
+This allows easy integration and sharing of the data in `lib/maintainers.nix`,
```
zimbatm: has now moved to `maintainers/maintainer-list.nix`

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,197 @@
...
+<!--  Why are we doing this? -->
+Currently we have two mechanism in place that explicitly describes who the owner is.
+It either is defined in the `meta.maintainers` in a `stdenv.mkDerivation`, or when it matches a
+pattern in `CODEOWNERS` file.
+
+The `meta.maintainers` only covers the packages, and some NixOS modules and NixOS tests.
+It is not used to determine reviewers on GitHub.
+
+The [`CODEOWNERS`](https://help.github.com/articles/about-codeowners/)
+is only used to help GitHub select a reviewer and is centrally managed in a single file.
```
FRidh: - it is used to define ownership over scopes that are not or cannot be set with `meta.maintainers`. Examples are additional scripts and structure of and routines for a package set, and thus also library functions.
- automatically selecting a reviewer is a nice benefit. Because the `mention-bot` had become unreliable and discussion on `MAINTAINERS` file was stalled, the `CODEOWNERS` file was adopted.

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,197 @@
...
+One of the problems is that this loses the data locality, unlike the `.gitignore` files, on which
```
FRidh: How so? Because we have two places we describe maintainers? What about nested `.gitignore` files?

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,197 @@
...
+the format is heavily inspired.
+This makes further delegation of maintainer responsibilities harder.
+
+These two systems do not integrate together.
+The meta data in `pkgs/*` is used by Hydra packages break, and the `CODEOWNERS` is only used by
```
FRidh: break?

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,197 @@
...
+GitHub for reviews.
```
FRidh: And by individuals who simply want/need to know who owns something?

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,197 @@
...
+
+The `CODEOWNERS` format is not extensible at all; there is no way to specify e.g. that some parts
+require one positive review of the maintainers and others might require more than one.
+
+Furthermore, GitHub is not suitable to handle the permission that the Nix community needs;
```
FRidh: This is quite a different topic. I agree with it, but is authorization also within the scope of this RFC? If not, drop it.

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,197 @@
...
+# Detailed design
+[design]: #detailed-design
+
+<!-- This is the bulk of the RFC. Explain the design in enough detail for somebody
+familiar with the ecosystem to understand, and implement.  This should get
+into specifics and corner-cases, and include examples of how the feature is
+used. -->
+
+## Requirements
+The input to the maintainers script is a list of files the output will be a machine readable list
```
FRidh: and the output

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,197 @@
...
+of maintainers. It will support a `--why` option to print out which decisions points were
+encountered, in order to debug why someone is a maintainer of a file.
+
+This script will read in a maintainers files, with support for nested delegation like what is
```
FRidh: What is the motivation for nested files?

rfcs/0019-maintainers.md
```diff
@@ -0,0 +1,197 @@
...
+[
+  {
+    paths = [ "pkgs/develpoment/ruby-modules/bundix/**/*" ];
+    maintainers = [ zimbatm manveru ];
+  }
+]
+```
+
+## Evaluation of rules
+The first rule int the list that matches is accepted.
```
FRidh: int

PR #21, opened 2017-11-06
[RFC 0021] Pull Request Testing on Hydra

globin:
[Rendered](https://github.com/mayflower/nixos-rfcs/blob/5c1723577e86e265c7ca14bdb2f6c6d0c60eda5a/rfcs/0021-pr-testing.md)
cc @edolstra @fpletz @vcunat @zimbatm @grahamc @domenkozar @rbvermaa

### DIFF
--- a/rfcs/0021-pr-testing.md
+++ b/rfcs/0021-pr-testing.md
@@ -1,0 +1,85 @@
+---
+feature: pr-testing
+start-date: 2017-11-06
+author: @globin
+co-authors: (find a buddy later to help out with the RFC)
+related-issues: https://github.com/NixOS/hydra/pull/516 https://github.com/mayflower/nixborg
+---
+
+# Summary
+[summary]: #summary
+
+In order to properly test Pull Requests and reduce the maintainers' workload
+we need automation to build PRs on Hydra. This RFC specifies an architecture
+to achieve this and contains a reference implementation.
+
+# Motivation
+[motivation]: #motivation
+
+We currently have peaks of ~70 PRs a month at release time, this consumes an
+immense amount of maintainers' time to test the PRs.
+
+Another issue is the difficulty of testing mass-rebuilds. Only a few people
+have resources for themselves to test such computation intensive builds and
+Hydra access is limited to only a small amount of people.
+
+Due to having already built all packages prior to merging this would reduce
+the need for staging.
+
+# Detailed design
+[design]: #detailed-design
+
+The general idea is to have a service which receives webhooks from github,
+then requires some acknowledgement of a maintainer to build/test a PR and
+subsequently schedules a jobset on Hydra.
+
+## Details of the reference implementation at https://github.com/mayflower/nixborg
+(to be moved to the NixOS github organisation if accepted)
+
+The current implementation consists of three seperate services:
+ - nixborg: flask app to receive github hooks and fill a queue (currently redis, probably switching to rabbitmq in the future)
+ - nixborg-workers: workers consuming the queue, doing the git handling, commenting on github, HTTP requesting to nixbot-receiver
+ - nixbot-receiver: stdlib-dependency-only python script to be run on Hydra that calls out to hydra-update-jobset (PR: https://github.com/NixOS/hydra/pull/516)
+
+1. the main service `nixborg` listens for `issue_comment` hooks containing `@nixborg build`
+2. `nixborg` passes the hook on to the `nixborg-workers`
+3. a `nixborg-worker` rebases the PR on the base branch of the PR (mainly master, staging or release-XX.XX)
+4. the rebased branch gets pushed to a `pr-XXXXX` branch
+5. a `nixborg-worker` HTTP calls out to the `nixbot-receiver` on Hydra with all necessary information for a jobset; this message is HMAC authenticated
+6. `nixbot-receiver` passes on the information to create/update a jobset `pr-XXXXX` with the newly implemented `hydra-update-jobset` CLI utility
+
+Some small remarks on the architecture:
+ - the rebasing is done in order to:
+   - make sure Hydra builds a commit that has been reviewed by a maintainer
+   - ensure an older PR is tested based on the current state of the base branch
+ - `nixbot-receiver` is implemented as a stdlib-dependency-only python script not doing any logic to have a stable interface to Hydra that doesn't need to be updated
+    often, when iterating on nixborg and to not have a large footprint (cpu, memory, disk space)
+
+# Alternatives
+[alternatives]: #alternatives
+
+ - The Hydra jobset management could be achieved by using declarative jobsets, this does not work out currently as jobsets cannot be compared across projects
+ - [bors](https://bors.tech/) is the bot the Rust developers use to automate their testing with buildbot. It would have to be adapted to call out to Hydra
+   and currently AFAIK is not that easy to run yourself, when not using heroku. A lot of nixborg is inspired by bors.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+ - Which "release.nix" should be used to test:
+   - We probably want to test aarch64 and x86_64-darwin, not only x86_64-linux, as only a little number of maintainers/pull requesters have access to these architectures.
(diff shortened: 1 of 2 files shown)


zimbatm:
Sorry for bringing this up again, can you expand on why we can't re-use bors? It would be nice to re-use existing software if we can. Basically can Hydra:
* build a set of fixed branches every time some code is pushed to them?
* report the build status as a github build status on the commit that has been built?

globin:
Currently it can do neither, I also have too many difficulties writing perl to be able to implement that properly..

Also we definitely don't want the bot merging PRs right now, this might be an option in the future but currently it should only give us an opportunity to test PRs more easily.

Profpatsch:
> Sorry for bringing this up again, can you expand on why we can't re-use bors? It would be nice to re-use existing software if we can.

Would be awesome to reuse the work of the rust team. Especially considering not very many people know (or want to write) perl these days, but lots of people want to learn Rust anyway.

moretea:
@zimbatm @Profpatsch AFAIK, the bors currently used by Cargo (package manager for Rust) is  [bors-ng](https://github.com/bors-ng/bors-ng) which is implemented in Elixir.

It appears that they don't use it yet for Rust itself (see bottom comments in [this thread](https://internals.rust-lang.org/t/proposal-move-to-bors-ng/5334))

globin:
Can we please keep discussing PR testing on hydra for now which is a small piece of work compared to changing our whole CI system.

Ericson2314:
I think we all agree that **eventually**, only CI should be changing master. Is it a good idea to tackle this in this RFC, or best to leave it for later? Relatedly, I'm a bit unsure on rebasing or merging PRs.

7c6f434c:
@Ericson2314 

This «eventually» requires more Hydra capacity than there seems to be right now. And more trust in the handling of transient failures than we currently have (and a solution probably requires that the transient failures, and not the capacity, are the bottleneck for large rebuilds). Also a clear idea of what breakage is acceptable at what points (this obviously depends on the previous points, because we need confidence that we find out about failures, and find out quickly). This RFC is a good step to try now (and see what needs fixing), and only-CI-moves-master (and even only-CI-merges-PRs) seems significantly out of scope.

globin:
I generally agree with most of the points mentioned but I wanted to keep this RFC rather small in scope to not reach a deadlock in discussion and planning while not moving forward :)

Ericson2314:
@7c6f434c the incremental step would be having CI able to push to master (if it's a fast-forward to the commit it just built successfully), without requiring CI to push to master. The extra step of going back to merge a PR after CI succeed is generally a waste of human time, too.

rfcs/0021-pr-testing.md
```diff
@@ -0,0 +1,85 @@
...
+Hydra access is limited to only a small amount of people.
+
+Due to having already built all packages prior to merging this would reduce
+the need for staging.
+
+# Detailed design
+[design]: #detailed-design
+
+The general idea is to have a service which receives webhooks from github,
+then requires some acknowledgement of a maintainer to build/test a PR and
```
7c6f434c: Re: «a maintainer» — maybe add something about supporting a flexible definition of a maintainer (any NixOS project member/Nixpkgs committer or the package maintainer who doesn't yet have access) to «Future work»? Not in the main body, because this discussion should not delay checking if Hydra can build all PRs.

rfcs/0021-pr-testing.md
```diff
@@ -0,0 +1,85 @@
...
+ - [bors](https://bors.tech/) is the bot the Rust developers use to automate their testing with buildbot. It would have to be adapted to call out to Hydra
+   and currently AFAIK is not that easy to run yourself, when not using heroku. A lot of nixborg is inspired by bors.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+ - Which "release.nix" should be used to test:
+   - We probably want to test aarch64 and x86_64-darwin, not only x86_64-linux, as only a little number of maintainers/pull requesters have access to these architectures.
+   - We probably want to also test NixOS module tests, especially for PRs touching these
+ - Can Hydra handle the additional workload by this, we probably have to start with a small number of PRs being tested to evaluate this.
```
7c6f434c: That, and maybe perform initial tests only `x86_64-linux`? How the Hydra capacity (or maybe scaling costs, if cloud builds are used regularly — I do not know) compare across targets?

rfcs/0021-pr-testing.md
moretea: See #19 also.

AFAIK the current implementation of that bot just has a hardcoded list of people who should be allowed access.

I think that people from the maintainers file + anyone who currently has commit access should be allowed to instruct the bot via @-mentions.

rfcs/0021-pr-testing.md
```diff
@@ -0,0 +1,85 @@
...
+1. the main service `nixborg` listens for `issue_comment` hooks containing `@nixborg build`
+2. `nixborg` passes the hook on to the `nixborg-workers`
+3. a `nixborg-worker` rebases the PR on the base branch of the PR (mainly master, staging or release-XX.XX)
+4. the rebased branch gets pushed to a `pr-XXXXX` branch
+5. a `nixborg-worker` HTTP calls out to the `nixbot-receiver` on Hydra with all necessary information for a jobset; this message is HMAC authenticated
+6. `nixbot-receiver` passes on the information to create/update a jobset `pr-XXXXX` with the newly implemented `hydra-update-jobset` CLI utility
+
+Some small remarks on the architecture:
+ - the rebasing is done in order to:
+   - make sure Hydra builds a commit that has been reviewed by a maintainer
```
grahamc: We should note that the bot compares the time the nixborg comment was made to when the last push was, to ensure there isn't a race condition between the comment and a malicious push.

rfcs/0021-pr-testing.md
globin: Currently "a maintainer" refers to a contributor with commit access, "member" in GitHub terminology, I'll clarify this.

rfcs/0021-pr-testing.md
globin: Will do

globin:
Added some minor clarifications, after the comments above and am happy that @grahamc is co-authoring this.

@edolstra do you have some comments on this?

Ericson2314:
It just occurred to me that in the failing case I wish the original branch was being built, so I'd have some cached binaries to help debug. But I suppose the PR could be updated to the nixborg's rebase itself.

Also, does it rebase on master or non-master when the target branch is non-master (e.g. staging)?

globin:
Says so in the RFC, too:
```
 3. a `nixborg-worker` rebases the PR on the base branch of the PR (mainly master, staging or release-XX.XX) 
```

zimbatm:
How much work left is required to make this work?

The sooner we have *any* sort of automation the better, and it's hard to predict the problems that might come up without deploying and running the code. So instead of mulling over the RFC too much, I'm in favor of getting something running and have a trial period, let's say 3 months. After that period we will review the state of things and decide if we want to keep it or not.

During that period, the implementer need to have access to the machines where the code is deployed so that they can fix and iterate.

The only big concern would be to overload Hydra. If that happens we either need a separate Hydra deployment or a way to pull the plug until a solution can be found.

dtzWill:
This may be incorrect or premature, but... hasn't ofborg filled this gap more or less?

7c6f434c:
Currently: not enough to do full tests as opposed to small samples, not
enough to test Chromium even on its own, not prebuilding the binaries to
be ready in the binary cache right after the merge.

But maybe the first two points are also arguments against overloading
Hydra with that.

zimbatm:
It's coming. The next step is going to be able to share the binary cache between ofborg and hydra so that hydra can basically become a noop.

edolstra:
In that case, we should just get rid of Hydra?

7c6f434c:
@zimbatm not yet close in power to what is needed 

@edolstra it's like a blockchain, if in the future Borg can perform a 51% buildrate attack against Hydra…

edolstra:
Closing after discussion in the RFC SC meeting: this is obsoleted by ofborg.

PR #23, opened 2018-02-20
[RFC 0023] Support musl libc

dtzWill:
WIP, but comments and discussion welcome throughout the process :).

[RFC 23 - support musl libc (rendered)](https://github.com/dtzWill/rfcs/blob/feature/musl-rfc/rfcs/0023-musl-libc.md)


shlevy:
:+1: from me.

7c6f434c:
Given that the only hardware platform that >90% have with root access is `x86_64`, and that we don't have a free software Darwin image that can be booted in a VM, doing portability fixes on `x86_64-linux` does sound appealing.

And with the current amount of effort towards cross-compilation, don't we have any cross-targets in use  by Nixpkgs users where musl is the preferred libc?

edolstra:
IMHO it's best to keep musl support in a branch or overlay so that the effort of maintaining musl-specific patches is not inflicted on package maintainers who don't care about musl.

(The same applies to RISC-V support BTW, which really should also be the subject of a RFC.)

coretemp:
I am in favor of a design where everything is in one repo where a broken `musl` doesn't impact `glibc` package users. I am not in favor of a branch based approach or overlay. `glibc` should only have a temporary special status in this regard, such that at some point we might even switch to `musl` as the default library.

zimbatm:
Overlays work best when the package set is different. In the case of musl it seems like it would be hard to reach into all the packages with overrides. Nix is quite composable but some times it's not possible to pry open a derivation enough to replace the inner parts (like when a function is hiding the scope).

How about accepting the musl patches but not making it a CI failure? It's not official but it's already the case that we have supported platforms and unsupported ones in nixpkgs.

zimbatm:
I mean as part of an unsupported platform we would set the expectations: it's okay to break the musl patches as part of the maintenance cycle. The musl team can have their own build and test infrastructure to sweep after breaking changes.

Ericson2314:
I'm super for it. And I'd like some cross platform to in fact be required by CI too. I think it's good that different platforms make packaging harder, it keeps us and upstream honest. Lastly, Musl, with it's minimal set of features, is a great test because if a package is built with musl, it can probably be built with anything.

Just to reiterate, the key to writing portable software is parametricity: don't add endless `#ifdef`s or conditionals in Nix, but abstract over platform differences as far upstream as possible so downstream need not care. With this principle, more platforms is not linearly more work.

7c6f434c:
@edolstra are you also against the cross-compilation support in the main repository?

Making a call on macOS failures is already needed, and with musl I can at least reproduce the problem locally and see if it is easy to fix; so it is just the question about the level of expected stability.

zimbatm:
@Ericson2314 it's like every platform. First it should go through a trial phase where compilation success is not required. Once it has been demonstrated as a viable target it can be promoted to the supported platforms.

The question is, who is responsible for deciding which platforms are supported or not :)

Ericson2314:
@zimbatm Agreed, I think a `release.nix` + `release-cross.nix` merge should be done by 18.03 with *no new* required-to-pass derivations, as exactly that second step.

zimbatm:
Alright let's do this. Any objections @edolstra ?

Ericson2314:
I should add that @dtzWill was a bit more to the point in the original RFC than me pontificating on parametricity above. I think prior portability efforts were a bit more "conditional-explosion-y" than the current ones: the patches we're capable of landing today really do *improve* the software in code quality too. Given the necessary ugliness of earlier efforts, I understand and sympathize @edolstra's reservations. Given the elegance we can get away with now, I don't think those concerns still apply.

grahamc:
I've noticed the RFC as-is doesn't go much in to how maintenance will work:

 - how we test and verify patches work
 - how well will musl be supported in the end
 - do most packages work out of the box with musl?
 - what impact does this have on Hydra?
 - what impact does this have on ofborg?
 - how do we detect if a patch is breaking musl support?

Another level is, this adds a bit more complexity to the cross / arch support. Should we be looking for a more grand unifying design to clean this up?

rfcs/0023-musl-libc.md
```diff
@@ -0,0 +1,158 @@
+---
+feature: musl-libc
+start-date: 2018-02-19
+author: Will Dietz
+co-authors: (find a buddy later to help our with the RFC)
```
grahamc: I'm concerned that a whole new libc only has one "sponsor" here.

shlevy: I'd happily cosponsor.

grahamc: The more the merrier, I think! This writes down who can be pinged for future musl problems.

AndersonTorres: Would it be a hard task?

Ericson2314:
@graham I think the idea is that initially, nothing is supported, but merely allowed to exist.

> how we test and verify patches work

The only way to is to build, but the package maintainer isn't obligated to do this.

> how well will musl be supported in the end

Depends how the experiment goes

> do most packages work out of the box with musl?

Yes. Packages that don't will probably be eventually broken on glibc as it becomes more standards-enforcing. (I didn't know this until I read the RFC. This is great news as too much of GNU doesn't require opt-in to extra GNU-specific features.)

> What impact does this have on Hydra / of borg?

Nothing. Well almost nothing; there are a few odd jobs in release-cross.nix.

> how do we detect if a patch is breaking musl support?

We don't need to for an unsupported experiment.

grahamc:
> the idea is that initially, nothing is supported, but merely allowed to exist.
> Depends how the experiment goes

I don't love the idea of having unsupported, unexercised code sitting in Nixpkgs. What is the point of having it _in_ nixpkgs, if it is allowed to be broken by people who don't even know it is being broken? At that level of "unsupported", there is very little reason to have it in the repo. Nixpkgs receives almost 500 commits a week, and a lot of rot can set in over that time.

I think if we're going to add major new support, part of the experiment should be at least diagnostics-level checking of the code per PR. 

If you disagree with this, please pretend we're a few months down the line and musl is a resounding success and I re-asked all those questions.

rfcs/0023-musl-libc.md
```diff
@@ -0,0 +1,158 @@
...
+* [issues matching "musl", newest first](https://github.com/NixOS/nixpkgs/search?o=desc&q=musl&s=created&type=Issues&utf8=%E2%9C%93)
+* [2015 libc discussion](https://github.com/NixOS/nixpkgs/issues/6221#issuecomment-116754223)
+
+# Detailed design
+[design]: #detailed-design
+
+I believe the linked implementation from current nixpkgs master is perhaps
+the best embodiment of "Detailed Design"?
+
+Linky: [bd7d5b365799a145717d31122ebfd24b51fea117](https://github.com/NixOS/nixpkgs/commit/bd7d5b365799a145717d31122ebfd24b51fea117)
```
grahamc: I suspect this link is wrong :)

rfcs/0023-musl-libc.md
```diff
@@ -1,8 +1,8 @@
 ---
 feature: musl-libc
 start-date: 2018-02-19
-author: Will Dietz
-co-authors: (find a buddy later to help our with the RFC)
+author: Will Dietz (@dtzWill)
+co-authors: Shea Levy (@shlevy)
```
grahamc: w00t

shlevy:
A new libc or port is simply not something that can realistically live out of tree, overlays etc. aren't expressive enough. Saying that we need a full support story before we can start work proving viability, attracting maintainers, and building userbase is effectively saying "go fork nixpkgs, this is never going to happen upstream."

Or if you disagree, how exactly do you propose we go about this kind of thing such that we have a realistic chance of ever getting it merged, given the above constraints?

shlevy:
The point of having it in nixpkgs is so that people can use it, and iterate on it. People undertaking these efforts understand there is burden on them to do whatever form of CI they need on their own resources until there is enough community support for more. If things bitrot, that's what `git rm` is for.

shlevy:
I hesitate to post this because I think it's a worse solution than allowing it upstream, but: if the ultimate answer here is no (and, setting this particular RFC aside, we don't seem to have any way to *get* to an ultimate answer in these...), then I'll (on my own fork) start `nixpkgs-next` à la `linux-next`, where all sorts of experiments are regularly merged on top of master and it's expected that the individual experiment maintainer makes sure things don't break for everyone else (enforced by being removed from the repo list until fixes are in) and users who want to take advantage can use a common place that's kept up-to-date with master.

To reiterate, this is a *worse* solution than finding a way to let this kind of work happen upstream. We're not the kernel community, and the last thing we need is further impaired discoverability, further divergence in possible user configurations, and more third-party infrastructure. But I'm tired of ideas being shot down because of potential maintenance burden without any viable alternative existing, so there's a viable alternative.

shlevy:
By the way, in retrospect the move to be much more compiler generic to the point where we can use clang for darwin was clearly a huge win, even for non-darwin users. Our wrappers are cleaner, our metadata makes more sense, etc. This seems similar to me.

Ericson2314:
@Shlevy hit all my major points first, thanks! I'll add that it sounds like, @grahamc, you are basically saying we should ofborg this too? I absolutely agree, and in practice that's so cheep and easy it's hardly a blocker. But I also stand by the principle that *low-impact* experiments should land without friction.

Ericson2314:
(@ofborg-ing just eval checks is the cheap and easy victory, to be clear.)

edolstra:
@shlevy Having a -next branch with experiments sounds good to me. I don't really understand why such experiments must be merged into master *right away*. Our history in cleaning up old experimental cruft is not good. Also, we don't mark actually mark things as "experimental" in Nixpkgs. E.g., is the RISC-V support experimental? If we ever decide that adding support for what is essentially a vaporware architecture is not a good idea, will it be removed? Who gets to make that decision? Will the burden of proof be on the people who want to keep it or the people who want to remove it?

> But I'm tired of ideas being shot down because of potential maintenance burden without any viable alternative existing

In the case of musl, there obviously is an alternative: glibc.

BTW, it's possible that musl support is pretty low impact, in which case I'm not super against it. But that's hard to judge at this point and also depends on the end goals. (The RFC does not address this, only stating that we should "add support for musl".) For example, is it to make a substantial part of Nixpkgs (or NixOS?) buildable with musl? If so, how many packages will need musl-specific patches? Doing musl support on a branch for a few months would make it much clearer what the impact is.

Another issue: adding musl support fractures the Nixpkgs user/development community: some people will run musl-based packages and some will run glibc-based packages. As a result all of Nixpkgs/NixOS will end up being less tested. it doubles the test matrix on Linux, after all.

grahamc:
I'm totally _pro_ this RFC. If the answers to most of those questions are "I don't know", that is fine with me, but I would really like to hear some speculation at least about what you _hope_ the answers to the questions are. 

I think the questions are valid, but thus far I feel the only responses I'm getting are that the questions aren't valid.

Take note dtz's response to me asking these questions was:

> I appreciate your comment! I'm surprised my quick draft got this far without needing to be refined hehe.  You're asking the right questions, and someone needs to :).   

So I guess I'm still holding out for @dtzWill's answers. Meanwhile, I'm disappointed that my questions were misinterpreted as having a negative subtext. I'll try harder in the future to not come off negatively when I'm just asking questions.

grahamc:
A follow-up question: what use cases might this support that we don't adequately support now?

shlevy:
@grahamc FWIW, my response was not to your questions, which are good ones that I'm waiting for @dtzWill before weighing in with my thoughts. They were a response to @edolstra and https://github.com/NixOS/rfcs/pull/23#issuecomment-367190629.

shlevy:
@edolstra I mean an alternative to proving viability and value, not an alternative to the feature in question.

shlevy:
This is getting pretty far out of scope of this RFC at this point, but since I'm not sure of another channel that's more appropriate and likely to get a satisfactory response, I have two general questions that I don't have a good answer to but seem fundamental to resolving this:

1. Under what circumstances, exactly, may we declare an RFC accepted or rejected? The status quo seems to be "consensus of those who have opinions and are fairly involved in the community", which effectively means most interesting (good or bad) RFCs will simply stagnate.
2. Addressed to whatever person, group, or process represents the answer to question 1: There are certain fundamental changes to our package set that require some amount of coverage from multiple people/use cases before they can be fully justified or rejected, and also can't realistically live out of tree due to limits in overriding. Is there any situation where such changes would be allowed to have a trial run on master to gain visibility, pick up users, and reduce the burden on the developer working on the feature, at the cost of possible bitrot?

I would really like a clear definitive answer to these questions, but I have no idea how to get to a point where I have them.

vcunat:
1.  This mainly depends on how _strong_ consensus we aim for, and it's an interesting question.  For example, changes concerning Nix itself would IMO typically need stronger consensus than those about NixPkgs.
In IETF (where I mainly follow [dnsop](https://tools.ietf.org/wg/dnsop/)) any drafts that don't "catch enough positive attention" just expire after a few months and they don't become RFCs, but the strength of the consensus and directing the discussion probably depends on the particular people chairing the WG.

edolstra:
@shlevy Well, this RFC has been open for just two days and there are several requests for clarification (e.g. about the scope of musl support) so it's a bit early to start a meta-discussion about process. However the answer to question 1 is in RFC 1: "After a number of rounds of review the discussion should settle and a general consensus should emerge. This bit is left intentionally vague and should be refined in the future. We don't have a technical commitee so controversial changes will be rejected by default."

vcunat:
I think this RFC's discussion might be dealing with two issues at once:
(A)  rules, maintenance responsibilities, etc. for "second-tier platforms".
(B)  having musl for Linuxes as a "second-tier platform".


##### My views:
**(A)**  Note that we de-facto have some second-tier platforms already, and even more discussion around them.  For example, some cross-compiled platforms are getting nontrivial attention lately.  We're now even [promoting `aarch64-linux`](https://github.com/NixOS/nixpkgs/issues/33745) (non-cross) to officially/better supported status, i.e. bad failures there will block some channel, I expect.  There was also a related discussion for [`i686-linux`](https://github.com/NixOS/nixpkgs/issues/31089) which was demoted, and proposals to try building everything with llvm, etc.

IMHO NixPkgs is getting very unique in the _ease_ of unified support for various targets (contrary to sprinkling lots of conditionals everywhere).  I attribute this mostly to having a language that can create abstractions much better, and community that is able to utilize that.  (Ever read a `*.spec` file for a complex package?)  My personal hunch is to allow pushing in this direction to exploit Nix advantages, but we certainly should address the concerns and limit the impact on the majority somehow (expressed mainly by Eelco for now).

Maintenance responsibilities are often an interesting point, but it seems there is consensus that second-tier platforms would be allowed to be untested/broken.  It actually seems a significant property of being second-tier.  This might lead close to Shea's suggestion &ndash; using a branch that only gets explicit updates from master (or better, from stable branch), _even if_ the support is in master already. (Or a separate channel, but that's almost the same thing.)

**(B)**  If there should be another libc in nixpkgs, I don't know a better choice, from what I've heard over the years.  I suppose your primary use cases are some resource-constrained environments (cross-compilation to embedded targets?)

shlevy:
I think the history of this repo, our mailing list, IRC, etc. shows pretty clearly that consensus does not in general emerge. But it's helpful to know that's the official policy.

edolstra:
It does say "should be refined in the future".

7c6f434c:
@edolstra I think without a meta-discussion the default assumption — given your general opposition to the goal of the RFC and not some specific details — is that it will not get merged in next, say, three months. The meta-question whether we are discussing medium-term goals or the current state and immediate goals seems significant for a productive discussion, so it might not be too early for a meta-discussion.

shlevy:
@edolstra By whom? How?

edolstra:
@7c6f434c @shlevy The meta-discussion is not specific to this RFC so it's best to have it in a separate issue.

@7c6f434c I'm not (necessarily) opposed to the goal of this RFC. The main issues are 1) what the goals actually are; and 2) whether this should be done on master right away.

vcunat:
BTW, I believe that larger pieces of future work need a "preliminary" consensus of this kind:
> If/when the work meets this list of (fuzzy) conditions, we shall (most likely) accept it to nixpkgs master.

That way the (prospective) implementors can have more confidence about what would be accepted **before** doing most of the work.

Ericson2314:
I think #18 is a good venue for 1, so I'll switch things back to 2.

Yes some experiments absolutely must be done in tree. My cross refactors (sort of making a distinction from musl and riscv which are cherries on top :)) have basically taken an entire year, and still much of that time was spent fixing things after rebase from earlier branches. I'm pleased with the outcome, and yet it still took longer than I'd have liked---and significant parts of it were done on work time when budgeting really counts. There's just no way it could have all been done out of tree.

@grahamc 

> What use cases might this support that we don't adequately support now?

Tiny statically-linked binaries (AFIAK glibc has trouble being statically linked). This is a good capability to have in general. But specific benefits are with non-nixos deployments. This will allow us to compete with Alpine Linux for Docker images on all fronts, which is good for community awareness, and also allow easy copying of relocatable binaries to non-nixos without needing to install Nix for closure copying. Both are good first steps for people/organizations that are nervous about Nix. I've also seen a situation where there was no root on the prod machine, so installing Nix (sans proot tricks)

> If so, how long does this experiment run?

I think it might help to give my ideal-case experiment pipeline.

1. Experiment can land in master. More ambitious PRs get more review, but there's no special hurdles.

2. The experiment is going well; specifically I deem it so if
   - Only a small portion of packages need tweaks (sub-linear work as a function of new succeeding builds)
   - the patches are generally clean: fixing importable code rather than slapping on new conditional cases.

   Then we add eval checking (presumably with ofborg). A few jobs on hydra, e.g. just building a stdenv and a handful of packages, are also a good idea

3. The experiment continues go to well, with tons of packages successfully building. Some sponsor maintains a binary cache, demonstrating this. The experiment should have more visibility however, so the sponsor instead commits some resources so `hydra.nixos.org` does the builds instead. This would be moving "tear 3" to "tear 2".

4. The platform is quite popular, some small set of breakages (like the ones that were first on the official hydra) now block channel success (probably `nixpkgs-unstable`).

5. The platform is a smashing success, and multiple parties beyond the experiment sponsor donate resources. The platform becomes tear 1 and blocks channel success with the same rough number of builds are our other tear 1 platforms.

A great example of an experiment that (to my knowledge) actually *hasn't* made it past step 1 is https://github.com/NixOS/nixpkgs/pull/26021, a cleanup of the Cygwin stuff. Even if everything just bit-rots again, it had nice diff-stats and I saw no reason to block it. Notice that the diff is from deleting patches from an earlier incarnation of the experiment. Had the experiment needed the same amount of patching as the previous attempt, I might have indeed been more reserved about merging it, but hey in 2017 even Cygwin needs fewer hacks! Conversely I could see somebody pushing this to step 2, but I think had this PR been blocked on that it might not have even landed.

Ericson2314:
Note that while RISC-V and musl are both brand new examples of cross (and native) compilation, I think that after a year of work cross compilation as whole is ready for step 2 under this plan. To that end, I am working on cleaning up `release-cross.nix` so that it can be integrated into `release.nix` for easier ofborg integration by 18.03. But we need some example platform for cross compilation, and I think these two are great candidates; better candidates than the older ones in fact.

rfcs/0023-musl-libc.md
dtzWill: I think it's right, it's just revision of master at time of writing.

Ericson2314: Maybe a diff or merge commit would be better?

Gabriel439:
Regarding the meta-discussion: I'm in favor of trunk-based development rather than long-lived branches.

In my experience long-lived feature branches lead to deferred integrations and merge conflicts which waste development resources or lead to the feature being abandoned.  If the feature ends up being a mistake then you can always just remove it from the trunk and no damage is done.

The other issue is that merging in a giant feature branch is difficult to roll back or `git bisect` if things go wrong.  Iterating in small self-contained changes is much less disruptive.

shlevy:
@Gabriel439 Can you bring that comment to #18?

dtzWill:
(...WIP still...)

7c6f434c:
I would say that separating into an overlay would come with a requirement to actually expand the nixpkgs support for overrides so that this using an overlay for `musl` becomes at least superficially plausible.

AndersonTorres:
Sorry by the off-topic, but, @shlevy what do you think about a separate RFC on that `-next`thing?

shlevy:
@AndersonTorres If that's the path this ends up going I'd much rather just create it on my own github and see how things go. People can use it or not :smile:

7c6f434c:
@shlevy are you considering to run it as a Cthulhu merge of branches or as a single integrated repository with some things going directly into it?

dtzWill:
> @shlevy are you considering to run it as a Cthulhu merge of branches or as a single integrated repository with some things going directly into it?

1) I'm not sure what a "Cthulhu merge" means, but judging by the name that sounds like the only way to go :wink: 
2) *ahem* please move conversation about a hypothetical tangentially-related repository's update policies elsewhere :innocent:.... and tell me where that is so I can subscribe :+1:.  Let's try to stay on topic please :).  And as @shlevy says "if that's the path this ends up going" -- which I'd like to amend by putting an emphasis on the "if".

Ericson2314:
Just going to say that I'm *really* hoping that if we'll accept this RFC, we'll do so by 18.03.

zimbatm:
I can see three main uses of musl: embedded (NixWRT?), smaller docker images, publishing static binaries (for non-nix targets). All of these seem like good things to have to increase Nix's popularity.

If there are 3+ people that pledge to create a self-managed musl team, I think that should be enough to accept the RFC.

shlevy:
@zimbatm Agreed on both counts. I'm willing to be part of that team in whatever capacity @dtzWill thinks makes sense.

domenkozar:
You can also count on IOHK helping maintain musl (as a consumer of the work).

About the integration: I'd start with one codebase (nixpkgs master) and decide otherwise if it's too much conditionalization once the prototype is ready for review.

shlevy:
OK, I think this is a go then. @dtzWill any modifications before we merge?

Ericson2314:
I'd say double check with @grahamc that his questions are addressed. That should be quick back and forth.

grahamc:
1. `"musl team" is reponsible, to be formed` please form that as part of this RFC. It sounds like you have the team here, it just needs to be written down. We could create a proper GitHub team handle that can be pinged, too.
2. `These changes usually can be safely applied on all platforms (although sometimes they are not for rebuild reasons)` I'd be inclined to just do it anyway, especially if `these changes are largely fixes improving compliance and correctness resulting in higher-quality programs"` holds true. Reducing the confusion and variability and chances some other patch only breaks on musl. Essentially, reducing the cyclomatic complexity of nixpkgs.
3. Please open an issue on https://github.com/nixos/ofborg with a details description (or even, a PR!) of how to add it to the evaluation checker. Here is probably the best place to look on how the eval checks are done cross-platform: https://github.com/NixOS/ofborg/blob/released/ofborg/src/outpaths.nix

Following these, as far as I'm concerned,

7c6f434c:
I would say that if a musl patch is changing just the assumption «Linux = Android xor glibc», it could be OK to delay its application on Darwin to reduce the number of rebuilds, but eventually applying it could indeed be a good idea.

telent:
As a non-contributor, I don't know if my opinion carries any weight at all, but here it is anyway: 

* I don't have a strong opinion on process or responsibilities, but if it can be made to work then as NixWRT wrangler I would find it absolutely marvellous: switching my (admittedly small) package set from glibc to musl was accomplished in a single evening and the only change needed in nixpkgs was actually a cross-compilation fix not a musl fix.  It opens up so many opportunities for using the Nix package collection in low-resource/embedded scenarios.

* Despite that, I would not be in favour of this if I thought that it was going to be a gateway to "combinatorial explosion hell" - adding ugly conditional fixes to packages to get them to build with a feature-impoverished or buggy C library.  But by all accounts it's not that and could even be looked on as a form of "linter" - improving the compatibility and standards-compliance of the code in the package collection.

tl;dr I can't speak to whether this approach to supporting musl-in-nixpkgs is the right one, but I can say I'd definitely be a user of musl-in-nixpkgs and so the need addressed by this RFC is (for me, anyway) real.

dtzWill:
Alright, I suppose it's ready? :innocent: 

@grahamc I liked the team idea so I put it in the RFC, hope that's alright :+1:.

rasendubi:
I'm interested in being the part of the musl team, too.

copumpkin:
> I would not be in favour of this if I thought that it was going to be a gateway to "combinatorial explosion hell" - adding ugly conditional fixes to packages to get them to build with a feature-impoverished or buggy C library. But by all accounts it's not that and could even be looked on as a form of "linter" - improving the compatibility and standards-compliance of the code in the package collection.

Hear, hear!

dtzWill:
(... I just switched my NixOS profile to use musl-built Nix 2 :grin: ...)

vcunat:
> Just going to say that I'm really hoping that if we'll accept this RFC, we'll do so by 18.03.

@Ericson2314: do you mean that you want the musl-support patches to keep getting cherry-picked to the stable 18.03 branch already?

Ericson2314:
@vcunat I guess I didn't think that far; correct me if I'm wrong @dtzWill, but I thought stream of musl fixes and slowed to a trickle :).

All I was thinking is if we're going to go through with this, it would be a cool thing to publicize (somewhat, is it is technically an experiment). If we end up going ahead but only decide after after 18.03, we miss a good window to give the experiment more visibility.

zimbatm:
Until we have a "nixpkgs team" to make a decision on this, I propose to delegate this decision to the release team: @vcunat and @fpletz

edolstra:
Merged this because it reflects the current situation.

nh2:
It seems the 1.1.19 -> 1.1.20 upgrade broke GCC for me: https://github.com/NixOS/nixpkgs/pull/46064#issuecomment-420126446

PR #26, opened 2018-03-05
staging workflow

vcunat:
Define a new workflow for the `staging` branch that can better accomodate the current and future influx of changes in order to deliver mass-rebuilds faster to master. As part of this new workflow an additional branch, `staging-next`, shall be introduced.

_More details in the RFC proposal itself now._


shlevy:
@vcunat Before going into the specifics of how we can change things on the workflow side, which I agree are probably needed regardless, I do want to point out that there seems to be an issue with hydra that exacerbates this. For at least two 5+-day periods this month, there was a case where some build was blocked on "sending inputs" for many days and that for some reason resulted in the queue not being pulled from properly. It's possible with that issue fixed we won't have to worry about this *quite* yet.

vcunat:
Yes, I know this.  If one is unlucky and a "mass-rebuild step" gets stuck this way...

shlevy:
So it seems like right now there are a bunch of different things we use staging for:

1. Validate that package X still builds with my change to package X itself
2. Validate that the problem motivating my change is fixed by my change (not applicable if this is just "bump the version because there's a new one")
3. Validate that my change satisfies basic sanity checks for my system
4. Validate that my change satisfies basic sanity checks for all systems
5. Validate my change satisfies a full suite of checks for all systems when integrated with the latest, and make binaries available.

This is over-engineered, but just to share the thought: I think in principle we could achieve guaranteed* eventual merging of correct code with a series of queues, possibly prioritized. At each stage, jobs are processed in the order they are evaluated for that stage. Each jobset only allowed one evaluation in the given queue at a time, if it needs a new commit to be tested it goes back to the back of the line at that queue. A jobset at stage N+1 must have as its base a commit that succeeded at stage N. If any commit at stage M causes a change to the output path of one of the builds of stage N < M, that change must go back to the earliest such N. The last integration stage is special: *merge* commits that change the output paths of earlier steps but don't break them are allowed, but regular commits aren't.

The expected workflow is that you'd start off of some commit in master, iterate at each stage fixing the failures without changing what already worked, and only at the last stage merging in the latest and doing fixups there. This guaranteed eventual merging falls down if every time we get to the "integration" stage a new change has been merged that fundamentally breaks all earlier stages. But this seems very unlikely to happen by accident. Maybe we can simply allow only one change in the final step at a time, and allow arbitrary iteration within that step.

peti:
Note that `staging` is assigned a ridiculously low [0.01% of Hydra's scheduling shares](https://hydra.nixos.org/jobset/nixpkgs/staging#tabs-configuration), i.e. it's building pretty much only if nothing else is. Maybe we should re-consider that jobset's priority?

vcunat:
The regular jobsets certainly should have a very large advantage against staging, but on the whole we might agree on some system how to set the shares.  It seems rather ad-hoc ATM.

shlevy:
@vcunat What do you think of this workflow:

In the normal case, most things can just go to staging and big changes (like a gcc bump) have to have their own release-small jobset pass first. At any time, a maintainer can branch off `staging-stabilization`, which then changes the workflow to: anything that fixes things broken on `staging-stabilization` goes directly there (or, if applicable, into `master` then merged into `staging-stabilization`) and is then merged into `staging`, any new mass-rebuild changes go into `staging`. Once `staging-stabilization` is ready, it is merged into `master` and we go back to the normal workflow. `staging-stabilization` can have a higher proportion of shares and, since only fixes can go into it, is much more likely to actually reach readiness in a timely manner than `staging`, without anyone having to wait to integrate their work in with others.

grahamc:
@vcunat how about we get this RFC written? :)

vcunat:
Well, I don't expect to really have time to "maintain" staging this year, with all other things I've "promised" to do, so I'm perhaps not a good person to act on this.  I believe Shea's or mine proposal or some hybrid might work reasonably. Perhaps it's better to first start using something and see how it goes, and document the resulting workflow retrospectively?  (Both of the changes really affect only Hydra's setup and workflow of "staging maintainer(s)" and not regular merges to staging.)

FRidh:
I also like the idea of two branches:
1. the current staging that we're working on getting into master (i.e. stabilizing)
2. the branch where other mass-rebuilds are staged that go into the next cycle

For 2) I would like to see aHydra job that corresponds to a small set, say `release-small` plus some additional packages such as `ghc`. Just as a check before we start rebuilding everything.

xeji:
I like the idea of `staging -> staging-stabilization -> master`. 
We could even take that a step further and make it the standard workflow for everything: 

- _All_ PRs go to  `staging` by default. `staging` gets high scheduling priority on Hydra.
- Only fixes to stabilize `staging-stabilization` or urgent fixes to `master` with limited rebuilds go there directly.
- `staging-stabilization` is (automatically?) merged to `master` (then freshly branched off `staging`) when a defined set of tests succeed and the number of Hydra failures is below some threshold. Pretty much like a channel today but with stricter requirements. So `master` always has whatever we define as  "release quality".

This approach would make it much easier to create and stabilize a release. It could also evolve into a rolling-release model in the future.

Ekleog:
I'm not sure about having *all* PRs to `staging`: this loses the current advantage of having non-mass-rebuilding changes merged fast into `master` and thus into NixOS unstable. In exchange it'd mean more effort in getting `staging` merged, though, so maybe it'd actually be a net win?

7c6f434c:
@Ekleog Hm, good question how to have a fast track and a slow track without having fast track catch the slow track in eternal catching-up and re-testing…

vcunat:
I think it's doable here.  Master especially shouldn't directly get any "risky" changes; if breakage happens, it should be quickly fixed or reverted ("moved" to another branch).  Mass-rebuild status is now auto-determined on PRs (reliably), and non-mass changes rarely break very many packages, so workflow with fixes and running nixos tests can typically be iterated much faster than on staging.

rfcs/0026-staging-workflow.md
```diff
@@ -0,0 +1,78 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+A potential drawback of this new workflow is that the additional branch may be considered complicated and/or more difficult to work with.
+
+# Alternatives
+[alternatives]: #alternatives
+
+What other designs have been considered? What is the impact of not doing this?
```
Ekleog: I think this would need to be removed / reworded before merging. Something like this?
```markdown
## Keep the statu quo

As explained above, this leads to issues with `staging` never managing
to stabilize due to the stream of mass-rebuilds.

## Merge all changes in `staging`

Also merging non-mass-rebuilds changes in `staging` would mean a
slower delivery of non-mass-rebuilding changes to `master`. Given
there is now a way to know, before merging a PR, whether it is a mass
rebuild, the benefit of having non-mass-rebuilding changes go to
`staging` are limited.

The most important one is duplicate work in case a package doesn't
build on `staging` while it has already been fixed in `master`. This can
be handled by regularly merging `master` in `staging`, as `master`
rebuilds should be almost free compared to `staging` rebuilds.
```

rfcs/0026-staging-workflow.md
```diff
@@ -0,0 +1,78 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- The exact amount of shares, which is something that has the be found out.
+
+# Future work
+[future]: #future-work
+
+-
```
Ekleog: Same here, maybe “No future work is expected from this RFC”? Or something like “Adjust hydra intervals and build shares depending on the load that will be observed”?

Actually, maybe reevaluating the choice of “small fixes in `master`” vs. “small fixes in `staging-next` too” could make sense after some experimentation, depending on how frequently we manage to merge `staging` into `master` with this new scheme.

rfcs/0026-staging-workflow.md
```diff
@@ -0,0 +1,78 @@
+---
+feature: (fill me in with a unique ident, my_awesome_feature)
+start-date: (fill me in with today's date, YYYY-MM-DD)
+author: (name of the main author)
+co-authors: (find a buddy later to help our with the RFC)
+related-issues: (will contain links to implementation PRs)
```
Ekleog: I think this should be filled in.

rfcs/0026-staging-workflow.md
```diff
@@ -0,0 +1,78 @@
...
+
+| Branch         | Interval | Shares | Jobset
+|----------------|----------|--------|-----------
+| `master`       | 4        | High   | release.nix
+| `staging`      | 12       | High   | release.nix
+| `staging-next` | 6        | High   | release-small.nix
+
+
+The check interval of `staging` is reduced from 24 hours to 12 hours. This can
+be done because only stabilization fixes shall be submitted and thus fewer
```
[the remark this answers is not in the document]
LnL7: I'm also slightly confused about what staging* branch is intended for pull requests vs stabilisation after reading this.

rfcs/0026-staging-workflow.md
vcunat: Done.

rfcs/0026-staging-workflow.md
```diff
@@ -0,0 +1,78 @@
...
+as possible.
+
+# Detailed design
+[design]: #detailed-design
+
+There shall be the following branches:
+- `master` is the main branch and all small deliveries shall go here;
+- `staging` is branched from `master` and mass-rebuilds and other large deliveries go to this branch;
+- `staging-next` is branched from `staging` and only fixes to stabilize and security fixes shall be delivered to this branch.
+
```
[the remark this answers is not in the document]
vcunat: Clarified.

rfcs/0026-staging-workflow.md
vcunat: Now "future work" is filled, basically in sense "put this RFC into action".

rfcs/0026-staging-workflow.md
vcunat: Filled.  I can't see any potential advantage of "Merge all changes in `staging`"  compared to the current workflow or the one proposed in this RFC, but it seems to be essentially the "Single branch" alternative in RFC draft now, if I understand it right.

rfcs/0026-staging-workflow.md
```diff
@@ -0,0 +1,78 @@
...
+rebuilds shall typically have to be performed. The `staging-next` shall have a
+short interval of only 6 hours. This is done because of the relatively small
+jobset, and to obtain a higher resolution to detect any troublesome deliveries.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+A potential drawback of this new workflow is that the additional branch may be considered complicated and/or more difficult to work with.
+
```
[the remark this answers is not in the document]
vcunat: Right, added.

rfcs/0026-staging-workflow.md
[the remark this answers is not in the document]
vcunat: Changed to `staging-next`.

Confusion: I believe that any fixes-only changes should go to `staging-next` (counting only problems present in the branch already).  I would hope that the current text makes it clear:
> - `staging-next` is branched from `staging` and only fixes to stabilize and security fixes shall be delivered to this branch. This branch shall be merged into `master` when deemed of sufficiently high quality.

but certainly feel free to suggest a clearer formulation.

The exact mechanism how they get there doesn't seem too important to me.  Ideally they get committed to `staging-next` directly, but if they might be in `staging` already (accidentally) and get cherry-picked to `staging-next`...

rfcs/0026-staging-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+Instead of multiple branches only a single branch, say `master`, could be kept
+for development. While this removes the issue of merge conflicts, it will result
+in continuous mass-rebuilds on `master`, slowing down the delivery of binary
+substitutes and thus development.
+
+## Reduce Hydra jobset size
+
+Reducing the size of the Hydra jobset would mean the iteration pace could be
+higher, but has the downside of testing fewer packages, and having fewer binary
+substitutes available.
```
7c6f434c: Another option here is to reduce the channel-blocking Hydra jobset size, but have a 48-hourly cycle of trying to build the long tail. Still probably increases amount of things broken in the channel, but might be a middle ground. (I am _not_ saying I prefer that compromise to the current proposal)

7c6f434c:
I think you should remove the first line of the initial message now.

rfcs/0026-staging-workflow.md
```diff
@@ -0,0 +1,100 @@
...
+for `staging`) to 12 hours. This can be done because only stabilization fixes
+shall be submitted and thus fewer rebuilds shall typically have to be performed.
+
+The `staging` branch shall have a short interval of only 6 hours. This is because
+of the relatively small jobset, and to obtain a higher resolution to detect any
+troublesome deliveries.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
```
7c6f434c: Looking at the proposal, I am not completely sure if it will in fact reduce or increase the overall number of jobs for Hydra to build. It would increase usefuk information for unit of build time, but it might increase the load.

This risk should either be mentioned as a possible drawback or there should be a discussion why this risk is not likely.

vcunat:
@7c6f434c: which part do you mean?

7c6f434c:
The text is not empty anymore.

rfcs/0026-staging-workflow.md
vcunat: The channel-update criteria contains waiting for all builds to finish, but the thing is doable by having two jobsets on the branch.  Personally I think the channel-blocking set should grow and not shrink, especially for the release branches.

vcunat:
I see, I was just looking at the RFC text :-)

rfcs/0026-staging-workflow.md
7c6f434c: I agree, but I also think it is good to have options enumerated and explicitly dismissed with a reason given.

rfcs/0026-staging-workflow.md
vcunat: I assume we will see how the load turns out in real life and adjust shares/intervals or even the whole workflow.  People actively working on staging stabilization should also be able cancel jobs and force evaluations.

In any case, I would like to push towards more power in Hydra over the long term, as I believe it will help with mass rebuilds and human time is more precious than machine time.

rfcs/0026-staging-workflow.md
vcunat: Added a sentence describing the approach.

rfcs/0026-staging-workflow.md
7c6f434c: Good point about Hydra access (this is a part of proposed workflow, right?), and I agree in principle about expanding Hydra checks.

I just want things mentioned explicitly.

I think such explicit discussion could also be a way to better understand the trade-offs relative to complicated alternatives: `staging (release-small)` -> `staging-next (release-small)` -> `staging-next (release)` (manually triggered, unless Hydra can have jobsets with preconditions).

vcunat: Well, I was always doing cancels and evals on Hydra when working with staging.  The right to manage Hydra jobsets doesn't seem more security-sensitive that push access to nixpkgs.  Overall, this text so far hasn't mentioned people &ndash; who will be doing this staging stabilization, etc.

Complicated alternatives: another thing to consider is that Hydra seems likely to be over-powered on x86 linux (relative to other platforms), and even for other reasons it makes sense for most changes to stabilize them for `x86_64-linux` first/deeper before others.

zimbatm:
Happy to merge if the release managers give me a :+1:

vcunat:
I'm certainly inclined to start using the process now(-ish).  It seems probably that we'll have to tweak it based on real experience, so here it depends:
- postpone merge after those tweaks, or
- merge now and _amend_ the RFC later, or
- merge now and _replace_ the RFC by another one later.

#### Immutable RFCs
I actually don't know if it's acceptable to do substantial changes to the RFC; I can't immediately see it [in the text](https://github.com/NixOS/rfcs/blob/master/rfcs/0001-rfc-process.md), though I'm certain we've touched it already in some discussion.  For example, IETF RFC's are strictly immutable, and even typo fixes are only published separately; that seems hard to use (I've implemented several RFCs) and IMO unnecessary when we have nice version control tools nowadays.

vcunat:
Yes, I recall we agreed on that, except for the immutability part.  I see the immutability makes little difference when the RFCs aren't meant do document anything (except that consensus-searching process); that's a significant difference from IETF.

I might consider looking into RFC 1 if explanation of this could be improved.

zimbatm:
The README should probably be updated to reflect that (which was mostly copied from RFC 1 actually)

dtzWill:
Please add link to rendered proposal for easier review. <3

dtzWill:
> @dtzWill: simple GitHub-rendering is okay?

Yep! I actually thought it was part of the RFC process since many of the early ones did this, but at a glance I don't see it mentioned?  Anyway that's all I was asking for, if you look at other RFC's they "usually" (biased by my interests and memory :P) include this in the PR text for everyone's reading convenience.

vcunat:
Let's start a bit at least: I now created a [staging-next](https://hydra.nixos.org/jobset/nixpkgs/staging-next) at a point that had relatively lots of binaries.  The influx of changes has been too fast for our darwin build slaves, so we haven't been able to merge for weeks.

FRidh:
I think we should also state that we want to maintain the branch topology and thus use `git merge --no-ff`. I recall I wasn't aware of that initially.

vcunat:
Probably:
- when fixes are merged to `staging-next` directly, merge `staging-next` to `staging`.  /cc https://github.com/NixOS/nixpkgs/issues/43651#issuecomment-405452217

FRidh:
Something like the following:

> New stabilization:
> - merge `master` into `staging`
> - merge `staging` into `staging-next`
> 
> During `staging-next` stabilization:
> - merge `master` into `staging-next`
> - merge `staging-next` into `staging`
> 
> Conclude stabilization:
> - merge `staging-next` into `master`

FRidh:
Just did the first cycle:

```
git fetch upstream
git checkout upstream/master
git merge --no-ff upstream/staging-next
git push upstream HEAD:master
git checkout upstream/staging
git merge --no-ff upstream/master
git push upstream HEAD:staging
git checkout upstream/staging-next
git merge --no-ff upstream/staging
git push upstream HEAD:staging-next
```
I need an alias for this :smile:

FRidh:
One downside I'm experiencing is that it typically means more merge conflicts and, worse, resolving the same conflicts multiple times. E.g., it happens when merging `master` into `staging-next` and then also `master` into `staging`. I notice I now refrain from merging `master` into `staging` until `staging-next` has landed in `master`.

timokau:
One possibility would be to forget about the nice-history and just hard-reset `staging-next` to `staging` when making the transition. Thats not ideal of course.

Why do we need to merge `master` into `staging-next` in the first place? Shouldn't it only be done the other way around?

FRidh:
> Why do we need to merge master into staging-next in the first place? Shouldn't it only be done the other way around?

Sometimes it can take some time before `staging-next` is ready. By that time, a lot of changes in `master` could have accumulated. To reduce the amount of rebuilding on `master`, it is merged first into `staging-next` so at least some rebuilds can be done there before merging the result back into the (newer) `master`.

Zimmi48:
> One possibility would be to forget about the nice-history and just hard-reset staging-next to staging when making the transition. Thats not ideal of course.

Hard-resetting is generally a very dangerous idea when several people are working on the same branch. It makes it really hard to track where you are compared to upstream.

orivej:
> One downside I'm experiencing is that it typically means more merge conflicts and, worse, resolving the same conflicts multiple times.

You could use `git config --global rerere.enabled true`. This makes git resolve repeated conflicts using the choices you have made the first time.

FRidh:
Another part I would like to add. When pushing to `staging`, first merge the current `master` into `staging`. That way, the person contributing a change is also solving merge conflicts their change may cause.

LnL7:
@FRidh That's a good idea, same with treewide changes that can go straight into master.

vcunat:
Right, the conflicts.  `rerere` is local to the machine, so it unfortunately won't help different people resolving the same conflicts again.  IIRC there are some problems when merges are done in a directed circle among three or more branches (two-branch case is specifically hacked around).  I could try to think about detailed consequences or find the source I have this from.  I suspect we want to avoid that and prefer two-way merges, i.e. `master <-> staging-next <-> staging`.

domenkozar:
Another thing I'd like to add is that staging merges back to base branch are done via PR and 1 day delay. This allows for community feedback before everything breaks.

vcunat:
I don't really understand that.  My point of view has been that the review (e.g. PR) is to be done *before* merging to staging.  There's typically a long delay before getting to staging and getting to master for those who don't catch up that fast (as a single day is rather short IMHO anyway).

domenkozar:
@vcunat I guess my complaint is that RFC doesn't specify how merge to master is done, I think it should go through PR and have defined a minimum time-span.

FRidh:
As soon as there is a new `staging-next` a PR could be made. That way there is a place for discussion/coordination. As it will take at least a day to build anyway...

ivan:
`staging-next` is a confusing name to me: it suggests that it's a less-stable `staging` that will be promoted to `staging`.  But that is not the case (it's really the "next" `master`).  I would suggest using a name that doesn't imply something about the `staging` branch.  If `staging-stabilization` is too long, maybe just `next` could work?

Left as-is, I think the existing naming could be a point of confusion for a good number of new people trying to figure out the branching scheme for the first time.

grahamc:
We've been following this for a bit now. Shall we merge?

Mic92:
I guess so.

Mic92:
Also we should change the name as @ivan said.

zimbatm:
Let me know when it's ready to merge

oxij:
Can't we rename `staging-next` to `testing`, `staging-stable` or whatever else after all? I'm sure everyone here, including me, got accustomed and stopped caring already, but I still think it is a very confusing naming scheme for new contributors.

xeji:
I agree. `staging-next` is a confusing name. Initially I thought it meant the next stuff to go into staging... 

> `staging-next` fits the `man gitworkflows` docs nicely, where `pu` (staging) -> `next` (staging-next) -> `master`.

Why not simply call it `next` then?

LnL7:
This should just be merged. The proposal is in use and nitpicking about the name seems pointless to me.

zimbatm:
Merging as it's implemented

jtojnar:
Is the following diagram correct?

```dot
digraph {
    "small changes" [shape=none]
    "mass-rebuilds and other large changes" [shape=none]
    "critical security fixes" [shape=none]
    "broken staging-next fixes" [shape=none]

    "small changes" -> master
    "mass-rebuilds and other large changes" -> staging
    "critical security fixes" -> master
    "broken staging-next fixes" -> "staging-next"

    "staging-next" -> master [color="#E85EB0"] [label=stabilization] [fontcolor="#E85EB0"]
    "staging" -> "staging-next" [color="#E85EB0"] [label=stabilization] [fontcolor="#E85EB0"]

    master -> staging [color="#5F5EE8"] [label="any time"] [fontcolor="#5F5EE8"]
}
```

[Open in Editor](https://edotor.net/#digraph%20%7B%0A%20%20%20%20%22small%20changes%22%20%5Bshape%3Dnone%5D%0A%20%20%20%20%22mass-rebuilds%20and%20other%20large%20changes%22%20%5Bshape%3Dnone%5D%0A%20%20%20%20%22critical%20security%20fixes%22%20%5Bshape%3Dnone%5D%0A%20%20%20%20%22broken%20staging-next%20fixes%22%20%5Bshape%3Dnone%5D%0A%0A%20%20%20%20%22small%20changes%22%20-%3E%20master%0A%20%20%20%20%22mass-rebuilds%20and%20other%20large%20changes%22%20-%3E%20staging%0A%20%20%20%20%22critical%20security%20fixes%22%20-%3E%20master%0A%20%20%20%20%22broken%20staging-next%20fixes%22%20-%3E%20%22staging-next%22%0A%0A%20%20%20%20%22staging-next%22%20-%3E%20master%20%5Bcolor%3D%22%23E85EB0%22%5D%20%5Blabel%3Dstabilization%5D%20%5Bfontcolor%3D%22%23E85EB0%22%5D%0A%20%20%20%20%22staging%22%20-%3E%20%22staging-next%22%20%5Bcolor%3D%22%23E85EB0%22%5D%20%5Blabel%3Dstabilization%5D%20%5Bfontcolor%3D%22%23E85EB0%22%5D%0A%0A%20%20%20%20master%20-%3E%20staging%20%5Bcolor%3D%22%235F5EE8%22%5D%20%5Blabel%3D%22any%20time%22%5D%20%5Bfontcolor%3D%22%235F5EE8%22%5D%0A%7D%0A)

Mic92:
I think we can append this graph to the RFC.

zimbatm:
Has the RFC workflow been copied to nixpkgs yet? RFC are just here to gain consensus, the live documentation should live somewhere else.

LnL7:
@jtojnar I guess it's not really mentioned in the RFC but I think in general even critical security fixes should go to staging-next first.

This obviously depends on the size of the rebuild and current state of staging-next, but in my experience having a large queue on master is detrimental for the stability there and often results in breakages/issues that hold back channel updates.

Ekleog:
It's [actually in the RFC](https://github.com/NixOS/rfcs/pull/26/files#diff-3c21165d2a89d60eb18398fd137a7406R39): (emphasis mine)

> `staging-next` is branched from `staging` and only fixes to stabilize ***and security fixes*** shall be delivered to this branch.

(though granted I can't find a place where it explicitly states that security mass-rebuilds don't go in `master`)

jtojnar:
I understood that as fixes to stabilize staging and fix vulnerabilities on staging. 

That interpretation is consistent with the doctrine that critical security fixes should go directly to the branch that is affected by the vulnerabilities. Not sure if this doctrine still applies with this RFC accepted.

jtojnar:
I bring you the [latest diagram](https://edotor.net/?engine=dot#digraph%20%7B%0A%20%20%20%20%22small%20changes%22%20%5Bshape%3Dnone%5D%0A%20%20%20%20%22mass-rebuilds%20and%20other%20large%20changes%22%20%5Bshape%3Dnone%5D%0A%20%20%20%20%22critical%20security%20fixes%22%20%5Bshape%3Dnone%5D%0A%20%20%20%20%22broken%20staging-next%20fixes%22%20%5Bshape%3Dnone%5D%0A%0A%20%20%20%20%22small%20changes%22%20-%3E%20master%0A%20%20%20%20%22mass-rebuilds%20and%20other%20large%20changes%22%20-%3E%20staging%0A%20%20%20%20%22critical%20security%20fixes%22%20-%3E%20master%0A%20%20%20%20%22broken%20staging-next%20fixes%22%20-%3E%20%22staging-next%22%0A%0A%20%20%20%20%22staging-next%22%20-%3E%20master%20%5Bcolor%3D%22%23E85EB0%22%5D%20%5Blabel%3D%22stabilization%20starts%22%5D%20%5Bfontcolor%3D%22%23E85EB0%22%5D%0A%20%20%20%20%22staging%22%20-%3E%20%22staging-next%22%20%5Bcolor%3D%22%23E85EB0%22%5D%20%5Blabel%3D%22stabilization%20ends%22%5D%20%5Bfontcolor%3D%22%23E85EB0%22%5D%0A%0A%20%20%20%20master%20-%3E%20%22staging-next%22%20-%3E%20staging%20%5Bcolor%3D%22%235F5EE8%22%5D%20%5Blabel%3D%22every%20six%20hours%22%5D%20%5Bfontcolor%3D%22%235F5EE8%22%5D%0A%7D%0A):

Staging workflow

Now the reverse flow (master → staging-next → staging) happens automatically every six hours thanks to https://github.com/NixOS/nixpkgs/pull/105153.

```
digraph {
    "small changes" [shape=none]
    "mass-rebuilds and other large changes" [shape=none]
    "critical security fixes" [shape=none]
    "broken staging-next fixes" [shape=none]

    "small changes" -> master
    "mass-rebuilds and other large changes" -> staging
    "critical security fixes" -> master
    "broken staging-next fixes" -> "staging-next"

    "staging-next" -> master [color="#E85EB0"] [label="stabilization starts"] [fontcolor="#E85EB0"]
    "staging" -> "staging-next" [color="#E85EB0"] [label="stabilization ends"] [fontcolor="#E85EB0"]

    master -> "staging-next" -> staging [color="#5F5EE8"] [label="every six hours"] [fontcolor="#5F5EE8"]
}
```

Ericson2314:
^ This is great, should go in the Nipxkgs manual.

jtojnar:
See https://github.com/NixOS/nixpkgs/pull/105986 :wink:

FRidh:
Commenting on @jtojnar https://github.com/NixOS/nixpkgs/pull/106302

> This is a security fix so it should go to master according to NixOS/rfcs#26. Actually, that says staging-next so I am confused because I kept hearing master and even repeated that myself.

Yes I think we need to revise/clarify what is written there. The rfc indeed says staging-next, however it lacks a motivation.
Regarding master versus staging-next, there are pros and cons to both. I don't recall what I was thinking about this at the time; the biggest issue back then was the influx of changes on staging all the time, making stabilization impossible.

### Critical security issues

I'd argue that critical issues should go to master, however if staging-next is as good as ready, that should merged right as well as well because that saves rebuilds. At that point it does not matter whether the fix goes to master or staging-next regarding the amount of rebuilds. It does matter regarding channel updates; during the rebuild there won't be any updates due to other smaller changes, but that is not very important. What is important is that `nixos-small` gets updated.

> This obviously depends on the size of the rebuild and current state of staging-next, but in my experience having a large queue on master is detrimental for the stability there and often results in breakages/issues that hold back channel updates.

This I agree with. So this is a trade-off, and I think in case of critical security fixes the delay of a day or two due to the rebuilds is worth it.

### Non-critical security fixes

These go through staging as usual, to reduce the amounts of rebuilds necessary.

PR #28, opened 2018-03-28
[RFC 0028] Nix Release Model

shlevy:
[Rendered](https://github.com/shlevy/rfcs/blob/nix-release-model/rfcs/0028-nix-release-model.md)

### DIFF
--- a/rfcs/0028-nix-release-model.md
+++ b/rfcs/0028-nix-release-model.md
@@ -1,0 +1,163 @@
+---
+feature: nix-release-model
+start-date: 2018-03-28
+author: Shea Levy <shea@shealevy.com>
+co-authors: Michael Raskin <7c6f434c@mail.ru> (any other Nix committers on-board?)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+A model of when, what, and how to make new releases of Nix.
+
+# Motivation
+[motivation]: #motivation
+
+We want to set clear expectations for users, provide guidance for
+developers, and provide both groups the benefits of easy availability
+of new improvements.
+
+# Detailed design
+[design]: #detailed-design
+
+The proposed process contains a number of interrelated components.
+I've numbered them for ease of reference and discussion, not to
+indicate any meaningful ordering. Any aspect of this can be suspended
+at the security team's discretion when handling a vulnerability, or at
+the core team's discretion for especially destructive bugs (e.g. data
+corruption).
+
+1. Before merging to `master`, require testing, documentation,
+   changelog entries, removal of WIP markers, etc. to keep `master`
+   somewhat stable, maintain bisectability, and shorten the release
+   preparation process. This includes both objective and subjective
+   components, and will always involve something of a judgment call.
+2. The Nix core team shall decide when a new release is warranted,
+   through its normal decision procedures. That decision will also
+   include designating a release manager for the release. The specific
+   conditions for a release being warranted are intentionally out of
+   scope of this RFC, but it is expected that the core team will
+   balance both the needs of the community at large and the needs of
+   developers working on Nix and the tradeoff between stability and
+   development progress. In particular, if an especially major change
+   has landed recently or is nearing readiness, the core team should
+   either mark the release point as being before the relevant change
+   or expect more stabilization work than normal.
+3. The release manager will branch off a `#-maintenance` branch
+   (e.g. `2.1-maintenance`) in advance of any mainline release. A
+   mainline release is a release that is a direct git ancestor of
+   `master` and is, in the relevant feature and bugfix senses, "ahead"
+   of all other releases at the time it is made.
+4. Changes aimed at getting the release ready (fixing bugs, improving
+   docs, etc.) should target the `#-maintenance` branch and then
+   be merged forward into `master`. No commit should exist on the
+   `#-maintenance` branch during the stabilization phase without very
+   soon thereafter being merged into `master` except under rare
+   circumstances. This implies that especially large changes ready
+   around a release should either be included before branch-off (with
+   extra time to stabilize the new feature) or should, if possible,
+   wait to be merged to `master` until after the release. Which path
+   is taken should be a collaboration between the release manager and
+   the developers of the change in question.
+5. New development unrelated to the new release can go directly into
+   `master` in parallel with a release stabilization.
+6. During stabilization, the release manager should update
+   `nixUnstable` to point to various commits of the `#-maintenance`
+   branch, to easily enable wider testing and integration.
+7. When the release manager determines readiness, the relevant commit
+   is announced to the community. Barring objections or last-minute
+   fixes judged valid by the release manager (or core team), the
+   commit is tagged and the branch goes into the maintenance phase.
+   Readiness is a judgment call, and should require increased
+   scrutiny/validation for a release with more complex/major changes.
+8. So long as no new feature work has yet happened on `master`,
+   `master` should be kept equal with the most recent `#-maintenance`
+   branch (through fast-forwards, not cherry-picks). Once new feature
+   work occurs, the semver minor (or major) versions are bumped and
+   normal development continues against `master`. This allows us to
+   maintain the "no patch release without a corresponding mainline
+   release" invariant (see point 12 below) while still allowing patch
+   releases to happen on the appropriate `#-maintenance` branch.
+9. Associated with each `NixOS` stable release is an associated
+   `maintained` series of `Nix`, with the same support lifetime as
+    the stable release.
+10. Any important bugfix can be cherry-picked from `master` to a
+    `#-maintenance` branch of a supported `maintained` series at any
+    committer's discretion.
+11. In rare cases, if a change truly only makes sense on a
+    `#-maintenance` branch (e.g. when we added support to `1.11` to
+    be able to work with a `2.0` database schema) it can be targeted
+    directly to the `#-maintenance` branch. This should be avoided if
+    possible and subject to especially careful testing and review.
+12. A maintenance release can be tagged off of the `#-maintenance`
+    branch at any time *after* there is a mainline release that
+    contains all of the fixes included in the desired maintenance
+    release, at the discretion of the Nix core team. This implies that
+    maintenance releases may be tagged off of commits behind the tip
+    of the maintenance branch, if the latest commits haven't been
+    included in a mainline release.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The main drawback to having a formalized model in general is our model
+may not match a reasonable productive process in practice and so may
+lead to obnoxious busy work or delays for no purpose.
+
+The drawbacks of this specific model are a bit hard to see without
+putting it into practice. One thing that seems apparent is there is a
+non-trivial amount of complexity (12 bullet points!) which could
+perhaps be reduced. It also may prove difficult under this model to
+develop especially large features, as without the right development
+practices (e.g. feature flags on `master`) this can tend to create
+new long-lived branches that can effectively become the new `master`.
+
+# Alternatives
+[alternatives]: #alternatives
+
+If we don't do anything at all, we are likely to have more long-lived
+unreleased branches and significant undertested divergence, as well as
+lack of guidance/clear policy for the Core team to handle release
+requests.
+
+This model is a synthesis of a [discussion] on the nix-core mailing
+list, some other concrete designs can be seen there. Additionally, it
+has evolved through the discussion on the [PR].
+
+[discussion]: https://groups.google.com/forum/#!msg/nix-core/9L7jZ9W8VGc/8LaBUc_tBQAJ
+[PR]: https://github.com/NixOS/rfcs/pull/28
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What specific criteria should the Nix core team use when deciding to
+start a release process?
+
+# Future work
+[future]: #future-work
+
+CI, both to ensure `master` is working and to ensure actively
+maintained series can build and install `master`.
+
+Streamlined responsive code review practices, ideally to the point of
+allowing low-friction branch protection.
+
+Interface descriptions and guarantees suitable to semver.
+
+We may in the future decouple maintenance support from the `NixOS`
+release cycle. Any work which involves splitting components out of
+`Nix` or breaking `Nix` itself up into components will also need to
+keep this in mind.
+
+Automated testing of semver bounds could be nice. Newer versions
+running older testsuites, alternating new and old versions running
+operations on the same store, old versions running all new versions
+testsuites except tests explicitly marked as being bugfixes since
+then, etc.
+
+Explicit guidelines/process around developing major features without
+long-lived branches or mainline instability.
+
+Some kind of policy around reversions, particularly when the process
+described here is accidentally violated, would be helpful.
--- a/rfcs/0028-nix-release-model.md
+++ b/rfcs/0028-nix-release-model.md
@@ -51,15 +51,14 @@ corruption).
    of all other releases at the time it is made.
 4. Changes aimed at getting the release ready (fixing bugs, improving
    docs, etc.) should target the `#-maintenance` branch and then
-   be merged forward into `master`. No commit should exist on the
-   `#-maintenance` branch during the stabilization phase without very
-   soon thereafter being merged into `master` except under rare
-   circumstances. This implies that especially large changes ready
-   around a release should either be included before branch-off (with
-   extra time to stabilize the new feature) or should, if possible,
-   wait to be merged to `master` until after the release. Which path
-   is taken should be a collaboration between the release manager and
-   the developers of the change in question.
+   be merged forward into `master`. Normally, during the stabilization 
+   phase the '#-maintenance' branch should be merged into `master` 
+   soon after each set of commits. This implies that especially large 
+   changes ready around a release should either be included before
+   branch-off (with extra time to stabilize the new feature) or 
+   should, if possible, wait to be merged to `master` until after the
+   release. Which path is taken should be a collaboration between the
+   release manager and the developers of the change in question.
 5. New development unrelated to the new release can go directly into
    `master` in parallel with a release stabilization.
 6. During stabilization, the release manager should update
--- a/rfcs/0028-nix-release-model.md
+++ b/rfcs/0028-nix-release-model.md
@@ -92,7 +92,9 @@ corruption).
 12. A maintenance release can be tagged off of the `#-maintenance`
     branch at any time *after* there is a mainline release that
     contains all of the fixes included in the desired maintenance
-    release, at the discretion of the Nix core team. This implies that
+    release, at the discretion of the Nix core team. If for there is
+    also a newer `#-maintenance branch`, it should also be released 
+    before (or simultaneously with) the older one. This implies that
     maintenance releases may be tagged off of commits behind the tip
     of the maintenance branch, if the latest commits haven't been
     included in a mainline release.


shlevy:
cc @NixOS/nix-core

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,127 @@
...
+of new improvements.
+
+# Detailed design
+[design]: #detailed-design
+
+The proposed process contains a number of interrelated components.
+I've numbered them for ease of reference and discussion, not to
+indicate any meaningful ordering. Any aspect of this can be suspended
+at the security team's discretion when handling a vulnerability. Items
+marked with a * need technical work to implement.
```
7c6f434c: Do you suppose that this technical work is a prerequisite for adopting the process, or can happen in parallel?

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,127 @@
...
+
+1. Keep `master` as close to release readiness as possible. This
+   includes testing, documentation, changelog entries, etc. being
+   ready *before* merge.
+2. *Use code review and automated checks, enforced through branch
+    protection, to ensure point 1 (and generally improve quality and
+    increase knowledge spread).
+3. Follow semver. In light of 1, this includes ensuring any necessary
+   semver bumps happen before a relevant change is merged.
```
7c6f434c: Maybe a future work item to test the semver claims at some point? I mean that a newer version should be able to run the entire test suite of an older version, unless a breaking change is declared; and `A B A B` operations by two claimed-compatible versions on the same store should work.

rfcs/0028-nix-release-model.md
shlevy: I think it can happen in parallel. I'll update the RFC, thanks.

rfcs/0028-nix-release-model.md
shlevy: Ah, nice idea!

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,127 @@
...
+4. Any community member may request a new release by opening an issue
+   on GitHub. If any Nix core team member (who isn't the requestor)
+   thinks a release is warranted, they can assign themselves the
+   ticket and assume release manager responsibilities for the release.
```
7c6f434c: Any reason to say that the release manager cannot be the requestor? Maybe better to say that there should be some time (a day? a week?) for any core team members who think release should not happen until [something] to explain their position?

If one core team member thinks it is time for release, but doesn't see anyone else in the core team willing to become a release manager, talking a random contributor on `#nixos` into opening a please-release ticket can happen even unconciously.

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,127 @@
...
+5. There will be at least one mainline release before each `NixOS`
+   stable release.
```
7c6f434c: I think this implies a weaker form of 4 to be useful. If no core team member is willing to initiate a release, maybe there shouldn't be one even if the specified number of months has elapsed?

shlevy: I think we should view this as a commitment by the core team that *someone* will be willing to step up. If development on Nix really slows down that much we can change this.

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,127 @@
...
+6. The release manager will branch off a `#-stabilization` branch
+   (e.g. `2.0.1-stabilization`) in advance of any release.
+7. Changes aimed at getting the release ready (fixing bugs, improving
+   docs, etc.) should target the `#-stabilization` branch and then
+   be merged forward into `master`. No commit should exist on the
+   `#-stabilization` branch without very soon thereafter being merged
+   into `master`.
```
7c6f434c: Do you imply that if `master` drops some code that is still present in the stabilisation branch, the release process has to be restarted?

rfcs/0028-nix-release-model.md
shlevy: Sure, that's probably better.

rfcs/0028-nix-release-model.md
shlevy: Hmm... The idea of stabilization is that it should be fairly quick. If it's taking long enough that master wants to diverge this much, we've gone wrong IMO. Can you give an example of where what you're suggesting might happen?

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,127 @@
...
+8. New development unrelated to the new release can go directly into
+   `master` in parallel with a release stabilization.
+9. During stabilization, the release manager should update
+   `nixUnstable` to point to various commits of the `#-stabilization`
+   branch, to easily enable wider testing and integration.
+10. When the release manager determines readiness, the relevant commit
+    is tagged and the branch receives no further commits. Normal
+    development against `master` continues.
```
7c6f434c: Maybe advise the release manager to announce that some commit is going to become a release, with a short period of time (a couple of days?) when community members can request an explicit acknowledgement that some PRs against stabilisation branch are not going to be included? It is hard to remember everything.

shlevy: Good call, thanks.

rfcs/0028-nix-release-model.md
7c6f434c: Well, there is not much that I can predict getting dropped… But, say, if derivation-files get dropped completely (which would be a pity for me but seems likely), I wouldn't be surprised if it happens soon after starting a release stabilisation branch. If by any chance it so happens that a new feature would be simpler to implement without derivation-files, there can be an incentive to drop in master before release is formally announced.

rfcs/0028-nix-release-model.md
7c6f434c: A raw idea that might be silly or might be a basis for something: if a test for a _new_ patchlevel release fails for the previous release, it must have an explicit reference to a bug report.

rfcs/0028-nix-release-model.md
shlevy: So you're imagining a) we branch off a stabilization branch b) we remove derivation files on master*, c) we need a fix on stabilization that involves derivation files? I think that kind of thing is unlikely enough that we can just tell people to wait for stabilization... Maybe we should revisit if/when this arises.

*: IMO if we do remove derivation files it should still be possible to serialize derivations somehow, just maybe not by default.

shlevy:
@7c6f434c Can I add you as coauthor?

rfcs/0028-nix-release-model.md
7c6f434c: If the question is someone stepping up, maybe add a provision that as a normal Nix core team decision (by the +2 -1 rule) a willing Nixpkgs committer can be granted release responsibilities?

I just want to say that if this point is ever relevant with the process as described, there is probably an unanimous agreement that it is hard/risky/pointless to do release right now (if the only problem is finding time, asking for help sounds like a good idea).

rfcs/0028-nix-release-model.md
7c6f434c: As written, (c) is guaranteed to happen: it will be the update to the documentation that replaces «is deprecated and will be dropped soon» with «will be dropped in the next release».

shlevy: Hmm... OK I'll add an exclusion for this, but I didn't intend for this to be *quite* so legalistic :grinning:

7c6f434c:
A global question given it is an RFC about discretionary releases.

If some significant change is foreseen (say, a feature is slowly approaching the state where it is built by default and needs only a runtime switch), should the policy be to release before or after the change?

rfcs/0028-nix-release-model.md
shlevy: OK, will clarify.

rfcs/0028-nix-release-model.md
7c6f434c: Well, I am trying to balance nitpickiness and specific-scenario-realism. I think an exclusion specifically for documenting dropped features is not a good idea per se.

Maybe the answer depends on the answer to my global question: before-large-changes releases have larger risk of immediate divergence (whatever the subject matter is), after-large-changes releases have larger risk of missed consequences.

shlevy:
@7c6f434c I think we can consider that out of scope/up to the judgment of the individual team members for now.

7c6f434c:
Re: releases-around-changes: it feels a bit strange to describe the small details of a process without even trying to find out if there is a consensus on a major pont.

Re: coauthor: I am not sure what are expectations (I haven't seen RFC process actually working as a process lately). I am not yet sure I will actually support all the points of the final version, for example. On the other hand, I guess you'd want to find a coauthor among people with Nix commit access?

shlevy:
OK, will update in response to the feedback later.

shlevy:
@7c6f434c Updated.

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,158 @@
...
+I've numbered them for ease of reference and discussion, not to
+indicate any meaningful ordering. Any aspect of this can be suspended
+at the security team's discretion when handling a vulnerability. Items
+marked with a * need technical work to implement, which needn't delay
+the rest of the process.
+
+1. Keep `master` as close to release readiness as possible. This
+   includes testing, documentation, changelog entries, etc. being
+   ready *before* merge.
+2. *Use code review and automated checks, enforced through branch
```
7c6f434c: Please close this italic

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,158 @@
...
+17. In rare cases, if a change truly only makes sense on a
+    `#-maintenance` branch (e.g. when we added support to `1.11` to
+    be able to work with a `2.0` database schema) it can be targeted
+    directly to the `#-maintenance` branch. This should be avoided if
+    possible and subject to especially careful testing and review.
+18. A maintenance release can be tagged off of the `#-maintenance`
+    branch at any time *after* there is a mainline release that
+    contains all of the fixes included in the desired maintenance
+    release, at the discretion of any Nix core team member.
+19. *Automated testing, ideally applied to all mainline PRs, should
```
7c6f434c: Please close this italic

shlevy:
@7c6f434c Fixed unintended italics, thanks

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,158 @@
...
+may not match a reasonable productive process in practice and so may
+lead to obnoxious busy work or delays for no purpose.
+
+The drawbacks of this specific model are a bit hard to see without
+putting it into practice. One thing that seems apparent is there is a
+non-trivial amount of complexity (19 bullet points!) which could
+perhaps be reduced. It also may prove difficult under this model to
+develop especially large features, as without the right development
+practices (e.g. feature flags on `master`) this can tend to create
+new long-lived branches that can effectively become the new `master`.
```
7c6f434c: Maybe add to the future work: guidelines for developing major new features without long-lived unmerged branches, but also without damaging the master branch stability?

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,158 @@
...
+    possible and subject to especially careful testing and review.
+18. A maintenance release can be tagged off of the `#-maintenance`
+    branch at any time *after* there is a mainline release that
+    contains all of the fixes included in the desired maintenance
+    release, at the discretion of any Nix core team member.
+19. \*Automated testing, ideally applied to all mainline PRs, should
+     ensure that all releases of all supported maintained series are
+     able to successfully install master if a release were cut. This
+     should include a number of definitions of "install", especially
+     `NixOS` upgrades.
```
zimbatm: The list could be split up to only include the release process

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,158 @@
...
+6. If no core team member is available to do the release work but the
+   team agrees (through its normal decision procedures) that a release
+   is warranted, a volunteer from the community can serve as release
+   manager.
+7. Under normal circumstances, there will be at least one mainline
+   release between each `NixOS` stable release. An explicit decision
+   *not* to release should be made by the core team if a stable
+   release is approaching and no new release has been made.
+8. The release manager will branch off a `#-stabilization` branch
+   (e.g. `2.0.1-stabilization`) in advance of any release.
```
zimbatm: if master is always in a releasable state (according to point 1), is this branch necessary?

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,158 @@
...
+at the security team's discretion when handling a vulnerability. Items
+marked with a * need technical work to implement, which needn't delay
+the rest of the process.
+
+1. Keep `master` as close to release readiness as possible. This
+   includes testing, documentation, changelog entries, etc. being
+   ready *before* merge.
+2. \*Use code review and automated checks, enforced through branch
+    protection, to ensure point 1 (and generally improve quality and
+    increase knowledge spread).
```
zimbatm: This seems to dictate having a CI to test the PRs which I am very much in favour of

rfcs/0028-nix-release-model.md
7c6f434c: Yes, because a «releasable state» is not a completely formal notion. Stabilisation process can attract wider testing (and wider reading of documentation) to make sure nothing has been missed.

rfcs/0028-nix-release-model.md
shlevy: Not sure what you mean?

rfcs/0028-nix-release-model.md
shlevy: Yeah, exactly. If nothing else, letting things cook in `nixUnstable` for a while is essential IMO.

rfcs/0028-nix-release-model.md
shlevy: Yep! Though as noted above I don't think we need to *wait* for that to adopt this, adopting this would more be seen as a commitment to move toward that.

rfcs/0028-nix-release-model.md
shlevy: Added, thanks!

7c6f434c:
@shlevy re: my doubts if we end up agreeing on everything: this point is withdrawn, after the discussion and edits I am completely happy to enthusiastically support the entirety of your vision.

7c6f434c:
OK with adding me as a coauthor (well, I have once implemented a feature in Nix that we still cannot get rid of), I guess that would mean removing specific attribution of some of the points to me (also OK with keeping this attribution if you actively prefer).

I guess putting a remark like «(also, any Nix committers willing to join as coauthors?)» in the coauthor field could be a good idea for now.

shlevy:
@7c6f434c Done, thanks!

rfcs/0028-nix-release-model.md
zimbatm: Right now the list contains general ideas like keeping master stable that don't need to be numbered. It makes the list more impressive than it has to be.

rfcs/0028-nix-release-model.md
zimbatm: To make it easier to release bug-fixes it might make sense to create `release-2.0` branch that contains all the tiny releases.

Master can keep evolving with major features without having to worry on impairing release.

If a bug fix release is needed, cherry-pick the fixes from master onto the `release-2.0` branch until it's "ready" and then tag the release from there.

If a new major or minor release is required, branch off from master and create the new `release-MAJOR.MINOR` branch. Test until it's stable and tag the `MAJOR.MINOR.0` release.

This would be closer to the nixpkgs channel releases where master is constantly evolving and patches get cherry-picked into the release channels.

7c6f434c: I think what you describe is contained in points 15–18 about maintenance branches. But if everything goes smoothly, they are not mandatory.

shlevy:
FYI, I have some simplifications and scope reductions in mind that I'll propose for this in the next few days.

rfcs/0028-nix-release-model.md
zimbatm: That's what I am saying, stabilization and maintenance branches are essentially the same thing.

7c6f434c: Well, maintenance branches are used for more than one release. I think minimising the divergence when possible is a worthy goal, so as long as it makes sense to just release the point-release from master branch, it is allowed.

zimbatm: In any case, as long as the nix-core team agrees on the process, all my input can be disregarded.

7c6f434c: Well, we don't know for sure the opinions of most of the Nix core team (only @shlevy), and maybe your comments would be used as a basis of an alternative approach.

Of course I agree that your approach has its benefits, it's all tradeoffs.

Ericson2314:
> If we do want stabilization, having separate `stabilization` and `maintenance` branches is superfluous

This makes sense to me. What we do for Nixpkgs is that and I see it applying here. I don't fully understand @shlevy's distinction so I might be missing something.

>  if master is supposed to be in a releasable state, there is no need for stabilization.

Well it would be just like in nixpkgs, branch off early so big changes not slated for that release can continue to hit master.

> However I'm skeptical that point 1 is a realistic goal. Major changes may need significant soak time before all the bugs are fixed

Tests, tests, tests. It should always be *as good as the test suite* and the maintenance period before the release is proportional to the inadequacy of the test suite.

> I'm rather meh on point 2 (code review / branch protection). Having to jump through a lot of hoops to commit on my own project will make me less likely to contribute.

Making everyone obey the same process is *crucial* to ensuring that the process isn't too cumbersome. In my experience, this is how every major open source project does things; and it doesn't entail a loss of the original author's authority/motivation.

@edolstra imagine, you could sketch out a new feature, and leave somebody else to write/fix the tests before the PR is merged. How nice is that? :D

shlevy:
@edolstra Thanks! These points all make sense and I'll push an update later taking them into account. To minimize the impact of the initial step I'll go with your suggestions completely, but for posterity, to explain why I made the original choices I did:

> It's unclear to me why stabilization branches are needed, given point 1: if master is supposed to be in a releasable state, there is no need for stabilization. However I'm skeptical that point 1 is a realistic goal. Major changes may need significant soak time before all the bugs are fixed, and they're unlikely to be perfected on a feature branch (where they're generally only tested by the feature's author).
>

Perhaps this wasn't worded clearly enough, but the idea is that point 1 means higher standards for `master`, including both easy to verify things like "everything has docs and tests" and harder to verify things like "the code looks clean". But, as you pointed out on the ml and again here, there's only so much you can do on a feature branch. The stabilization period gives us guaranteed soak time for all features as well as providing a natural place for the broader community to start testing.

Ultimately I think these goals serve related but slightly different purposes. At the very least, point 1 would mean less work for stablizers and less likelihood of master being broken for other developers. Ideally it would also make for better bisectability, which has been a pain in bug hunting in the past.

> 
> If we do want stabilization, having separate stabilization and maintenance branches is superfluous, I think. We can just make a single <version>-maintenance branch for both. This would make life a bit simpler by removing the need to change Hydra jobset configurations, etc. So the first release (e.g. 2.1.0) would be tagged on the 2.1-maintenance branch once it is deemed to be in a releasable state. Subsequent bug fix releases (2.1.1, etc.) would be made on the same branch.
>

Reusing the names is fine, but I do want to point out that the workflows for stabilization vs maintenance are very different. Stabilization has an operational end date ("when everything is working") whereas maintenance has a support window, stabilization gets fixes *first* and gets merged into master while maintenance gets cherry-picks, maintenance releases should be preceded by mainline releases with the same fixes while stabilization releases *are* mainline releases, etc.

> I'm rather meh on point 2 (code review / branch protection). Having to jump through a lot of hoops to commit on my own project will make me less likely to contribute.

I'd hope we could reach a point where this is streamlined, but it's a fair concern. I do think code review is extremely helpful for spreading knowledge and making sure multiple perspectives are taken into account, though, so I hope we can address that some way in the future.

> 
> Semver is probably a good idea, though realistically it will mean that every non-bug-fix release will be a major version bump (since they pretty much always break/remove something).

Agreed. Version numbers are cheap, though :smile: 

> It should also be decided what is covered by the scope of semver (e.g. is the Nix C++ interface included?).
>

I think to start we should only treat the documented CLI and the daemon protocol as covered, and expand as we go if need be. I'll add that clarification.
 
> Point 4 is kind of superfluous: people may already request anything, which however does not mean it will happen.

Yes, the goal here was to give an official way to make the request, that we could document for those who don't already know it.

> Regarding point 5, it might be better to make whether to do a release a core team decision, rather than an individual decision by any core team member.

Sounds good.

> 
> I don't think point 7 is necessary: if there haven't been sufficient new features since the last NixOS release to warrant a new Nix release, then there is no reason to make a new release. Users probably appreciate the stability of not having to deal with a Nix upgrade :-)

Ah OK, I thought you had said you wanted to ensure one release each time.

7c6f434c:
Is an explicit revert policy for `master` something that could replace point (2) about master-protection?

Also, as Nix is a lower-traffic project than Nixpkgs, having a channel-tracking branch and bisecting over it could be a suitable replacement for pre-review.

shlevy:
I modified the RFC according to feedback, see https://github.com/shlevy/rfcs/pull/3

@7c6f434c Can you either approve, give feedback, or let me know if you want to be removed as coauthor?

@edolstra The RFC has been largely shaped by your feedback, would you like to be listed as coauthor as well?

7c6f434c:
@shlevy I am failing at doing GitHub (as usual): how can I read the entire resulting RFC text and comment on specific subsections? I can either get a rendered version, or a proposed diff to the current proposal.

The point where I started to want line comments is semver: if a feature of the `nix` command is not documented, but it is used in a stable release of NixOS, does it deserve semver consideration?

shlevy:
@7c6f434c I'm not sure there is one such place, sadly :frowning: Can you try to comment at the closest appropriate place on the diff?

In response to your specific question: Yes, and it also should be documented :wink: Do you think that warrants explicitly calling out?

shlevy:
@7c6f434c 

> re: comments ondiff: Won't we split discussion to a point of impossibility to follow?

Probably, yeah... Maybe just commenting here is the best we can do :sob: Any github experts with a better idea?

> re: documentation: I think nix copy is currently considered to have experimental syntax, and it is used in a stable release of NixOS. Nix semver policy should have some prescription on handling this grandfathered-in situation.

I'm with @Ericson2314... Ideally we wouldn't be using this at all. If it's infeasible to switch back to `nix-copy-closure` or similar, then I think we should keep track of and limit the usage, and avoid it moving forward. But that's more nixpkgs than Nix policy IMO...

> I think revert policy should be mentioned at least in the «future work»

Good call, updated

7c6f434c:
I am not completely sure that submitting a PR to your own PR is not making things too complicated.

Well, we have an experimental fact: there is currently no working NixOS policy against using unstable features of stable Nix. I think semver decisions in Nix should have an explicit reference to this fact.

shlevy:
> I am not completely sure that submitting a PR to your own PR is not making things too complicated.

It almost certainly is :grin: I just didn't want to merge it directly given how significant a change it is and that you're co-author

> Well,  we have an experimental fact: there is currently no working NixOS  policy against using unstable features of stable Nix. I think semver  decisions in Nix should have an explicit reference to this fact.

Concrete suggestion for how to incorporate this?

7c6f434c:
Re: point 1 (master stability): maybe admit that there are objective (and almost objective) and subjective requirements? Master should always satisfy automatically verifiable requirements for a release, but there are also requirements that require human judgement and we recognise the need for stabilisation.

Re: point 3 (release decision): I am not sure if the considerations to balance are usage vs development; maybe it is more of stability vs improvements?

I think the point 4 (branching the maintenance branch) is quite unclear. Why is it separate from the next point? What is a mainline release — a minor release, and point releases should come from maintenance branches?

Re: point 13 (maintenance-branch releases): I think this effectively implies that a patchlevel release can be tagged as a revision which is behind the tip of the maintenance branch. Should this be explicit?

shlevy:
@7c6f434c Thanks, all good points, will update later.

shlevy:
OK, will do that.

7c6f434c:
I think that in the ideal world automated semver testing should include compatibility tests for Nixpkgs and NixOS, and any change in the Nix interface that is known to break a stable release of NixOS/Nixpkgs should be considered a major version bump. (Future work: co-stabilisation of Nix features that provide critical benefits for NixOS/Nixpkgs from both Nix and Nixpkgs and NixOS sides)

Ericson2314:
For the moment that's good, but let's limit nixpkgs's use formally and quickly. Too much of "just don't break downstream" quickly gets us in a Postel's law hell where anything observable could implicate stability, and evolving Nix is like traversing a minefield.

7c6f434c:
@Ericson2314 I do not say this shouldn't be broken, I just say it should be explicitly recognised as a breaking change. Nixpkgs master should only have to package the latest Nix release on every major branch.

Even without the downstream provision it is still likely that Nix release process will not produce any minor releases untill Nix 5.0.

Ericson2314:
That sounds good. Certainly undocumented parts of `nix-` *should* be considered part of the public interface for just those reasons. I guess as long as we allow changing Nix to keep compat *or* changing nixpkgs to not rely on the undocumented feature, when these situations arise, I'm good.

7c6f434c:
I think what I say would imply a need to fix known bugs (preferably by backporting fixes) in more maintenance branches, but not a limitation on the rate of change in the master branch.

Ericson2314:
@7c6f434c I just want to be able to break the interface new `nix` commands while they are declared unstable.

shlevy:
@edolstra Thoughts on the proposed changes? How do you want to incorporate your semver concerns?

edolstra:
I would drop the semver for now. Rule 1 of the semver spec is "Software using Semantic Versioning MUST declare a public API" which we don't have at the moment. We would first have to decide what our public interface is.

Otherwise it looks good to me.

shlevy:
Great, thanks! I'll have an update with all the feedback next week.

vcunat:
Semantically, our public API is the "old" set of `nix-*` cli commands, and we already have documentation for its behavior. (I agree with Ericson2314)  Eelco is thinking public _library_ API, and there we probably don't have/want a stable one yet.

shlevy:
Updated

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,157 @@
...
+   commit is tagged and the branch goes into the maintenance phase.
+   Readiness is a judgment call, and should require increased
+   scrutiny/validation for a release with more complex/major changes.
+8. So long as no new feature work has yet happened on `master`,
+   `master` should be kept equal with the most recent `#-maintenance`
+   branch (through fast-forwards, not cherry-picks). Once new feature
+   work occurs, the semver minor (or major) versions are bumped and
+   normal development continues against `master`. This allows us to
+   maintain the "no patch release without a corresponding mainline
+   release" invariant (see point 13 below) while still allowing patch
```
7c6f434c: Point 11 in the new order.

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,157 @@
...
+develop especially large features, as without the right development
+practices (e.g. feature flags on `master`) this can tend to create
+new long-lived branches that can effectively become the new `master`.
+
+# Alternatives
+[alternatives]: #alternatives
+
+If we don't do anything at all, we are likely to have more long-lived
+unreleased branches and significant undertested divergence, as well as
+lack of guidance to the Core team for how to handle release requests.
```
7c6f434c: Given that the Nix Core team has to approve this policy first, I would say «lack of a clear policy» instead of «lack of guidance»

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,158 @@
...
+   somewhat stable, maintain bisectability, and shorten the release
+   preparation process. This includes both objective and subjective
+   components, and will always involve something of a judgment call.
+2. The Nix core team shall decide when a new release is warranted,
+   through its normal decision procedures. That decision will also
+   include designating a release manager for the release. The specific
+   conditions for a release being warranted are intentionally out of
+   scope of this RFC, but it is expected that the core team will
+   balance both the needs of the community at large and the needs of
+   developers working on Nix. In particular, if an especially major
```
7c6f434c: Not sure if the tradeoff is between developers and users; I would says that both groups can be split w.r.t. the real tradeoff: stability vs. development progress.

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,158 @@
...
+10. Any important bugfix can be cherry-picked from `master` to a
+    `#-maintenance` branch of a supported `maintained` series at any
+    committer's discretion.
+11. In rare cases, if a change truly only makes sense on a
+    `#-maintenance` branch (e.g. when we added support to `1.11` to
+    be able to work with a `2.0` database schema) it can be targeted
+    directly to the `#-maintenance` branch. This should be avoided if
+    possible and subject to especially careful testing and review.
+12. A maintenance release can be tagged off of the `#-maintenance`
+    branch at any time *after* there is a mainline release that
```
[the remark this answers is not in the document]
shlevy: Ah right, updated.

rfcs/0028-nix-release-model.md
shlevy: Ah right, updated.

rfcs/0028-nix-release-model.md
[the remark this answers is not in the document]
shlevy: Yes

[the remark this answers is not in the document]
shlevy: The idea is we never want the maintenance series to be "ahead" of the mainline series in any meaningful way.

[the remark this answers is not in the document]
7c6f434c: I understand the idea, but does that mean that «2.0.2» might be expected to wait for «2.1.0» release in case some feature development happens?

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,159 @@
...
+developers, and provide both groups the benefits of easy availability
+of new improvements.
+
+# Detailed design
+[design]: #detailed-design
+
+The proposed process contains a number of interrelated components.
+I've numbered them for ease of reference and discussion, not to
+indicate any meaningful ordering. Any aspect of this can be suspended
+at the security team's discretion when handling a vulnerability.
```
7c6f434c: Maybe we want to follow the good part of Linus Torvalds' position on security bugs and say that the same is expected to happen in case of data-loss bugs…

rfcs/0028-nix-release-model.md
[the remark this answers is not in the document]
shlevy: Yes, that's the intention. That requires a certain level of stability on master, which I think is one of the goals here to encourage :grinning:

rfcs/0028-nix-release-model.md
shlevy: Fixed.

rfcs/0028-nix-release-model.md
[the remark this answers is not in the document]
7c6f434c: Maybe define the mainline release explicitly then? Because «2.0.2» in my example could be mistaken for a mainline release (if there is no «2.1.0» or «3.0.0»).

Maybe just remove the word «mainline» and say something like «there is a release from a commit that diverges with the master branch after all of the master-branch versions of the fixes included in the desired maintenance release»?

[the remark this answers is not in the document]
shlevy: Does the updated point 3 help here?

rfcs/0028-nix-release-model.md
```diff
@@ -0,0 +1,163 @@
...
+   developers working on Nix and the tradeoff between stability and
+   development progress. In particular, if an especially major change
+   has landed recently or is nearing readiness, the core team should
+   either mark the release point as being before the relevant change
+   or expect more stabilization work than normal.
+3. The release manager will branch off a `#-maintenance` branch
+   (e.g. `2.1-maintenance`) in advance of any mainline release. A
+   mainline release is a release that is a direct git ancestor of
+   `master` and is, in the relevant feature and bugfix senses, "ahead"
+   of all other releases at the time it is made.
```
7c6f434c: If some stabilisation is needed the release itself will not be a direct git ancestor of master.

rfcs/0028-nix-release-model.md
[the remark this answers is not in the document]
7c6f434c: Not too much…

rfcs/0028-nix-release-model.md
7c6f434c: Ah, my claim also assumes some minor feature development on master, but that can happen, too…

shlevy: It will be, because we merge forward during stabilization

7c6f434c: So basically 
> No commit should exist on the `#-maintenance` branch during the stabilization phase without very soon thereafter being merged into `master` except under rare circumstances.
means just that «normally, maintenance branch should be merged into master frequently»?

shlevy: Yep... what else would it mean?

7c6f434c: Well, when you focus on commits being merged, there are interesting things that can happen with commits. Like a branch with a couple of fixes being merged into both maintenance and master.

shlevy: > Changes aimed at getting the release ready (fixing bugs, improving docs, etc.) should target the #-maintenance branch and then be merged forward into master.

This seems clear enough to me. If someone merges a branch twice instead of merging into the maintenance branch and then merging maintenance into master, that's not the expected workflow but it's also not actually a big deal... So there's a merge commit that they differ by, trivial to fix up with a merge later if we need to.

7c6f434c: I just think that if you define things like «mainline release» in a way that assumes the expected workflow, and you have clear expectations, maybe a more explicit description is a good idea. It is not even longer anyway.

shlevy: Can you open a PR with your proposed rewording?

rfcs/0028-nix-release-model.md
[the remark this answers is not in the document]
7c6f434c: Also, if there are 2.0.8, 2.1.6, 2.2.3 with 2.2.3 still being mainline, and a minor bug is found, does 2.0.9 normally require 2.1.7 to happen first/simultaneously or only 2.2.4?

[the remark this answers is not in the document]
shlevy: Currently we only have provisions for a single maintained series at any given time:

> Associated with each NixOS stable release is an associated maintained series of Nix, with the same support lifetime as the stable release.

But IMO if we ever do have more than one we should always update the newer ones first.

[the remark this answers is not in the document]
7c6f434c: I think .09 is still maintained-ish all April and .03 all October. At least for things that are annoying and simple to backport.

shlevy:
@NixOS/nix-committers @NixOS/nix-core @edolstra Any further thoughts/concerns here?

7c6f434c:
I think we addressed the issues raised by him. 

If we believe the Nix Core team experiment, maybe we should try to talk some other Core Team person into reviewing it (maybe @vcunat?), then — if we get to the state of two Core Team members approving the new version — we could say it is ready for a Core vote?

shlevy:
Re the core team: The setup was meant to be robust against different numbers of people participating, so if I'm OK and Eelco's OK (which I'd like confirmation from him on) then we can move forward.

7c6f434c:
I know; I just have an impression of «generally OK with it after the changes» more than «definitely for it», and getting release management advice from another NixOS release manager could be useful.

shlevy:
@edolstra ping?

grahamc:
@edolstra is following this for the release of 2.1 -- happening now -- so maybe we can merge this soon?

shlevy:
@edolstra Do we want to follow this moving forward as well?

Ericson2314:
:(

PR #29, opened 2018-07-28
[RFC 0029] Backports team

samueldr:
[Rendered](https://github.com/samueldr/rfcs/blob/backports-team/rfcs/0029-backports-team.md)

* * *

#### TODO

 * [x] Finish the draft *Detailed design* section from the rough notes.
 * [ ] Find a co-author.
 * [ ] Rebase into one commit when finished. <sup>(Keeping history meanwhile for those playing at home)</sup>

* * *

Quick note, I may seem critical into the RFC text, but I am mainly terse. I know and understand that *it's a lot of work*. That is, especially true when considering how it involves the release managers.

* * *

## Additional motivation

The tipping point making me write up this RFC is how stale current-stable seems to become. I was particularly incensed by spotting [#43675](https://github.com/NixOS/nixpkgs/issues/43675), opened on July 17th 2018, while the issue was fixed on April 25th 2018 on master, but the fix not backported. It seems that the expectation that authors will backport their own fixes is not universally true. I hope to help by creating processes, tooling and a community around backports.

### DIFF
--- a/rfcs/0029-backports-team.md
+++ b/rfcs/0029-backports-team.md
@@ -1,0 +1,132 @@
+---
+feature: backports_team
+start-date: 2018-07-28
+author: samueldr
+co-authors: ...
+related-issues: ...
+---
+
+Summary
+=======
+[summary]: #summary
+
+This intends to formalize the backports process and formalize the formation of
+a team managing backports.
+
+
+Motivation
+==========
+[motivation]: #motivation
+
+The way it is currently handled makes it so too often fixes for lesser-used
+software are not backported to the *current-stable* revision of NixOS.
+Furthermore, even some more-often used software sometimes lag behind since the
+updates were simply not cherry-picked.
+
+Backports is not only a security feature, but also a user-experience feature.
+Upstream software is getting updated; it would be best if current-stable does
+not get the reputation of being stale.
+
+By formalizing the process, it is expected to create a good hygiene that will
+keep NixOS current-stable fresher, instead of presenting stale software to the
+end-users.
+
+Finally, by appointing a team, even if fluid, it is expected that the
+responsibilities given to the users will _nudge them_ toward doing the work
+required of maintaining the backports. The work of integrating backports should
+be the release manager's, as per [RFC 0015][rfc0015]. Through [dubious
+stats][stats_release], it seems that for the 18.03 release their work has
+been minimal, and instead filled-in by ad-hoc work by other members.
+
+
+Detailed design
+===============
+[design]: #detailed-design
+
+> *Currently being worked on from misc. notes.*
+
+<!--
+
+TODO!
+
+Main points:
+
+## Team
+
+ * Building a team with multiple people
+ * Prevent burnouts by distributing across people
+
+## Process / Tooling
+
+ * Make the effort distributed
+ * Make the process bite-sized
+ * Brain activity not required for most parts
+ * Standardize to expedite backports
+
+-->
+
+
+Drawbacks
+=========
+[drawbacks]: #drawbacks
+
+Formalizing a process always reeks of red tape. This may turn off some
+volunteers away from doing backports.
+
+It may well happen that once the team formed, nothing changes and the backports
+are not maintained any more than they are.
+
+Writing and maintaining tooling may slow down or halt progress toward actually
+maintaining backports.
+
+
+Alternatives
+============
+[alternatives]: #alternatives
+
+Continuing as-it-is, with ad-hoc updates, sometimes missed. This has proved not
+to be the most successful way to manage backports.
+
+It could also be possible to only implement parts of the RFC. Either the team
+or the tooling. Both are of equal value and generally independent. They would,
+though, work best if working together.
+
+
+Unresolved questions
+====================
+[unresolved]: #unresolved-questions
+
+ * Actual team organization (if any).
+ * Specifying processes for all software updates and fixes.
+
+
+Future work
+===========
+[future]: #future-work
+
+ * Specifying collaboration with automated updates (@r-ryantm)
+
+
+Definitions
+===========
+[definitions]: #definitions
+
+### Backport
+
+Taking fixes and features from a newer version and applying them to an older
+maintained version to provide the equivalent fix or feature.<sup>
+[[1]][1]
+[[2]][2]
+[[3]][3]
+</sup>
+
+### Current-stable
+
+A version-agnostic way to specify the current stable branch of NixOS.
+
+
+[1]: https://en.wikipedia.org/wiki/Backporting
+[2]: https://en.wiktionary.org/wiki/backport
+[3]: https://access.redhat.com/security/updates/backporting
+[rfc0015]: https://github.com/NixOS/rfcs/blob/master/rfcs/0015-release-manager.md
+[stats_release]: https://gist.github.com/samueldr/7ec402f71d3bb2ac2e059f33d29d95bb
--- a/rfcs/0029-backports-team.md
+++ b/rfcs/0029-backports-team.md
@@ -38,6 +38,9 @@ be the release manager's, as per [RFC 0015][rfc0015]. Through [dubious
 stats][stats_release], it seems that for the 18.03 release their work has
 been minimal, and instead filled-in by ad-hoc work by other members.
 
+Additionally, the formalized process will be built upon the [general
+guidelines][grahamc_backport] previously written.
+
 
 Detailed design
 ===============
@@ -98,6 +101,8 @@ Unresolved questions
 
  * Actual team organization (if any).
  * Specifying processes for all software updates and fixes.
+ * Keeping patch-releases up-to-date when a major is out at upstream.
+ * Finding sources for patches (e.g. debian) when
 
 
 Future work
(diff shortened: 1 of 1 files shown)


rfcs/0029-backports-team.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Examples
+
+An example of broken software needing a major update to be backported is
+Spotify, which regularly breaks with old versions.
+
+Don't backport if the patch is just for Darwin, they use nixpkgs-unstable not a
+stable branch.
+
+> `FIXME` Is this true?? I see nixpkgs-18.03-darwin in the channels list.
```
LnL7: No, darwin related changes should follow the same rules now that we have stable channels. Including darwin only fixes if applicable.

FRidh:
- Request for `ofborg` to support backporting https://github.com/NixOS/ofborg/issues/133
- Request for @ryantm's bot to backport patch updates https://github.com/ryantm/nixpkgs-update/issues/65

rfcs/0029-backports-team.md
```diff
@@ -0,0 +1,256 @@
...
+   no patch-level or minor-level update is made available.
+ * Any update when the current stable version is completely broken.
+ * Extremely security-sensitive software, in particular web browsers,
+   mail user-agents and the kernel, can and should be updated to the version
+   expected by upstream to be secure.
+
+### Examples
+
+An example of broken software needing a major update to be backported is
+Spotify, which regularly breaks with old versions.
```
timokau: I'm hoping to fix this with https://github.com/NixOS/nixpkgs/pull/44123.

rfcs/0029-backports-team.md
samueldr: Exactly! I have taken this from the linked [general guidelines](https://gist.github.com/grahamc/c60578c6e6928043d29a427361634df6#what-to-backport) and was a bit worried that this wasn't true. Removing.

rfcs/0029-backports-team.md
samueldr: Awesome, said example was gleaned from the [general guidelines](https://gist.github.com/grahamc/c60578c6e6928043d29a427361634df6#what-to-backport) linked earlier. I'm sure as a group we can either abstract away the exact software which break as any one of those, or think up a small list of "glassware" (often-breaking software).

vcunat:
This isn't in any manual (yet), but I suggested a different approach (since 18.03) than forming a separate team, described as _## Maintenance workflow_ in https://groups.google.com/forum/#!topic/nix-devel/3KxPNwxDV9E.

In particular I would stress:
 - the efficiency point - knowledge of what to do isn't much about being a "backport person" but about knowing that particular package well (e.g. understanding how it's used, what exactly changed in a version);
 - motivation point - I suspect it might be hard to find many people who need bugfixes of _everything_ on 18.03.

... maybe I'm wrong and a team catching everything will appear.  In general I think we need more contributors that are more focused on some packages; years ago it was a necessity that contributors touched very many packages, but the community has been growing significantly, and hopefully we could afford to move in this direction.

- - -

Then there's the usual maintenance problem, regardless of process (and not NixPkgs-specific): most packages don't really maintain older versions.  Very simple patches (bug/security) tend not to be too difficult to backport with some confidence, but not even all security fixes fall into that category and then you may:
 - keep the old buggy version; or
 - update to new version on stable as well (bringing some incompatibilities); or
 - somehow backport without understanding it well (including some other distribution doing it and us piggy-backing).

Fortunately our release cycle is short, which helps this.

vcunat:
In nixpkgs almost everything will just pass CI when cherry-picked blindly (as it passes on unstable as well), but the SW may contain some kind of incompatible changes, including new bugs.

Ah, yes, so what clearly seems missing:
 - [ ] add "what to backport" to manual
 - [ ] add link to that from the PR template (and yes/no/bordeline checkboxes), possibly even if we decide to choose the "backport team" way instead

_And some more feedback/discussion about the backporting workflow._

samueldr:
> In nixpkgs almost everything will just pass CI when cherry-picked blindly (as it passes on unstable as well).

I find this oddly entertaining how `release-18.03` had broken eval after a [backport merge](https://github.com/NixOS/nixpkgs/pull/44270) where the failure <sup>❌</sup> was ignored.

> the SW may contain some kind of incompatible changes, including new bugs.

This is *also* relevant on `nixpkgs-unstable`, and is (in my opinion) the one big issue with the automated upgrades some packages received from a friendly robot.

When software contains incompatible changes (let's ignore bugs now), *unstable* users may receive them at any moment, uncoordinated or not. This is not inherently bad, but should probably be added as a note somewhere about the channels "software will update through major revisions without notification, this could introduce new behaviours, features and issues".

Additionally, bugs are an issue, when introduced with updates. An update to software that *also* introduces bugs could go from mildly annoying to infuriating depending on the scope of the software and the bugs.

What those last paragraphs are about, is that I don't know how this situation should be handled. Software changes, it is expected. There is a dire lack of *changelog updates* with our contributions. Not just for big updates. It may be of interest (possibly out of scope, but touching this RFC) to also figure out a more in-depth **update policy** (and explore those of other distributions). Such a policy could include figuring out a way to include upstream CHANGELOG somehow, and force *somehow* all packages to include some update log. Not that the nixos manual should mirror upstream software's own changelogs, but that it *is* necessary to be a bit more comprehensive and try to be proactive into noting changes that are of interest to the end-users.

Then, there may be a need for package-level testing integration, even if not integrated with the complete nixos test suites. Packages opting-in into those could be more reliably automatically updated, while those *not* opting into those tests will need some eyballing. This is true even for unstable.

vcunat:
We might make some Borg checks more fatal/visible/mandatory.

- [ ] We can explicitly state somewhere that `nixos-unstable` is... unstable?  (Well, document what it means &ndash; the policies for both branches.)  But in my experience people tend to be rather surprised that so many people are running on unstable, as presumably they're reminded of Debian or others.

Later we can think of improving the branch policies, but so far they seemed to have worked reasonably well, perhaps partially thanks to the ease of rolling back (and using other versions in general).

IIRC there's some WIP around package testing; there was even a presentation on the last NixCon... but in any case the main work there is to write those tests :-)  except for the easy cases like just a missing `doCheck = true;`.

Ekleog:
I'd have a question about the “what to backport” instructions included here.

In https://github.com/NixOS/nixpkgs/pull/43889#issuecomment-409790889 (see the linked post for full details), we wonder whether to backport an update to software that have been declared as unstable by upstream. Basically, upstream offers no support in debugging what could potentially go wrong with any version that is not the latest one, and doesn't care about backporting fixes.

Given the fact that upstream is officially unstable (as embodied by a v0.* version), I wonder whether exactly the same rules for backporting fixes should be applied, or whether we should be more lenient in our backporting of these, as users know they're using unstable software anyway and it'd take too much resources handling backport of all fixes.

For a comparison:
 * Debian has 0.33.0 in testing (but has no matrix-synapse in any stable version).
 * Fedora 28 has 0.31.2, last updated 2 months ago. I think the notable fact though is that they have upgraded from 0.27.3 to 0.31.2 (sorry for the web proxy link, pkgs.fedoraproject.org appears to use an invalid certificate with HSTS from here) during the lifecycle of Fedora 28.

So, I would think:
 * If we aim for the stability level of Debian, maybe we should just not accept declared-unstable-by-upstream packages into our stable version (though I'm not sure whether that was a conscious choice by Debian or whether they just had no stable version rollout since their having a package for synapse)
 * If we aim for the stability level of Fedora, maybe we should accept potentially-breaking update backports of declared-unstable-by-upstream packages

It may be important to make a distinction between unstable software used online (like synapse) and unstable software entirely offline, though, as unstable software used online can be broken by updates to *other* software on other machines (as has been my experience with synapse, see the linked post).

Anyway, I would think the guidelines could be made a bit more explicit around the “what to backport” decision, for on- and off-line declared-unstable-by-upstream software. :)

What do you think about it?

Zimmi48:
> If we aim for the stability level of Debian, maybe we should just not accept declared-unstable-by-upstream packages into our stable version

This is incompatible with the way branching the stable branch from the unstable one works in nixpkgs.

samueldr:
I am open for every crazy, impossible and different comments about the branching model and backports. Even if they are not possible to implement, knowing about them is useful, and could ideas or concept that weren't thought of.

Thanks for all your comments.

7c6f434c:
@vcunat for interesting package tests we need a solution to «testing with X11 needs partial duplication of NixOS test functionality, and debugging NixOS tests with all their overhead is way too annoying».

@Ekleog well, software used offline but mainly for handling some maliciously developed data formats (like .doc) could be effectively broken if the format updates quickly gain use.

rfcs/0029-backports-team.md
Baughn: Anything that can't be hosted on nixos.org is a candidate for this. `grep` can find more than a couple. (It also finds things like tor-browser-bundle, which we probably could rehost but haven't, and which is a prime candidate for backporting.)

rfcs/0029-backports-team.md
```diff
@@ -0,0 +1,251 @@
...
+Alternatives
+============
+[alternatives]: #alternatives
+
+Continuing as-it-is, with ad-hoc updates, sometimes missed. This has proved not
+to be the most successful way to manage backports.
+
+It could also be possible to only implement parts of the RFC. Either the team
+or the tooling. Both are of equal value and generally independent. They would,
+though, work best if working together.
```
Baughn: It would be possible to add "should be backported" as a check-box on PRs, automating the discovery part.

globin:
Closing for now due to no activity, after communicating with @samueldr.

PR #31, opened 2018-08-23
[RFC 0031] Support ppc64(le) architecture

CrystalGamma:
In nixpkgs#45474, @edolstra made me aware that this warrants an RFC.
@dtzWill and @Ericson2314 had displayed interest earlier in nixpkgs#45340 (maybe one of you is interested in co-authoring?).

[rendered view](https://github.com/CrystalGamma/rfcs/blob/master/rfcs/0030-ppc64-platforms.md)

### DIFF
--- a/rfcs/0031-ppc64-platforms.md
+++ b/rfcs/0031-ppc64-platforms.md
@@ -1,0 +1,86 @@
+---
+feature: ppc64-platforms
+start-date: 2018-08-23
+author: CrystalGamma
+co-authors: (find a buddy later to help out with the RFC)
+related-issues: https://github.com/NixOS/nixpkgs/pull/45340 https://github.com/NixOS/nixpkgs/pull/45472 https://github.com/NixOS/nixpkgs/pull/45474 https://github.com/NixOS/nixpkgs/pull/45475
+---
+
+# Summary
+[summary]: #summary
+
+Bring support for ppc64 platforms, initially especially ppc64le, to nixpkgs/NixOS.
+
+# Motivation
+[motivation]: #motivation
+
+NixOS/nixpkgs has long prided itself to be a general-purpose OS/package repository, allowing its packages to be used across a broad range of systems, and providing more packages than ever.
+
+With alternative architectures such as OpenPOWER and RISC-V gaining traction, now is the time to add support for those architectures.
+Specifically motivating this request is the OpenPOWER, or powerpc64le, architecture—and its embodiment in the POWER series of microprocessors from IBM.
+Fully open source commercially available systems have recently become widely available for this platform and there are active Nix users who desire support.
+
+# Detailed design
+[design]: #detailed-design
+
+The purpose of this RFC is to create consensus for general acceptance of ppc64(le) related pull requests (as long as they are technically reasonable).
+A general commitment for contributions to not break ppc64(le) is not necessary as long as the architecture stays in only niche use.
+This especially means that building a Continuous Integration infrastructure for ppc64le is not required in the near future.
+
+Bringing support for ppc64(le) platforms will be done as follows:
+
+The first step for implementing support is extending lib/ to recognize powerpc64(le) as a CPU architecture and allowing bootstrap files for the architecture to be (cross-)built and a stdenv based on those bootstrap files to be built.
+This has already been merged in [nixpkgs#45340](https://github.com/NixOS/nixpkgs/pull/45340).
+
+The author of this RFC is already running a NixOS-based system on his POWER9-based computer, so for a basic system without graphical interface, only a handful of packages are known to require changes to build a working system:
+
+* mesa: currently assumes that any non-ARM system is x86. Requires restructuring the driver selection, but to the author's knowledge (since no actual graphical software was tested yet) it should not require any changes to the actual build description. ([nixpkgs#45474](https://github.com/NixOS/nixpkgs/pull/45474))
+* strace: also assumes that any non-ARM system is x86 and so tries to use the m32 personality. Requires checking the available machine personalities for all non-x86 architectures. ([nixpkgs#45472](https://github.com/NixOS/nixpkgs/pull/45472), already merged)
+* TeX Live: builds LuaJIT as part of mfluajit, which doesn't work because LuaJIT doesn't support 64-bit Power architectures yet; seems to be possible to disable on that platform without too much impact. ([nixpkgs#45475](https://github.com/NixOS/nixpkgs/pull/45475))
+* Bootloader tooling:
+    * the internal option *system.boot.loader.kernelFile* is currently set to the kernelTarget attribute of the host platform definition.
+      PowerPC kernels however use `zImage` as build target, yet produce a `vmlinux` file.
+      A one-line change would allow the option to default to the host platform definition's kernelFile attribute, if set.
+    * A typical bootloader for OpenPOWER systems is Petitboot. It can read a variety of bootloader configuration files, including those of GRUB 2 and SYSLINUX.
+      Using GRUB, even with *boot.loader.grub.device* = "nodev", as it is, implies building GRUB for the host architecture, which fails for ppc64le (currently because it requires soft-float).
+
+      However, the generic-extlinux-compatible bootloader module can be modified to provide options to store the configuration files in a path that Petitboot will consider and use absolute paths starting from the filesystem root, so that Petitboot can load the kernel/initrd images.
+
+Beyond a headless system, the biggest obstacle for ppc64(le) support seems to be Rust, because a lot of desktop software has it as a (direct or transitive) build dependency.
+In particular, a lot of GNOME software has a dependency on librsvg (which has parts written in Rust) via their documentation tools.
+The author of this RFC plans to create a way to build bootstrapping binaries for Rust on ppc64le (either via cross-building from x86 or from source via mrustc).
+
+Some packages, like OpenBLAS, will need target descriptions for these platforms, but many don't require changes to the actual build process.
+
+Support for Big Endian ppc64 might require changes in a few packages, though the majority of changes are assumed to be upstreamable (see [Unresolved Questions][unresolved]).
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Supporting another architecture might introduce additional maintenance overhead for nixpkgs.
+Generally this is restricted to packages that ship (JIT or AOT) compilers or do some feature selection based on architectures.
+
+If a new Rust bootstrap is created for ppc64le, as long as it isn't adopted for the other platforms, it will be a second rustc bootstrap that may cause maintenance effort.
+
+If a build infrastructure of ppc64le systems should be established, it will cause maintenance effort as well as costs for either procurement of hardware, or use of public cloud services (e. g. IntegriCloud), unless an organisation is willing to sponsor build machines. 
+
+# Alternatives
+[alternatives]: #alternatives
+
+For ppc64(le) in general:
+
+* keep the already-merged stdenv/bootstrapping code, maintain package changes as an overlay
+* revert stdenv/bootstrapping support, maintain ppc64le as a nixpkgs fork
+* not have ppc64le support in any NixOS-based distribution (barely an alternative)
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Should the nixpkgs/NixOS project host bootstrap files for ppc64le?
+
+Should Big Endian be in-scope for this RFC?
+
+# Future work
+[future]: #future-work
+
+If Big Endian support on ppc64 is decided to be out of scope for this RFC, it might be of interest for a future project.


justinlynn:
I am willing to donate significant on-going build-farm resources on POWER9 hardware (multiple Raptor CS dual 18-core Talos 2 machines with around 128GiB of RAM each) to this effort.

dtzWill:
Not sure if needed, but I'd be happy to host bootstrap tarballs, assuming they are known to at least somewhat work ;).  This would match how the musl tarballs are provided today.... for better or for worse. Just bug me if that'd be useful and I'll add it to the list! But hosting is easy these days but thought I'd offer :).

Can't test things really (other than QEMU haha) but generally in favor of this!

I don't think adding another URL to the rust bootstrap expression is a big deal, but I suppose someone will need to generate them.  I take it upstream doesn't do so already?

Do you know the story re:GHC?

rfcs/0030-ppc64-platforms.md
```diff
@@ -0,0 +1,85 @@
...
+
+* mesa: currently assumes that any non-ARM system is x86. Requires restructuring the driver selection, but to the author's knowledge (since no actual graphical software was tested yet) it should not require any changes to the actual build description. ([nixpkgs#45474](https://github.com/NixOS/nixpkgs/pull/45474))
+* strace: also assumes that any non-ARM system is x86 and so tries to use the m32 personality. Requires checking the available machine personalities for all non-x86 architectures. ([nixpkgs#45472](https://github.com/NixOS/nixpkgs/pull/45472))
+* TeX Live: builds LuaJIT as part of mfluajit, which doesn't work because LuaJIT doesn't support 64-bit Power architectures yet; seems to be possible to disable on that platform without too much impact. ([nixpkgs#45475](https://github.com/NixOS/nixpkgs/pull/45475))
+* OpenBLAS: pulled in by LVM. Needs a target description for ppc64le, but no changes to the build itself. <!-- I ran into a problem with a broken test a while ago, but it seems to be fine right now -->
```
7c6f434c: OpenBLAS needed by LVM sounds very strange. LLVM maybe? It doesn't seem to be so on x86_64, but I wouldn't be surprised if POWER is different enough.

CrystalGamma: I think it is at least (transitively). It was definitely pulled in by something related to file systems.

CrystalGamma: Strange. It seems it's not required anymore for the system derivation (maybe the dependency was broken somewhere along the line since last month?). Should I remove the bullet point then?

7c6f434c:
re: ofBorg and POWER9: ppc64 will run Linux with normal namespace-based sandboxing, so trusted status is not needed. Unless there are some problems with Linux namespaces on these architectures…

rfcs/0030-ppc64-platforms.md
7c6f434c: Well, if you want scientific calculations (Sage, Julia…) you will need OpenBLAS soon, but you might want to deprioritise it towards the end of the list.

I am not a ppc64 user (neither current nor near future), so I am not in position to give advice on priorities.

CrystalGamma:
@dtzWill I haven't looked into Haskell myself, but looking at @justinlynn 's comment it seems to be manageable.

Re: Rust, I know upstream are running some builds for ppc64(le) … (whether they can be used as bootstrap, I don't know)

Generally I'd really like for Rust to be built from source, but maybe that should be a separate RFC?

Ekleog:
@7c6f434c I was meaning, as they're all connected to the same AMQP, maybe a builder can send commands to another builder, and then justinlynn could send commands to LnL7's builder, which would require LnL7 to trust justinlynn :)

@CrystalGamma I think that, whether or not it deserves a RFC, building Rust from source is a matter that should not be discussed in this RFC, as RFCs that are not completely consensual have a high likelihood not to go anywhere :)

rfcs/0030-ppc64-platforms.md
CrystalGamma: OK, I guess I will put it after Rust, as something like "Some packages, like OpenBLAS, will need target descriptions for these platforms, but generally don't require changes to the actual build process."

CrystalGamma:
@Ekleog so remove that sentence, too? I'm not too keen on spending a lot of time on a solution (binary bootstrap) I don't want to use, though … :(

grahamc:
I'm not sure we should support POWER9 at this time. Nixpkgs and NixOS are being drawn in a large number of directions, and I think we might should focus on more pressing issues (memory use, existing architectures we do support) before expanding its scope.

hlandau:
@grahamc I think a distinction needs to be made between official support and agreeing to merge patches that provide unofficial support.

Certainly, it would be wholly premature for NixOS as an entity to provide official builds or support for ppc64le. There are other architectures which can currently be built but for which official builds aren't provided. OTOH, patches which make ppc64le should be merged if people submit them. There are people willing to do the work here; moreover, the rate of change of nixpkgs is fast enough that trying to maintain these patches out-of-tree has caused difficulty.

IMO, adding ppc64le support to the extent that persons interested can build it themselves is unlikely to negatively affect people working on nixpkgs who don't care about ppc64le support, or, to the extent that it does affect those people, it will be in a way that enhances the ability of nixpkgs to target multiple platforms and architectures generally.

Moreover, there doesn't need to be a commitment *not to break* ppc64le (e.g. someone who doesn't test on ppc64le merges a patch that breaks it) at this stage; it could be organised as a best-effort initiative by people interested in ppc64le, with no requirement that patches not break ppc64le and subsequent fixes for ppc64le maintained by people interested in it.

What I think people interested in a ppc64le port need at this time is a willingness by nixpkgs to merge ppc64le patches, not to make any sort of official commitment or impose CI requirements. This RFC was specifically created because it was requested in response to such a PR; the ability to get such PRs merged is all that's really necessary right now. Discussing anything else seems premature.

7c6f434c:
I think @grahamc makes a good point that in the current state of affairs ofBorg support for POWER9 is indeed increasing annoyance and premature, so the discussion of ofborg in context of POWER9 should be delayed for a long time.

(I do not agree with the point about build errors on inaccessible architectures as long as there is Darwin support — in the case of Darwin I actually don't know how to run a build locally, in the case of Aarch64/POWER9/RISC-V I only need someone to publish a bootable Nix-containing Linux image for QEMU)

Specifying that there is no expectation to avoid breaking `ppc64le` support of packages even when it is already present **should** probably be a part of this first-stage POWER9 RFC.

Re: RAM requirements — the work is not fully fungible in that case, POWER9 support seems to attract new contributors and people already working on cross-builds and multi-target support. RAM usage requires (probably) something about NixOS module system. Not everyone risks to touch the module system, and one needs quite a bit of reputation to get the necessary reviewer efforts from the few and busy people who are expected to review fundamental changes in the module system.

(disclaimer: I gave up on NixOS but stay fully dependent on Nixpkgs, so my interest is about Nixpkgs being nice, flexible, and preferably not shattered into a dozen of forks with various subsets of fixes)

7c6f434c:
@CrystalGamma re: Rust: I think any solution would include you publishing a Rust binary suitable for bootstrapping on POWER9, be it via cross-bootstrapping or `mrustc`, be it as the only way to bootstrap or as a proof that the toolchain works. So if you promise to work on building this specific deliverable, it would be good enough for the scope of the current RFC. Whether your work will result in something that can be used on other platforms is not important for this RFC.

dezgeg:
I don't personally see having POWER support in nixpkgs having a big effect on maintainability given it should be a mature architecture by now (though maybe not as popular as say, aarch64). I expect it will need way less amounts of patches/tweaks than RISC-V, and perhaps even less than aarch64.

Regarding rust, according to https://forge.rust-lang.org/platform-support.html ppc64le (unless I mixed this up with some other architecture) is in the same support tier as both aarch64 and armv7 which means they do provide prebuilt compilers, so if all goes well getting it to work is like 5 lines of changes, similar to what was done in https://github.com/NixOS/nixpkgs/commit/222f905fd9042f80d4cdef65bb7d2801fdb63429

zimbatm:
My prediction for the future: this RFC will go the same route as https://github.com/NixOS/rfcs/pull/23 where the RFC gets stalled and then merged because some people will have added POWER9 support into nixpkgs. I don't know if it's good or bad but it reflects the anarchist approach our community has to software development. In the absence of clear leadership and direction the next best thing to do is probably to nurture the ground for independent activities to happen.

timokau:
I think that reflects our current RFC process until https://github.com/NixOS/rfcs/pull/18 moves forward: If there are absolutely no disapprovals after some time, just do it. If there are any disapprovals, the RFC is essentially doomed for eternal limbo.

Regarding this RFC I think "just do it" is fine: As long as any changes are clearly commented with explanations about their purpose and we don't make any commitments, it would be easy to remove support specific workarounds again if momentum dies. I don't see any major cost to this.

ryantm:
@CrystalGamma It looks like this should get renamed to RFC 31

CrystalGamma:
@ryantm I suppose. The previous PR didn't carry a number in the title, so I didn't think it would be a new RFC.

globin:
Per meeting of the @NixOS/rfc-steering-committee this RFC is open for shepherding nominations.

CrystalGamma:
OK. Anything I as author need to do (except reading/updating the RFC again, since it's been a while)?

shlevy:
RFC steering committee nominates @grahamc, @ryantm, @Ericson2314, and @zimbatm as shepherds, as they should cover the interests of ofborg, hydra, stdenv, and general package maintainership well.

Ultimately this is up to the shepherd team and the RFC author, but we recommend shifting this RFC to lay out the general criteria/process for accepting a new architecture as supported (or removing one from the supported list).

Please let us know if you accept!

ryantm:
I accept the nomination.

zimbatm:
happy to help!

zimbatm:
@shlevy did you get any reply from @grahamc or @Ericson2314 ?

rfcs/0031-ppc64-platforms.md
```diff
@@ -0,0 +1,86 @@
...
+[design]: #detailed-design
+
+The purpose of this RFC is to create consensus for general acceptance of ppc64(le) related pull requests (as long as they are technically reasonable).
+A general commitment for contributions to not break ppc64(le) is not necessary as long as the architecture stays in only niche use.
+This especially means that building a Continuous Integration infrastructure for ppc64le is not required in the near future.
+
+Bringing support for ppc64(le) platforms will be done as follows:
+
+The first step for implementing support is extending lib/ to recognize powerpc64(le) as a CPU architecture and allowing bootstrap files for the architecture to be (cross-)built and a stdenv based on those bootstrap files to be built.
+This has already been merged in [nixpkgs#45340](https://github.com/NixOS/nixpkgs/pull/45340).
```
zimbatm: The more architectures we support, the harder it's going to be to upgrade gcc and other heavily-patched packages. Maintainers don't have access to the more esoteric platforms.

How do you propose that patches be handled if for example gcc has to be upgraded quickly for security-related reasons?

zimbatm: Maybe lib/systems/platforms.nix should be amended with an extra "officiallySupported" flag.

rfcs/0031-ppc64-platforms.md
```diff
@@ -0,0 +1,86 @@
...
+
+      However, the generic-extlinux-compatible bootloader module can be modified to provide options to store the configuration files in a path that Petitboot will consider and use absolute paths starting from the filesystem root, so that Petitboot can load the kernel/initrd images.
+
+Beyond a headless system, the biggest obstacle for ppc64(le) support seems to be Rust, because a lot of desktop software has it as a (direct or transitive) build dependency.
+In particular, a lot of GNOME software has a dependency on librsvg (which has parts written in Rust) via their documentation tools.
+The author of this RFC plans to create a way to build bootstrapping binaries for Rust on ppc64le (either via cross-building from x86 or from source via mrustc).
+
+Some packages, like OpenBLAS, will need target descriptions for these platforms, but many don't require changes to the actual build process.
+
+Support for Big Endian ppc64 might require changes in a few packages, though the majority of changes are assumed to be upstreamable (see [Unresolved Questions][unresolved]).
```
zimbatm: How about keeping track of this progress in the nixos.wiki ?

While these information give an idea of how much work is left to do, the RFC should also cover other aspects of adding new platforms like the added cost of nix evaluation and maintenance overhead that it creates inside of nixpkgs. I would like to see things addressed like:

* what are the rules of package maintainers in interaction with the non-official platforms?
* where is the list of official and unofficial platforms listed?
* what are the rules for adding and dropping a platform?
* for non-official platforms, are they keeping their own build farm? if yes, how can this be listed?
* where should the stdenv bootstrap files be stored?

zimbatm: One thing that I would like to see explored, is if it's possible to store the stdenv out of the main nixpkgs tree, along with the binary cache and all the platform-specific information. Using the @nix-community organisation would be OK for that type of activity if you want.

rfcs/0031-ppc64-platforms.md
```diff
@@ -0,0 +1,86 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Supporting another architecture might introduce additional maintenance overhead for nixpkgs.
+Generally this is restricted to packages that ship (JIT or AOT) compilers or do some feature selection based on architectures.
+
+If a new Rust bootstrap is created for ppc64le, as long as it isn't adopted for the other platforms, it will be a second rustc bootstrap that may cause maintenance effort.
+
+If a build infrastructure of ppc64le systems should be established, it will cause maintenance effort as well as costs for either procurement of hardware, or use of public cloud services (e. g. IntegriCloud), unless an organisation is willing to sponsor build machines.
```
zimbatm: Did you try and talk to IntegriCloud if they are willing to sponsor hardware for open source projects? I can arrange for the NixOS Foundation to talk to them if that helps.

One thing that you have to gage is how much people are interested in using this architecture, and how much effort you are willing to put into that project. This is really on your side I'm afraid, but it would be good to know. Obviously if the port is successful it might attract more users in the future and this hard to foresee. Do you already have a sense of today's user base that you would have?

rfcs/0031-ppc64-platforms.md
```diff
@@ -0,0 +1,86 @@
...
+For ppc64(le) in general:
+
+* keep the already-merged stdenv/bootstrapping code, maintain package changes as an overlay
+* revert stdenv/bootstrapping support, maintain ppc64le as a nixpkgs fork
+* not have ppc64le support in any NixOS-based distribution (barely an alternative)
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Should the nixpkgs/NixOS project host bootstrap files for ppc64le?
```
zimbatm: The informal NixOS hardware team can provide some file hosting if necessary but it can't look official.

rfcs/0031-ppc64-platforms.md
```diff
@@ -0,0 +1,86 @@
...
+
+Should Big Endian be in-scope for this RFC?
```
zimbatm: Probably not, one thing at a time.

rfcs/0031-ppc64-platforms.md
nh2: > One thing that I would like to see explored, is if it's possible to store the stdenv out of the main nixpkgs tree

Would this make Continuous Integration less easy?

One thing I'd definitely want to have as a maintainer is swift feedback on whether changes I make break things for some architecture.

zimbatm: how would you be notified that your change broke an architecture that is not checked by ofborg and the main hydra?

let's say I am the maintainer of POWER, I can:
* add power to the main nixpkgs architectures
* create my own nixpkgs-power repo. that repo contains the stdenv, overrides and power-specific packages
* create my own hydra instance that runs whenever nixpkgs master or nixpkgs-power changes => and publishes my binary cache

grahamc:
I should have posted this ages ago, but yes I support the nomination to be part of the shepherd team :) however, I can't commit to be its leader at this time.

ryantm:
I like the idea of Official vs Unofficial. Then people could start experimentally adding support for architectures without affecting other people as much. Based on @zimbatm's feedback, it seems like he wants to go in the direction of making this RFC about supported architectures generally instead of specifically ppc64. @CrystalGamma Are you willing to accept this change of scope on your RFC?

7c6f434c:
>I like the idea of Official vs Unofficial. Then people could start experimentally adding support for architectures without affecting other people as much. Based on @zimbatm's feedback, it seems like he wants to go in the direction of making this RFC about supported architectures generally instead of specifically ppc64. @CrystalGamma Are you willing to accept this change of scope on your RFC?

Just in case — I tried to summarise the questions to answer for proper support tiers and there are a lot of details we might want to fill in for a full RFC (I currently maintain a list of concerns at https://gist.github.com/7c6f434c/431c8233793571cad313e77356d2f7db in a gist).

I am not sure we want to pivot this RFC in a true and full solution to support tiers; maybe just say that the support is unofficial, no testing promises are made, and for the initial time ppc64-specific fixes are accepted either when they are generic cleanups with potential to benefit unrelated minority platforms, or when they are small enough and change a package that naturally depends on architecture (like compiler having a list of compilation target names).

CrystalGamma:
@ryantm I would be OK with 'retargeting' this RFC to be more about how to deal with new platforms/architectures in general. That said, I don't have sufficient maintenance experience to actually make any useful technical suggestions.

zimbatm:
a few things I suggest can be done independently from this RFC:
* create a unofficial architectures wiki page on the unofficial https://nixos.wiki . This page can include documentation on how to create new targets
* create a pack for the POWER arch with a list of interested parties and link to PRs and binary cache
* create a PR to add a "supported" flag to all supported architectures

7c6f434c:
>* create a PR to add a "supported" flag to all supported architectures

I am afraid adding a «supported» flag would require determining what «supported» means — and I think we already have some sliding scale, just without codifying it.

zimbatm:
exactly, that's why it's a good idea to add the flag and force us to clarify things. The most supported platform is "x86_64-linux" and we can start with that.

zimbatm:
pkgs/top-level/release.nix says:
```
supportedSystems ? [ "x86_64-linux" "x86_64-darwin" "aarch64-linux" ] 
```

7c6f434c:
Right, so the flag is already there but definition is not!

(On the other hand, as I mentioned in my gist, the system name doesn't fully describe the platform — we have code for `x86_64-linux` with `musl`, and it is arguably a platform, but not a fully supported platform)

domenkozar:
@NixOS/rfc-steering-committee is ready to move this forward if @Ericson2314 thumbs up means we have the following shepherd team:  @ryantm @zimbatm @grahamc @Ericson2314

Ericson2314:
Yeah I can shepherd.

domenkozar:
Who would like to be the leader (so far, Graham declined)?

shlevy:
@ryantm @Ericson2314 Are either of you available to lead the shepherd team?

ryantm:
@shlevy I think @Ericson2314 would be more appropriate given his knowledge of cross-compilation. But if he isn't up for it, I can try to do it.

Ericson2314:
Sure I can do that. Hopefully I'll learn more about the process first hand with https://github.com/NixOS/rfcs/pull/13 meeting soon.

edolstra:
@ryantm @zimbatm @grahamc @Ericson2314 We just discussed in the RFC steering committee whether this RFC should be kept open, or whether it makes sense to close it in light of RFC #46 (Platform Support Tiers), which addresses the more general issue of platform support. What do you think?

Ericson2314:
Yes. I approved of the early plan of changing this to be like #46 had that one not been opened anyways.

CrystalGamma:
I second the proposal to close this PR in favor of #46 because

* discussing levels of platform support in general will yield better criteria that can be applied for future platforms.
* I am not the author this RFC needs because I am not familiar enough with the Nix(OS) community to make good proposals regarding maintenance and infrastructure.

zimbatm:
:+1: for superseding this RFC with #46

ryantm:
Yes, let's close this.

PR #32, opened 2018-08-29
[RFC 0032] Phase running changes for better nix-shell use

dezgeg:
Making some minor tweaks to how phases are run in the stdenv to improve the UX of nix-shell.

Rendered view: https://github.com/dezgeg/rfcs/blob/phase-changes/rfcs/0032-run-phase-changes-for-better-nix-shell-use.md

### DIFF
--- a/rfcs/0032-run-phase-changes-for-better-nix-shell-use.md
+++ b/rfcs/0032-run-phase-changes-for-better-nix-shell-use.md
@@ -4,6 +4,8 @@ start-date: 2018-08-26
 author: @dezgeg
 co-authors: TBD
 related-issues: https://github.com/dezgeg/nixpkgs/tree/better-run-phases
+shepherd-team: Samuel Dionne-Reil, John Ericson, Linus Heckemann
+shepherd-leader: Linus Heckemann
 ---
 
 # Summary
--- a/rfcs/0032-run-phase-changes-for-better-nix-shell-use.md
+++ b/rfcs/0032-run-phase-changes-for-better-nix-shell-use.md
@@ -9,7 +9,9 @@ related-issues: https://github.com/dezgeg/nixpkgs/tree/better-run-phases
 # Summary
 [summary]: #summary
 
-Making some minor tweaks to how phases are run in the stdenv to improve the UX of nix-shell.
+The intent of this proposal is to tweak a bit how the stdenv runs phases during a build to achieve these two goals:
+1. Improve the UX of nix-shell by making it easier to manually run phases inside a nix-shell
+2. Improve the consistency of pre/post hooks by always running such hooks instead of only running them in non-overridden phases
 
 # Motivation
 [motivation]: #motivation
--- a/rfcs/0032-run-phase-changes-for-better-nix-shell-use.md
+++ b/rfcs/0032-run-phase-changes-for-better-nix-shell-use.md
@@ -1,8 +1,8 @@
 ---
 feature: run-phase-changes-for-better-nix-shell-use
 start-date: 2018-08-26
-author: @dezgeg
-co-authors: TBD
+author: @globin
+co-authors: @dezgeg
 related-issues: https://github.com/dezgeg/nixpkgs/tree/better-run-phases
 shepherd-team: Samuel Dionne-Reil, John Ericson, Linus Heckemann
 shepherd-leader: Linus Heckemann
--- a/rfcs/0032-run-phase-changes-for-better-nix-shell-use.md
+++ b/rfcs/0032-run-phase-changes-for-better-nix-shell-use.md
@@ -17,21 +17,27 @@ Making some minor tweaks to how phases are run in the stdenv to improve the UX o
 The primary goal of this RFC is to make it easier to run build phases manually inside a nix-shell.
 Currently, it's a bit annoying because to run say, `buildPhase` the only invocation that works correctly in all cases is:
 
-````
+````sh
 eval "${buildPhase:-buildPhase}"
 ````
 
 The goal is to be able to replace the above with the obvious:
-````
+````sh
 buildPhase
 ````
 
 The secondary goal is to change how hooks (like `preBuild`, `postBuild` etc.) interact with overridden phases.
 For example a derivation `foo-package` doing,
-````
+````nix
+stdenv.mkDerivation {
+    # ...
+    
     buildPhase = ''
         ./my-funky-build-script.sh
     '';
+    
+    # ...
+}
 ````
 causes `preBuild` and `postBuild` not to be called anymore.
 In 99% of the cases this isn't a problem, but it can cause hidden annoyances when using `.overrideAttrs`, for instance:
@@ -50,7 +56,7 @@ Thus, to counter this inconsistency, this RFC proposes that those hooks will be 
 [design]: #detailed-design
 
 All the 'phase' functions in Nixpkgs need to be reworked a bit. Conceptually, the following diff will be applied to each of them:
-````
+````diff
 -buildPhase() {
 +defaultBuildPhase() {
 -    runHook preBuild
@@ -84,7 +90,7 @@ and the function responsible for the default implementation is now renamed to `d
 Then, the `runHook` calls are pulled up from the default phase implementation to the new `buildPhase` function itself.
 
 The actual logic is abstracted to helper function I've named `commonPhaseImpl` (bikeshedding on the name welcome). Thus the implementation of `buildPhase` presented above will be this one-liner:
-````
+````sh
 buildPhase() {
     commonPhaseImpl buildPhase --default defaultBuildPhase --pre-hook preBuild --post-hook postBuild
 }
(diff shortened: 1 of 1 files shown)


samueldr:
Formatting tip with github-flavoured markdown:

 ### Highlighting

Given a backticks-guarded block of code <tt>```</tt> you can [append a language name to the first set of backticks](https://help.github.com/articles/creating-and-highlighting-code-blocks/); giving it e.g. [`diff`](https://github.com/github/linguist/blob/7ff508e0cfec4086852b39eb59ec79d87be1d196/lib/linguist/languages.yml#L1051) to highlight a patch.

#### `nix` addendum

Parts of attrsets will want their braces.

```nix
  foo = "bar";
```

```nix
{
  foo = "bar";
}
```

<sup>I will collapse this comment once acknowledged with either changes or a reaction</sup>

samueldr:
>  given that the issue being fixed is a pure UX issue, not fixing it wouldn't prevent any other work from happening.

Good UX is important in *everything*. Anything that is clumsy will be fumbled with, and possibly be hated by the end-user, that's not a Nix-specific statement! Don't sell the desire for improvement short.

 * * *

Is there anything preventing your `runPhase` alternative to be implemented right now? If it were implemented, (I think) it could be also made compatible with the implemented RFC. This would already be a net increase in UX over running `eval ...`. The only drawback I see is that using `runPhase` would clobber the `runPhase` name if one would implement a `runPhase` as a phase.

 * * *

As for the RFC itself, other than the generic baseline deprecation concerns, the idea seems sound enough. In reality the "new way" isn't that much of a hassle compared to the status quo. This mostly affects the implementation for *a default phase*; the user overriding a phase won't need to bother, right?.

As a new~ish user and developer running hooks also makes more sense than clobbering the `pre` parts by setting a custom phase. That is, in addition to making it easier to run phases properly.

dezgeg:
> Is there anything preventing your runPhase alternative to be implemented right now? 

No, it could be added straight away. But I am kind of hopeful it won't be needed.

> In reality the "new way" isn't that much of a hassle compared to the status quo. This mostly affects the implementation for a default phase; the user overriding a phase won't need to bother, right?.

Correct. You can see the changes needed in this commit: https://github.com/dezgeg/nixpkgs/commit/87959f3d98a55b8c86de283d84c7c110ee3f09ea

dezgeg:
> As more future work (out of scope of this RFC, which is good as-is):

I actually experimented with implementing such tool at some point: https://github.com/dezgeg/nix-debug-shell but I am not particularly good with things like documentation or finishing stuff that I've started, so it's in a kind of forgotten state and potentially bitrotted by now. Maybe it could be moved to https://github.com/nix-community/ if there are people willing to work on it.

Ericson2314:
Nice proposal. I definitely want this.

https://github.com/ryantrinkle/nix-build-inplace is another such debug build helper.

FRidh:
For Python (`mk-python-derivation.nix`) we need this
```
    buildPhase = ''
        buildPhase
        (cd foo; buildPhase)
        (cd bar; buildPhase)
    '';
```
I'm glad to see this is covered.

This RFC looks good to me, though I think the Summary can be improved. Running the hooks independent of the phases has nothing to do with the UX of `nix-shell` usage and so that change should already be mentioned there.

dezgeg:
> For Python (mk-python-derivation.nix) we need this

@FRidh do you have a pointer on where is there is such code in Python infra that need changes? I have found (and fixed) some such cases in my branch but none in Python.

> [...] I think the Summary can be improved. Running the hooks independent of the phases has nothing to do with the UX of nix-shell usage and so that change should already be mentioned there.

Yeah, the hook changes are an afterthought that came up after I started writing the description... I will reword it a bit.

Ericson2314:
This is probably off topic, but I'd like if we made the `findInputs` dependency stuff in a phase, which like `preHook` was run by `nix-shell` by default for backwards compat. It seems more consistent to do all "real work" in phases.

FRidh:
@dezgeg `fixupPhase` in `pkgs/development/interpreters/python/mk-python-derivation.nix`

dezgeg:
> Actually, reading it again it is a bit different.

Yep, it's a different case. The sort of Nix-level code in `mk-python-derivation.nix` doesn't require any changes. Changes are only needed if inside a `fixupPhase = '' ... '';` block (or equivalently, `postFixup`) there is a call to the Bash function `fixupPhase`.

FRidh:
@dezgeg do you have a branch with this change? I would like to try it out, especially because of https://github.com/NixOS/nixpkgs/pull/45898/commits/9333c1bd0ea30faa5f7fcb904e6883b3fe571eb4

dezgeg:
The branch is mentioned under related-issues: https://github.com/dezgeg/nixpkgs/tree/better-run-phases

timokau:
Since there is nobody to decide and there hasn't been any opposition, should we merge this?

zimbatm:
It seems like a good target for 19.03. How much pain will it create for out-of-tree nix code?

lheckemann:
I usually use `runHook buildPhase`, is there anything wrong with that?

shlevy:
@dezgeg Are you ready for us to open nominations for a shepherd team as per the [new RFC process](https://github.com/NixOS/rfcs/tree/d9aad9291186cc10375afa5aaefed2b01357cf7e#process-from-creation-to-merge)?

shlevy:
Closing for now. Feel free to reopen when this is ready for general review and shepherding.

globin:
As there has been active discussion on this I'd like to revive this (maybe included in the structured attrs work, maybe seperately, depending on what the rest of the community prefers). @dezgeg would you want to move this forward? If not does anyone else want to jump on this? Otherwise I'd be open to try and carry this on.

globin:
This is open for shepherding nominations!

@FRidh @Ericson2314 @samueldr would you be interested?

Ericson2314:
You bet!

FRidh:
Thanks, but I don't have time to shepherd this RFC.

domenkozar:
ping @samueldr

lheckemann:
I nominate myself to co-shepherd this RFC.

Mic92:
Do you want to shepherd this one? @matthewbauer, @timokau

timokau:
Unfortunately my todo list is overflowing right now, so I won't be taking on a new responsibility.

samueldr:
I must have missed the notifications in the background noise.

I would be interested in shepherding.

Mic92:
We decided on @lheckemann as a shepherd leader on this RFC.

Mic92:
@lheckemann Do you think this RFC should have a meeting?

lheckemann:
@dezgeg, @globin has offered to take over primary authorship of the RFC in case you don't currently have the capacity to continue. If you have any objections, let us know :) otherwise we'll proceed with him as author.

Mic92:
@lheckemann Do you still plan to make a meeting for this?

lheckemann:
So, I've finally arranged the meeting with @Ericson2314 and @samueldr. Results:

- This RFC looks good generally
- Unresolved question "This needs some testing to see if this works with the ancient Darwin Bash." — this should, as far as we know, not be a problem: the darwin bootstrap uses a more recent bash from the bootstrap tarball, not the macOS system bash; and nix-shell uses `bashInteractive` from `<nixpkgs>`, which is also recent enough for this not to be a concern.
- Unresolved question regarding elegance of the way that backwards compatibility is ensured: we consider this an implementation detail which should not be an issue for the RFC in general.
- Some general discussion on backwards-incompatible changes to stdenv. We decided not to bike-shed this RFC on that subject :)
- We're not sure who will implement this. Both @dezgeg and @globin haven't been very active recently and may not have the capacity/time currently.
- The last concern notwithstanding, we decided to **call the Final Comment Period on this RFC, with disposition to accept**. Speak within the next 10 days, or forever hold your peace!
- Implementation need not follow immediately, but we believe accepting this RFC is helpful to indicate the community's general acceptance of the idea.

rfcs/0032-run-phase-changes-for-better-nix-shell-use.md
```diff
@@ -0,0 +1,180 @@
...
++
++    if [ -n "$buildPhase" ]; then
++        eval "$buildPhase"
++    else
++        defaultBuildPhase
++    fi
++
++    runHook postBuild
++}
+
```
roberth, line 90: Has this not been implemented yet, or implemented differently?

Ericson2314: Haven't seen @dezgeg in a while, don't think this ever got implemented.

roberth: Fwiw, https://github.com/NixOS/nixpkgs/issues/175605 suggests a perhaps less breaking solution that's opt-in. Opting in is in itself messy though, so that's a reason to prefer this RFC if feasible.

PR #35, opened 2018-10-02
[RFC 0035] Default `name` from `pname`

Synthetica9:
[Rendered](https://github.com/Synthetica9/rfcs/blob/default_name_pname/rfcs/0035-default-name-from-pname.md)

Currently looking for a co-author


FRidh:
> I disagree, different source versions might require different steps to build, install, etc. I feel it would be strange if this wouldn't be reflected in the derivation.

The version attribute does not consider that, but it is of course considered in the computation of the store path hash.

To be clear, I am not saying version should not be included in the derivation name, because I think it is good to have it there for convenience from a user point of view given the current tooling and when looking at store paths directly. However, it does not really make sense to have it as an attribute of the build artifact, because its an attribute of one of its dependencies: its source derivation. Same goes for license and homepage. The reason for including these in the derivation is again convenience/tooling support.

Anyway, what I state here fits I think better in the discussion/proposal of writing derivations differently (lost the link).

Synthetica9:
Ah, sure, I can see that. However, this being such a simple RFC, I don't think it would do any good to drag in this big a change into it.

Ekleog:
> Care should be taken […] to add an assertion that the generated name is consistent with the actual name if all three attributes are present (implemented in e0d2348).

Do the whole of `nixpkgs` still evaluate with this added assertion? I would fear that some packages may have eg. `name = "${pname}-bis-${version}` or similar stuff.

Synthetica9:
@Ekleog Good point! The original PR eventually switched to `lib.strings.isSuffix` instead of `(==)`. This was due to the fact that Python packages defined their `name` as `"${libversion}-${pname}-${version}"`, for example: `"python2.7-setuptools-40.2.0"` (which is currently the first example of such a failing package). 

I see two options here:

- **Use the less common `baseName` instead of `pname`.** I tested this, this seems to work out of the box. However, it would require some more refactors, and it is inconsistent with the currently official practice of using `pname` for Python packages.
- **Use `hasSuffix`.** This is a less thorough consistency check, but seems to work as well. 

P.S.

I couldn't get the entirety of nixpkgs to evaluate even unmodified, when I do `nix eval "(import ./. {})"` I get `error: enum-0.4.4 not supported for interpreter python3.6m`. Any pointers?

Profpatsch:
Not really a fan, 700 places doesn’t look like a lot for a major source of implicitness (because yes, as you say, it’s going to be non-trivial to find out why the `name` is what it is, and where it is set.

Synthetica9:
> Not really a fan, 700 places doesn’t look like a lot for a major source of implicitness (because yes, as you say, it’s going to be non-trivial to find out why the `name` is what it is, and where it is set.

Keep in mind, there is precedent: `buildPythonPackage` already does this.

Ericson2314:
Also, I eventually want `pname` and `version` to be separate meta fields, and Nix itself to name derivations. This is a good step in that direction.

zimbatm:
`pkgs.bundlerApp` has a similar mechanism as well so it's not only python

Synthetica9:
> `pkgs.bundlerApp` has a similar mechanism as well so it's not only python

Does this also only add a prefix to the standard `pname-version` scheme?

timokau:
@zimbatm
> Not really a fan, 700 places doesn’t look like a lot for a major source of implicitness (because yes, as you say, it’s going to be non-trivial to find out why the `name` is what it is, and where it is set.

I personally prefer that style and would have used it more often but was discouraged from doing so because it was not yet a standard. So the 700 may be a low number. If I could go back in time I would even use `name` instead of `pname` and give a longer name to whats currently called `name` since its generally (but not always) an implementation detail.

Also I think the consistency with python is very valuable.

Synthetica9:
@edolstra @zimbatm @shlevy Looks like all the discussion that is going to happen on this has happened, could I get a verdict?

Profpatsch:
From what I read in the comments, the general opinion is in favor of this change. I’d say “go forth and implement”.

shlevy:
Go for it IMO

edolstra:
This PR doesn't really explain *why* this change should be made. Presumably it's to allow tools like `nix-env` to unambigously distinguish between the name and version of a package, but it doesn't discuss what the impact on such tools is.

Zimmi48:
AFAIU it's just about simplifying the writing of a derivation, that's all.

Synthetica9:
> This PR doesn't really explain why this change should be made. Presumably it's to allow tools like nix-env to unambigously distinguish between the name and version of a package, but it doesn't discuss what the impact on such tools is.

@Zimmi48 is correct, it is just a simplifying change to make it easier to write packages with less repeated "boilerplate".

Synthetica9:
> Go for it IMO

> From what I read in the comments, the general opinion is in favor of this change. I’d say “go forth and implement”.

@shlevy @Profpatsch https://github.com/NixOS/nixpkgs/pull/49398

dezgeg:
There are/would be other advantages besides simplifying expressions for derivations:

1. This approach won't break if a package's name contains a dash followed by a number. E.g. you currently can't have the `name` be something like `ntfs-3g-${version}` because that will be misinterpreted .
2. Often there's a need to attach some extra suffix to a package's `name`. For example the `tp_smapi` package (which is a kernel module) has:
     ````
     name = "tp_smapi-${version}-${kernel.version}";
     version = "0.43";
    ````
    and it's sometimes convenient to be able to know just the version and not the extra suffix.

It's worth noting that `lib.getVersion` from Nixpkgs already uses the `version` attribute if that's specified and only then falls back to `builtins.parseDrvName`

timokau:
After adding a new kernel module I just noticed that repology will be confused by the naming scheme @dezgeg mentions. It will list the kernel module version as the kernel version. The current PR wouldn't allow setting `pname`, `version` and `name`. Should we maybe change the kernel module naming scheme instead?

timokau:
Thats true, but it is how it is commonly done. Not sure who came up with it first and why. I'd support a naming scheme change.

matthewbauer:
I believe it was @fridh who said that version really shouldn't be a part of the derivation at all. It really should be part of "src".

I don't doubt that there are advantages to splitting name and version up, but I'm not sure if it outweighs the cost of churn here. The ```name = "hello-1.0``` idiom has been there for over 10 years (https://nixos.org/~eelco/pubs/nspfssd-lisa2004-final.pdf). I don't think there's a compelling enough reason to change it yet.

timokau:
I don't see any negatives here. What do you mean by churn?

7c6f434c:
The proposed change doesn't forbid direct name specification, it just reduces redundancy in case a better way (which is already also an idiom…) is used.

dtzWill:
Don't suppose this could work for things like override or even overrideAttrs? O:)

7c6f434c:
I think working with `override` is very unlikely to be broken (it happens before the call to `mkDerivation` anyway), am I missing something?

dtzWill:
Oh, I was thinking it's be "convenient" to only need to specify new
version/src when `overrideAttrs`'ing, with name auto-rederived from
pname and (new) version.  Not worth complicating or delaying for
but came to mind when adding an override earlier.
(since setting `name` again is a bit tedious boilerplate, NBD)

7c6f434c:
>Oh, I was thinking it's be "convenient" to only need to specify new
>version/src when `overrideAttrs`'ing, with name auto-rederived from
>pname and (new) version.  Not worth complicating or delaying for
>but came to mind when adding an override earlier.
>(since setting `name` again is a bit tedious boilerplate, NBD)

If `overrideAttrs` can handle cross-compilation, `pname` is easy, and
I don't see any point in it if it cannot…

edolstra:
Closing since this is already in Nixpkgs.

Ericson2314:
If something already exists matching the RFC (as opposed to conflicting), I rather see it merged. IMO this makes the outcome clearer at a glance, and is harmless because RFCs should specialize an idempotent course of action.

edolstra:
Merging it would imply that it was approved via the RFC process, which is not the case here.

Ericson2314:
Could the committee come to a decision retroactively? (keep vs revert)

7c6f434c:
@Ericson2314 I think the current status allows you some ways to promote `pname` usage without competing for the RFC processing bottleneck.

dtzWill:
Not to put too fine a point on the issue,
but I suspect it's not entirely regretted that there are
pain points on the "implement around the RFC process" workflow ;).
And more specifically, limited motivation for discussing/fixing such
pains.

(see the musl RFC for another example guilty of this :/:()

That said, I imagine there's room for positive discussion here!
This now seems more an issue of how to put together an RFC
that seeks to codify existing practices...
but I'm not sure how that's best accomplished offhand
(new seems simplest? dunno).

Which reminds me, I should go refresh myself with all our RFC's.... O:)

rkoe:
I'm not very happy with automagically adding variables in a non-obvious way.
It might save a few lines-of-code, but makes it less readable and so I doubt that it's worth it.
Explicit is better than implicit.

This RFC proposes that "name" should magically be added.
This would make package-definitons less readable/understandable, since then ``name`` exists with some value, but is not defined in the package-definition, and "comes out of nowhere".

In addition, ``pname`` isn't mentioned in the documentation at all; and this behaviour also isn't mentioned at all in the Nixpkgs-manual. So, if you want to go this way, PLEASE document it clearly in the Nixpkgs-manual, BEFORE promoting it.

zimbatm:
I noticed that overriding packages has become more complicated.  When pname + version is being used, overriding a package with overrideAttrs now requires to fix both the version and the name.

Mic92:
> I noticed that overriding packages has become more complicated. When pname + version is being used, overriding a package with overrideAttrs now requires to fix both the version and the name.

Can we automate this in `overrideAttrs` or is this too much magic?

gloaming:
Where a package provides `pname` and `version` but no `name`, `overrideAttrs` already works fine in my testing.

The only problem I can see is when a package sets an explicit `name` using those, because the interpolation happens without a fixpoint. I presumed this was rare, but now I see that all python  packages have it. It doesn't work for haskell packages either, even though they don't have a prefix.

> Can we automate this in overrideAttrs or is this too much magic?

What automation do you propose? String substituting the old and new `pname` and `version`? It seems like a kludge, but it's a very nice kludge.

Alternatively, we could allow the `name` to be specified as a function `{ pname, version, ...}: ...` but that seems like too much of a special case.

We could also require nixpkgs authors to fetch the `pname` and `version` from the fixpoint before using them in the name, but this doesn't help people writing packages outside of nixpkgs.

I guess the general solution would be for `stdenv.mkDerivation` to provide a fixpoint, and use `with self` instead of `rec`:

```nix
stdenv.mkDerivation (self: with self; {
  pname = "hello";
  version = "1.0";
  name = "prefix-${pname}-${version}";
}
```

Is there a reason this isn't done already? It does increase boilerplate.

PR #25, opened 2018-02-28
[RFC 0025] Nix Core Team

grahamc:
Introducing the Nix Core Team.

[Rendered](https://github.com/grahamc/rfcs/blob/4fb9343dff52137cccd8e5b017181ac476d7b13b/rfcs/0025-nix-core-team.md).

# Voting

Please register your approval or disapproval with the "Review Changes" feature. Obviously:

 - Approval = Approval
 - Request Changes = ... Not ... approval ...

and place any final feedback in the body of that review.

PR #36, opened 2018-11-22
[RFC 0036] Improving the RFC process 

globin:
[Rendered RFC](https://github.com/mayflower/nixos-rfcs/blob/improve-rfc-rfc/rfcs/0036-rfc-process-team-amendment.md)

This is the result of the discussion at the hackday at NixCon 2018.

Thanks to @grahamc for co-authoring it, @willibutz mainly for the illustrations, but also sanity-checking and proof-reading, @ekleog for the previous PR(#18) trying to improve the process and working with us on this RFC
Also @zimbatm, @domenkozar, @fpletz, @edolstra, @shlevy, @garbas, @mic92, @flokli

*Implementation PR will follow ASAP*


rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+forming an RFC Shepherd team from the available nominations on each RFC. This
+team also names the leader of the Shepherd team. This has to happen within 1
+week after the PR has been opened. Until then the Steering Committee is
+responsible for guiding the discussion. In case of the Shepherding Team not
+doing its work the Steering Committee shall encourage them or step in and assign
+new Shepherds. They also are in charge of merging accepted and rejected RFCs.
+Generally by these expectations they should find time to meet once a week for
+about an hour.
+
+##### Shepherd
```
FRidh:
```suggestion
##### Shepherd Team
```

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+A team of 3-4 community members defined unanimously by the RFC Steering
+Committee, responsible for accepting or rejecting RFCs. This team is created per
```
FRidh:
```suggestion
Committee, responsible for accepting or rejecting a specific RFC. This team is created per
```

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+RFC from community members nominated in the discussion on that RFC.
+
+This team should be people who are very familiar with the main components
+touched by the RFC.
+
+##### Leader
```
FRidh:
```suggestion
##### Shepherd Leader
```

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+expand on that to flesh out the process further. During this effort a lot of
+inspiration has been taken from [Rust's RFC
+process](https://github.com/rust-lang/rfcs#what-the-process-is) which works well
+and we have adapted to our needs.
+
+
+# Detailed design
+[design]: #detailed-design
+
+## Glossary
```
zimbatm: A Glossary is usually a section that is at the end of the document and that contains definitions that can be "glossed" over. For example the FCP acronym is introduced in the middle of the document, of one skiped it, it's useful to be able to go to the end and quickly look it up.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+
+*In short, to get a major change included in Nix or nixpkgs, one must
+first get the RFC merged into the RFC repository as a markdown file under the
+`accepted` directory. At that point the RFC is accepted and may be implemented
+with the goal of eventual inclusion into Nix or nixpkgs.*
+
+0. Have a cool idea!
+1. Find a co-author. A co-author is critical to making sure your RFC is viable
+   and will receive support. Your co-author helps flesh out the RFC, and should
+   also support the RFC.
```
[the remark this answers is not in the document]
globin: One can still write up the initial draft on one's own, but then before PR'ing a co-author should be found, who will probably have some remarks to flesh it out further.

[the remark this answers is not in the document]
7c6f434c: That's true; but watching RFC repository was previously a reasonable way to notice the discussion on still-raw RFCs. Now it will spill in all the different directions…

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+
+In cooperation and discussion with Eelco Dolstra and all nominees the proposal
+for the first iteration of members of the RFC Steering Committee are:
+
+ - Eelco Dolstra (edolstra, niksnut)
+ - Shea Levy (shlevy)
+ - Domen Kožar (domenkozar)
+ - Jörg Thalheim (Mic92)
+ - Robin Gloster (globin)
+
```
zimbatm: For convenience, it would be good to start mapping teams to github teams. Eg: @NixOS/rfc-steering . Not a blocker but I would like to see this happening.

Related, I should be removed write access on the rfcs repo since I haven't be selected in the initial team.

grahamc:
I propose we manage this RFC under the process suggested by this RFC.

7c6f434c:
I have a feeling that in some cases a non-biased Shepherd team has a risk of not being able to reach unanimity. SC stepping in might not have a consensus about the outcome either. Should this be explicitly mentioned as grounds for an explicit «postpone» decision?

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+A team of people defined by _this_ RFC and stays consistent until the team
+members are changed via a follow-up RFC. This committee is responsible for
+forming an RFC Shepherd team from the available nominations on each RFC. This
+team also names the leader of the Shepherd team. This has to happen within 1
+week after the PR has been opened. Until then the Steering Committee is
+responsible for guiding the discussion. In case of the Shepherding Team not
+doing its work the Steering Committee shall encourage them or step in and assign
+new Shepherds. They also are in charge of merging accepted and rejected RFCs.
+Generally by these expectations they should find time to meet once a week for
+about an hour.
```
grahamc: I think we should add a paragraph about how to handle replacing members of the steering committee, in case they're not able to participate in the meetings. This would, ideally, include guidelines on _when_ to replace members: for example, if they stop showing up.

zimbatm: It's important to clearly outline the responsibilities of the team. In it's own paragraph or bulleted list.

If we want this to work we need the members to be accountable for something.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+
+##### Shepherd Team
+A team of 3-4 community members defined unanimously by the RFC Steering
+Committee, responsible for accepting or rejecting a specific RFC. This team is created per
+RFC from community members nominated in the discussion on that RFC.
+
+This team should be people who are very familiar with the main components
+touched by the RFC.
```
zimbatm: same, what are the responsibilities of the RFC sheperds?

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+forming an RFC Shepherd team from the available nominations on each RFC. This
+team also names the leader of the Shepherd team. This has to happen within 1
+week after the PR has been opened. Until then the Steering Committee is
+responsible for guiding the discussion. In case of the Shepherding Team not
+doing its work the Steering Committee shall encourage them or step in and assign
+new Shepherds. They also are in charge of merging accepted and rejected RFCs.
+Generally by these expectations they should find time to meet once a week for
+about an hour.
+
+##### Shepherd Team
```
zimbatm: wording: "Team" is associated to a durable thing whereas "Commitee" is more temporary.

* RFC Steering Committee -> RFCs Team
* RFC Sheperd Team -> RFC Sheperd Commitee

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+## Glossary
+
+##### RFC Steering Committee
+A team of people defined by _this_ RFC and stays consistent until the team
+members are changed via a follow-up RFC. This committee is responsible for
+forming an RFC Shepherd team from the available nominations on each RFC. This
+team also names the leader of the Shepherd team. This has to happen within 1
+week after the PR has been opened. Until then the Steering Committee is
+responsible for guiding the discussion. In case of the Shepherding Team not
+doing its work the Steering Committee shall encourage them or step in and assign
```
shlevy: "doing its work" is as of this RFC not super well defined. IMO this can evolve over time and be specified as things settle down, but it needs to be understood that *someone* (presumably the steering committee) is ultimately responsible for the meta-task of keeping things running by some process or other.

rfcs/0036-rfc-process-team-amendment.md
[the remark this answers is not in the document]
shlevy: I don't have a strong opinion here, but FWIW I've never quite understood why a co-author is critical nor, in the case where there is clearly a single driving force, what the co-author is actually expected to do.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+0. Have a cool idea!
+1. Find a co-author. A co-author is critical to making sure your RFC is viable
+   and will receive support. Your co-author helps flesh out the RFC, and should
+   also support the RFC.
+2. Fill in the RFC. Put care into the details: RFCs that do not present
+   convincing motivation, demonstrate understanding of the impact of the design,
+   or are disingenuous about the drawbacks or alternatives tend to be
+   poorly-received. You might want to create a PR in your fork of the RFCs
+   report to help you flesh it out with a few supporters or chat/video
+   conference with a few people involved in the topic of the RFC.
```
shlevy: :100: In general, we as a community *seriously* underutilize synchronous voice/video calls.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+3. Submit a pull request. As a pull request the RFC will receive design feedback
+   from the larger community, and the author should be prepared to revise it in
+   response.
+4. For the nomination process for potential members of the RFC Shepherd Team,
+   that is specific to each RFC, anyone interested can either nominate another
+   person or themselves to be a potential member of the RFC Shepherd Team. This
```
shlevy: Presumably other-nominees have to accept?

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+![Review Process](./0036-review-process.png)
+
+
+## The RFC life-cycle
+
+Once an RFC is accepted the authors may implement it and submit the feature as a
+pull request to the Nix or nixpkgs repo. Being accepted is not a rubber stamp,
+and in particular still does not mean the feature will ultimately be merged; it
+does mean that in principle all the major stakeholders have agreed to the
+feature and are amenable to merging it.
```
shlevy: Should we explicitly state that a positive RFC switches the default to "no reply means no go" to "no reply means go ahead?" i.e. if some feature goes through the RFC process, then the implementation is PR'd and sits for a reasonable amount of time with the appropriate owners pinged but no substantive negative feedback and otherwise stalled conversation, should anyone with the appropriate commit bit feel free to merge?

rfcs/0036-rfc-process-team-amendment.md
shlevy: Agreed about github teams (though naturally that should wait until after this RFC is accepted, if indeed it is).

Ambivalent about the commit bit question. So long as we trust everyone with the appropriate permissions to follow the relevant policies (including not accidentally running afoul of them), I don't think we need to be overly strict on access control.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+If the Steering Committee were too biased, it might select a biased Shepherding
+Team. We are hoping for them and believe them to commit to doing their work in
+the interest of the community. Also this RFC introduces more process and
+bureaucracy, and requires more meetings for some core Nix/nixpkgs contributors.
+Precious time and energy will need to be devoted to discussions.
+
+# Alternatives
+[alternatives]: #alternatives
+
+The current state, which hardly ever results in an RFC being accepted.
```
shlevy: Other alternatives include well defined owners for any particular domain, with the rights to decide on the process relevant to their domain and the responsibility to follow through on that. An extreme example oft his case is a BDFL with responsibility for all final decisions.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+
+None, as of now.
+
+# Future work
+[future]: #future-work
+
+Work on auto-labeling RFCs and automation of parts of the process that either do
+not need human intervention or to remind people to continue their work.
+
+Define how the Steering Committee is picked in the future.
```
shlevy: A lot of formalization of steering and shepherding responsibilities should be fleshed out as we go IMO.

shlevy:
> I propose we manage this RFC under the process suggested by this RFC.
 
:+1: let's do it!

So I guess the first thing is: Nominations are open for the shepherding team for this RFC. So if you think you know someone (possibly yourself) who should be primarily driving and responsible for the outcome of this RFC, make a note here!

@edolstra @domenkozar @Mic92 @globin I'll reach out separately to coordinate an initial meeting of the steering team with respect to this RFC.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Process from Creation to Merge
+
+*In short, to get a major change included in Nix or nixpkgs, one must
+first get the RFC merged into the RFC repository as a markdown file under the
+`accepted` directory. At that point the RFC is accepted and may be implemented
+with the goal of eventual inclusion into Nix or nixpkgs.*
+
+0. Have a cool idea!
+1. Find a co-author. A co-author is critical to making sure your RFC is viable
```
shlevy: Author and co-author should maybe be recused from steering team and ineligible from shepherding team... Which is a shame because I'd really like to nominate @globin and @grahamc for this shepherding team :grin:

domenkozar:
I propose @zimbatm since he gets stuff done :)

shlevy:
I nominate @edolstra as a shepherd, if he has bandwidth.

shlevy:
> I have a feeling that in some cases a non-biased Shepherd team has a risk of not being able to reach unanimity. SC stepping in might not have a consensus about the outcome either. Should this be explicitly mentioned as grounds for an explicit «postpone» decision?

I think this is potentially one of the roles of the shepherd leader.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+or postpone).
+    * This step is taken when enough of the tradeoffs have been discussed that
+      the RFC Shepherd Team is in a position to make a decision. That does not
+      require consensus amongst all participants in the RFC thread (which is
+      usually impossible). However, the argument supporting the disposition on
+      the RFC needs to have already been clearly articulated, and there should
+      not be a strong consensus against that position outside of the RFC
+      Shepherd Team. RFC Shepherd Team members use their best judgment in taking
+      this step, and the FCP itself ensures there is ample time and notification
+      for stakeholders to push back if it is made prematurely.
```
shlevy: Might be good to reference https://tools.ietf.org/html/rfc7282 as a guideline, or even a formal expectation.

rfcs/0036-rfc-process-team-amendment.md
7c6f434c: Well defined owners plus a process for recognition of new domains, probably. That process can be either well defined or indeed explicitly be a judgement call by a leader.

rfcs/0036-rfc-process-team-amendment.md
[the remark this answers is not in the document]
gilligan: I also don’t quite understand the motivation behind the co-author. Is it just to signal that more than a single person believe the suggestion is a good idea? Or is it more about increasing the quality of the initial draft by having a second pair of eyes looking over it before a bigger audience gets to review it? 

Depending on the actual motivation I wonder if it might end up being just a formality without too much added value?

7c6f434c:
> >    I have a feeling that in some cases a non-biased Shepherd team has a risk of not being able to reach unanimity. SC stepping in might not have a consensus about the outcome either. Should this be explicitly mentioned as grounds for an explicit «postpone» decision?

>I think this is potentially one of the roles of the shepherd leader.

As RFC is written right now, Shepherd Committee Leader is responsible for making sure the process is not neglected, and the process calls for unanimity of FCP declaration.

I would expect that if the Shepherd Leader is responsible for tie-breaking, it might be clear in advance (the RFCs often appear after a few less formal discussions) for every reasonable nominee what is the likely final decision. In a way this is probably a worse policy, because we might not want the Steering Committee to be in a situation where they need to (de facto) pre-determine the outcome unanimously — that work is supposed to be offloaded to the specific Shepherd Committee.

rfcs/0036-rfc-process-team-amendment.md
7c6f434c: I think many conferences have Steering Committees which are responsible for long-term continuity (and they are indeed durable).

shlevy:
> > > I have a feeling that in some cases a non-biased Shepherd team has a risk of not being able to reach unanimity. SC stepping in might not have a consensus about the outcome either. Should this be explicitly mentioned as grounds for an explicit «postpone» decision?
> 
> > I think this is potentially one of the roles of the shepherd leader.
> 
> As RFC is written right now, Shepherd Committee Leader is responsible for making sure the process is not neglected, and the process calls for unanimity of FCP declaration.
> 
> I would expect that if the Shepherd Leader is responsible for tie-breaking, it might be clear in advance (the RFCs often appear after a few less formal discussions) for every reasonable nominee what is the likely final decision. In a way this is probably a worse policy, because we might not want the Steering Committee to be in a situation where they need to (de facto) pre-determine the outcome unanimously — that work is supposed to be offloaded to the specific Shepherd Committee.

Yeah, that's a fair point. I'm generally of the inclination that we should avoid front-loading process decisions as much as possible... If we're treating the steering team as having meta-responsibility for ensuring this process works maybe we should just leave it open for them to step in and make necessary callls (and propose necessary formal adjustments) if the process is not meeting its intended outcomes?

7c6f434c:
>I'm generally of the inclination that we should avoid front-loading process decisions as much as possible... If we're treating the steering team as having meta-responsibility for ensuring this process works maybe we should just leave it open for them to step in and make necessary callls (and propose necessary formal adjustments) if the process is not meeting its intended outcomes?

I think when the process has a clear potential for a deadlock, it is good both to create more chances to avoid the deadlock after discussion, and to prepare a clear path of recognition of a deadlock. After all, we already know we are bad at this recognition.

shlevy:
> I'm generally of the inclination that we should avoid front-loading process decisions as much as possible... If we're treating the steering team as having meta-responsibility for ensuring this process works maybe we should just leave it open for them to step in and make necessary callls (and propose necessary formal adjustments) if the process is not meeting its intended outcomes?
> I think when the process has a clear potential for a deadlock, it is good both to create more chances to avoid the deadlock after discussion, and to prepare a clear path of recognition of a deadlock. After all, we already know we are bad at this recognition.

Fair. Can you recommend a specific patch?

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+    * For RFCs with lengthy discussion, the motion to FCP is usually preceded by
+      a summary comment trying to lay out the current state of the discussion
+      and major tradeoffs/points of disagreement.
+    * Before actually entering FCP, all members of the RFC Shepherd Team must
+      sign off the motion.
```
7c6f434c: Maybe add something like:

«
If the Shepherd Committee reaches unanimous agreement that the discussion is not producing new major substantial points, but cannot reach unanimous agreement about the disposition, it is normally advised that an FCP motion of postponing the RFC due to unresolved disagreement is proposed. It can be adopted by unanimous agreement of either the Shepherd committee or the Steering Committee.
»

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+10. The FCP lasts ten calendar days, so that it is open for at least 5 business
+days. It is also advertised widely, e.g. in NixOS Weekly and through Discourse
+announcements. This way all stakeholders have a chance to lodge any final
+objections before a decision is reached.
+11. In most cases, the FCP period is quiet, and the RFC is either merged or
+closed. However, sometimes substantial new arguments or ideas are raised, the
+FCP is canceled, and the RFC goes back into development mode.
+12. In case of acceptance, the RFC Steering Committee merges the PR into the
+`accepted`, in case of rejection into the `rejected` directory.
```
7c6f434c: What happens in case of postponing?

Maybe in case of postponing all the active participants of the discussion (including the authors and the Shepherd Committee) are encouraged to share if there is any example of evidence that would surprise them and decrease their confidence in the current position?

7c6f434c:
> Fair. Can you recommend a specific patch?

Commented inline.

rfcs/0036-rfc-process-team-amendment.md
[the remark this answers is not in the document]
zimbatm: The co-author was introduced in the initial RFC process to play some of the role of the sheperd's team; help shape the RFC and clean it before submitting the RFC for wider consumption. There was also the idea that if the author can't find at least one co-author then it will be hard motivating more people that the RFC is worth considering.

7c6f434c:
>This idea predates the improvements GitHub did on the review system where it was easy to drown a review in typos.

I think inline comments about typos get hidden as reviews on obsolete lines, and that has worked for many months.

zimbatm:
@7c6f434c related to that, it would be best not to reply by email since GitHub doesn't know how to inline the reply properly and it pollutes the main thread.

oxij:
Oh, a deliberative democracy system. I approve of the idea. Here go some comments.

A) I, too, feel like formal co-author requirement just adds bureaucracy. I would change

-1. Find a co-author...
-2. Fill in the RFC...

to something like the following, as it is usually how it really happens anyway:

+1. If your RFC is a technical proposal, you should implement a minimal working example (MWE) first. A line of code can be worth a thousand words. Don't expect other people to implement your idea for you.
+2. Write a draft of the RFC. Put care into the details: RFCs that do not present convincing motivation, demonstrate understanding of the impact of the design, or are disingenuous about the drawbacks or alternatives tend to be poorly-received.
+3. Show it to someone who, you think, should be aware on the issues you are trying to solve. Advertising in related Nixpkgs issues and PRs might help. You might even get a co-author or two in which case it is recommended to create a PR in your fork of the RFCs repository to help you flesh it out with a few supporters or chat/video conference with a few people involved in the topic of the RFC.
+4. Iterate on MWE and RFC with the supporters using the feedback you now have.

The rest are fine, just need to be renumbered.

+5. Submit a pull request. ...

`0036-rfc-process.png` needs to be redrawn.

As it is now, getting a co-author before you yourself got a clear idea of what you are trying to do (which usually means MWE and at least a draft of an RFC) is strange.

B) I agree with @edolstra that this should obsolete RFC #25.

C) What happens if Steering Committee/Team and Shepherd Committee/Team disagree on the result? Are the decisions of the Shepherd Committee binding?

E.g. imagine this: somebody makes an unconventional proposal to solve a problem nobody knew they had, Shepherd Committee is selected, it investigates in good faith and comes to the conclusion that the author is right, the problem is severe, and his/her change is a good idea. Does this mean that the code change author proposed automatically gets merged?

What happens if that change gets reverted by somebody else with commit access? Especially interesting case is when somebody from the Steering Team does it.

rfcs/0036-rfc-process-team-amendment.md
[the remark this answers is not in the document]
zimbatm: Please vote with thumbs on this comment: remove the co-author from the RFC

shlevy:
> What happens if Steering Committee/Team and Shepherd Committee/Team disagree on the result?

IMO it's not the role of the steering committee as such to have opinions on the result at all. Their job is to ensure the process overall is moving forward and to identify shepherds who have the knowledge and attitude to make a good decision, whatever that may be.

Individuals on the steering team may have opinions, of course. They can express them just like any other community member can, as part of the discussion period and/or by trying to get named to the shepherd committee.

> Are the decisions of the Shepherd Committee binding?

In normal cases, I'd think that a technically sound PR implementing an accepted RFC should be merged, and anyone disagreeing with the acceptance should go through the RFC process itself. However, we shouldn't forget that this formalism is ultimately just an agreement between people with the appropriate authority to voluntarily constrain their behavior in certain ways; in practice, anyone with the commit bit and the right amount of respect/reputation within the community can overrule an RFC, and while such actions should definitely not be taken without serious consideration and attempts at alternative resolutions (and doing so will and should carry reputational costs), it is IMO a *good* thing that we do have an escape hatch if things somehow go awry.

Note, though, that the steering team itself has no special overruling power, and shouldn't.

oxij:
@shlevy Sounds good, but I think that both need to be addressed in the RFC itself. If NixOS is going the committees route then I'd like to see both rights and responsibilities for the committee members clearly outlined. #25 did that (more or less, it stated what the core team would not do, see "What this team is not" section), current version of this RFC is pretty vague on that point.

Also, I think the process section should outline a predicate that can be used to decide what should go via RFC process and what should not.

To be more blunt: while I like and support the idea of introducing a deliberative democracy system to reach and implement decisions (as many nontrivial decisions require the voters to first research the subject to make an informed decision and there needs to be a pre-commitment and some enforcement system for making people stick to those, else they will bail out as soon as things stop being black-and-white and the whole thing will stall) I'm still unconvinced that the RFC bit (as "a highly formalized document in a separate special repository", not "an issues that needs discussion") itself is at all necessary or even useful for technical decisions. (E.g. IETF's protocol RFCs which Rust's RFCs and hence this repository try to emulate usually disagree with the actual running code, sometimes comically, e.g. nobody really follows any of SMTP/MIME-related RFCs except for the most general structure of the protocol and message format. If that process doesn't work too well for protocol specifications, why do you expect it to work for specifying the actual code? To IETF's credit, they now, at least sometimes, follow a better approach of evolving actual implementations and RFCs together, e.g. see changes between TLS v1.3 drafts.)

To me pushing technical discussions to this repository with all these formalized processes seems like a futile attempt to escape the unmanageable mess that Nixpkgs PRs and issues have now become.

For instance, ask yourself, wouldn't it be more logical for Nixpkgs-related technical RFCs to live within the PRs that are supposed to implement them? Why would you ever want discuss anything technical without the attached proof-of-concept code? Or, vice versa, is it ok for every nontrivial technical non-merged PR to become an RFC?

From my point of view the push for technical RFCs looks like an attempt at solving the "drowning in PRs while unable to make decisions" problem by requiring authors to submit more PRs. IMHO, that is counterproductive. Just look at both open and closed https://github.com/NixOS/rfcs/pulls, it looks exactly like https://github.com/NixOS/nixpkgs/pulls, only, for now, smaller. Also, the only accepted technical RFC is #23, which was accepted after it was long implemented in https://github.com/NixOS/nixpkgs/pull/34645 and related PRs.

In short, I would like to see

- a description of a formal structure for reaching and implementing decisions in NixOS organization in general (not just for RFCs),

- a description of what should go via the RFC process and what should not (and I'd argue that most for all? technical decisions it adds no value, but adds friction).

To be more constructive, I feel like this would do:

- a Steering Committee/Team with clearly outlined scope that appoints Shepherds and forces them to make decisions on publicly raised issues in a timely manner (and does nothing else),

- a public forum (RFCs would work, but just an ML would also do, but those are now out of fashion, I know) for discussing meta-topics like governance, appointments to the Steering Committee and etc,

- a public forum (an ML would also do...) where people could raise issues with links to technical discussions and PRs they think are unjustly ignored/unmerged. I think requiring the PRs to be in a mergeable state and a half-page "abstract" discussing "what and why" should be enough.

rfcs/0036-rfc-process-team-amendment.md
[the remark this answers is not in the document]
globin: @grahamc and I decided to keep it mainly for the reason of keeping up the standard of opened PRs. For poorly written our thought out RFCs that probably won't be accepted, it would add unnecessary work for the RFC SC. A second set of eyes would hopefully ensure that there is support of at least a few people and obvious fallacies can be addressed before a larger audience is let loose on it.

rfcs/0036-rfc-process-team-amendment.md
globin: We moved it to the top as we had complaints that it comes too late in the text, but the naming of that probably can be improved. Any ideas?

rfcs/0036-rfc-process-team-amendment.md
globin: I agree. Also as @shlevy has said I don't necessarily thing that your commit bit should be removed. I don't see you trying to cause any harm etc.

7c6f434c:
@oxij I think by now many people are good at packaging the arguments into any predefined format, and we are trying to solve the problem of reaching collective decisions recognised as collective decisions. So we need to experiment with decision-making processes in restricted cases, and however bureaucratic we make creation of proposals, it would still be worth it if we get a better process for _some_ cases out of the experiment.

And I would call RFC #26 also technical.

I probably agree that making it an explicit option to attach a PoC could be a good idea. I am afraid that trying to make a process for fully binding RFCs tied to PRs to merge might lead to less useful information about processes than semi-binding RFCs as described here.

Also, maybe this RFC could recommend to post a note to Discourse in the very beginning, to involve the interested people earlier? (I think Discourse is an I think people who want a forum find Discourse an acceptable forum, and I use it in ML mode and find it an OK ML, and browsing its archive with elinks works fine. MI think people who want a forum find Discourse an acceptable forum, and I use it in ML mode and find it an OK ML, and browsing its archive with elinks works fine. Macceptable implementation of an ML with archives and an acceptable implementation of forum, so forum/ML distinction is not too important in the current context)

rfcs/0036-rfc-process-team-amendment.md
globin: I'll try to add that

rfcs/0036-rfc-process-team-amendment.md
globin: Yes, I thought about that when opening the PR, but haven't had the time of adding it, yet.

7c6f434c:
Re: Nix Core team — for context — have any decisions been considered under the Nix Core Team Process?

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+---
+
+# Summary
+[summary]: #summary
+
+This RFC proposes an RFC Steering Committee who decide on a group of RFC
+shepherds for each RFC who guide the discussion to a general consensus and then
+propose a motion for a "Final Comment Period" (FCP) with a disposition for
+acception, rejection or postponing. (see Glossary for a short definition of
+these terms)
```
7c6f434c: I think that «postponing» is never defined, and «FCP» and «disposition» are not mentioned in the Glossary/Terminology section.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,202 @@
...
+
+
+# Motivation
+[motivation]: #motivation
+
+A lot of RFCs have stalled and already an [RFC has been submitted exactly on
+this topic](https://github.com/NixOS/rfcs/pull/18), which ironically has not
+been decided on either. This new RFC takes the above into account and tries to
+expand on that to flesh out the process further. During this effort a lot of
```
7c6f434c: Does it make any prescription for preexisting proposals (except closing #18 as superseded by the current one, I guess)?

globin: If this is accepted we will go through the past RFCs that have been decided on and follow the process as defined in the RFC. (i.e. #18 will most probably be either closed before we reach that point or else probably should be a rather easy one to reach consensus on closing)

globin:
I'd like to additionally propose @ekleog and @samueldr as Shepherds.

rfcs/0036-rfc-process-team-amendment.md
7c6f434c: I think if we expect such a change of perception, it needs to be explicit in the RFC. If this amendment gets direct objections, at least we will know what is the situation…

oxij:
@7c6f434c

> it would still be worth it if we get a better process for _some_ cases out of the experiment

Sure, but I don't see how doing the same thing over and over again could change the result.

> And I would call RFC #26 also technical.

IMHO it's meta. It is also kinda sad (and funny) how the worst possible name got stuck when all was needed to fix it was `sed 's/staging-next/next/g'` and `git branch -m staging-next next`. Do we now need a separate RFC for running two shell commands?

> Discourse != ML

- Sets wrong header fields. It should keep the "From:" and add "Sender:" or "Resent-*" fields, not mangle the "From:", see RFC5322. Anything that mangles "From:", by definition, is not an ML.
- No downloadable archive.
- No way to create a topic in a sublist (e.g. "Announcements") without going via the http.
- Sends bb-code (or whatever that format is) instead of using "> " quotes.

But at least it preserves threading, unlike GitHub, which I am thankful for.

7c6f434c:
There are some subtle changes that might change the outcome. We learn more by trying smaller steps than by changing everything then finding out it also doesn't work.

(Your points about Discourse — I guess I should create a Discourse thread about the parts where there is a hope to improve things)

rfcs/0036-rfc-process-team-amendment.md
vcunat: I think it's clear that one shouldn't get into a shepherd team without agreeing with that :-)  We could write it down explicitly, but it could be thought as implicitly being part of step 5 (i.e. SC choosing reasonably).

rfcs/0036-rfc-process-team-amendment.md
vcunat: I'd expect the implementer to @mention the corresponding shepherd team, and as a result we should have at least one review of the implementation.  RFCs are supposed to be about more significant changes, and I wouldn't like to merge those without at least a second pair of eyes on the code changes.

Yes, there might be some exceptions, but as today &ndash; it ultimately depends on the judgement of the one who pushes/merges a change without any reaction.  For those "trivial" RFC implementations, it should be easy enough to get another opinion like _"I haven't really looked, but this kind of change seems OK without review"._

rfcs/0036-rfc-process-team-amendment.md
vcunat: I don't think that's really a "hard alternative".  I expect that SC will be only glad for anything that makes the shepherd choice straightforward.  We have CODEOWNERS now already, although perhaps under-used, and the file probably can't really cover completely everything, even if we have it in every repo.

To be clear, I do believe that we need to move a bit in direction of maintainership/ownership (at least in nixpkgs).

rfcs/0036-rfc-process-team-amendment.md
vcunat: Yes, it won't hurt to at least link what we mean by "consensus", and I believe this particular link is a good one.  I expect longer-term participants from a similar community to mostly have a very similar notion of what it means, but still...

rfcs/0036-rfc-process-team-amendment.md
shlevy: > it ultimately depends on the judgement of the one who pushes/merges a change without any reaction.

I think our community has shown time and time again that that judgment will almost always err in favor of inaction. We're a respectful and cautious bunch, if we don't explicitly say somewhere that it's OK to go ahead in most cases we won't. If that's the desired outcome that's one thing, but if the expectation is that an accepted RFC will generally mean that a sound PR will get merged, then we need to be clear about that.

vcunat: > [...] the expectation is that an accepted RFC will generally mean that a sound PR will get merged [...]

I agree with the expectation, though it's hard to say how strong it should be.  I'd say it might be individual for each RFC, judged from the discussion around it.  Sometimes it's difficult to anticipate all important information _before_ making an implementation and yet you may want an RFC to gather some more formal opinion before undertaking full implementation.

The current text does express something like what you write, though I expect you'd prefer a stronger wording:

> Being accepted is not a rubber stamp, and in particular still does not mean the feature will ultimately be merged; it does mean that in principle all the major stakeholders have agreed to the feature and are amenable to merging it.

7c6f434c: I would say that given the project's track record on reaching a decision or at least finding out the reason why decision has not been reached, this wording doesn't really promise that RFCs mean anything.

I would have more hopes for a text like 

> Of course, an implementation PR for an accepted RFC can still be rejected due to implementation-specific concerns, or due to some substantial issues not discussed in the context of RFC that become apparent in the context of an implementation.

I guess that would be too optimistic:

> If substantial new (or implementation-specific) issues are not raised in a reasonable time, it would be acceptable to merge for any willing person with commit access after a single positive detailed implementation review (by a person who didn't participate in the implementation).

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,200 @@
...
+
+0. Have a cool idea!
+1. Find a co-author. A co-author is critical to making sure your RFC is viable
+   and will receive support. Your co-author helps flesh out the RFC, and should
+   also support the RFC.
+2. Fill in the RFC. Put care into the details: RFCs that do not present
+   convincing motivation, demonstrate understanding of the impact of the design,
+   or are disingenuous about the drawbacks or alternatives tend to be
+   poorly-received. You might want to create a PR in your fork of the RFCs
+   report to help you flesh it out with a few supporters or chat/video
```
Zimmi48:
```suggestion
   repo to help you flesh it out with a few supporters or chat/video
```

rfcs/0036-rfc-process-team-amendment.md
bennofs: Maybe there should be a deadline for review in the implementation PR, just like there is for the RFC PR? The sheperds could be responsible for this: say a implementation PR musst enger final review after x days of no new comments, and then must either be closed or merged after 10 days. This should prevent stalled PRs at least.

rfcs/0036-rfc-process-team-amendment.md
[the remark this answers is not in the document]
bennofs: An idea: replace requirement for the co-author with "find at least one person that agrees to be nominated as part of the shepherding Team"?

So you have confirmation that at least one other person considers this important. Also, a possible shepherding member is usually more experienced so can give advice during rfc draft process.

rfcs/0036-rfc-process-team-amendment.md
globin: We haven't added that on purpose, since the discussion should continue without time limit as long as it is constructive, new points are brought up and the RFC is iterated on. If this is the case no longer, then the Shepherd Team shall step in with a motion for FCP.

rfcs/0036-rfc-process-team-amendment.md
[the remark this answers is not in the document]
globin: Removed the necessity for a co-author and fleshed out the request to the author, to gather feedback at an early stage/before opening the PR.

rfcs/0036-rfc-process-team-amendment.md
globin: I think I've addressed this.

rfcs/0036-rfc-process-team-amendment.md
globin: I think I've addressed this.

rfcs/0036-rfc-process-team-amendment.md
```diff
@@ -0,0 +1,233 @@
...
+created per RFC from community members nominated in the discussion on that RFC.
+
+This team should be people who are very familiar with the main components
+touched by the RFC. The author cannot be part of the Shepherd Team. In addition,
+at most half of the Shepherd Team can be part of the RFC Steering Committee.
+
+The resposibility of the team is to guide the discussion as long as it is
+constructive, new points are brought up and the RFC is iterated on and from time
+to time summarise the current state of discussion. If this is the case no longer,
+then the Shepherd Team shall step in with a motion for FCP.
```
7c6f434c: Is the main decision point  accept/reject the FCP motion disposition? Maybe it would be a good idea then to add that the Shepherd Team also has to unanimously decide, after studying all the points raised during the discussion, whether to recommend accepting the RFC?

rfcs/0036-rfc-process-team-amendment.md
7c6f434c: I think the idea of postponing is still unclear.

Maybe it would be better just to consider postponing a rejection, and say explicitly that new context or new tools or new knowledge can justify submitting a later RFC similar to an earlier rejected one?

7c6f434c:
Re: "stalled discussion" — I think there are situations where failure to reach unanimity even in a group of three-four people with diverse points of views is a real risk; if there is no such risk, things usually get decided even without an RFC…

rfcs/0036-rfc-process-team-amendment.md
Ekleog: Even though the idea is more or less the same, I think separating postponement and rejection is useful to avoid strong wording, that can have negative effects on both sides of the discussion (people not wanting to use it and unconsciously preferring to let a RFC stall, people feeling rejected). IMO postponement is “we want that but are not ready for it” (eg. a RFC to make nixpkgs use `builtins.encryptString` everywhere), and rejection is “as of now, we don't think we want that” (eg. a RFC to remove overlays).

Now, agree with you that they are currently not well-defined-enough in the `Terminology` section and that it could use some wording to make it explicit that RFC choices can be revisited if the context|tools|knowledge changes.

Ekleog:
Just had an idea coming from the Rust process: experimental RFCs. These are RFCs accepted on a “let's try this” basis. In Rust this is technically enforced by not allowing stabilization of those without another RFC to validate them. In Nixpkgs, it could be “we'll try this but promise nothing about whether it stays in”. I'm thinking it could be useful for things like “Support `<arch>`”, where we might want to say “OK let's try, but not promise anything”.

I'm not totally sure whether this should get in this RFC or in a separate one, though.

Unrelatedly, as this RFC amends RFC0001, I think a link to it should be added towards the top of RFC0001.

rfcs/0036-rfc-process-team-amendment.md
7c6f434c: I am in favour of postponement (and of encouraging the shepherds to specify their personal — non-unanimous — expectations of what has to happen first for a real decision); but if there is a reason not to define it (or its consequences), I think omitting it might be still better than having an unclear hanging notion.

7c6f434c:
Re: experimental RFCs: I guess with our process we have less separation of stable releases from the experimental development. I think it would be a risky scope creep for the current RFC.

globin:
@Ekleog 
> Just had an idea coming from the Rust process: experimental RFCs. These are RFCs accepted on a “let's try this” basis. In Rust this is technically enforced by not allowing stabilization of those without another RFC to validate them. In Nixpkgs, it could be “we'll try this but promise nothing about whether it stays in”. I'm thinking it could be useful for things like “Support `<arch>`”, where we might want to say “OK let's try, but not promise anything”.
> 
> I'm not totally sure whether this should get in this RFC or in a separate one, though.
> 

I'd prefer looking at this in a seperate RFC, I want to keep this rather small and uncontroversial.

> Unrelatedly, as this RFC amends RFC0001, I think a link to it should be added towards the top of RFC0001.

I'll add that in the upcoming implementation PR.

shlevy:
Shepherd team for this RFC is @zimbatm, @edolstra, @Ekleog, and @samueldr, with @zimbatm serving as shepherd leader.

globin:
I think I've addressed the postponing/rejected issue that was the largest remaining discussion point.

WRT the Nix Core Team RFC, do all of you @NixOS/nix-core agree that this supersedes it? If this needs more discussion I'd like to postpone that issue to another RFC as it only partly fits into this RFC and I'd like to keep this as small and uncontroversial as possible.

Are there any further open issues I've missed?

shlevy:
@globin Personally happy to count this as superseding the core team if there are no objections.

oxij:
Looks good enough to me.

rfcs/0036-rfc-process-team-amendment.md
shlevy: Fine leaving this for now, but would be worth returning to if/when we hit more contentious RFCs

vcunat:
@globin me too.  This RFC doesn't really touch any workflow around the Nix repo, but I'm afraid I've been so far unable to find significant amount of time for it anyway.

domenkozar:
Python seems to have taken a similar path: https://www.python.org/dev/peps/pep-8016/

Ekleog:
I guess this fits in the “bugfix” category and thus doesn't require cancelling the FCP for a fix.

I think rewording the paragraph into the following would work:

> ##### Shepherd Leader
> The person in charge of the RFC process for a specific RFC, and responsible for ensuring the process is followed in a timely fashion. The Shepherd Leader has no special resposibility with regard to moving an undecided Shepherd Team to a certain decision.

Additional bonus: this brings this paragraph's style in line with the other definition paragraphs by starting it with a nominal sentence.

vcunat:
_For others, the above is just_ `s/He/The Shepherd Leader/`.

domenkozar:
I've finally had the time to read what Python is doing. I miss two failure prevention mechanisms in the current form:

a) [Conflict of interest](https://www.python.org/dev/peps/pep-8016/#conflicts-of-interest) - putting an upper cap of 2 persons to work for a single employer
b) [Vote of no confidence](https://www.python.org/dev/peps/pep-8016/#vote-of-no-confidence) - vote to disassemble the whole team or a single member

I'm happy to include those if there are no objections.

Ekleog:
@domenkozar Thank you for reading it! However, I think such a change would need to suspend the current FCP so that it could be put to debate correctly, as it's non-trivial from a content point of view.

I am personally neutral on the question, but would consider these to be problems we could address either now or in a later revision of this RFC once we have gained more experience in the in-practice process. If delaying to later is the chosen path forward (ie. if no one is strongly willing to discuss this now -- still have 5 days to express such an opinion), then this should be put into the unresolved questions at least so that we indeed come back to it in the next round of discussion.

domenkozar:
I realize that we'd have to step backward, but given what we've learned with Nix Core team, I think at least vote of no confidence is needed, so that status quo can be resolved.

I'm +0 to include those, so I'm interested to hear from others.

globin:
> I've finally had the time to read what Python is doing. I miss two failure prevention mechanisms in the current form:
> 
> a) [Conflict of interest](https://www.python.org/dev/peps/pep-8016/#conflicts-of-interest) - putting an upper cap of 2 persons to work for a single employer
> b) [Vote of no confidence](https://www.python.org/dev/peps/pep-8016/#vote-of-no-confidence) - vote to disassemble the whole team or a single member
> 
> I'm happy to include those if there are no objections.

I'd propose to move this to a new amending RFC, so that this RFC and the process can get going, especially wrt other stalled RFCs. I don't think we'll have problems with this until we follow up with that. But I definitely agree that these points should be added at some point. So far we already had made sure to check that no more than 2 people of either Mayflower or Tweag would be members of the RFC Steering Committee.

shlevy:
Re @domenkozar 's suggestion, I think these are things we should be thinking about but should not be overeager on up front process. IMO process should evolve in response to real needs facing our community... e.g. we do have quite a few industrial backers of Nix but so far most of them appear to seriously be engaging in appropriate community-centered behavior in good faith. Back to an earlier comment of mine, we should remember that at the end of the day there are several informal escape hatches from the formal structure we build here if things truly get out of hand.

vcunat:
The a) and b) would be for the Committee, right?  a) might also be a softer recommendation when choosing each Shepherd Team, but I wouldn't be so strict there, leaving that on the good judgement of the Committee and b) is there already IIRC.

domenkozar:
@shlevy I have a few objections with that logic, but before we go too deep I'd like to ask @zimbatm, @edolstra, @Ekleog, and @samueldr for comments.

samueldr:
IMHO, (a) and (b) can and should either cause this RFC to go to *Refine Idea*, or those be added through RFC that amend this one. I **strongly suggest going through RFCs**, either one for both, or one per. One per is probably better for the fact that issues with either of the proposed amendments will not block the other, but at the same time, could cause RFC fatigue.

(Opinion is in bold only to make it obvious.)

Ekleog:
@vcunat both (a) and (b) are currently left as future work as I understand it:
> Define how the Steering Committee is picked in the future and how to replace members thereof if they are not able to participate in the meetings, including guidelines on when to replace members. (a timeline, not being active, etc.)

@domenkozar My opinion is summed up at https://github.com/NixOS/rfcs/pull/36#issuecomment-447306795. Honestly, if you feel like this is important, I think we should interrupt the FCP, discuss the ideas, amend the RFC and trigger a new FCP. I'd guess if the rate of discussion is as active as it is currently being, this would delay the merging of this RFC by 2-3 weeks. However, I am personally neutral on this question.

An alternative option would be to define the RFC process for other RFCs *before this RFC is merged* (and the RFC process is actually formally defined), and consider that when the current FCP ends (ie. in a few days) the current state of this RFC should be used to handle other RFCs, until this RFC is finalized and merged, at which point the merged RFC will become reference.

This way:
 * This RFC handles all the cases
 * We actually abide by “However, sometimes substantial new arguments or ideas are raised, the FCP is canceled, and the RFC goes back into development mode.” -- I think @domenkozar's argument can be considered substantial, but as it is currently explicitly left as future work there is room for debate about whether it is new
 * Other RFCs can move forward in the mean time.

Deciding to go this way would require approval of the Steering Committee, though, given it will affect how all RFCs are handled until this RFC is merged.

If the Steering Committee is in favor of this, I am in favor of cancelling the FCP and polishing this RFC. Otherwise, I think this RFC should continue its way forward so that we have an agreed-upon basis, but will not try to defend this thought should anyone strongly want to examine these ideas (because of the aforementioned sentence in the RFC text about “substantial new ideas”)

Currently the FCP is ticking, so absent answer of the Steering Committee agreeing to this idea or someone expressing strong will to examine these ideas within the context of this RFC (and not subsequent ones), if the FCP runs to its end this RFC will be approved.

domenkozar:
I have only seen Python proposal two days ago otherwise I'd be happy to flag myself as slow to respond in this particular case.

If we're building rules for the next few years, I'd say 2-3 weeks is a minor bump and is expected  - otherwise what's the point of RFC process having FCP be broadcasted? RFC process has substantial overhead though - we need another meeting of steering committee, pick shephard team again, etc. At the end it's going to be more work.

> Honestly, if you feel like this is important, I think we should interrupt the FCP, discuss the ideas, amend the RFC and trigger a new FCP

I'm still +0 to address these - @shlevy raised a valid point that this is a bigger issue for nixos foundation than the steering committee itself.

Shephards should vote if my concern is valid (maybe the process needs to be amended how FCP can actually be  "canceled, and the RFC goes back into development mode"?).

shlevy:
@Ekleog Not sure I understand, are you proposing that we start using the steering/shepherd model for other RFCs now before this one is merged, and in the mean time take this one out of FCP to visit the issues @domenkozar raises? If so, I'd personally prefer we get this one in and spend the polishing time on a separate amendment RFC, but I'm willing to participate in steering team work if we decide to go that route.

shlevy:
(Note that the "personal preference" above only applies if people are anxious about getting started shepherding soon. I'm personally fine with this taking longer as well).

Ekleog:
@domenkozar The FCP being cancelled, in my understanding, just means that we do like it never existed, it doesn't mean that the RFC is rejected and starts from scratch.

As for shepherds voting, a single shepherd willing to stop the FCP can do it, so there is no need for a vote (as entering the FCP is made by unanimity, it is logical that leaving it can be done by a single person, as the idea is that the end result must be unanimous).

Currently, my reading is @samueldr is -1, and I am -0 (and ready to switch to +1 should someone voice a strong desire to discuss these now and not in amendment RFCs)

@shlevy That's exactly what I am saying indeed, as an alternative to either going the amending-RFC or the second-FCP route. Without a strong preference for this new idea either, but I wanted to point out that we could still handle @domenkozar's concerns now while not delaying further pending RFCs.

zimbatm:
@domenkozar I would rather us go forward at this point. The RFC is clear enough for me to follow as a process. I don't think that it needs to be treated like a legal document. Now the next thing to do is to actually do the work and exercise that process and then we can adjust it with real data.

zimbatm:
The 10 days period has elapsed with no major disagreement. Declaring the @NixOS/rfc-steering-committee to be official!

7c6f434c:
Now some of the «future work» can start without creating confusion.

Does anyone want to start the discussion (on Discourse, probably?) about long-term plans for SC composition update rules, or should I start something then see where it goes?

Should this discussion include the Foundation positions too?

grahamc:
Merry Christmas!

PR #37, opened 2018-12-21
add x86_32 rfc

goertzenator:
Add support for i386-i586 cpus.


rfcs/0037-x86_32.md
```diff
@@ -0,0 +1,57 @@
...
+
+Once the above change is agreed upon then the addition of `i[345]86` CPUs becomes straight forward.  Predicates `isi586`, `isi485`, and `isi386` will be added.  Usages of `isi686` in various packages will be flipped to `isx86_32` as needed to make them work on these CPUs.
+
+Pragmatically, I will be looking to enable i586 support on several common packages including linux kernel, libc, and openssh.  I'm not interested in i486 and i386, but it is trivial to add these CPU definitions alongside i586 should someone in future want to take a crack at it.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- i586 and predecessors are already uncommon and use continues to diminish.
```
copumpkin: The usual issue we have with things like this (minority architectures/platforms) is that someone strongly motivated will propose something, do a lot of cross-cutting work to support it across all of nixpkgs, then lose steam and then the maintenance burden of all the cross-cutting changes falls upon the community.

So I'd phrase the drawbacks in terms of maintenance burden and degree to which this support can be confined to a small region of nixpkgs. The more "compact" such support is, the easier it is for me to believe it's practical to put into nixpkgs. If it's straggly (all sorts of random switches in all sorts of random builds) it becomes a much harder sell to me. I've already been on the fence about the cost/benefit ratio of our i686 support, not because I don't want more people to benefit from Nix, but because being the only maintainer of a niche platform within a niche package manager is not fun and leads to burnout when everyone's patches are constantly breaking your stuff. The biggest example we have of this is macOS, which I think has achieved "critical mass" (to the point of being sustainable), often had me questioning it along the way. Other Linux users probably still do question the value of the occasional `.isDarwin` switch floating around their packages that they can't test and can't engage meaningfully on.

goertzenator:
In practice I do not forsee any i[345]86-specific hackery landing.  90% of the stuff I tried on i586 just worked, and the remainder lacked generic cross compile support that would be needed for any cross target.  I do expect some correction of precision errors (`isx86_32` instead `isi686`), but that won't cause maintenance burden.  i[345]86 is a subset of i686, so some packages may require disabling of i686-specific features; but the package would already have that if it supported ARM and other targets.

7c6f434c:
Are there any CPUs below i586 level manufactured today for PC-acrhitecture applications?

I think if the entire question is likely to be just about a few cases where target architecture needs to be given to the build, it might be not that bad; adding an architecture unlikely to ever be useful with Nix (i386) still sounds weird.

Maybe a good question is also the following: how much needs to get into the mainline Nixpkgs to make an overlay significantly less painful?

Ericson2314:
@copumkin on the contrary I think the only RFC-worthy issue is the breaking change of fixing the naming. We already parse far more ARM variants than we test (with next to no special cases within 32-bit ARM), and this hasn't caused any problems or complaints to my knowledge. I don't see why this would be any different.

goertzenator:
I've made the RFC and implementation PR more conservative.  Implementation is little more than allowing `i[345]86` to parse (like the various [arm flavors](https://github.com/NixOS/nixpkgs/blob/master/lib/systems/parse.nix#L71-L83)) and some build disambiguation between `isi686` and `isx86_32`.  I see no breaking changes because `isi686` will continue to work as before on i686 hosts.

Ericson2314:
My bad via breaking change.

shlevy:
@goertzenator Are you ready for us to open nominations for a shepherd team as per the [new RFC process](https://github.com/NixOS/rfcs/tree/d9aad9291186cc10375afa5aaefed2b01357cf7e#process-from-creation-to-merge)?

goertzenator:
The [implementation PR](https://github.com/NixOS/nixpkgs/pull/52634) was already merged, so I'm not sure what should happen.  Maybe this topic wasn't quite RFC worthy to begin with, but there is useful content and discussion here for reference purposes.

PR #39, opened 2019-01-16
RFC-0039: unprivileged maintainer team

grahamc:
Package maintainers who are not able to commit directly to Nixpkgs
don't have adequate tools to attentively maintain their package.
OfBorg requests reviews of maintainers it can identify. GitHub only
allows requesting a review of a Collaborator of the repository.

This RFC bridges that gap, and allows OfBorg to request reviews of
maintainers.

Rendered: https://github.com/NixOS/rfcs/pull/39/files?short_path=68b0938#diff-68b0938e7e9a24cb0cc24ac6a70950b9

### DIFF
--- a/rfcs/0039-unprivileged-maintainer-teams.md
+++ b/rfcs/0039-unprivileged-maintainer-teams.md
@@ -70,7 +70,7 @@ The existing Nixpkgs maintainer list already contains a structured
 attribute set of per-maintainer details, including GitHub account
 names. Automation will sync this list of GitHub handles with the
 team's membership, automatically adding and removing people to/from
-the team as the list changes.
+the team as the master branch's maintainer list changes.
 
 GitHub handles can change from one user to another, and so we will
 change the maintainer list to include the GitHub user *ID* as well as
@@ -90,12 +90,18 @@ will be written in Rust with the hubcaps library. It will run on the
 NixOS infrastructure with limited credentials, with only sufficient
 permission to manage the team.
 
-The automation will fetch a fresh version of Nixpkgs, extract the
-maintainer information, and update the team. It will support a dry-run
-option.
+The automation will fetch a fresh version of Nixpkgs's master branch,
+extract the maintainer information, and update the team. It will
+support a dry-run option.
 
 New members of the team will receive an invitation to join the GitHub
 organization.
+
+## Changes to Reviewer/Maintainer Behavior
+
+Reviewers and maintainers should use GitHub's review tools (Approve,
+Request Changes, etc.) to clearly communicate their feedback about the
+pull request.
 
 ## OfBorg changes
 
@@ -135,10 +141,8 @@ will continue to 100%.
  - Package maintainers who do have a GitHub account, but do not wish
    to use 2 factor authentication will not benefit from this change.
 
- - If a GitHub user is a real jerk on the internet, it can potentially
-   reflect negatively on the NixOS ecosystem. Someone who is banned
-   from the NixOS GitHub organization is not allowed to be a package
-   maintainer.
+ - Someone who is banned from the NixOS GitHub organization is not
+   allowed to be a package maintainer.
 
 # Alternatives
 [alternatives]: #alternatives
@@ -147,6 +151,18 @@ Mentioning people in GitHub comments is the main alternative. This has
 the major down-side of not receiving the support of [GitHub's UI
 for requested reviews](https://github.com/pulls/review-requested).
 
+
+# Reesolved questions
+[resolved]: #reesolved-questions
+
+ - Is it possible for the automation to spam a user who doesn't want
+   to be part of the team with invitations?
+   No.
+ - Do maintainers want to be part of this team?
+ - Will the requirement of 2FA cause a significant number of people to
+   not want to participate?
+ - How will we handle people who have been invited, but have not
+   accepted the invitation?
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/rfcs/0039-unprivileged-maintainer-teams.md
+++ b/rfcs/0039-unprivileged-maintainer-teams.md
@@ -1,0 +1,176 @@
+---
+feature: unprivileged-maintainer-teams
+start-date: 2019-01-16
+author: Graham Christensen <graham@grahamc.com>
+co-authors: zimbatm <zimbatm@zimbatm.com>
+related-issues: https://github.com/NixOS/ofborg/pull/303
+---
+
+# Summary
+[summary]: #summary
+
+Package maintainers who are not able to commit directly to Nixpkgs
+don't have adequate tools to attentively maintain their package.
+OfBorg requests reviews of maintainers it can identify. GitHub only
+allows requesting a review of a Collaborator of the repository.
+
+This RFC bridges that gap, and allows OfBorg to request reviews of
+maintainers.
+
+# Motivation
+[motivation]: #motivation
+
+The goal of this RFC is to involve package maintainers in reviewing
+pull requests against their packages. This RFC does not grant
+maintainers the ability to merge pull requests against their own
+package.
+
+Maintainers take a responsibility for their package, and want to know
+about updates to their package's expression. However, Nixpkgs receives
+over 1,000 pull requests each month and subscribing to them all is not
+a reasonable requirement to maintain a package.
+
+The ideal outcome is package maintainership means a more active role
+in reviewing and approving changes to Nixpkgs.
+
+# Detailed design
+[design]: #detailed-design
+
+Package maintainers will be a member of a GitHub team, allowing OfBorg
+to request a review.
+
+## The Team
+
+We will create a GitHub team under the NixOS GitHub organization
+called "Nixpkgs Maintainers" which only grants "read" access to
+Nixpkgs.
+
+This team will not grant any privileges to the Nix ecosystem
+repositories which non-members already have. They will not be able to
+close other people's issues or PRs or push branches. Experimentation
+and documentation shows this will only grant access to a team
+discussion board on GitHub.
+
+Being a member of this team will let the user mark themselves as a
+public member of the organization. This will show the NixOS logo on
+their GitHub profile, and people will see "Member" next to their
+account name when browsing issues.
+
+In order to be a member, each user will enable 2FA on their GitHub
+account, since [the GitHub organization requires 2FA of all
+members](https://github.com/NixOS/nixpkgs/issues/42761).
+
+See
+https://help.github.com/articles/permission-levels-for-an-organization/
+for more information about what this will grant.
+
+## Changes to `maintainers/maintainer-list.nix`
+
+The existing Nixpkgs maintainer list already contains a structured
+attribute set of per-maintainer details, including GitHub account
+names. Automation will sync this list of GitHub handles with the
+team's membership, automatically adding and removing people as the
+list changes.
+
+GitHub handles can change from one user to another, and so we will
+change the maintainer list to include the GitHub user *ID* as well as
+their handle. When syncing, the automation will validate the user ID
+matches. GitHub User IDs are easily found at
+`https://api.github.com/users/«username»`.
+
+If a user ID's GitHub handle changes, the maintainer should remain
+part of the team under their new name. The repository's nickname
+should be updated to reflect their new nickname.
+
+## Team Automation
+
+The team must be automatically updated at least once a day to ensure
+the maintainer list is fresh and up to date. The automation for this
+will be written in Rust with the hubcaps library. It will run on the
+NixOS infrastructure with limited credentials, with only sufficient
+permission to manage the team.
+
+The automation will fetch a fresh version of Nixpkgs, extract the
+maintainer information, and update the team. It will support a dry-run
+option.
+
+New members of the team will receive an invitation to join the GitHub
+organization.
+
+## OfBorg changes
+
+OfBorg will identify PRs which are approved by their maintainers, and
+add a special label `approved-by-maintainer`.
+
+## Roll-Out Plan
+
+1. Write an explanatory post on Discourse about the what-and-why of
+   this plan.
+2. Select a small group of maintainers who are not committers to be
+   part of the first round, and manually run the tooling, and pause
+   half a week to see what changes.
+3. Automate the tooling on the infrastructure.
+4. Expand the group to one quarter of the maintainers, and pause a
+   half a week to gauge response.
+5. Expand the group to one half of the maintainers and wait one week.
+6. Expand the group to all of the maintainers.
+
+If we receive no major feedback or problems during the rollout, we
+will continue to 100%.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+ - Putting each maintainer in a read only team will display
+   maintainers as "member", without specifying which team they are a
+   member of. This gives the impression of authority which maintainers
+   don't already receive. This is a pro and a con.
+
+ - A mistake in the automation, or in the admin panel of GitHub could
+   grant the team write access to Nix ecosystem repositories.
+
+ - Package maintainers who do not wish to have a GitHub account will
+   not benefit from this change.
+
+ - Package maintainers who do have a GitHub account, but do not wish
+   to use 2 factor authentication will not benefit from this change.
+
+ - If a GitHub user is a real jerk on the internet, it can potentially
+   reflect negatively on the NixOS ecosystem. Someone who is banned
+   from the NixOS GitHub organization is not allowed to be a package
+   maintainer.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Mentioning people in GitHub comments is the main alternative. This has
+the major down-side of not receiving the support of [GitHub's UI
+for requested reviews](https://github.com/pulls/review-requested).
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+ - Is it possible for the automation to spam a user who doesn't want
+   to be part of the team with invitations?
+ - Do maintainers want to be part of this team?
+ - Will the requirement of 2FA cause a significant number of people to
+   not want to participate?
+ - How will we handle people who have been invited, but have not
+   accepted the invitation?
+
+# Future work
+[future]: #future-work
+
+ - Writing the automation program.
+ - Adding UIDs to every maintainer.
+ - Creating the GitHub team
+ - Updating the NixOS Org Configurations repository to run the
+   automation with credentials on an automated basis.
+
+# Future Potential RFCs
+The following topics are explictly _not_ part of this RFC.
+
+ - Allowing maintainers to merge pull requests against their packages
+   without having commit access.
+ - Requiring all maintainers to have a GitHub account with 2FA.
--- a/rfcs/0039-unprivileged-maintainer-teams.md
+++ b/rfcs/0039-unprivileged-maintainer-teams.md
@@ -152,23 +152,16 @@ the major down-side of not receiving the support of [GitHub's UI
 for requested reviews](https://github.com/pulls/review-requested).
 
 
-# Reesolved questions
-[resolved]: #reesolved-questions
+# Resolved questions
+[resolved]: #resolved-questions
 
  - Is it possible for the automation to spam a user who doesn't want
    to be part of the team with invitations?
    No.
- - Do maintainers want to be part of this team?
- - Will the requirement of 2FA cause a significant number of people to
-   not want to participate?
- - How will we handle people who have been invited, but have not
-   accepted the invitation?
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
- - Is it possible for the automation to spam a user who doesn't want
-   to be part of the team with invitations?
  - Do maintainers want to be part of this team?
  - Will the requirement of 2FA cause a significant number of people to
    not want to participate?
--- a/rfcs/0039-unprivileged-maintainer-teams.md
+++ b/rfcs/0039-unprivileged-maintainer-teams.md
@@ -46,7 +46,7 @@ called "Nixpkgs Maintainers" which only grants "read" access to
 Nixpkgs.
 
 This team will not grant any privileges to the Nix ecosystem
-repositories which non-members already have. They will not be able to
+repositories which non-members don't already have. They will not be able to
 close other people's issues or PRs or push branches. Experimentation
 and documentation shows this will only grant access to a team
 discussion board on GitHub.
@@ -56,8 +56,8 @@ public member of the organization. This will show the NixOS logo on
 their GitHub profile, and people will see "Member" next to their
 account name when browsing issues.
 
-In order to be a member, each user will enable 2FA on their GitHub
-account, since [the GitHub organization requires 2FA of all
+In order to be a member, each user will need to enable 2FA on their
+GitHub account, since [the GitHub organization requires 2FA of all
 members](https://github.com/NixOS/nixpkgs/issues/42761).
 
 See
@@ -69,8 +69,8 @@ for more information about what this will grant.
 The existing Nixpkgs maintainer list already contains a structured
 attribute set of per-maintainer details, including GitHub account
 names. Automation will sync this list of GitHub handles with the
-team's membership, automatically adding and removing people as the
-list changes.
+team's membership, automatically adding and removing people to/from
+the team as the list changes.
 
 GitHub handles can change from one user to another, and so we will
 change the maintainer list to include the GitHub user *ID* as well as
@@ -79,8 +79,8 @@ matches. GitHub User IDs are easily found at
 `https://api.github.com/users/«username»`.
 
 If a user ID's GitHub handle changes, the maintainer should remain
-part of the team under their new name. The repository's nickname
-should be updated to reflect their new nickname.
+part of the team under their new handle. The user's entry in
+`maintainer-list.nix` should be updated to reflect their new handle.
 
 ## Team Automation
 


rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+to request a review.
+
+## The Team
+
+We will create a GitHub team under the NixOS GitHub organization
+called "Nixpkgs Maintainers" which only grants "read" access to
+Nixpkgs.
+
+This team will not grant any privileges to the Nix ecosystem
+repositories which non-members already have. They will not be able to
```
7c6f434c: missed «don't»

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+close other people's issues or PRs or push branches. Experimentation
+and documentation shows this will only grant access to a team
+discussion board on GitHub.
+
+Being a member of this team will let the user mark themselves as a
+public member of the organization. This will show the NixOS logo on
+their GitHub profile, and people will see "Member" next to their
+account name when browsing issues.
+
+In order to be a member, each user will enable 2FA on their GitHub
```
benclifford: will *need* to?

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+
+# Summary
+[summary]: #summary
+
+Package maintainers who are not able to commit directly to Nixpkgs
+don't have adequate tools to attentively maintain their package.
+OfBorg requests reviews of maintainers it can identify. GitHub only
+allows requesting a review of a Collaborator of the repository.
+
+This RFC bridges that gap, and allows OfBorg to request reviews of
```
benclifford: to be clear (I think the later context says so) you mean, on-github reviews of PRs, not some other review mechanism?

and so if I am a maintainer, I'll get a regular looking github review and I can hit approve/request changes just like on any other project i'm a full super-full-powers member of?

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+See
+https://help.github.com/articles/permission-levels-for-an-organization/
+for more information about what this will grant.
+
+## Changes to `maintainers/maintainer-list.nix`
+
+The existing Nixpkgs maintainer list already contains a structured
+attribute set of per-maintainer details, including GitHub account
+names. Automation will sync this list of GitHub handles with the
+team's membership, automatically adding and removing people as the
```
7c6f434c: Do we want to remove people automatically or by request?

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+list changes.
```
benclifford: adding and removing people from the team or from maintainer-list.nix? I think the former?

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+   reflect negatively on the NixOS ecosystem. Someone who is banned
+   from the NixOS GitHub organization is not allowed to be a package
+   maintainer.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Mentioning people in GitHub comments is the main alternative. This has
+the major down-side of not receiving the support of [GitHub's UI
+for requested reviews](https://github.com/pulls/review-requested).
```
alyssais: How important is the GitHub UI for requesting reviews? Addresses most of the drawbacks, and I'm not sure the UI is that valuable in comparison.

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Changes to `maintainers/maintainer-list.nix`
+
+The existing Nixpkgs maintainer list already contains a structured
+attribute set of per-maintainer details, including GitHub account
+names. Automation will sync this list of GitHub handles with the
+team's membership, automatically adding and removing people as the
+list changes.
+
+GitHub handles can change from one user to another, and so we will
```
benclifford: this is weird phrasing for what i think you mean is that a github handle doesn't, over time, uniquely identify a user and might be reused?

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+change the maintainer list to include the GitHub user *ID* as well as
+their handle. When syncing, the automation will validate the user ID
+matches. GitHub User IDs are easily found at
+`https://api.github.com/users/«username»`.
+
+If a user ID's GitHub handle changes, the maintainer should remain
+part of the team under their new name. The repository's nickname
+should be updated to reflect their new nickname.
```
benclifford: new name? new handle?


> The repository's nickname should be updated to reflect their new nickname.

what does this mean?

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+ - A mistake in the automation, or in the admin panel of GitHub could
+   grant the team write access to Nix ecosystem repositories.
+
+ - Package maintainers who do not wish to have a GitHub account will
+   not benefit from this change.
+
+ - Package maintainers who do have a GitHub account, but do not wish
+   to use 2 factor authentication will not benefit from this change.
+
+ - If a GitHub user is a real jerk on the internet, it can potentially
```
7c6f434c: Do we want to phrase it this way as opposed to «non-constructively aggressive in project discussions» or something?

Both from the point of view of not using vague insults, and to put the focus on interactions where the maintainer status actually comes up.

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+   reflect negatively on the NixOS ecosystem. Someone who is banned
+   from the NixOS GitHub organization is not allowed to be a package
+   maintainer.
```
benclifford: >  Someone who is banned  from the NixOS GitHub organization is not allowed to be a package  maintainer.

i don't understand quite what this means?

If i'm a jerk, you will ban me from the github organisation and then I will have to maintain my packages in the same way as: "Package maintainers who do not wish to have a GitHub account will not benefit from this change." - i.e. jerks will not benefit from this but will not otherwise be harmed?

rfcs/0039-unprivileged-maintainer-teams.md
vcunat: Creating the team seems fairly cheap to me.  /cc comments will create unnecessary noise every time, similarly to what borg did before being moved to the "checks" tab.

alyssais: Another alternative, not mentioned here, is to just automatically send mail to maintainers. This would require a custom program to be written, but so would the proposal here. No GitHub features required.

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+part of the team under their new name. The repository's nickname
+should be updated to reflect their new nickname.
+
+## Team Automation
+
+The team must be automatically updated at least once a day to ensure
+the maintainer list is fresh and up to date. The automation for this
+will be written in Rust with the hubcaps library. It will run on the
+NixOS infrastructure with limited credentials, with only sufficient
+permission to manage the team.
```
Ekleog: Does GitHub allow credentials to be limited enough to only add/remove members to a read-only team? I would have guessed it required `Owner` access to the organization.

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+Mentioning people in GitHub comments is the main alternative. This has
+the major down-side of not receiving the support of [GitHub's UI
+for requested reviews](https://github.com/pulls/review-requested).
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+ - Is it possible for the automation to spam a user who doesn't want
+   to be part of the team with invitations?
```
kragniz: I think this would work better as an opt-in process.

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+ - Do maintainers want to be part of this team?
+ - Will the requirement of 2FA cause a significant number of people to
+   not want to participate?
+ - How will we handle people who have been invited, but have not
+   accepted the invitation?
```
7c6f434c: Intuitively, there are committers, maintainers in the team, maintainers not on GitHub, and maintainers who have received but not accepted the invitation. If such lists are kept and re-synchronised, for the people with invitation sent the bot should simply wait for acceptance? If someone rejects and then changes the mind, we can probably handle it manually.

rfcs/0039-unprivileged-maintainer-teams.md
Ekleog: Additionally, it'd work for people who are not using GitHub. The only drawback (compared to this RFC) would be that reviews from maintainers wouldn't show up in color, but in gray -- but if ofborg adds `approved-by-maintainer` anyway that's likely not an issue ; esp. as otherwise reviews from maintainers of one package would appear in color for other packages too.

rfcs/0039-unprivileged-maintainer-teams.md
alyssais: I think it's reasonable to expect people who sign up to maintain packages to receive notifications about proposed changes to those packages. That's really all the maintainer role is for, isn't it?

rfcs/0039-unprivileged-maintainer-teams.md
Synthetica9: It could also be relevant in other places on GitHub, since one could display the NixOS badge under their name

rfcs/0039-unprivileged-maintainer-teams.md
7c6f434c: Well, they will also not be able to participate in GitHub discussions, and submit PRs (which are the main way of non-committer maintainers updating their packages)

rfcs/0039-unprivileged-maintainer-teams.md
7c6f434c: I think «Team maintainer» for «Maintainer team» might be a bit lower; it still seems to be able to give the team write access, though.

rfcs/0039-unprivileged-maintainer-teams.md
7c6f434c: Is it realistically bearable to be an active Nixpkgs package maintainer without Github account? I don't see many patches posted to mailing list, at least.

rfcs/0039-unprivileged-maintainer-teams.md
7c6f434c: Not sure I have ever seen it done with NixOS badge (as we are not writing the rules for banning for behaviour while pretending to represent NixOS organisation, I think focus should be on what happens usually)

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: I think automatically. If they are removed from the maintainer list, they either are being forced out, or have requested removal.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: I think requesting reviews is much better than a `cc` comment. The expectation is the requestee will review it. A `cc` comment doesn't convey such an expectation. Additionally, by using the tooling specific for this purpose, we get for free all the workflow improvements GitHub provides.  I don't think it is worth optimizing for the few users who doesn't want to have a GitHub account, while sacrificing the user experience of the many who do.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: I'd like to keep this case vague, as it isn't intending to side-load a code of behavior. The first time I brought this up  people were concerned with what if a real jerk is part a maintainer, and I think the general solution is we'll handle it however we handle it, and them being a member of this read-only team doesn't really change anything.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: Right.

rfcs/0039-unprivileged-maintainer-teams.md
zimbatm: You could add another Check that sends emails and reports back to the PR but it wouldn't be as prominent as adding the reviewers to the list. It's quite nice when submitting a PR to know who is going to review your work.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: Updated (will push shortly.)

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: I think API tokens can be created specifically for modifying teams, but maybe not a specific team. I tried to make this vague enough to say "as limited as possible, but still able to do the job."

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+
+The team must be automatically updated at least once a day to ensure
+the maintainer list is fresh and up to date. The automation for this
+will be written in Rust with the hubcaps library. It will run on the
+NixOS infrastructure with limited credentials, with only sufficient
+permission to manage the team.
+
+The automation will fetch a fresh version of Nixpkgs, extract the
+maintainer information, and update the team. It will support a dry-run
+option.
```
LnL7: Should membership be based on just master or also other branches like the active release?

grahamc:
I have addressed several questions in the latest push.

shlevy:
Seems good to me.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: a great question! 

It should probably be the sum of all the maintainers in the branches we release channel updates for: https://github.com/NixOS/nixos-org-configurations/blob/33967236a51f41fe4af1d1a321b505f61ed4f246/nixos-org/hydra-mirror.nix#L44-L57

any other release-branches should probably no longer accept PRs.

However, the new maintainer format was introduced in 18.03, so I don't think we can include 17.09. Additionally, the old branches should receive a patch including the user IDs of each maintainer.

This is also pretty complicated if a user changes their name, hmmm...

Synthetica9: Maybe ping based on the branch that the PR is against?

grahamc: OfBorg already does that, I think the question is what branches does the team's member list comprise of?

rfcs/0039-unprivileged-maintainer-teams.md
timokau: @alyssais the maintainers may think bothering with 2fa isn't worth the benefits. In that case they may still do their maintainer duties but shouldn't be spammed by invitations.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: The minimum authority we can grant to a token is org:admin, unfortunately. This means it can add anyone to any team.

grahamc: To me, this means the token would need to be well guarded, not that we shouldn't do this.

rfcs/0039-unprivileged-maintainer-teams.md
timokau: Realistically I would say just master is fine. If someone removed themselves as a maintainers since the last release, they probably won't be willing to review anyways.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: We can pretend it means they can't write, since it doesn't authorize that (https://developer.github.com/apps/building-oauth-apps/understanding-scopes-for-oauth-apps/) but it does mean they can add their account and then write.

shlevy: @grahamc In this case I think we should make the code/node that has access to the token separate from the rest of the code and manage its lifecycle independently, exposing just the team managemnt as a locked down interface.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: Any spamming of invitations is a bug, probably on both GitHub's side, and definitely on our side.

rfcs/0039-unprivileged-maintainer-teams.md
timokau: I realize there are infrastructure difficulties preventing this, but ignoring those:

ofBorg could aggregate all its information into one comment. So after its done with eval and builds, it posts a comment like this:

> I have successfully build <pkg> on [linux](www.link-to-log.tld) and [darwin](www.link-to-log.tld). Aarch64 was not attempted because it is marked as unsupported.
> @MaintainerXYZ, you are listed as the maintainer for this package. Please review this change.

That way the "spam" is contained. In fact I would prefer it to no message at all. It also scales well for future ofBorg improvements. It could also add messages like

> !Warning!: I have determined that this change will result in more than 500 rebuilds on linux but is targeted to the `master` branch.

grahamc: I don't think this is on topic for this RFC.

rfcs/0039-unprivileged-maintainer-teams.md
LnL7: I can't think of a good argument to query multiple branches, just asking in case somebody does.

rfcs/0039-unprivileged-maintainer-teams.md
alyssais: We already use review requests as more of a notification mechanism than assigning work to somebody. So, I think that showing review requests is at best not useful and at worst misleading.

grahamc: I think we are on the path of changing behavior, and making better use of our tools. I don't think we should get too bogged down in how we currently misuse features when considering how we can make better use later.

alyssais: > we get for free all the workflow improvements GitHub provides.

Can you explain what these are? I'm not sure that's been addressed so far in this RFC.

grahamc: a note about why we should use review requests:

This takes advantage of a structured way to express a review is requested. This data has specific meaning to GitHub and all API clients of GitHub. By declaring yourself a maintainer, you have effectively taken ownership of the thing being changed. Changing something owned by someone else, ultimately (in the future where maintainership in Nixpkgs means more than it has in the past) should get a review by the owner. Taking advantage of the semantically reasonable mechanism to do this means GitHub can make improvements based on what they know that data to mean.

---

Specific examples include the linked page listing PRs where your review is requested (but you haven't replied yet). The future workflows they could build are unknowable. There are also third-party dashboards people use (though I can't remember their names :|) which display review requests differently.

grahamc: I think we shouldn't use review requests for notifications. We should use `cc` comments for that.

alyssais: Fair enough.

rfcs/0039-unprivileged-maintainer-teams.md
Ekleog: Can this go into the drawbacks section? It's more or less present with “a mistake in the automation”, but the offensive security aspect (ie. people willingly trying to break it) are only suggested here.

I was personally ambivalent before this information (unsure of both the advantages and the drawbacks), so adding potential attack surface for it makes me half-heartedly opposed to the change.

Maybe an option would be to have a cron mailing nixos owners when an invitation/ejection is required, and this could work without granting automation too much rights? If there are not too frequent changes to the maintainers list, then I think this could reasonably work.

rfcs/0039-unprivileged-maintainer-teams.md
Ekleog: While a page listing PRs where one's review is requested without answer yet is an argument, I don't think we should choose our workflow based on hypothetical improvements GitHub could do later on. First, make the tooling, then decide the workflow accordingly.

Now, whether a page listing PRs where the review is requested and/or third-party dashboards are an advantage enough to warrant this change is another question. I personally don't use them and would assume people who maintain just one or two packages would see even less use to these, but may be wrong here :)

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+If we receive no major feedback or problems during the rollout, we
+will continue to 100%.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+ - Putting each maintainer in a read only team will display
+   maintainers as "member", without specifying which team they are a
+   member of. This gives the impression of authority which maintainers
+   don't already receive. This is a pro and a con.
```
Zimmi48: FTR, in the Coq organization, we invite lots of external contributors (in practice to a team granting write access, not just read access, but our branches are protected, so this is just about giving them rights to triage issues), but we wish that only the core developers are shown as members of the organization. Therefore, as an administrator, I can change the visibility status of these members from Public to Private (and not the converse). This could be automated by the bot.

7c6f434c: I think in case of Nixpkgs it is _also_ convenient for members to see if the person commenting is a committer or not (basically, should I say «nice, mention me when this is ready to merge») — but yes, as a partial solution it might be better than nothing.

Of course, it creates a question whether we want to enforce public visibility of membership for committers.

rfcs/0039-unprivileged-maintainer-teams.md
7c6f434c: I think it is enough if in the 72h after adding the first maintainership it is rare that any packages maintained by the new maintainer receive a non-maintainer change…

rfcs/0039-unprivileged-maintainer-teams.md
7c6f434c: I think given:
 (a) various interesting corner cases in Github behaviour we hit from time to time,
 (b) our story of migrating basically everything exchanging one featureset to another one
we shouldn't really look too much at what Github _intends_ things to mean and ask which set of the side-effects we like better.

(I do probably like side-effects of review requests better than of a comment, unless there is a unified comment with everything ofBorg wants to say about the PR)

Separately, I think that making maintainership mean much more (even if desirable) has time horizon larger than replacing at least one part of the tool collection; so discussing how multiple tools (Github, ofborg, this team maintenance bot) interact can assume nothing big changes in the process for low-impact packages.

rfcs/0039-unprivileged-maintainer-teams.md
timokau: I would actually like if there was a way to highlight maintainers. Gives a sense of membership and responsibility and gives their reviews more weight.

However making it completely indistinguishable from committers might be a problem. That information already is hard to get for nixpkgs.

Zimmi48: Actually, there is another way to know if they have write-access: whether there approved request appears as a green tick or a gray one (you have lots of examples of gray ticks in this very PR).

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: I agree we shouldn't bank on GitHub making big changes. I do think the existing tooling is very good already. I think changing behavior is much faster than you say -- small nudges have far reduced direct pushes within a short period of time of having ofborg do reasonable checks and validation.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: I think it would be an option to run it manually, but I think it is less good.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: > Therefore, as an administrator, I can change the visibility status of these members from Public to Private (and not the converse). This could be automated by the bot.

This can be done, sure, however there is nothing preventing them from marking themselves publicly immediately after.

rfcs/0039-unprivileged-maintainer-teams.md
7c6f434c: I think that works for tooling adoption — and high standards you uphold for ofborg help — better than for shifting general notions (especially in the direction that creates more single-person-of-delay situations).

Anyway, I agree that in the current situation review requests make the most sense.

7c6f434c:
I would like to ask @ckauhaus for permission to nominate.

If that permission is granted, I want to nominate @ckauhaus as a person with an ongoing experience of both running an important part of semi-automated Nix*-related tooling and being a non-committer maintainer.

grahamc:
This link is now visible: https://discourse.nixos.org/t/new-rfc-39-open-for-shepherd-nominations-unprivileged-maintainer-team/1933/1

I'd be interested in nominating @alyssais, given her being a fairly new member.

rfcs/0039-unprivileged-maintainer-teams.md
7c6f434c: About the frequency of maintainer list changes, just in case the tradeoff discussion turns to estimating the amount of effort: git log reports ~1–2 commit/day on normal months (of course, there are merge commits in the count, so I would estimate the number of additions as half that):
```
     50 Date:   2018-04
…
     32 Date:   2018-09
     41 Date:   2018-10
     27 Date:   2018-11
     14 Date:   2018-12
```

rfcs/0039-unprivileged-maintainer-teams.md
alyssais: Having thought about this more, this isn’t actually mutually exclusive with an option that did not depend on GitHub, especially since implementing an email-based system would require no priveleges whatsoever, and could be done as an opt-in system by any community member. While I’m among the (probable) minority of community members who would like Nixpkgs development to move away from GitHub, this doesn’t actually increase our lock-in at all, and would be strictly better than what we have now.

rfcs/0039-unprivileged-maintainer-teams.md
alyssais: Should we perhaps use user IDs instead?

rfcs/0039-unprivileged-maintainer-teams.md
alyssais: I think having it public would actually be a good thing. It gives people a stake in the community, and it probably puts even more impetus on us to remove bad actors from our community.

rfcs/0039-unprivileged-maintainer-teams.md
grahamc: I think that is what this paragraph is saying we should do?

7c6f434c: Just for fun I tried to find out my own user-id on GitHub via UI and failed; I guess it is only available in some API query…

grahamc: https://api.github.com/users/7c6f434c you're #1891350 https://github.com/NixOS/rfcs/pull/39/files#diff-68b0938e7e9a24cb0cc24ac6a70950b9R78

7c6f434c: Thanks, indeed API-only, but at least easy to access.

7c6f434c:
@ckauhaus I also hope that your being on the record as Caring About Security _and_ Caring About Tooling will be useful whe co-moderating the security-vs-usability tradeoff discussions.

lheckemann:
I'd like to self-nominate for shepherding this RFC.

globin:
Per meeting of the @NixOS/rfc-steering-committee: Shepherding Team is @lheckemann  as leader, @ckauhaus and @alyssais.

globin:
I think this RFC is a nice example of bringing our tooling forward to improve the use of features github can provide, without impeding current workflow and also enabling a wider audience to help review pull requests, which as I and other release managers have pointed out previously, is rather necessary!

I very much welcome this and would like to encourage anyone opposing this or having doubts about this to step up and voice their opinion, as otherwise if I haven't missed anything all issues brought forward earlier have been addressed.

rfcs/0039-unprivileged-maintainer-teams.md
7c6f434c: Given that GitHub allows spamming with review requests for same changes, I would say we cannot expect any extra line of defense…

7c6f434c:
(I am not opposed to the RFC as a whole, and not strongly opposed to any part of it)

I am not sure the following comment threads are fully resolved:

1) bot access — given the rate of maintainer changes (an entry between every day and every two days by my rough estimate), and the currently apparently insufficient GitHub access control options, should invitations be automated (via a token that has essentially complete org access) or offered to an administrator to perform?

2) visibility — do we want to recommend the maintainers to set their membership to public, not do anything and expect private membership to be the default, recommend «public for committers private for maintainers», enforce private membership? (public membership cannot be enforced or centrally automated)

3) ignored invitations — do we want unaccepted invitations to timeout eventually? to be reissued under some conditions?

Ekleog:
Something that may have a weight in the visibility debate: maintainers reviewing PRs that are *not* maintained by them would have their reviews showing up with a green tick instead of the grey tick if the visibility is public. And I'm not 100% sure whether if the visibility is private the tick would show as green or as grey?

This might be confusing for committers who may assume that the PR has already been reviewed by a committer who just didn't merge yet because they were waiting on ofborg results or something, and said committer would merge the PR without checking it properly.

7c6f434c:
@Ekleog I think the UI situation is slightly different; for example, this very PR has a grey checkmark for an approval by an organisation member who doesn't have write access to this repository.

Of course, without being able to add a separate badge type any solution will be somewhat confusing in some situations…

Melkor333:
I really like this idea! Another possible future RFC building on this one could be a package based "reading list". E.g. if I want to know of any future PRs on the package `emacs`, i can add myself to its reading list without having to add myself as maintainer.

But there are two questions that seem unanswered to me:
- If a PR is marked as approved by a maintainer, does that mean that it may be merged without any further revision or will the Collaborator who merges the PR have to take another look at the changes?
- Does a PR need to be approved by at least one maintainer before it will be merged?

I guess the second question will probably be a *no*. But if PR's don't *need* to be reviewed, I don't see where maintainers gain any responsibility?

7c6f434c:
>I guess the second question will probably be a *no*. But if PR's don't *need* to be reviewed, I don't see where maintainers gain any responsibility?

Well, being reliably called into discussion does increase the level of encourangement to participate in everything related to the package in question.

And while the committer is still supposed to take a look at the changes, maintainer approval does raise the baseline confidence that the changes make sense.

timokau:
For the committer it could be the difference between multiple feedback cycles with the contributor or a quick glance over the diff followed by merge.

If there are still things to discuss after the maintainer approved, the maintainer may learn from that. A sort of internship program for future committers :)

globin:
> @Ekleog I think the UI situation is slightly different; for example, this very PR has a grey checkmark for an approval by an organisation member who doesn't have write access to this repository.

I actually have write access, so I guess it's green if one has write access and gray otherwise.

Zimmi48:
>I actually have write access, so I guess it's green if one has write access and gray otherwise.

They were not talking about you, I think. Anyway, your conclusion is the right one.

rfcs/0039-unprivileged-maintainer-teams.md
lheckemann: One potential side effect of "spamming" could be that people remove their maintainership to avoid the notifications. I'd consider this probably a good thing, as it promotes "freshness" of maintainership info and packages actually being maintained.

7c6f434c: Spamming in my comment meant asking me to review the exact same commit I had previously approved (I evaluated it faster than ofborg, I guess). Spamming in this thread is about resending team invitation before the person in question reacts.

One mention per actual change is not spamming and makes sense etc.

7c6f434c:
Re: status visibility policy: if we want the gray/green checkmarks to bear information about commit access, we might also want to encourage everyone to use checkmarks/change requests instead of mere comments.

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,176 @@
...
+maintainer information, and update the team. It will support a dry-run
+option.
+
+New members of the team will receive an invitation to join the GitHub
+organization.
+
+## OfBorg changes
+
+OfBorg will identify PRs which are approved by their maintainers, and
+add a special label `approved-by-maintainer`.
```
7c6f434c: 1. Are there drawbacks to do it independently from the rest of the RFC?
2. Should ofBorg also autolabel PRs _authored_ by corresponding maintainers?
3. Should there be special labelling for _change requests_ from corresponding maintainers?

7c6f434c:
To make Shepherds' job of recapping te discussion easier: does anyone think that there are unresolved issues being discussed that are not related to the points in my comment https://github.com/NixOS/rfcs/pull/39#issuecomment-458534763 ?

Also, is there anyone who objects against the RFCin principle, i.e. for any combination outcomes on the issues still being discussed?

rfcs/0039-unprivileged-maintainer-teams.md
ckauhaus: A certain amount of races will be inevitable. We should, of course, try to minimize time windows where eager human action is not detected.

zimbatm:
To recap the concerns:
1. the bots gets sudo access on the nixos org opening a new front for privilege escalation; what can we put in place to tighten this as much as possible?
2. the subtle changes in feedback that github gives to contributors; is it going to be clear that green review checkmarks don't necessarily mean merge access?

What we could do is write a script that implements (2), try it out for a bit and get feedback. For me it's difficult to foresee all the corner cases and if it's going to cause issues before trying this out. We don't change contributors that often that it needs to be run as a background task yet, there could be a big fat warning in the contributors list to ping @grahamc when the list changes :)

7c6f434c:
@zimbatm 

Re:visibility

Green checkmarks _do_ imply write access (right now… an UI update can change it)

But member badges will be even more confusing — not everyone who merges PRs sets membership to public, and now there will be people shown with a member badge to other members, and maybe some of them will make membership public and some will not (the current RFC doesn't advise anyway).

I think experimenting with (2) is OK, but we should explicitly document what visibility settings are recommended. And maybe ask people to use GitHub-UI approvals more. (One cannot hide commit access from an intentional adversary, because GitHub shows who merged and who pushed commits).

Re: list updates

I think update volume right now is high enough to be worth a cronjob per se, but (in my personal opinion — I do not have necessary access to do the work) low enough for the job to be a no-privilege process that obtains the list of maintainers, excludes already-seen entries, and provides a list for a trusted human to feed into actual invitation 2FA-ed UI.

Zimmi48:
>But member badges will be even more confusing — not everyone who merges PRs sets membership to public, and now there will be people shown with a member badge to other members, and maybe some of them will make membership public and some will not (the current RFC doesn't advise anyway).

I'm not sure if it was implicit in what you wrote but note that the public status of the membership only affects what non-members see.

7c6f434c:
I was implying that with the «to member»/«public» distinction, but thanks for stating it clearly — I wrote something confusing there, sorry.

zimbatm:
can I request a demo? add one read-only user to the org. then have that user review and approve one PR.

grahamc:
Sure, here @lheckemann is a member of a read-only team: https://github.com/eb6cb83a-cf75-4e88-a11a-ce117467d8ae/4ec54bd9-c707-4238-bc46-48596d6d7659/pull/4

if you're a member of the org, you see:

image

Zimmi48:
In this case, you see the same thing if you are not a member of the org.

zimbatm:
Hovering the check marks also tells: "only reviewers with merge access count towards mergability". So the read-only members are available as reviewers but not confuse people into thinking that the PR can be merged by the participant. Seems perfect to me.

In terms of design of the bot, it would be nice if it wasn't accessible from the internet. Running a hourly cron from a highly secured network is probably good enough and wouldn't leave the token too exposed.

7c6f434c:
Re: bot design: I guess in case full automation is chosen there are some hardening measures that could be considered and declared worth it or not.

I would start a completely non-exhaustive list with the following points:
1. Cronjob that pulls data from GitHub, machine doesn't accept any inbound connections except reasonably secured SSH.
2. Work split between two UIDs, first UID grabs the maintainer list, processes it, outputs a list of maintainer handles as newline-separated alphanumeric strings; the second UID has the token and refuses to touch it if it gets a list not matching '([-_a-z0-9]+\n+)+'.
3. GitHub TLS certificate is pinned, with manual OK for updates.

I would mildly prefer not to create a bot with powerful access in the first iteration.

I strongly oppose a fully automated implementation without (1), I think (2) is nice to have (but if it requires too much extra work, it is not a must-have), I am not yet sure if (3) is net positive even if implementation is cheap.

7c6f434c:
My position on the other two topics I listed as apparently not fully settled.

On ignored invitations: I think the ideal first iteration would have a list of sent invitations with dates somewhere, send only on invitation to each account, and let resending in special cases be handled manually.

The invitation should be revoked in the same cases where someone would be removed from the read-only team. (And I guess neither revocations nor removals would be performed automatically in the first version…)

On visibility: My preference is asking committers to make membership public (while respecting the opinion of those who will refuse), to not make any official recommendation on public/private membership in the read-only team, and to ask people to use approval features more, especially for the packages they maintain.

lheckemann:
# Shepherd team summary

After discussing the various issues raised in comment threads here, @grahamc, @ckauhaus, @alyssais and I have decided on these questions and would like to call the Final Comment Period.

## Automatic removal of team members when removed from maintainer list for package?
https://github.com/NixOS/rfcs/pull/39#discussion_r248378859

Read-only maintainer team membership is based on master and synced — so yes, team members are removed automatically.

_OK by all._

## Use of review requests (vs cc comments)?
https://github.com/NixOS/rfcs/pull/39#discussion_r248379378

We have decided to leave this as is. GitHub's review request mechanism communicates clearly what is desired of maintainers. Lock-in isn't really a major issue, the action taken can easily be swapped out. If desired, additional "backends" could be implemented by anyone through an event reexporter similar to events.nix.gsc.io.


## "Real jerk" phrasing?
https://github.com/NixOS/rfcs/pull/39#discussion_r248381162

We don't like jerks, but also don't feel that it's something that should be addressed directly in this RFC. This is why we have decided to remove the "real jerk" sentence, and tie exclusion simply to banning, the semantics of which we may or may not wish to discuss in a different RFC.

## Does GitHub send an Invitation repeatedly, or just on the first invite? (invitation spam)
https://github.com/NixOS/rfcs/pull/39#discussion_r248382160

Invitation emails are only sent once, so they are easy to ignore.

## Security impact of org:admin token?
https://github.com/NixOS/rfcs/pull/39#discussion_r248382022

We have concluded that a bot performing these changes is still the best solution.

Steps that will be taken to mitigate some of the risks:
 - Thorough code review
 - Bot will run on NixOS Foundation infrastructure, which is security-critical anyway due to the large-scale reliance on the official binary cache
 - If possible, membership change notifications. It seems [this was at one point a GitHub feature which could not be disabled, which has since been removed](https://github.community/t5/How-to-use-Git-and-GitHub/New-unstoppable-notification-email-about-new-org-team-members/td-p/3781).

## Branch scope of maintainership?
https://github.com/NixOS/rfcs/pull/39#discussion_r248396123

Membership of the unprivileged maintainer team will be synchronised with the master branch. Review requests are determined based on maintainership in the target branch. If no package maintainers are members of the GitHub team, reviews should be requested manually.

## Membership visibility?
https://github.com/NixOS/rfcs/pull/39#discussion_r248579822

People who are added as maintainers should be able to display their affiliation with Nixpkgs and feel proud of it.

7c6f434c:
(I understand the announcement as a plan to declare FCP after the updated version is added in this PR)

> We have decided to leave this as is. GitHub's review request mechanism communicates clearly what is desired of maintainers.

Given that in the current practice the majority of discussions happen without using GitHub's approvals and change requests, I would appreciate if the aim to use them more could be reflected maximally explicitly in the final version of the RFC.

> People who are added as maintainers should be able to display their affiliation with Nixpkgs and feel proud of it.

Do we want any reminder that the badge is only visible to members unless they explicitly set visibility to public in settings, or is it out of scope?

Zimmi48:
>If possible, membership change notifications

Should be doable with a webhook:

org-webhook

grahamc:
I have incorporated the meeting's results in to the latest commit. I think we're ready for FCP, what say you @lheckemann, @ckauhaus, @alyssais?

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,192 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+Mentioning people in GitHub comments is the main alternative. This has
+the major down-side of not receiving the support of [GitHub's UI
+for requested reviews](https://github.com/pulls/review-requested).
+
+
+# Reesolved questions
+[resolved]: #reesolved-questions
```
7c6f434c:
```suggestion
[resolved]: #resolved-questions
```

rfcs/0039-unprivileged-maintainer-teams.md
```diff
@@ -0,0 +1,192 @@
...
+   to be part of the team with invitations?
+   No.
+ - Do maintainers want to be part of this team?
+ - Will the requirement of 2FA cause a significant number of people to
+   not want to participate?
+ - How will we handle people who have been invited, but have not
+   accepted the invitation?
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
```
7c6f434c: This seems to list the same questions as the previous, «resolved» section. Should this even exist? Are some questions unresolved and misplaced into the previous section?

grahamc:
Thanks @7c6f434c sorry I pushed that. I've pushed a fixup commit.

7c6f434c:
Thanks for the quick fix!

Pushing all the substantial adjustments without blocking on five-fold re-proof-reading _is_ a good idea in my opinion. Especially if you incorporate the proofreading suggestions so quickly.

ckauhaus:
> I have incorporated the meeting's results in to the latest commit. I think we're ready for FCP, what say you @lheckemann, @ckauhaus, @alyssais?

Go

lheckemann:
We would like to **accept** this RFC. Final Comment Period is [Go!](https://www.youtube.com/watch?v=BHIo6qwJarI)

Barring any objections, the RFC will be accepted on **Monday, 11 March**. :tada:

oxij:
> - Will the requirement of 2FA cause a significant number of people to
>    not want to participate?

IMHO this warrants a discussion of any of the alternative CLI interfaces https://github.com/search?o=desc&q=command+line+interface+github&s=forks&type=Repositories for GitHub.

From my perspective, requiring 2FA is just a security theater (because nothing prevents maintainers from leaking their shared secret plaintexts in their dotfiles repos anyway and you can't force everyone use hardware tokens more than you can force them to use random passwords) but it is a huge pain in the ass to use when all you browser instances run in separate ephemeral VMs or similar (which is the only reasonable thing to do, but almost nobody does it). Arguably, anything involving encrypted random passwords with something like `pass` is "2FA" anyway: `pass` stores the correct random password, I remember passphrase from the corresponding `gpg` key.

Using a CLI interface seems to solve that particular problem for me, but I don't use the official `hub` because it's bloated and messes with `git`, but `git-hub` from the above list is already packaged in `pkgs/applications/version-management/git-and-tools/git-hub`, it doesn't mess with `git`, and it seems fine at the first and second glances through the code.

See also a short abstracts on related projects in https://github.com/sociomantic-tsunami/git-hub#id19.

In short, if you want to keep 2FA for the web interface and accepting an invitation will require enabling it, I think listing more usable alternatives in this RFC should be mandatory.

> - How will we handle people who have been invited, but have not
>   accepted the invitation?

This also needs discussion, IMHO. I planned to reject the invitation until I learned about `git-hub` tool for 2FA reasons (now I need to play with it first to know for sure), but other people might have other reasons.

lheckemann:
Discussion on the subject of 2FA is welcome; I don't see it as a reason to cancel the Final Comment Period, however, as you are the first to raise this concern and it still provides a benefit as long as there are people who don't mind using 2FA. I'd say this should probably be considered during the technical implementation.

7c6f434c:
>>  - How will we handle people who have been invited, but have not accepted the invitation?

I think commitment to avoid invitation-spamming people who do not accept invitation (see «Resolved questions» section) and lack of plans to change anything for those who do not accept makes it almost-no-downside change? This RFC does mandate slow and gradual roll-out that should be stopped if a mass problem appears. The plan is «give those who want some extra notification channels; allow the rest just filter their GitHub email by maintained package names; see what happens».

> 2FA

1. 2FA can be considered either a security theater, or password manager enforcement. I agree that the recommended way of using 2FA (even with HW tokens!!) makes it no more secure than proper usage of password managers, but makes a password manager the least annoying solution. We had this discussion in the 2FA issue, and I agree this RFC is a good reason to open an issue on documenting 2FA use approaches.
2. Ephemeral VMs require you already to relay data from password manager in, so just add oathtool to the mix.
3. Plain Git with SSH keys, or plain Git with a GitHub-issued push token used instead of password work with 2FA enabled.

(I agree that both points by @oxij deserve better documentation, but agree with @lheckemann that this documentation should be a part of implementation PR or a separate change, not an FCP-breaking concern).

grahamc:
FCP has passed!

FRidh:
GitHub added a permission level called "Triage" that allows members to apply labels and close PR's. https://help.github.com/en/articles/repository-permission-levels-for-an-organization

I propose we amend this RFC to modify the permission level from Read to Triage.

zimbatm:
RFCs are only meant to record a decision, not as the source of data. Once the RFC is merged we update documents / implement the decision. If the change doesn't have deep implications I think we can also skip the RFC process.

grahamc:
The first todo is merged (NixOS/nixpkgs#66361)

> add GitHub IDs to the maintainer list

grahamc:
Second TODO is done, adding the nixpkgs-maintainers team https://github.com/orgs/NixOS/teams/nixpkgs-maintainers/edit

grahamc:
Third todo is done, my tool can now sync maintainers.nix to a github team https://github.com/grahamc/rfc39/commit/9b5ebdc73751a3fca56592f0cfe6845d40851315

grahamc:
>  Write an explanatory post on Discourse about the what-and-why of this plan.

Done.

grahamc:
I've selected a very small initial group:

 - symphorien
 - asymmetric
 - volth
- evanjs
- thefloweringash
- Hodapp87
- dasJ
- matthiasbeyer
- petabyteboy
- tiplner
- Taneb

I am inclined to add these to the list today, and then possibly wait less than a half week to add another small sample, before moving on to

>  Expand the group to one quarter of the maintainers, and pause a half a week to gauge response.

ryantm:
That group looks good to me, and I think you should go for it!

grahamc:
I'm delighted to say that the permissions grantable for the new team management APIs are much finer grained than before, requiring a GitHub Application instead. The credentials used for this implementation are now restricted to a much finer subset than we could have done when this RFC was authored. The scope it has can be read about here: https://developer.github.com/v3/apps/permissions/#permission-on-members

grahamc:
First group has been invited:

image

with revision d76b4e2eb0f17793b305f746c8fbd8747c6cc46f of https://github.com/grahamc/rfc39/

grahamc:
So far 26 people have been invited, and 20 have joined. One person contacted me confused about what the invitation was for. We may want to send an email to people before inviting them, but I'm inclined to wait and see how the next ~20 people go.

grahamc:
As of now there are 30 members of the team, and I just invited another ~~25.

grahamc:
There were 41 members with 22 invitations, and I invited more so now there are 43 members and 41 invitations.

grahamc:
I've gotten cold feet on going in 250 person invitation sprees,  so I've been doing 25-person groupings.

alyssais:
another data point: I heard from somebody who was at some point aware of this RFC that they were confused why they’d received the invitation.

alyssais:
> At 148 members and 138 invitations I've started to get an error:

Is that happening consistently?

grahamc:
We're within ~110 people of having everyone invited. Currently at 460 members, 453 pending invitations.

7c6f434c:
Currently ~700 members (out of ~900 target size?). Is it true that everyone has been invited?

Zimmi48:
> Currently ~700 members (out of ~900 target size?). Is it true that everyone has been invited?

Not everybody decides to accept the invitation. The 2FA requirement in particular could turn off some people.

7c6f434c:
@Zimmi48 I understand that; and 700 out of 900 is exactly the number where both «not everyone agrees to deal with 2FA/not everyone has got around to accepting» and «the slow rate-limit-fighting process has not converged yet» are plausible

grahamc:
As of now, all the to-do items on this RFC are complete.

* the code is now owned by NixOS: github.com/nixos/rfc39
* the infra is automatically running the update tooling every 30 minutes: https://github.com/NixOS/nixos-org-configurations/commit/ef48be95f5d30880f41f481c62148ac302d3604a
* there are graphs for its status: https://status.nixos.org/grafana/d/w8HHOCBWz/rfc39-maintainer-sync?orgId=1&from=1577147422732&to=1577156793325

As of now, the team has 733 members, with 376 pending invitations. About 100 new invitations were sent this evening, catching up with new maintainer entries over the past several months. Several people cannot be invited:

* 166 are a missing a `githubId` values (note: do not blindly add these, see note later)
* 43 are missing a `github` value (note: do not blindly add these, see note later)
* 2 people seem to have declined the invitation or blocked the repo or something where GitHub says we haven't invited them, but also we can't invite them.
* 1 person is blocked from being invited by GitHub due to U.S. trade controls law restrictions.
* Additionally, I accidentally removed (and re-invited -- sorry!) about 10 people as they didn't have a `githubId` set. I added their ID back.

### Adding a `github` or `githubId` value after the initial addition:

You can't just go look up GitHub handles and assume it is right. Instead:

 1. Find a maintainer without a githubId
 2. Look in the git blame output and find that section
 3. Look up that commit on GitHub and see who GitHub thinks authored it.
 4. Find that user's ID (https://api.github.com/users/«username»)

This is because a surprising number of GitHub users change their account names!

Mic92:
Thanks a lot for all the hard work!

alyssais:
> * 1 person is blocked from being invited by GitHub due to U.S. trade controls law restrictions.

Yikes.

grahamc:
This is requiring more work, as GitHub is now allowing us to send additional invitations after the first one. They have now purgedthe database of invited users twice in two months now. I have stopped the sync until I can add some sort of database.

image

grahamc:
A PR using a database of invited people has been created and merged. We're doing automated invitations again. If you didn't want to join, this should be the last invitation you receive. See: https://github.com/NixOS/nixpkgs/issues/107815

PR #40, opened 2019-02-06
[RFC 0040] "Ret-cont" recursive Nix

Ericson2314:
"Ret-cont" recursive Nix is a restricted form of recursive Nix, where one builds a derivations instead of executing builds during the builds. This avoids some platform-specific contortions relating to nested sandboxing. More importantly, it prevents imperative and overly linear direct-style build scripts; easy to write but throwing away the benefits of Nix.

[Rendered](https://github.com/Ericson2314/nix-rfcs/blob/ret-cont/rfcs/0000-ret-cont-recursive-nix.md)


Profpatsch:
I don’t entirely understand the proposal.

From experience, naïvely recursive systems are usually extremely constrained, so I can see that problem. You should split out the problem with the current proposal from “Motivation” into its own section and start with that.

Can you frame the idea in terms of already existing systems? Scheme’s `call/cc` comes to mind, but from the description it is unclear to me whether the ideas are similar. We shouldn’t implement systems like that without looking at previous work and theoretical underpinnings first.

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,102 @@
...
+Additionally, see https://github.com/edolstra/nix/commit/1a27aa7d64ffe6fc36cfca4d82bdf51c4d8cf717 for Eelco's draft implementation of recursive Nix, and the Darwin sandboxing restriction.
+Sandboxing and Darwin are crucial to Nix today, and we shouldn't sacrifice either of them.
+With "ret-cont" recursive Nix, actual builds are never nested, so we don't need any fancy constraints on the derivation "runtime" (i.e. the code that actually performs and isolates builds).
+
+# Detailed design
+[design]: #detailed-design
+
+Derivations building derivations should have some special attribute indicating this, and two outputs, "store" and "drv".
+"store" would be a local Nix store limited to just drvs and fixed output builds.
+"drv" would contain a symlink to one the derivations in the store, the root.
```
Mic92:
```suggestion
"drv" would contain a symlink to one of the derivations in the store, the root.
```

Mic92:
So in a nutshell instead of importing nix files from a derivation,
this will import derivation files from a derivation?
The idea sounds to me the equivalent to autotools/automake, where before a release the `configure` and `Makefile.in` files are generated and put in the release tarball.

edolstra:
It would be good to add an example and/or some use cases where this approach would apply. For example, I can see it working in the case where you want to build a single derivation (e.g. building Hydra from the `release.nix` extracted from the Hydra tarball). But it wouldn't address the scenario we have at Tweag where we want to be able to call Bazel from Nix and then have Bazel call Nix to fetch some packages.

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,102 @@
...
+Just as we can CPS-transform programs, reifying the context of a function call as another function passed as an argument, so we can imagine splitting the derivation in two at this blocking point.
+This gives the "continuation" part of the name.
+
+But whereas the CPS transformation makes the continuation an argument, the Nix *derivation* language is first order.
+Instead, we can produce a derivation which has the callee as a dependency, and continuation drv downstream depending on it.
+Since the outer derivation evaluates (builds) the inner derivation rather than calling anything, I deem that it returns the derivation.
+This gives the "return" part of the name.
+
+I've always been concerned with the ease of which someone can just "nix-build ...; nix-build ...; nix-build ..." within a derivation with recursive Nix.
+This creates a linear chain of dependencies, which isn't terribly performent: shorter critical paths are crucial for parallelism and incrementality and this fails with both.
```
siddharthist:
```suggestion
This creates a linear chain of dependencies, which isn't terribly performant: shorter critical paths are crucial for parallelism and incrementality and this fails with both.
```

Ericson2314:
@Profpatsch 

> From experience, naïvely recursive systems are usually extremely constrained, so I can see that problem. You should split out the problem with the current proposal from “Motivation” into its own section and start with that.

Yeah, sometime I swear have to open the PR before I can see all the places I dropped in media res to no one's benefit. I'll rewrite things in the next draft to fix that.

> Can you frame the idea in terms of already existing systems? Scheme’s `call/cc` comes to mind, but from the description it is unclear to me whether the ideas are similar. We shouldn’t implement systems like that without looking at previous work and theoretical underpinnings first.

There's a notion of applicative vs monadic build systems. The DRV langauge today is strictly applicative:
```haskell
data Drv a -- derivation which makes 'a'
type CCompiler = CCode -> Binary
iCanWriteThis :: Drv (CCode -> Binary) -> Drv CCode -> Drv (CCode -> Binary)
```
This proposal makes it monadic by defining join:
```haskell
-- | switch to inner after building outer
join :: Drv (Drv a) -> Drv a
```
Maybe I should make note of this in the RFC :).

-------

Fun fact, operationally, that `join` is also a lot like [Reflex FRP's `switchHold`](http://hackage.haskell.org/package/reflex-0.5/docs/Reflex-Class.html#v:switchHold). (Ignore the final `m`, which could be thought of as an ambient interpreter.)

Ericson2314:
@Mic92 Yes, you can imagine trying to run Nix to the point where there are no drv-producing drvs left, which is a lot like that process those autoconf projects do before release.

Ericson2314:
@edolstra 

> It would be good to add an example and/or some use cases where this approach would apply.

Sure, I'll do that with the rephrase @Profpatsch was thinking of.

> For example, I can see it working in the case where you want to build a single derivation (e.g. building Hydra from the `release.nix` extracted from the Hydra tarball).

Indeed it would.

> But it wouldn't address the scenario we have at Tweag where we want to be able to call Bazel from Nix and then have Bazel call Nix to fetch some packages.

Well, with this + intentional store, you all would hopefully have no need for Bazel anymore :). But upstream projects will still use it and so I'll go over how we can build them more officially in Nixpkgs.

For starters, you can have a `bazel2nix` that will convert the (1st order) Bazel build graph to Nix derivations. This is just like instantiating Hydra's `release.nix`, but with more hoops to jump through first (`bazel2nix`).

(A brief aside: A naive `bazel2nix` would get GCC for whatever from the PATH, like a configure script, which means we'd have to build GCC first before any bazel2nix step, which reduces parallelism. Much better is to feed `bazel2nix` GCC's drv file, so it can just inject that as a dependency in the derivations produced from individual Bazel build nodes that use GCC. Then we get maximal intra-project parallelism. Both of these possible `bazel2nix`s could be written today, this RFC only means that either implementation would be usable from within a derivation.)

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,102 @@
...
+everything in the previous paragraph just describes marking derivations and post-processing their build results.
+Because drvs and produce plans of drvs producing more drvs ad-infinitum, it's possible to never terminate but that's the user's fault.
+We can detect simple cycles analogous to black holes in thunks: if a derivation produces a redirected derivation depending on the original, a cycle is effectively recreated even though we don't have a hash fixed point.
+\[Instead, the drv->drv substitution would never terminate.]
+Nix can should raise an error rather than looping, but either behavior is permissible.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+ - The opinionated nature may put of those who think Nix is too hard to learn already, and think simple recursive "nix-build" is good for newcomers.
```
globin:
```suggestion
 - The opinionated nature may put off those who think Nix is too hard to learn already, and think simple recursive "nix-build" is good for newcomers.
```

globin:
Per meeting of the @NixOS/rfc-steering-committee this RFC is open for shepherding nominations.

shlevy:
I'd like to be considered for the shepherd team.

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,102 @@
...
+
+ - Use a socket to talk to the host daemon.
+   https://github.com/edolstra/nix/commit/1a27aa7d64ffe6fc36cfca4d82bdf51c4d8cf717, a draft implementation of full recursive Nix, has done this and we can take the details from that.
+   This might sightly more efficient by reducing moving files, but is conceptual overkill given this design.
+   No direct access to the host daemon rules about a bunch of security concerns, and simplifies the interface for non-Nix tools producing derivations.
+   The latter I very much hope will happen, just as Ninja is currently used with CMake, Meson, etc., today.
+
+ - Full recursive Nix (builds within builds)
+
+ - Important from derivation.
```
siddharthist:
```suggestion
 - Import from derivation.
```

volth:
To the list of alternatives: ```builtins.exec``` producing either derivations or vendored .nix files (Pros: easy and straighforward; no problem with nested sandboxing; Drawbacks: not frienly to parallel and remove builds)

Ericson2314:
@volth Your `builtins.exec` sounds like import from derivation as it exists today. Is that the case, or am I missing something?

Ericson2314:
I rewrote the motivation, adding examples as proposed. Next I'll expand the detailed design with some formality.

volth:
> @volth Your `builtins.exec` sounds like import from derivation as it exists today. Is that the case, or am I missing something?

The main difference is the "build" code executed as user, not as "nixbld1" and so never in the sandbox

Ericson2314:
@volth Oh right. I forgot that it already exists. I barely consider it an alternative since it's wildly impure and has all the same eval-lengthening issues as import-from-derivation, but I'll add it anyways.

Ericson2314:
Also, the detailed design is rewritten with some half-baked formalism. I'm sure it could be improved, but it hopefully it already helps.

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,245 @@
...
+```
+This further simplifies the implementation.
+Derivations remain built exactly as today, with only logic *between* building steps that is entirely platform-agnostic changing.
+
+# Detailed design
+[design]: #detailed-design
+
+Derivations today build outputs, and are associated to those outputs.
+We extend the derivation language by allowing a derivation to indicate their output is more derivations, and ultimately be associated with one of *those* derivations's associated outputs.
+Derivations that that do so indicate this with some special attribute, say `__recursive`.
```
Mic92:
```suggestion
Derivations that do so indicate this with some special attribute, say `__recursive`.
```

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,245 @@
...
+Such derivations must have two outputs, `store` and `drv`.
+`store` would be a local Nix store limited to just drvs and fixed output builds.
+`drv` would contain a symlink to one of the derivations in the store, the root.
```
shlevy: Note that ideally for recursive drvs you also have the actual outputs of the final derivation specified, so the nix-lang usage of them can just reference outputs appropriately

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,245 @@
...
+## Design Notes
+
+There's a few things we can call out from the faux-formalization.
+
+ - `isTrustlessStore` is called that because the restricted on the contents—fixed output builds / plain data and drvs—is fully and cheaply verifiabled.
+   This is in contrast to normal builds,
+   where the relationship between the derivation and build can only be verified by redoing the build,
+   and where even then there's no way to know whether to blame the output for being actually malicious, or the derivation for merely being non-deterministic.
+
+ - The substitution of drvs in a downstream derivation reminds me of the substitution of drvs for content hashes with the intensional store.
```
shlevy: Yes, this was one of my comments on that PR. We should figure out how to handle reductions more generally: In the continuation-based recursive nix case, we have `foo.drv → foo'.drv` whereas in the intensional store case we have `(foo.drv)!out → cashash-foo`, but the principle is the same.

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,245 @@
...
+   To me though, this is a feature not a bug; I don't want to go in that direction just yet.
+
+# Alternatives
+[alternatives]: #alternatives
+
+ - Don't allow fixed-output builds.
+   All data can be stuck inside the drv file, so this can be cut without limiting expressive power.
+   But this is much less efficient, and more cumbersome for whatever produces the data.
+
+ - Use a socket to talk to the host daemon.
```
shlevy: I'm *hugely* in favor of this approach. Using nested stores and all that is kind of hacky, and we have lots of reasons besides actually nested builds to want recursive nix (e.g. store querying, dynamic dependency fetching, etc.). In general I'm in favor of moving toward a much more structured way for builds that are nix-aware to talk to the store; builds that *aren't* can then use generic functionality built on top of that.

rfcs/0000-ret-cont-recursive-nix.md
Ericson2314: Ah yes that's true.

Ericson2314:
Yes I should have put somewhere that no variation of recursive Nix is truly new, @shlevy has implemented them all at some point :D.

rfcs/0000-ret-cont-recursive-nix.md
Ericson2314: Yeah I don't really care soooo much on this point. I guess I am more interested in pointing out that we *can* skip the daemon socket, as proof of the simpler computational model that is this version, than insisting that we *must* skip the daemon socket.

shlevy:
Eelco has been nominated and accepted the nomination. @dtzWill @copumpkin are you guys game for joining as well?

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,250 @@
...
+The normal variant in the examples above might be termed "direct-style" recursive Nix.
+Consider what happens with the recursive "nix-build" in those examples:
+the outer build blocks while the inner one builds, and then the other one continues.
+Just as we can CPS-transform programs, reifying the context of a function call as another function (which is passed as an argument), so we can imagine splitting the derivation in two at this blocking point.
+This gives the "continuation" part of the name.
+But whereas the CPS transformation makes the continuation an argument, the Nix *derivation* language is first order.
+Instead, we can produce a derivation which has the callee as a dependency, and continuation drv downstream depending on it.
+Since the outer derivation evaluates (builds) the inner derivation rather than calling anything, I deem that it returns the derivation.
+This gives the "return" part of the name.
+Both differences together, the first example becomes something like:
```
ocharles: But the first example had two outputs - this one has only one. Does this mean I can only recurse once?

Ericson2314: You can recurse as many times as you want. That one output is the derivation itself, not some output of it. After you've been rewritten into the final derivation (i.e. a non-recursive one), you produce all of it's outputs.

ocharles: Ok, I suppose my question more meant: can you provide behaviour equivalent to your first example. If so, could that be demonstrated in this RFC?

Ericson2314: The later ones are equivalent. Perhaps my typo I fixed in https://github.com/NixOS/rfcs/pull/40/commits/5e56f21125faa6e92ff8269e8fd35655ad9ab958 was causing confusion?

Mic92:
I would like to nominate @layus as shepherd for this one.

shlevy:
I'd like to nominate @vcunat if he's available

vcunat:
Uh, I'm not sure; I'll try to consider during this weekend.

globin:
The @NixOS/rfc-steering-committee has chosen @edolstra, @shlevy, @copumpkin and @fpletz as shepherds for this RFC

domenkozar:
@fpletz is the leader

Mic92:
I am really looking forward to this RFC. PRs like https://github.com/NixOS/nixpkgs/pull/60788 worries me in term of large nix expressions.

fpletz:
How about an initial video call on Thursday at 1230 UTC, after the RFCSC meeting?

Ericson2314:
Well, to me this is surely worth waking up a little bit early for :).

edolstra:
@Mic92 BTW, something like NixOS/nixpkgs#60788 seem like a great use case for flakes. I.e. mastodon would be its own flake (outside of nixpkgs) while still being easy to use in a NixOS configuration. Same for tools like yarn2nix.

Ericson2314:
@edolstra I wouldn't want to commit generated code in any repo, not merely avoid committing it in nixpkgs.

rfcs/0000-ret-cont-recursive-nix.md
```diff
@@ -0,0 +1,250 @@
+---
+feature: ret-cont-recursive-nix
+start-date: 2019-02-01
+author: John Ericson (@Ericson2314)
+co-authors: (find a buddy later to help our with the RFC)
```
domenkozar:
```suggestion
co-authors: (find a buddy later to help our with the RFC)
shepherd-leader: Franz Pletz
shepherd-team: Franz Pletz, Eelco Dolstra, Daniel Peebles
```

Mic92:
@edolstra While it is nice for many reasons to have flakes, I would rather prefer to have mastodon in nixpkgs itself. One of the strength of a mono-repo is that coordination when updating packages is easier. Even when we not always can ensure that everything builds (also we try hard), at least it evaluates. Sure, one can always pin nixpkgs, but then you security updates becomes a hassle, we already have docker for that...

shlevy:
@fpletz Can you include me in the next chat? I'm not on IRC much anymore :frowning:

domenkozar:
@Ericson2314 could you merge my suggestion to add shepherds? That way I can write the tooling to generate RFC status report.

domenkozar:
Thanks <3

Profpatsch:
What’s the status of this?

Ericson2314:
After last meeting, it was agreed I should write up some small examples for the use-cases. I've just been super busy working on GHC as far as my programming free time goes, and haven't gotten back to this :(.

Ericson2314:
Oh, I should add to the proposal that Ninja, of all things, added something like this: https://github.com/ninja-build/ninja/pull/1521 . This was needed to do upcoming C++ modules.

Whereas before I advocated ninja2nix with IFD, and depfiles (a lousy dynamic dependency tool where "initial builds" and "rebuilds" are unavoidably distinguished), I would now advocate using ninja2nix. I think there is a real opportunity to gain acclaim from the C++ modules fiasco by beating Bazel, Ninja, Shake, and the rest to do both sandboxing and dynamic dependencies; that's a whole lot of developers discovering a need for the far corner of the "package manager cube" at once.

Ericson2314:
OK! I finally added the examples, expanded future work (including reference to @edolstra's goal of caching pure eval), and threw in the Ninja/C++ news for good measure. Let's go!

layus:
@Ericson2314 Thanks for the examples. This is way more clear.
In a sense, this whole ret-cont setup is a huge improvement for the IFD mechanism. With automatic "outer" derivations, no-one would ever need to write `__recursive = true` by themselves, much like no-ones ever writes `derivation {...}`. This makes the "opinionated nature of [ret-cont]" a no-brainer for users "who think Nix is too hard to learn already".

Overall, I am very enthusiast about this RFC, and it seems to be relatively easy to implement compared to recursive Nix. The implementation could be a quick win indeed :-)

Finally, there is a small annoyance having to re-evaluate nixpkgs.path from source everytime. And possibly some need for isolation between the imported derivation and the outer context. Or at least some recursion limit.

Ericson2314:
@layus Thanks for the enthusiasm! Though, I must caution that I don't quite know how to do the automatic IFD convesion so that part isn't in this RFC proper.

> Finally, there is a small annoyance having to re-evaluate nixpkgs.path from source everytime.

So for automatic IFD we could cheat and generate a random drv path for the stuck term, which we impurely evaluate using the current heap. I don't love this, as it's a complete 180° from trying to make evaluation less special and less trusted, but I should point it out.

> And possibly some need for isolation between the imported derivation and the outer context.

They are fully isolated by construction with the "`store` output" method proposed by this RFC. With the "talk to the the daemon" alternative method based on https://github.com/NixOS/nix/pull/3205, I think the restrictions from that PR are probably sufficient, but yes one must be careful. Daemons tempt!

> Or at least some recursion limit.

We can also do cycle detection (a la thunk black-holing). I should add both of those to the RFC, but to be clear I think it's fine if the initial implementation has neither as long as by the time Nixpkgs can use it, it does.

domenkozar:
Now that https://github.com/NixOS/nix/pull/3205 is merged, what's going to happen with this RFC?

edolstra:
Well, NixOS/nix#3205 is experimental so it shouldn't stop work on alternatives. Ret-cont Nix addresses different use cases I think (NixOS/nix#3205 is primarily useful for things like nix-ccache where you have a tool like make calling nix as a caching layer).

Ericson2314:
That's fine with me. Should we might more about this or ought we to have a prototype implementation / do we even need RFCs for experimental features or only when they become stable.

I suspect the answer is "no", so a lack of meeting on #56 is actually holding me back more than a lack of meeting in this.

Ericson2314:
Nix finally has some competition with https://www.usenix.org/conference/atc19/presentation/fouladi .

The deal with dynamism exactly as this RFC proposes. In their words "a thunk can write another thunk as its output"; just substitute "thunk" for "derivation", and you have this RFC.

volth:
@Ericson2314 https://discourse.nixos.org/t/thousands-of-lambdas/3532

Ericson2314:
Thanks @volth

Mic92:
@fpletz did you to create a new meeting for this one?

Ericson2314:
@Mic92 While I am happy to talk about this again, I was just gong to implement it after I work on CA derivations. The basic idea for both is clear to me, but in doing the programming little intricacies arise that I would never need to anticipate.

I suppose the main thing I'd like to discuss would be a new process RFC ---- getting PRs merged for Nix has been incredibly slow because it is unclear who can review and merge besides Eelco, and Eelco is busy with many other things. We need a "chain of command" for Nix, deciding who is responsible for reviewing what sort of changes.

spacekookie:
@fpletz @edolstra @shlevy @copumpkin Considering there hasn't been any activity on this RFC for a while, maybe you should consider appointing a new shepherd leader for this RFC to get things going again

Ericson2314:
In addition to what I said above about doing this after CA derivations, there are a few things I want to revise here and an intermediate feature/RFC we could do. Given all that, I am going to close this for now to be reopened when the time is ripe.

roberth:
I like that ret-cont avoids the problem with fixed-output derivations in recursive Nix (as implemented). See https://github.com/NixOS/nix/pull/3205#issuecomment-879975406

roberth:
> This proposal makes it monadic by defining join:
> 
> ```haskell
> -- | switch to inner after building outer
> join :: Drv (Drv a) -> Drv a
> ```

I believe it is more like `Category` and `Arrow`. Let me explain.

There's no Nixlang Functor for `Drv` because its type parameter `a` has a libstore kind, not a libexpr kind. The truly Monadic operation on derivations is import from derivation, which works for any Nixlang expression `expr`. Its type *in* the language is `import :: forall a :: Expr. Drv a -> a`, which, combined with Nix's implicit goal of producing derivations in the end gives the interpreter the ability to `Drv (Drv a) -> Drv a`. (`import` shouts Comonad, but let's skip that tangent*)

Ret-cont recursive Nix does not allow IFD. If it were implemented in the language, like RFC 92, it would look like `outputOf :: forall a :: StorePath. Drv2 (Drv2 a) -> Drv2 a`, where `Drv2` is an awkward type that is more about what happens in the sandbox than what happens in the language. What happens in the sandbox is not actually accessible to the language proper, so it is better modeled as an opaque type with some operations that define its behavior, so we get `derivation : Build a b` where you can think of build as something akin to `type Build a b = List a -> Drv b`. Now we're really in `Arrow` territory. We can even define `arr` by running the evaluator in the sandbox, no wonder there's a separate class hierarchy, because `Build` just joins the list of cases where `arr` is not desirable. What we really want to see is `(.)` of `Category`, which is `Build a b -> Build b c -> Build a c`. Applying our hand-wavy type alias definition, we get `(List a -> Drv b) -> (List b -> Drv c) -> List a -> Drv c`, which after a function application becomes `Drv b -> (List b -> Drv c) -> Drv c`, so we have to build path `b` (`build : forall p :: Path. Drv p -> p`) in order to get at `Drv c`. This is what ret-cont does, all at the derivation level, not the Nixlang level.

The power spectrum works out the same way. By caller power we have `Functor (f b) < Applicative (f b) < Arrow f < Monad (f b)` and similarly `traditional Nix < ret-cont recursive Nix < import from derivation`. Note that callee features are inversely related to caller power. In Haskell, in-language static analysis ends at Arrow (and/or Selective); in Nix build-free evaluation ends at ret-cont recursive Nix*.

*: I'm not including full recursive Nix, I should probably look into Comonad some more and all my steps are too hand-wavy. However, my comment ends here.

Ericson2314:
@roberth I think `a` could be the "type" of the derivation outputs. At least with subtyping `Drv <= Bytes`, `Drv (Drv Bytes) -> Drv Bytes` makes some sense. But yes it doesn't have the same super-linearizing feel as Monad as used in practice in Haskell, and I do think arrows are underrated if only we'd fix them up like as in https://arxiv.org/abs/1007.2885

PR #41, opened 2019-03-10
[RFC 0041] SELinux Support

e-user:
As suggested by @Mic92 in NixOS/nix#2670, I would like to propose this [RFC](https://github.com/NixOS/rfcs/blob/892f3f3e328f1858baef1f7e66844de6bf27e560/rfcs/0041-selinux-support.md) for implementation in Nix and NixOS.

@fishilico @etbe @mstone for your consideration, as well.

### DIFF
--- a/rfcs/0041-selinux-support.md
+++ b/rfcs/0041-selinux-support.md
@@ -1,0 +1,159 @@
+---
+feature: selinux-support
+start-date: 2019-03-10
+author: Alexander Kahl (@e-user)
+co-authors: (none yet)
+related-issues: NixOS/nix#2374, NixOS/nix#2670, NixOS/nixpkgs#56965
+---
+
+[NixOS/nix#2374]: https://github.com/NixOS/nix/issues/2374
+[NixOS/nix#2670]: https://github.com/NixOS/nix/pull/2670
+[NixOS/nixpkgs#56965]: https://github.com/NixOS/nixpkgs/pull/56965
+
+# Summary
+[summary]: #summary
+
+Add support for [SELinux] to Nix and NixOS. This means that Nix will be
+able to write SELinux file contexts on enabled systems and NixOS will be
+provided with a base policy that supports as many use cases as possible. New
+options in NixOS make it possible to build Linux with SELinux support enabled.
+
+[SELinux]: https://selinuxproject.org/page/Main_Page
+
+# Motivation
+[motivation]: #motivation
+
+Security Enhanced Linux is a Linux kernel security feature providing mandatory
+access controls. It confines processes to access files with corresponding labels
+for specific purposes, such as reading and writing.
+
+There is currently one family of Linux distributions with SELinux enforcing
+active by default: [Fedora] and derivatives, which includes [Red Hat Enterprise
+Linux], [CentOS], Scientific Linux and others. Several other Linux distributions
+have optional support for SELinux, including [Arch], [Debian], [Ubuntu], CoreOS,
+and [Gentoo].
+
+Without explicit file contexts written to files in the Nix store and Nix
+profiles, Nix-provided daemons currently cannot integrate with SELinux-enforcing
+systems, as trying to run them from systemd or interacting with them with
+systemctl results in kernel-level denials. This includes related issue
+[NixOS/nix#2374].
+
+Additionally integrating support for SELinux in NixOS will bring an additional
+level of hardening to the distributions, which increases its aptitude as a
+server platform but also as a professional workstation distribution.
+
+[Fedora]: https://fedoraproject.org/wiki/SELinux_FAQ
+[Red Hat Enterprise Linux]: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/selinux_users_and_administrators_guide/index
+[CentOS]: https://wiki.centos.org/HowTos/SELinux
+[Arch]: https://wiki.archlinux.org/index.php/SELinux
+[Debian]: https://wiki.debian.org/SELinux
+[Ubuntu]: https://wiki.ubuntu.com/SELinux
+[Gentoo]: https://wiki.gentoo.org/wiki/SELinux
+
+# Detailed design
+[design]: #detailed-design
+
+The RFC can be broken down into two pieces. One is support for SELinux in Nix,
+which solves the problems outlined above and serves as the enabler for the
+second piece, SELinux support in NixOS.
+
+Enabling Nix to write SELinux file contexts has already been put up as a PR,
+[NixOS/nix#2670]. Before writing any new file to the Nix store or creating a new
+profile, Nix checks for the designated file context and directs the SELinux
+kernel interface to write that label through SELinux library calls when creating
+the file. This corresponds to that other tools with support for SELinux are
+doing, including package managers and systemctl. All calls to SELinux-related
+functions are optional and dependent on compile-time presence of the
+corresponding SELinux libraries.
+
+For SELinux itself in order to provide sensible file contexts to Nix,
+[modifications] to its userspace libraries and tools have to be
+undertaken. These changes will enable the existing set of base policies for
+SELinux to be re-applied to the Nix store. Careful collaboration with SELinux
+maintainers will ensure the success of this endeavour. One crucial base policy
+change for Nix is the adoption of `/bin` for all binaries which are commonly
+installed under `/sbin` in [FHS-compliant] distributions, such as `httpd`.
+
+[modifications]: https://lore.kernel.org/selinux/7853167.K65cXu0y11@neuromancer/T/#u
+[FHS-compliant]: https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.pdf
+
+Lastly, while the above will make sure Nix can fully integrate with other
+SELinux-enforcing Linux distrutions, NixOS itself will have to be provided with
+a base version of the SELinux policy as well as options to enable SELinux
+compilation support for its Linux builds. This requires building the policy
+using common utilities and installing it into `/etc`. All the required libraries
+and userspace utilities for SELinux are already present in Nix but will need to
+be [updated].
+
+[updated]: https://github.com/NixOS/nixpkgs/pull/56965
+
+Test suits will also be put into place for Nix and NixOS, respectively, to make
+sure file contexts are applied and are enforced correctly.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Obtaining SELinux labels and writing them to files slows down Nix considerably
+in the current iteration of the PR. This might or might not be due to incorrect
+implementation details, such as lack of re-use of context structs. Other than
+that, the only way to mitigate this is to write SELinux labels only on enabled
+systems.
+
+Turning NixOS into an SELinux-supporting Linux distribution is not a one-shot
+project but a continued endeavour, occasionally requiring close collaboration
+with upstream developers. While not a drawback per se, it means that overall
+long-term maintenance of the distribution will be increased. One possible
+contigency plan can be to drop SELinux support in NixOS at any time but keep
+the support in Nix.
+
+While re-using and mending the existing SELinux base policy eases the initial
+adoption greatly, it is not particularly idiomatic for Nix. The policy includes
+corner-cases for every supported distribution of Linux and is built using M4
+macro processing. Turning the build process into something that can be steered
+from a Nix expression might be viable, but is a very long shot which would
+alienate NixOS from SELinux upstream development.
+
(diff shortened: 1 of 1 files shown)


rfcs/0041-selinux-support.md
```diff
@@ -0,0 +1,159 @@
...
+active by default: [Fedora] and derivatives, which includes [Red Hat Enterprise
+Linux], [CentOS], Scientific Linux and others. Several other Linux distributions
+have optional support for SELinux, including [Arch], [Debian], [Ubuntu], CoreOS,
+and [Gentoo].
+
+Without explicit file contexts written to files in the Nix store and Nix
+profiles, Nix-provided daemons currently cannot integrate with SELinux-enforcing
+systems, as trying to run them from systemd or interacting with them with
+systemctl results in kernel-level denials. This includes related issue
+[NixOS/nix#2374].
```
zimbatm: As I understand it, the real motivation of this RFC is to allow to run Nix on SELinux-enabled distributions. This is what I am inferring here but it might be worth writing it down explicitly.

e-user: That was also my original motivation, indeed - it's however just another step from there to full SELinux support in NixOS, which is why the RFC proposes to be split up for the second step, as an alternative.

e-user: @zimbatm does the new addition help to clarify the motivation?

zimbatm: thanks it does

Ekleog:
I have doubts on the idea of running Nix on SELinux-enabled distributions. What I currently understand appears to imply that binaries will be trusted based on their path relative to the root of their derivation: `/nix/store/1234567890abcdef-httpd/bin/httpd` would take into account only `/bin/httpd`.

This is insecure, as any user with access to the nix daemon could request such a file to be built with whatever content.

As I guess I misunderstood something, would it be possible to make the threat model and the currently-envisioned solution explicit?

Currently, I think I can see how to make NixOS interoperate cleanly with SELinux (by generating the policy as part of the module system), but I cannot see yet how to make Nix interoperate cleanly with a SELinux-enabled system.

edolstra:
I've never used SELinux, so take the following with a grain of salt, but it's my understanding that SELinux is a poor fit for the immutable model of the Nix store. If I understand https://github.com/NixOS/nix/pull/2670 correctly, it applies file contexts to store paths after they have been built, which breaks immutability. Alternatively, file contexts could be set by derivations at build time, but then file contexts should be recorded in NAR files.

Some previous discussion on SELinux on NixOS: https://nixos.org/nix-dev/2013-May/011091.html (the conclusion being that AppArmor is a better fit for NixOS).

e-user:
@Ekleog 

> I have doubts on the idea of running Nix on SELinux-enabled distributions. What I currently understand appears to imply that binaries will be trusted based on their path relative to the root of their derivation: `/nix/store/1234567890abcdef-httpd/bin/httpd` would take into account only `/bin/httpd`.

This is correct for the solution proposed by @etbe, which is also the primary one described in the RFC. However the alternative solution would be to assign contexts for each build, individually, which comes with all the drawbacks enumerated.

> This is insecure, as any user with access to the nix daemon could request such a file to be built with whatever content.
> 
> As I guess I misunderstood something, would it be possible to make the threat model and the currently-envisioned solution explicit?

Yes, there is a misunderstanding, here. Currently, binaries installed by Nix on an SELinux-enabled system are automatically labelled `unconfined_t` in lack of a policy covering anything in `/nix`, which is *least* restrictive on what these binaries can do. There is an exception however, which is that systemctl can't operate on unconfined files so it can't symlink services files through Nix profile directories, but that's about it. A user with root rights can still create said symlinks to service files manually or copy them so Nix binaries will ultimately be run through systemd in the least restricted manner, i.e. completely unsecured.

SELinux is _usually_ used only for system services requiring root privileges and restrains processes to the exact privileges they need to operate with. SELinux _can_ be tailored to operate in a more locked down, white-listing fashion, even affecting what regular users can do as well, but this far less common and no distribution I'm aware of is built this way. Maybe the SELinux folks can shed some more light on these use cases, like @etbe or @fishilico.
 
> Currently, I think I can see how to make NixOS interoperate cleanly with SELinux (by generating the policy as part of the module system), but I cannot see yet how to make Nix interoperate cleanly with a SELinux-enabled system.

Any level of integration will currently only take away privileges from Nix-owned binaries, so interoperability can't possibly worsen. SELinux does not grant additional privileges overriding default POSIX security measures like file ownership, so I can't possibly think of any benefit when a user tries to maliciously build a binary with a different name. If they can run it as root, they can do whatever they want anyway, including manually relabeling the binary to be unconfined. 

A system which is designed to be as locked down as possible and where SELinux is used for this purpose, wouldn't allow users to install and run arbitrary binaries to begin with. 

Again, @etbe and @fishilico please correct me if I'm wrong.

e-user:
@edolstra 

> I've never used SELinux, so take the following with a grain of salt, but it's my understanding that SELinux is a poor fit for the immutable model of the Nix store. If I understand [NixOS/nix#2670](https://github.com/NixOS/nix/pull/2670) correctly, it applies file contexts to store paths after they have been built, which breaks immutability. Alternatively, file contexts could be set by derivations at build time, but then file contexts should be recorded in NAR files.

SELinux can do anything AppArmor can do, just in a more fine-grained way and with more control over the process. Making SELinux and Nix a good match is just a matter of work.

NixOS/nix#2670 is indeed at odds with immutability and originally only aimed at fixing the very issue of providing basic interoperability on SELinux-enabled systems. With a bit of work, I could change the PR to instead derive file contexts from a policy derivation coming from the store itself which gets applied during build time and written to the NAR archive, as well. This would, I believe, supersede @etbe's patches to SELinux itself, but I might be wrong.

I understand that this may be the only way forward for actual NixOS support, whereas the proposed original solution only works as an ephemeral workaround to make Nix work again on SELinux-enabled systems in multi-user mode.

> Some previous discussion on SELinux on NixOS: https://nixos.org/nix-dev/2013-May/011091.html (the conclusion being that AppArmor is a better fit for NixOS).

I don't know AppArmor as well as I know SELinux, but I think this is a fallacy. SELinux is just more complex than AppArmor, but there isn't a single distinguishing feature or property in AppArmor that makes it more well-suited for NixOS. On the contrary, AppArmor operates on paths only as I understand it, which means only SELinux labels can be written to NAR archives in order to maintain immutability, even though this is isn't part of the current iteration of the RFC. 

I would like to wait for comments from @etbe and @fishilico as they are from SELinux upstream, but in the meantime, I could work on a proof-of-concept prototype for the aforementioned, immutability-adhering solution.

Mic92:
I would suggest that in the shepherd team at least one person from @NixOS/nix-core should be included.

shlevy:
@edolstra has been nominated. @Ekleog @aszlig available to join?

e-user:
I'm very happy to hear @edolstra will be the shepherd for this RFC!

How does the process continue, from here?

Ekleog:
@shlevy Given my upcoming schedule, I unfortunately think I won't be able to shepherd this RFC for at least the coming month.

As a side-note, I'd suggest inviting interested people from SELinux in the meeting, so that discussion could also use their insight.

globin:
I additionally nominate @joachifm @xeji and @thoughtpolice, who all have worked on hardening/SELinux PRs/issues in nixpkgs

joachifm:
I'd be happy to contribute to this, to the extent that I am able.

thoughtpolice:
I'm in the middle of many things right now (job hunt, house work, planned travels) so I'm afraid I can't make any kind of commitment at this moment to see things through, at least for several more weeks or so once things shake out -- and I also must admit I'm unfamiliar with SELinux. (I appreciate the nomination, nonetheless!)

shlevy:
@e-user Do you have any thoughts for others we can nominate to the team? If not, are you OK with having a smaller-than-usual shepherd team to facilitate the discussion and make the final decision?

asymmetric:
Can we have a link to the [rendered](https://github.com/NixOS/rfcs/blob/892f3f3e328f1858baef1f7e66844de6bf27e560/rfcs/0041-selinux-support.md) RFC in the top comment? :)

e-user:
@shlevy I'm fine with anything as long as we can move forward.

domenkozar:
@thoughtpolice what about now, did situation improve? :)

joachifm:
If the changes to Nix are acceptable in principle, maybe it'd be easier to reduce the scope of the RFC to only cover those and leave NixOS for later?

shlevy:
@edolstra Should we move forward with a meeting with the reduced shepherd team?

thoughtpolice:
Sorry for the delay here. I've had a recent job change and things have settled down a bit for me, so I can accept the nomination to help shepard this. Please note that I have little insight into SELinux itself, but am interested in helping steer things to ensure it can work for our users if that's what they desire, to the greatest extent I can (and I can surely get a bit more up-to-speed in the mean time.)

Regarding @joachifm's point, personally I think a single RFC is fine. But the larger, long-term timelines here (upstreaming SELinux changes, Nix changes/review cycles, cutting a new Nix stable with them, releasing to users, then making Nixpkgs changes) mean some people may feel differently. I think there are reasonable positions for both, but this RFC has lingered a bit -- so I think we should probably begin moving forward first, and scope discussions should probably be a starting point.

edolstra:
@e-user @shlevy @thoughtpolice @joachifm Do you have time for a call next week to discuss how to proceed with this RFC? Most times will work for me. A time in the afternoon (CET) would be best to allow @shlevy to join.

joachifm:
@edolstra that works for me

thoughtpolice:
@edolstra Sure, please let me know. Note I am also in Central Standard Time (as is Shea -- or at least he's close), so sometime afternoon in Europe would be nice for me as well.

edolstra:
@e-user @shlevy @thoughtpolice @joachifm Would Thursday 16:00 CET (== 10:00 CT) work for you?

joachifm:
@edolstra probably works for me.  I'm easily confused by time zones, so just to be clear, that is taking summer time into account (CEST/GMT+2)?

thoughtpolice:
@edolstra Yes and to be unambiguous: I am available tomorrow, Thursday the 18th, at 10:00am Central Time. Note that I AM NOT available Thursday, the 28th at 10:00am Central Time (yet, in a strange twist of fate, that exact timeslot is taken precisely so I can give a talk about Nix at work!)

edolstra:
@joachifm Yes, 14:00 UTC.

domenkozar:
I can't find who's in the shepherd team for this RFC.

globin:
AFAICS shepherd team is: @edolstra, @shlevy, @thoughtpolice, @joachifm

rfcs/0041-selinux-support.md
```diff
@@ -0,0 +1,162 @@
+---
+feature: selinux-support
+start-date: 2019-03-10
+author: Alexander Kahl (@e-user)
+co-authors: (none yet)
```
domenkozar:
```suggestion
co-authors: (none yet)
shepherd-team: Eelco Dolstra, Shea Levy, Austin Seipp, Joachim F.
shepherd-leader: Eelco Dolstra
```

e-user:
Hey everyone,

my work life has currently caught up and I don't have a lot of capacity left to work on this, but this might change sometime soon, when I can spend some of my spare time. I can definitely make some room for the call, just `@`-ping me with the details.

shlevy:
Shall we try to organize another call?

edolstra:
@shlevy @thoughtpolice, @joachifm @e-user Would you have time for a call some time next week?

thoughtpolice:
Yes, I can make time for a call -- details sooner better than later, of course!

e-user:
Sure

edolstra:
How about Friday 18th at 14:00 UTC (16:00 CET)?

joachifm:
> How about Friday 18th at 14:00 UTC (16:00 CET)?

Works for me.

e-user:
> How about Friday 18th at 14:00 UTC (16:00 CET)?

Alright. Will someone share the call details with me?

joachifm:
Sorry, some work related stuff has come up; I won't be able to make it.

e-user:
After a very productive call, @edolstra and I decided to close the RFC with the option of re-opening at a later time. Instead, we'll be focusing on a minimum viable solution to run Nix store binaries from systemd on SELinux-enabled systems, which most likely only requires changes to Nix' installer and is not intrusive towards the rest of the overall Nix and NixOS architecture.

Melkor333:
Is it possible to share some details why this decision was made for eventual later reference? :)

PR #45, opened 2019-04-28
[RFC 0045] Deprecating unquoted URL syntax

7c6f434c:


rfcs/0045-deprecate-url-syntax.md
```diff
@@ -0,0 +1,65 @@
...
+might not be worth making a specific decision.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Do nothing; get PRs from time to time that make homepages uniformly quoted
+strings or uniformly unquoted.
+
+Decide to use unquoted URLs for all URLs without special characters or variable
+expansion.
```
grahamc: Mind making these two items a list?

rfcs/0045-deprecate-url-syntax.md
```diff
@@ -0,0 +1,65 @@
...
+
+Add a note in the Nix manual that the special unquoted URL syntax is
+deprecated.
+
+Add a note in the Nixpkgs manual that the unquoted URL syntax is deprecated,
+changes to Nixpkgs should not increase its use, and it is recommended to
+convert URLs to quoted strings when changing them.
+
+Accept PRs that convert unquoted URLs to quoted strings if such PRs are
+submitted.
```
grahamc: I wonder if there is a way ofborg could detect newly introduced URL values?

rfcs/0045-deprecate-url-syntax.md
7c6f434c: Definitely done in the PR branch, I have no idea why GitHub doesn't show this yet…

rfcs/0045-deprecate-url-syntax.md
7c6f434c: Maybe your profiling patch to Nix is in the right direction for that…

joepie91:
Strongly in favour. It's never been clear to me why this was part of the syntax, and more than once I've been left with analysis paralysis because I wasn't sure whether there was a reason to use the special syntax or not, and there appeared to be no documentation on that.

And from a tooling implementor POV: removing this in the long run would reduce the complexity of a Nix (parser) implementation.

samueldr:
Quick thoughts about deprecating language/interpreter features:

As far as detection goes, deprecating features in the interpreter could be made strict with a list of "new" features, e.g. `--option strict-eval no-unquoted-urls` so that the interpreter gates everything related to the unquoted URLs under conditionals, making it fail. Tracing could additionally be added at the same locations to warn, hopefully with a flag to disable tracing those.

Though, this is all at the cost of adding more conditionals, thus more possibilities for weird behaviours.

Gating it in the main interpreter with an option would also allow end-users to set it globally, allowing new deprecation-free evals to be forced on their systems, hopefully helping weed out the leftover features deprecations.

7c6f434c:
Thanks for the suggestion of one more reason, let's have motivation section to be the longest one!

rfcs/0045-deprecate-url-syntax.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+Add a note in the Nix manual that the special unquoted URL syntax is
+deprecated.
+
+Add a note in the Nixpkgs manual that the unquoted URL syntax is deprecated,
+changes to Nixpkgs should not increase its use, and it is recommended to
+convert URLs to quoted strings when changing them.
```
zimbatm: Is it possible to be more specific? How would everybody be notified about this new rule?

7c6f434c: 1. I think the mere fact of an RFC passing is a noticeable effect event for a large fraction of active contributors. (It is visible in this repository, announced via Discourse, gets into the «weekly», is likely to be discussed a bit on IRC etc.)

For this specific RFC just the announcement «RFC on Deprecation of URL syntax passed» conveys enough information.

2. It will be noted in two out of three manuals.

3. Once we have an RFC that something «should not» happen, a request for cosmetic cleanup in a PR (and such requests happen) is more likely to contain a mention of quoting URLs (if relevant for the package in question). This also spreads the knowledge.

I do have an impression that quite a few people mentioned that they are disappointed they cannot refer to a policy that quoted URLs are better, so I expect the review channel of information dissemination to perform well.

Appendix A: future work — maybe the tool gets implemented, then we can open a countdown issue and maybe make ofborg check that PRs do not make things worse).

My plan is indeed just these points. I think that these things do not really need any additions to the text of RFC, and I think these mechanisms will be enough to distribute the information.

If you think that any of these four channels benefit from an addition to RFC text, or that there are other information distribution channels that should be used (and mentioned in the RFC text) please give some details.

zimbatm: It doesn't have to be formalized in the RFC but it's good to have an accompanying discussion.

For example you mentioned "they are disappointed they cannot refer to a policy". Would the RFC act as such policy? If not, maybe we should have a list of official policies in place.

7c6f434c: I think having a deprecation notice in the manual (added in accordance with an explicit bullet point in an approved/accepted RFC) is close enough to official policy for the purpose of asking people not to do the deprecated things.

I didn't start the discussion myself, because I thought that (1) is assumed, (2) and (A) are explicitly mentioned, and (3) is a manual variant of (A) in a sense.

rfcs/0045-deprecate-url-syntax.md
```diff
@@ -0,0 +1,70 @@
...
+# Summary
+[summary]: #summary
+
+Gradually deprecate the use of unquoted URL syntax in Nix code and especially
+Nixpkgs.
+
+# Motivation
+[motivation]: #motivation
+
+Nix language has a special syntax for URLs. However, unlike paths, URLs do not
```
FRidh:
```suggestion
The Nix language has a special syntax for URLs even though quoted strings can also be used to represent them. Unlike paths, URLs do not
```

rfcs/0045-deprecate-url-syntax.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Currently none.
+
+# Future work
+[future]: #future-work
+
+In case of a future major changes in the Nix syntax, removal of special URL
```
FRidh:
```suggestion
In case of future major changes in the Nix syntax, removal of special URL
```

rfcs/0045-deprecate-url-syntax.md
```diff
@@ -0,0 +1,70 @@
+---
+feature: deprecate_url_syntax
+start-date: 2019-04-28
+author: Michael Raskin
+co-authors: 
+related-issues:
```
FRidh: https://github.com/NixOS/nixpkgs/pull/27809
https://github.com/NixOS/nixpkgs/pull/46378

7c6f434c: I am not sure: the RFC template seems to imply this is for PRs implementing the RFC, not for the changes the RFC intends to undo.

7c6f434c:
@FRidh thanks for the proofreading.

edolstra:
We should not break backwards compatibility in such a massive way. I mean, deprecation is only useful if you eventually want to remove it, at which point it becomes impossible for Nix to build old Nix expressions.

In principle, the flakes "epoch" feature would make it possible to make such language changes in a backward compatible way (e.g. unquoted URLs could be deprecated in epoch 2020). I'm not convinced it's actually useful, though. The case of `x:x` falls under the "just don't do that" header IMHO.

7c6f434c:
There are some features that we have chosen not to use in Nixpkgs anymore and we do not recommend to use these features if they can be avoided at all. Additional tooling intended for aiding Nixpkgs development can be useful without supporting such features, and easier creation of tooling is good.

The RFC definitely does not argue for actually dropping the support unless some other breaking changes happen. It will be a pity if an epoch-guarded syntax change for something major like a change in overrides happens without a cleanup of minor annnoyances.

domenkozar:
In 2022 we can parse Nix identifier without first parsing a URI with a regex, which will speed up the parser.

edolstra:
@domenkozar I think URIs, identifiers etc. are all recognized using a single finite state machine, so it shouldn't matter much for performance.

Ericson2314:
Let http://blog.chriszacharias.com/a-conspiracy-to-kill-ie6 serve as a reminder for the benefits of deprecation even without removal!

rfcs/0045-deprecate-url-syntax.md
```diff
@@ -0,0 +1,70 @@
...
+
+Currently none.
+
+# Future work
+[future]: #future-work
+
+In case of future major changes in the Nix syntax, removal of special URL
+syntax might be considered.
+
+Explore options for automated tracking of the number of unquoted URLs in
```
Infinisil: If we deprecate this syntax, we should not use it anymore in nixpkgs. so I'd say it should be part of this RFC that all URL's be changed to strings. Writing a program that does this conversion shouldn't be hard, I could write one with hnix.

7c6f434c: RFC already has a commitment to accept any change like that; I did not put a commitment to make such a change immediately because I was not sure if anyone is willing to commit to write such a tool right now. This is that kind of tool that shouldn't be hard to write, but might end up being annoying to debug, and if `nix-fmt` happens it should do this as a side-effect anyway…

Infinisil: I'm giving it a shot, I've written a tool that does something very similar before

7c6f434c: Thanks! I guess if the tool works, it should be reasonably easy to replace «edit» with «count» and deploy it as an ofBorg check, too?

Infinisil:
@7c6f434c Alright done, here it is: https://github.com/infinisil/nix-quote-urls. Should work fairly well :)

7c6f434c:
Updated the text

edolstra:
I gotta ask: what's the point? If we're not actually removing URL syntax (and we can't, because of compatibility), then why not use it? Removing it seems like busy work to me.

7c6f434c:
Well, for some reason `builtins.toPath` is kept in Nix but deprecated…

Even if URL syntax is supported by Nix, not having an exception in the grammar actually used by Nixpkgs is a good thing for the tooling (it is necessary to support Nixpkgs to check if a tool is useful, but supporting the entire Nix can wait).

And using URL syntax means worse diffs (once I want to use string interpolation, I need quote the strings, and now the entire URL is highlighted as a change) and more annoyance when copying the URLs without interpolation from terminals (URLs contain `;` more often than `"`, so by default quoted URLs will get selected correctly and unquoted URLs will get selected wrong) or the editors (where having «copy contents of double-quoted string» is quote likely).

edolstra:
`toPath` is rather different. It's deprecated because nobody really knows what it's supposed to do :-)

Maybe we should add string interpolation to URLs?

zimbatm:
> Maybe we should add string interpolation to URLs?

Do you remember what was the reasoning for adding the URL notation for string literals to the syntax?

I remember as a newbie to be confused because path literals were returning paths but the URL syntax was returning string. I was expecting it to fetch the content, just like the path literal copies the content into the store.

The path literal in the syntax is genius. Is it possible that after that success the URL notation was added without too much thought because URLs are also important for a build system?

7c6f434c:
> Maybe we should add string interpolation to URLs?

For what gain? It solves diffs and guarantees even less tooling…

edolstra:
I don't see the tooling issue. Supporting URL literals requires *one* regexp in the lexer...

7c6f434c:
I think that if the definition of URIs refers to and (IETF) RFC but Nix URI syntax does not implement the definition of URI from that RFC, it is a bad enough sign.

grahamc:
My personal opinion is I think URL syntax offers very little value (possibly negative value, if you consider newbies being confused by it.) I don't really think extending URL syntax to include interpolation is a nice idea, as even the `foo.${bar} = baz;` is very strange already.

domenkozar:
@edolstra there is certain overhead with regexing each identifier. 

About the motivation to depreciate: cleaning the language and getting rid of features that provided little value is good on the long term. You can release a new version of Nix in a couple of years that would modernize the old patterns. But you have to start this process quite early on.

shlevy:
I nominate @grahamc and @edolstra to the shepherd team. Anyone else in this thread available for shepherding?

Ericson2314:
As the Schemers say:

> Programming languages should be designed not by piling feature on top of feature, but by removing the weaknesses and restrictions that make additional features appear necessary.

There was no weakness and restriction (in the human or the language) making URL syntax necessary to begin with!

dtzWill:
On Thu, 09 May 2019 06:59:38 -0700, John Ericson <notifications@github.com> wrote:
> As the Schemers say:
> 
> > Programming languages should be designed not by piling feature on top of feature, but by removing the weaknesses and restrictions that make additional features appear necessary.
> 
> There was no weakness and restriction (in the human or the language) making URL syntax necessary to begin with!

I'd just like to speak up for unquoted URL syntax.

I've always been a fan, and frankly I'm not sure there's
cause for removing it, and very unsure there's sufficient cause.
The bar should be high for this sort of breaking change IMO.

As for the value of the syntax, I think it's a useful notation
for a common (every single file has at least one or two!) "type".
There's value in making these literals and not expressions.
(although I will say I am a fan of /new/ syntax for things like
${version}, but those can't be "clicked"/followed trivially
and despite being useful are somewhat ugly. pname/pname-version.sadface.

Replacing useful types with strings hardly seems like "progress"...?

There are some quirks, but I'd much rather we fought to improve things
here-- replacing it with a string seems like we give up on any attempt
to distinguish these values from those used for 'name' or for a shell
script.

Anyway I don't really feel all that strongly about it.

Mostly, let's be sure this is worth whatever pain will result, please
:).

7c6f434c:
>I've always been a fan, and frankly I'm not sure there's
>cause for removing it, and very unsure there's sufficient cause.
>The bar should be high for this sort of breaking change IMO.

Well, the plan is to stop using it in Nixpkgs starting from some time
and replace existing uses with quoted string syntax, so the things won't
actually break.

>As for the value of the syntax, I think it's a useful notation
>for a common (every single file has at least one or two!) "type".
>There's value in making these literals and not expressions.

What do you mean — not expressions? All three syntax versions of string
literals are literals.

Or do you want to have fetchurl as a _syntactic_ construction, not just
a builtin?

>(although I will say I am a fan of /new/ syntax for things like
>${version}, but those can't be "clicked"/followed trivially

From the clicking point of view, strings are better than unquoted URL
literals, because the semicolon in the end is explicitly separated by
the quotes. Clicking on unquoted literals and getting semicolon included
in the URL (as semicolons have a recommended semantics in the URI RFC, 
of course they are included when click-selecting URLs…) happens all the
time…

>and despite being useful are somewhat ugly. pname/pname-version.sadface.
>
>Replacing useful types with strings hardly seems like "progress"...?

Right now it is not a type. The manual explicitly says that the unquoted
URL syntax is a syntax for string literals.

>There are some quirks, but I'd much rather we fought to improve things
>here-- replacing it with a string seems like we give up on any attempt
>to distinguish these values from those used for 'name' or for a shell
>script.

Well, there is no language support for it now, most URLs we do use are
strings anyway (because updating versions without using version 
expansion is annoying), no proposals what to do with a separate URL
type, and I think no reported hard to debug issues that would be 
prevented by such separation.

So, _what_ would the attempts to distinguish try to accomplish?

Mic92:
I would like to nominate @edolstra, @zimbatm and @Infinisil for this RFC.

zimbatm:
Happy to help.

I think what needs to happen next is to scope how the changes to nixpkgs should happen. Are we going to do a mass-rewrite or only accept clean PRs? How are we going to let everybody know about the new policy? Can we enforce some of it using ofborg?

edolstra:
Well, step 1 is to decide whether this should be done at all. To me it's a low gain, high pain change.

Ekleog:
To reduce the pain one option might be to not actually try to enforce anything, and let the information about this RFC percolate its way through the community -- a significant number of the people with commit bit should be watching more or less closely the RFC repository anyway, which means information should be pretty fast to spread, at least for new PRs. OTOH, doing a mass-rewrite would IMO be too high a pain due to instantly bit-rotting a lot of open PRs.

Synthetica9:
Maybe we could do a mass rewrite simultaneously with the enforcement of `nixfmt`?

edolstra:
No, I don't want to create a hard dependency on `nixfmt`.

Infinisil:
Well I did write https://github.com/infinisil/nix-quote-urls, which can fixup unquoted urls. I can extend that to be usable by ofborg, reporting when it found unquoted URLs.

I have to agree with @edolstra though, doing this change has *very* low gain. The time spent on discussing this probably already outweighs the gains this could ever get us. We could use this time instead to focus on more important problems, of which there are a lot tbh.

Infinisil:
@Mic92 I'll accept the nomination for this RFC

7c6f434c:
>I have to agree with @edolstra though, doing this change has *very* low gain. The time spent on discussing this probably already outweighs the gains this could ever get us. We could use this time instead to focus on more important problems, of which there are a lot tbh.

Well, there are discussions like that from time to time when someone tries to do a mass-change in either direction, I hoped that RFC would allow to have the last one…

Ekleog:
I have just noticed https://github.com/NixOS/nix/blob/b0fc5bcee9f74c717d8ca564c193a5ad7846e5c7/doc/flakes/design.md ; with this syntax for flakes the “it simplifies the parser” argument becomes much more compelling, as currently there might be ambiguity between “parse as a flake” and “parse as a URL” -- so this change might actually be necessary for the flakes change to land.

7c6f434c:
@Ekleog it looks like the flake IDs can be just treated as strings on the Nix parser level

joepie91:
I don't agree that this is a low-gain change, at all. Seemingly duplicated language features are a common source of frustration - and therefore a common timesink - for beginners. *We* might not see much gain from this change, but everybody who's just starting out to use Nix, will.

Because instead of spinning their wheels for an hour trying to figure out why there's special URL syntax, they will now never have to think about it, as a URL is simply represented as a string (as is the case ~everywhere else). The combined gain from all these people never having to figure out this oddity, is significant.

edolstra:
Numerous languages have multiple types of string literals (e.g. Python has 4 or so). I don't think there is any evidence that this is a major impediment to learning the language. There are much bigger issues with the Nix language (e.g. that it has functions, thus allowing people to create arbitrarily complex abstractions).

domenkozar:
That's arguing with [eastern logical fallacies](https://en.wikipedia.org/wiki/Whataboutism).

The question at hand is only if the cognitive overhead of another type is worth its usefulness.
And another is if deprecations help (which I hope we've so far established they do).

joepie91:
@edolstra When other languages have multiple types of string literals, generally at least one of the following is true:

1. All variants *look* like string literals, ie. there is some sort of quotation mark involved.
2. The different string literals have a notably different function.
3. The different string literals all support arbitrarily-formatted contents.

The problem with the URL literals in Nix is that they look significantly different from regular strings (there are no quotation marks), while being *no functionally different* from them, and also arbitrarily restricting the valid formatting.

This produces a literal syntax that very strongly *looks* like it should have a special meaning, but it doesn't. *That* is where the unnecessary cognitive overhead comes from. Other languages do not have this; either it looks different and works different, or it neither looks different nor works different.

7c6f434c:
@domenkozar I think it is wrong to deny that experience (and complaints or lack of complaints) of users of other languages is a kind of evidence, even if we do not agree about the interpretation (I agree with @joepie91 here — and I consider nixos/nix#2893 a part of confusion)

domenkozar:
It's impossible to measure and compare these things.

It's like saying Haskell shouldn't have static types because dynamic ones in JavaScript seems to work quite well. 

Existence of something is not a proof that cognitive load doesn't exist.

Infinisil:
@joepie91 

> Because instead of spinning their wheels for an hour trying to figure out why there's special URL syntax, they will now never have to think about it

I really doubt anybody has ever spent an *hour* on figuring out URL syntax. Most people will probably just have seen `https://example.com` in nixpkgs and thought "Oh cool, I can do that". Do you have any evidence of people actually being confused about URL syntax? I've never seen any instance of that.

globin:
@Infinisil not being confused about URL syntax itself, but I've seen multiple cases of people being confused about not being able to use interpolation in non-quoted URLs.

joepie91:
> Do you have any evidence of people actually being confused about URL syntax?

Yes. Myself, and various people I've spoken to.

I'm not sure that "but can you *prove* that people have an issue with this" is the right baseline to be following here, for a language feature that adds precisely zero benefits (as far as I can tell anyway). That's a really good way to ensure that people eventually just give up on pointing out ergonomics issues, and leave, because it's exhausting to have to spend a ton of energy and time for every little issue.

Infinisil:
There is however a difference between adding such a feature initially and removing it when it existed for a while, and only the latter comes with the trouble of having to deprecate it. For a new language I'd fully agree with you.

Anyways, I'm not particularly against this change, I can see there's some benefit to removing URL syntax, but there's also a non-zero amount of work needed to deprecate it, which seems to almost balance the benefit of removing it. So I'm very neutral to this change. And considering there's much more important problems in Nix, I'd rather not waste time on this.

joepie91:
Sure, fully *removing* it, from the language and runtime, involves a bunch of work. But soft-deprecating it by explicitly recommending against its usage and updating current usage in nixpkgs, does not, as far as I can tell - especially considering that the fixup tool already exists.

And ultimately, if we never tackle these sort of issues (there are *always* problems that seem more important in software projects), they will just keep compounding into worse and more widespread ergonomics issues. At least this is an issue where it's fairly clear what the path forward should be, and so all that remains would be the implementation.

Much of the time spent on this discussion so far seems to be of an "I don't see why we need to do this" nature, rather than any concrete arguments *against* doing it. *That* is the bigger waste of time here, in my opinion...

7c6f434c:
@Infinisil does _the manual_ being confused count? (There is therefore a non-deprecation cost — I guess it is OK to be slightly wrong about deprecated feature, but if its use is not discouraged, we need to figure out how to write a useful correct description)

edolstra:
> any concrete arguments against doing it

The concrete arguments against it are:

* Incompatibility with old Nix expressions.
* A huge amount of work. For example, this will change almost every file in the Nixpkgs repo, making it much harder to cherry-pick commits between branches. 

That kind of change had better be worth it, but the only arguments in favor are highly subjective or dubious (e.g. that users waste hours being confused by this). It reminds me of the `gcc` -> `cc` rename in Nixpkgs which caused random breakage everywhere for a year or so. And this would be a much bigger change.

7c6f434c:
This is a more predictable change than `gcc` → `cc` because the updated file is parsed to the exact same expression, and references to compiler are varied and sometimes unexpected, so the name matters.

Deprecating the use of the syntax in Nixpkgs doesn't obligate us to break the old code outside Nixpkgs.

The lines with the unquoted URLs are exactly the URLs that don't ever change, source URLs are quoted anyway because they use either `${version}` or `${name}`.

oxij:
> * Incompatibility with old Nix expressions.

In the short run, deprecation does not make incompatibility. In the long run, sure. But most out of tree expressions use `stdenv` anyway and evaluating old expressions against modern `stdenv` does not always work either (e.g. if you use some attr that `stdenv` didn't know about and now `stdenv` uses the same attr for something else).

If anything, given the existence of the tool to make this automatic I don't see the problem here. People can just run it over their old sources. In fact, we could also run it over all old nixpkgs branches...

> * A huge amount of work. For example, this will change almost every file in the Nixpkgs repo, making it much harder to cherry-pick commits between branches. 

... which would then solve this point completely for free.

But even if we don't, for some mysterious reason, at most, changes to master will multiply cherry-picks by two: for every change with a hunk having a recently converted URL you would need to cherry-pick its parent.

Unsurprisingly, I'm :+1:. Simpler is better.

The fact that semicolons do not work as IETF RFC says they should and so parsing of nix-URLs in tty emulators is broken is annoying enough to warrant this change by itself, IMHO.

joepie91:
> Incompatibility with old Nix expressions.

The argument here is for *soft* deprecation. Hard deprecation is a different discussion entirely, and for the future. Soft deprecation does *not* produce incompatibility, especially considering that this is supposedly "just another string literal", and so there are no type incompatibilities caused either (since the only thing that changes is the construction syntax, not the type of the resulting value).

arianvp:
I did an impact analysis because people keep asking for data. Conclusion, I think purging naked urls from nixpkgs isn't that big of a diff to be honest.

First of all. I ignored all occurrences of `homepage =` as they have no impact on code breaking.

Here are all naked expressions:
https://gist.github.com/arianvp/37d4b141a76f070ec691c24d68aa7d9b

There are **3681** occurrences. Almost all occurrences are in this single file for game assets in unreal engine: https://github.com/NixOS/nixpkgs/blob/92a047a6c4d46a222e9c323ea85882d0a7a13af8/pkgs/games/ue4/cdn-deps.nix

Filtering those we are left with **1925** matches 
https://gist.github.com/arianvp/dda7bac24774f57510937afd59028af1


Some of these nix expressions are autogenerated by `x2nix` scripts, and can thus be ignored. Filtering those out leaves us with: https://gist.github.com/arianvp/eceb1c80419feea7a6b0395c7877ec12

which is **1326**  matches which isn't really all that much, and can be done in small patches for the mentioned packages.   Also note that there are even still some false postitives (URLs in multine comments etc) so the number might even go down more.

Infinisil:
> First of all. I ignored all occurrences of `homepage =` as they have no impact on code breaking.

I don't get this. If we deprecate URL syntax, how is it justified to ignore it for some instances?

arianvp:
With _ignore_ i meant.  "It would be trivial to change them, without scary things happening."

I think we should change them . However they will give a big diff if we do them all at once. Maybe we should split up that change somehow

Infinisil:
Here is what the diff would look like: https://github.com/Infinisil/nixpkgs/commit/184fea6f95039de0ac67529947ed45ab16f3069f

This was done with https://github.com/infinisil/nix-quote-urls, so there shouldn't be any false positives or negatives.

That commit changes over half of all files in nixpkgs, not saying that's a bad thing, but there are the proportions if we want to do this.

Mic92:
Since shepherd was not explicitly announced before. For this RFCs @edolstra is the leader, while @zimbatm and @Infinisil are the other member on the shepherd team.

zimbatm:
Count me in.

rfcs/0045-deprecate-url-syntax.md
```diff
@@ -0,0 +1,72 @@
+---
+feature: deprecate_url_syntax
+start-date: 2019-04-28
+author: Michael Raskin
+co-authors:
```
domenkozar:
```suggestion
co-authors: 
shepherd-leader: Eelco Dolstra
shepherd-team: Eelco Dolstra, Jonas Pfenniger, Silvan Mosberger
```

rfcs/0045-deprecate-url-syntax.md
```diff
@@ -0,0 +1,74 @@
+---
+feature: deprecate_url_syntax
+start-date: 2019-04-28
+author: Michael Raskin
+co-authors: 
+shepherd-leader: Eelco Dolstra
+shepherd-team: Eelco Dolstra, Jonas Pfenniger, Silvan Mosberger
```
lheckemann: @zimbatm is your name correct here?

zimbatm: no it's "zimbatm" :)

Infinisil:
Considering that the vast majority of people are for this deprecation, which also does provide some benefits, I'm alright with it as well.

However I don't think we should actually enforce this right now, this syntax is very widely used. I'd much rather have this just be marked as deprecated (like with `isNull`, whose deprecation I'm not a fan of btw), so that people can be informed over time. Nixpkgs can slowly be cleaned of non-quoted urls over time as well, such that eventually we can enforce it for PR's once the majority of people already know about it. And eventually, in a long time, this can actually be removed from the Nix grammar.

This then shouldn't distract us much and we can keep focusing more important problems while dealing with this one slowly over time. How does this sound? Especially asking @edolstra

rfcs/0045-deprecate-url-syntax.md
lheckemann: @domenkozar or @7c6f434c would you care to fix this?

7c6f434c: Ah right, sorry.

zimbatm:
Let's move forward with this. It's not too controversial but also not worth spending too much time on it. We'll amend the nix and nixpkgs manuals, maintainers and reviewers can refer to this RFC and we'll enforce it more strictly over time.

edolstra:
I'm not in favor of moving forward with this. It's a huge amount of work for little to no benefit.

> And eventually, in a long time, this can actually be removed from the Nix grammar.

We can never do that because reproducibility requires being able to build old Nix expressions.

zimbatm:
What is a huge amount of work?

I think the main interest is to enforce that on nixpkgs and avoid having new users having to learn about the URL syntax. We don't need to change any of the nix implementation for this RFC to be useful.

arianvp:
Can we move this forward / or not based on a consensus vote?

Ekleog:
@arianvp The RFC process does not suggest a consensus vote. That said, it includes the following two statements:
- “all members of the RFC Shepherd Team must sign off the motion,” meaning things technically cannot move forward until every shepherd has signed off on @zimbatm's FCP motion (assuming that's what https://github.com/NixOS/rfcs/pull/45#issuecomment-521727562 is)
- “there should not be a strong consensus against that position outside of the RFC Shepherd Team,” meaning I do not think an FCP for rejection would be legitimate, as my reading of this thread is that there is a strong consensus in favor of this RFC, though there are quite strong arguments against some “Future work” items (and some discussion about whether making a mass-edit to remove all unquoted URL syntax should be done -- maybe moving that to “Future work” might make sense to smooth the process?)

FRidh:
Just agreeing we do not use the unquoted syntax in Nixpkgs could be a useful middle-ground. Third parties can choose themselves whether to use that syntax or not. No change in tooling is needed. For new tooling one can choose whether to support it or not. This may lead to fracturing though. I don't have a strong opinion on this matter.

gloaming:
Well, it seems that people are behaving like this has already been passed, and "fixing up" `meta.homepage` urls in their update PRs.

Personally, I would prefer that we do that at once in a treewide, if we're going to do it, and not make additional noise.

edolstra:
Just to be clear: this PR has *not* been approved yet so any changes to Nixpkgs that introduce massive code churn are premature.

Ekleog:
Data point I can't remember seeing in this thread: if we want to have *only one* of either quoted or unquoted URL syntax, we can only choose quoted URL syntax without changing semantics: a quick search returned all these, which each have a semicolon in the URL, and thus can't be unquoted URL syntax.
https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/system/fio/default.nix#L37
https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/networking/uqmi/default.nix#L22
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/libraries/nss_wrapper/default.nix#L16
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/libraries/resolv_wrapper/default.nix#L15
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/libraries/socket_wrapper/default.nix#L15
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/libraries/uid_wrapper/default.nix#L15
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/libraries/libubox/default.nix#L19

Would it make sense to have a shepherd meeting, so that a proper way forward is found? This RFC is currently stalling without any way forward for non-shepherds, as all arguments and opinions appear to have been raised.

domenkozar:
I'd like to gather some information here to move this forward - it's a bit of a bikeshedding change, but I think it's still important to pay attention to even little details in a language.

# Arguments against

## It's a lot of work

I agree, however there are people who are willing to do the work, there is automation and the work can be done gradually (as we have removed a bunch of things from nixpkgs in the past).

## Cherry-picking between branches is going to be harder

Do the automated switch just before NixOS branch-off. The number of cherry-picks to old NixOS releases will be small and usually won't affect `meta`.

## deprecation is only useful if you eventually want to remove it

I disagree, for beginners, each feature of the language is something they have to learn,
so if people stop using a feature all over the place, that cognitive overhead is gone. Although small,
it exists.

# Phases

I propose it's rolled out in phases where 1,2 and 3 can be done one after another, then with some pause between 4 and 5 (best to write down in RFC when exactly those should happen).

1. Release Nix documenting it's a deprecated feature, go through docs to get rid of the feature there
2. Replace all uris with quoted syntax before next NixOS release branch-off
3. Introduce CI check for nixpkgs
4. Release Nix with a warning when the feature is used
5. With the next edition of flakes, prohibit the feature

jb55:
I've seen some confusion on PRs because of this: with people asking people to quote urls on PRs even though there is precedent for removing them from https://github.com/NixOS/nixpkgs/pull/27809. I have a feeling this is going to come up again and again.

I'm in favor of quoting urls as the standard *convention*, just so we can avoid this discussion over and over. No deprecation needed.

nrdxp:
I am actually using this feature in my configuration; found a case where it simplifies the code i.e. in  an assert:

`foo.nix`
```
{ lib, withFoo ? ../somePath.nix }:
let
  ...
in
assert assertOneOf "withFoo" withFoo [ ../somePath.nix ./someOtherPath.nix ];

...
```

Now say I am importing the above code from a location a couple levels deeper in the file system:

`other.nix`
```
...

import ../../foo.nix { withFoo = ../../someOtherPath.nix; inherit lib; };
...
```

Since the relative paths get translated into full path strings (I am assuming) the assert still passes. I am actually using this code in my current configuration so my vote is against deprecation. 

Though I see nothing wrong with recommending quoted paths where possible, I am sure situations like the above will arise and unquoted paths are actually a fairly elegant solution in my opinion.

edolstra:
@7c6f434c @Infinisil @zimbatm Do you guys want to have a call sometime next week to talk about how to proceed with this RFC?

zimbatm:
sure, I am available most of monday, tuesday and wednesday

7c6f434c:
Sounds like a good idea. I should be able to participate on Monday, Tuesday, Friday,
or Thursday afternoon (Central Europe time).

zimbatm:
Ok let's meet Mondat 15:00 UTC.

* Hangouts link: https://meet.google.com/eoh-wqpt-xtg
* Calendar entry: <a target="_blank" href="https://calendar.google.com/event?action=TEMPLATE&amp;tmeid=NnV1azcwZGJsMWlucG5nNmF0a2IzNGRvMzYgemltYmF0bUB6aW1iYXRtLmNvbQ&amp;tmsrc=zimbatm%40zimbatm.com"></a>

7c6f434c:
>Ok let's meet Mondat 15:00 UTC.

2019-10-14 15:00 UTC, fine with me.

>* Hangouts link: https://meet.google.com/eoh-wqpt-xtg

That's pretty funny. Is Google the only videochat solution by now that works at least in Chromium (unlike Skype), but breaks if privacy.resistFingerprinting is enabled in Firefox?

(Well, if I toggle this option, it seems to work — of course I cannot really know in advance if something is broken…)

zimbatm:
We can fallback to https://whereby.com/zimbatm if it doesn't work for you. Let's all be on `#nixos-dev` to synchronize at the time.

7c6f434c:
>We can fallback to https://whereby.com/zimbatm if it doesn't work for you. Let's all be on `#nixos-dev` to synchronize at the time.

I agree re: `#nixos-dev`, I currently hope that either in Firefox or in Chromium Hangouts will agree to work.

orivej:
I have to say that even though I understand and can agree with this change, I cannot make myself like it because, from one perspective, it is about what constitutes a clickable URL, and it replaces an explicit syntax rule that defines a literal URL by an implicit guesswork of making everything that looks like an URL clickable (although admittedly this works better in terminals). (It would not improve my experience because I do not view or edit `.nix` files in a terminal.) The RFC says that "URLs do not have any special support that would make the difference useful", but this is not exactly true: Nix mode for Emacs supports URL literals and displays them as blue URLs rather than orange strings.

7c6f434c:
> an explicit syntax rule that defines a literal URL by an implicit guesswork of making everything that looks like an URL clickable … Nix mode for Emacs supports URL literals and displays them as blue URLs rather than orange strings.

On the other hand, why perfectly fine URLs inside meta.longDescription should not be clickable?

And the explicit syntax rule is still not perfect: fakeroot patches (not the only example) cannot have a literal-URL syntax because the URL contains a semicolon. So as a clean solution, making clickable strings that happen to match IETF URL definition and not contain `${` is likely to improve accuracy overall.

(The glorious future when some kind of Nix language server will provide on-the-fly evaluation for interpolated URLs would be nice, but definitely out of scope)

zimbatm:
I hereby motion that this RFC is ready for the next phase :)

FRidh:
Initially I did not have any strong opinion about this RFC but now I do.

Nix emphasises reproducible builds. That requires first of all a reproducible evaluation. Unlike derivations, which inputs may be pinned (pick a specific version of Nixpkgs e.g.), we do not specify a version of Nix that is needed to perform an evaluation (flakes may though in the future). This is a promise (certain) users want to rely on. If we break that promise already over such a little thing as the URL syntax, then that won't create much trust.

In my opinion we could *choose* not to use the syntax. I don't see anything against that, although it may lead to fragmentation as other tools could also *choose* not to support the URL syntax. Regardless, I think we should first document and version the grammar of the language.

joepie91:
> Unlike derivations, which inputs may be pinned (pick a specific version of Nixpkgs e.g.), we do not specify a version of Nix that is needed to perform an evaluation (flakes may though in the future). 

We don't, but we do have implicit requirements for this. For example, some two(?) major releases ago, before Nix 2.0 I believe, it became impossible to evaluate newer nixpkgs versions with the then-current Nix version because of newly-added language features being used.

That doesn't mean we should be breaking things willy-nilly, but we should be realistic and acknowledge that the "never break anything" ship has sailed a while ago.

> If we break that promise already over such a little thing as the URL syntax, then that won't create much trust.

This proposal is for soft deprecation, which will merely *mark* the feature as deprecated and discourage its use. It will not actually break any builds. Actual potential removal from the language is listed as future work.

Once a proposal lands to actually *remove* this feature from the language, sure, an assessment should probably be made as to whether the breakage is justifiable or not, ideally grouped in with some other breaking changes that have accumulated over time.

But I don't think it's useful to block a soft deprecation on concerns around a hard deprecation that are not yet relevant today.

> Regardless, I think we should first document and version the grammar of the language.

I would be strongly in favour of this as well, but that would be a separate RFC, I'd imagine.

FRidh:
> We don't, but we do have implicit requirements for this. For example, some two(?) major releases ago, before Nix 2.0 I believe, it became impossible to evaluate newer nixpkgs versions with the then-current Nix version because of newly-added language features being used.

That's the other way around. Of course if you use new features you cannot use a version that does not *yet* have those features. Here it is about *removing* functionality, thus breaking expressions that used to evaluate just fine. 

> Actual potential removal from the language is listed as future work.

It says in the RFC

> and will be eventually removed.

I could instead say

> and may eventually be removed.

Regardless, that will lead to the question: when? When saying you will or may remove something, then give a time plan or at least describe that it will be announced x months before actual removal so it is known how long people have to make the necessary changes.

joepie91:
> That's the other way around. Of course if you use new features you cannot use a version that does not yet have those features. Here it is about removing functionality, thus breaking expressions that used to evaluate just fine.

I'm not sure there's a significant difference from an end user perspective. Not all systems on earth (or even managed by one person) will be running the same Nix version on all of them, which means that newer features are going to be used while older versions of Nix are still in use.

That means that either way, you still end up with a situation where you cannot assume that Nix code that evaluates on one system, will also evaluate on another system. What's the difference between whether that's due to backwards-compatibility or forwards-compatibility issues, when the outcome of breaking that assumption is the same?

In a broader sense, I'm also not sure it's useful to consider "reproducibility" to include "guaranteed to build in the future". There are a lot of factors outside our control that already make this not the case, like eg. source tarballs no longer being available. I think a more useful definition - that reflects the current state of Nix better - is "guaranteed to build *with functionally the same result*, or not at all".

> It could instead say and may eventually be removed.

That could be amended in the RFC, sure. But I think it's important to avoid the word "may" in the actual documentation - in my experience, when it's said that something *may* be removed (vs. *will* be removed), people will often not heed the warning and continue using it as if nothing changed.

> Regardless, that will lead to the question: when? When saying you will or may remove something, then give a time plan or at least describe that it will be announced x months before actual removal so it is known how long people have to make the necessary changes.

I don't think it's realistic at this point in time to come up with a timeline for permanent removal. I do think it would be reasonable to guarantee that an eventual removal would always have X months notice.

alyssais:
> It says in the RFC
>
>> and will be eventually removed.
>
> I could instead say
>
>> and may eventually be removed.

I think that any mention of removal should be left to a future RFC.  I
don't think we can have it both ways -- saying that this RFC doesn't
address the removal of URL syntax from the language, while also stating
that it _will_ be removed.

> That could be amended in the RFC, sure. But I think it's important to
> avoid the word "may" in the actual documentation - in my experience,
> when it's said that something *may* be removed (vs. *will* be
> removed), people will often not heed the warning and continue using it
> as if nothing changed.

HTML is an example of a language where various features were marked as
deprecated, and have essentially fallen out of use as a result, even
though they have to be supported forever -- when did you last see a
<font>, <big>, or <u>?

7c6f434c:
The initial plan was to have it discouraged in general, (eventually) forbidden in Nixpkgs, and removed if any breaking changes happen. 

In the shepherd meeting I was told that one way or another there will be a way to version Nix language versions (be it flake editions or something similar), and then the new and proper Nix edition can drop plain URL syntax. It is hard to say whether there will be notice period or just an option to use an older syntax for one's code while importing Nixpkgs from a flake with a newer edition.

Speaking of using fresh Nix with old code, I doubt we will never drop MD5 support in Nix (version used by Nixpkgs) completely

alyssais:
> In the shepherd meeting I was told that one way or another there will
> be a way to version Nix language versions (be it flake editions or
> something similar),

That's a bold assertion to have been made with no accepted RFC on the
topic yet...

> and then the new and proper Nix edition can drop plain URL syntax. It
> is hard to say whether there will be notice period or just an option
> to use an older syntax for one's code while importing Nixpkgs from a
> flake with a newer edition.

In that case it would be fair enough, but I think it would be premature
for an accepted RFC to say that, because we currently have no way of
versioning the syntax.

zimbatm:
The main benefit of this RFC is to clarify the use of the URL syntax in nixpkgs as those tend to be mass-rewritten in both ways. Let's just stop using that literal syntax in nixpkgs and trust that our future selves will do the right thing regarding the actual removal, with the new context available then.

These RFCs are mainly useful to create a discussion and reach consensus. If the wording is not clear then it should be fixed but the important part is the overall consensus. And also to move forward.

edolstra:
I would suggest removing the line

> In case of future breaking changes in the Nix syntax, it would be nice to make the x:x snippet parse to the identity function, and not to an URL.

since it's actually not particularly desirable to make `x:x` parse as the identity function: any Nix formatter will presumably rewrite it to `x: x`.

Also "will be removed" should read "may be removed in a future edition of the Nix language" (and removed the "Future work" section because it's superfluous). Of course, the latter depends on there *being* an edition mechanism, so we won't remove unquoted URLs before we have editions.

Other than that, I would like to propose moving to FCP.

Synthetica9:
> since it's actually not particularly desirable to make `x:x` parse as the identity function: any Nix formatter will presumably rewrite it to `x: x`.

I'm not sure I understand what you are saying here.
Nix formatters would ideally not alter the parsing of any expression, right?

edolstra:
What I mean is that `x:x` will never be acceptable style so there is no reason to make it supported.

JohnAZoidberg:
So in the case of removing `x:x` as a URL, you'd propose to make it invalid syntax? @edolstra

edolstra:
Yes.

7c6f434c:
Once I started editing the language about removal, I noticed I had forgot to qualify my claims about lack of special support. 

@FRidh @alyssais is the current wording about next edition enough, or should I put some extra text about either editions being runtime-selectable, or removal being announced a full NixOS release cycle in advance?

FRidh:
Simply mentioning deprecation and not mentioning removal at all would have my preference.

7c6f434c:
If we ever start making incompatible changes to Nix for other reasons, bundling URL syntax removal to the change seems natural to me. Do you disagree to this statement, or do you disagree to mentioning that possibility?

Note that Nix language has relatively recently got an incompatible change: `builtins.fetchurl` has stopped accepting `md5` argument.

alyssais:
> If we ever start making incompatible changes to Nix for other reasons,
> bundling URL syntax removal to the change seems natural to me.  Do you
> disagree to this statement, or do you disagree to mentioning that
> possibility?

I neither agree nor disagree with the statement.  But I don't think this
RFC is the place to decide it -- we can decide later when the
possibility arises.

7c6f434c:
>> If we ever start making incompatible changes to Nix for other reasons,
>> bundling URL syntax removal to the change seems natural to me.
>I neither agree nor disagree with the statement.  But I don't think this
>RFC is the place to decide it -- we can decide later when the
>possibility arises.

What extra knowledge for making a better informed decision could we
expect to be available at that future time? I think if we can already
find out the outcome of this future decision making, we should let the
users know (and if we want to keep the feature indefinitely even in case
there are syntactically incompatible editions anyway, we should document
what exactly the feature does).

alyssais:
> What extra knowledge for making a better informed decision could we
> expect to be available at that future time? I think if we can already
> find out the outcome of this future decision making, we should let the
> users know (and if we want to keep the feature indefinitely even in case
> there are syntactically incompatible editions anyway, we should document
> what exactly the feature does).

Well, for one thing we'll know if we have adopted an edition system or not.

7c6f434c:
>Well, for one thing we'll know if we have adopted an edition system or not.

I think if the only thing that keeps us from dropping the URL syntax 
from the language is a lack of language-versioning, it is OK to decide 
so (and the decision «if there are editions, bundle dropping the URL
syntax to the next incompatible ones» is made with the knowledge that it
is only applicable _if_ there are editions, so this piece of information
can be used in the decision making)

I don't want to drop this decision from the scope without recording some
objections to the conditional decision «if we make a noticeably
incompatible language version, we also attach the URL syntax drop», as 
there were requests to include some decision on this into the RFC

edolstra:
> I neither agree nor disagree with the statement.  But I don't think this RFC is the place to decide it -- we can decide later when the possibility arises.

IMHO this RFC *is* the place to decide, after all why would we go through this whole process again when we can decide now?

@Infinisil Do you approve of the RFC in the current state? @zimbatm and I have so if you agree we can move to the FCP.

edolstra:
The Final Comment Period for this RFC has started and, barring any blocking issues, will be merged after 2019-11-24. Your opinions, comments, and approvals are welcome!

Ericson2314:
[I'm late to this conversation, so hopefully I am not missing something.]

> Add an ofBorg check that verifies that no new unquoted URLs have been added in
a PR.

I would like this to be through via Nix, even if ofborg and not hydra is doing the CI. The reasons are:

- I would like to also vet my own repos, via my own CI, for not containing URL literals.

- I want a single, canonical source of truth of the formal language that is "Nix expressions without URLs, as allowed in Nixpkgs". If we have one parser evaluating Nixpkgs for CI, and another separate parser validating Nixpkgs to not contain URLs, I imagine we could end up with weirdness that is hard to replicate in a 3rd implementation that intends to play both roles.

We now have experiment features. I think a good implementation of such a check would to have unquoted URLs be an "on by default" feature, which may at some point later become off by default or banned.

Re @alyssais's point. Hopefully we can at least ret-con this proposal into talking about a "lowecase-'e' 'edition'", without bias towards any one specific releasing or versioning scheme? N.B. in my view anything that is mutable over time (or variable over space, i.e. having multiple, unverified implementations) implicitly has multiple versions. This view conveniently de-politicizes the issue. If this interpretation sounds good to everyone, maybe we should furthermore replace "edition" with "version", as I think the latter word is less salient due and doesn't connote specific schemes.

7c6f434c:
@Ericson2314 

We already have Nix language versions and they are kind of expected to be one-way compatible (but not always truly are) so this would be a confusion. 

Not all parser generators make it easy to have individually configurable syntax features. So we will have a separate check first and then see what is the most efficient way forward…

Ericson2314:
@7c6f434c 

> We already have Nix language versions and they are kind of expected to be one-way compatible (but not always truly are) so this would be a confusion.

I would think that even rust style versions are within editions, so two versions could be of different divisions and therefore both ways incompatible. Therefore I don't think it's fair to say versions are always 1-way compatible.

> Not all parser generators make it easy to have individually configurable syntax features. So we will have a separate check first and then see what is the most efficient way forward…

Based on what @edolstra said about not wanting `x:x` to be a lambda, I think this won't be so bad, we don't need to generate multiiple parsers, just conditionally blow up in the same parser. (In other words, we are not doing things which adjust the boundaries of naturally-ambiguous stuff.)

7c6f434c:
>@7c6f434c 
>
>> We already have Nix language versions and they are kind of expected to be one-way compatible (but not always truly are) so this would be a confusion.
>
>I would think that even rust style versions are within editions, so two versions could be of different divisions and therefore both ways incompatible. Therefore I don't think it's fair to say versions are always 1-way compatible.

That incompatibility is controlled on the level of editions, though.

>> Not all parser generators make it easy to have individually configurable syntax features. So we will have a separate check first and then see what is the most efficient way forward…
>
>Based on what @edolstra said about not wanting `x:x` to be a lambda, I think this won't be so bad, we don't need to generate multiiple parsers, just conditionally blow up in the same parser. (In other words, we are not doing things which adjust the boundaries of naturally-ambiguous stuff.)

Well, if there is only one optional feature, it is indistinguishable from having two parsers; and in the general case there is a risk things become complicated.

edolstra:
Merged. Thanks everybody for your contributions!

Infinisil:
Note that I have https://github.com/infinisil/nix-quote-urls implemented to do the URL replacement in nixpkgs (when we eventually do it) and I'll also add a mode for it to be usable by ofborg soon.

Ericson2314:
Well, I certainly don't mind `hnix` being used!

PR #46, opened 2019-04-28
[RFC 0046] Platform Support Tiers

7c6f434c:
I hope to have some meta-discussion here so that #31 gets a language to discuss its specific merits.

[render link](https://github.com/7c6f434c/rfcs/blob/platform-support-tiers/rfcs/0046-platform-support-tiers.md)

### DIFF
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -2,7 +2,7 @@
 feature: platform_support_tiers
 start-date: 2019-04-28
 author: Michael Raskin
-co-authors:
+co-authors: Matthew Bauer
 related-issues:
 ---
 
@@ -49,43 +49,140 @@ such discussions more efficient.
 
 ## Platforms
 
-* `x86_64-linux`, native builds with `gcc` and `glibc`. Status: default
-  platform, the only truly first-tier platform. All packages not specifically
-  defined in terms of OS-specific or CPU-specific tooling for other platforms
-  are expected to work if they work anywhere.
+### Tier 1
 
-* `aarch64-linux`, native builds with `gcc` and `glibc`. `x86_64-darwin` as a
-  name for macOS with `clang` compiler. Status: second-tier platforms. Many
-  packages are supposed to work, the main Hydra puts a lot of binary packages
-  into the binary cache, there is tooling support to check builds on these
-  platforms and some level of effort is expected to be spent on investigating
-  new failures after update.
+Aims: all packages should work (unless they don't make sense)
 
-* `i686-linux`, `armv7l-linux`, `x86_64-linux` with `musl`, static builds.
-  Status: cross-compilation targets in different meanings of these words. No
-  binary cache available, checking a cross-build via ofBorg is possible but
-  complicated, no expectations on upgrade. Fixes not necessary on upper-tier
-  platforms are expected to be either localised inside `stdenv` dependencies
-  and other compilers/build tools, or to be general cleanups that just happen
-  to be optional on upper-tier platforms.
+A platform-specific fix is expected to be applied in `master`
 
-* Embedded systems like `arm-embedded`, `armhf-embedded` or even `avr`.
-  Status: only cross-compilation targets, no tooling support, no expectations
-  of attention from non-users. Platform-specific fixes are expected to only
-  touch immediate compiler toolchain.
+Support: good binary cache coverage, full support in tooling
+
+Developer/user base: most of the Nix developers/users
+
+* `x86_64-linux`, `gcc`+`glibc`
+
+### Tier 1.5
+
+Same aims and tooling support
+
+Fewer developers and users, less testing — significantly more broken packages
+
+* `aarch64-linux`, `gcc`+`glibc`
+
+* `x86_64-darwin`, `clang`+Darwin/macOS
+
+### Tier 2-ε
+
+Aims: most of the popular packages work
+
+Platform-specific things for arbitrary packages should not be too complicated
+
+Support: native bootstrap tools are available, cross-build toolchains in the
+binary cache, partial tooling support
+
+Package updates might break build on the platforms of this tier and lower
+
+* `i686-linux`, `gcc`+`glibc` — `ofBorg` builds via `pkgsi686Linux`, binary
+  cache contains `wine` dependencies
+
+### Tier 2
+
+Aims: most of the popular packages work
+
+Support: native bootstrap tools are available, cross-build toolchains in the
+binary cache
+
+* `armv{6,7,8}*-linux`, `gcc`+`glibc`
+
+* `mipsel-linux`, `gcc`+`glibc`
+
+* `x86_64-linux`, `gcc`+`musl`
+
+* `x86_64-linux`, `clang`+`glibc`
+
+### Tier 3
+
+Aims: some packages are expected to work
+
+Platform-specific fixes limited to general cleanups of non-standard
+assumptions in the upstream code and basic toolchain fixes
+
+* `aarch64-none`
+
+* `avr`
+
+* `arm-none`
+
+* `i686-none`
+
+* `x86_64-none`
+
+* `powerpc-none`
+
+* `powerpcle-none`
+
+* `x86_64-mingw32`
+
+* `i686-mingw32`
+
+* `x86_64-linux`, `gcc`+`musl` — static
+
+* `x86_64-linux`, `clang`+`glibc` — `llvm` linker
+
+* `x86_64-linux` — Android
+
+* `aarch64-linux` — Android
+
+* `armv{7,8}-linux` — Android
+
+### Tier 4
+
+Aims: none
+
+Support: none or accidental
+
+Platform definitions present
+
+A small amount of packages seems to work, maybe by luck
+
+* `x86_64-linux`, `gcc`+`glibc` — static
+
+* `x86_64-linux`, `gcc`+`glibc` — `llvm` linker
+
+### Tier 4.5
+
+Work ongoing to provide/merge Tier 4 support
+
+* `wasm-wasi`
+
+* `powerpc64le-linux`, `gcc`+`glibc`
+
+### Tier 5
+
+Aims: none
+
+Support: none
+
+No current support, but previous support of clear path to add support
+
+* `aarch64-darwin`
+
+* `i686-darwin`
+
+* `x86_64-freebsd`
 
 ## Adding a new platform
 
-It is expected that embedded-level support (platform definition, basic support
-for the target, maybe a toolchain in the binary cache) can added as long as
-the support code works and is contained in the expected places. If the
-platform users want to set higher expectations, a motivation should be
-provided. User interest, platform perspectives and development effort
-commitment are usually taken into account.
+It is expected that Tier-4 support can be added freely, and Tier-3 support is
+added once enough packages are tested and sustained development happens.
+Tier-2 support (and higher tolerance to platform-specific fixes in
+non-toolchain packages) is generally linked to higher user interest and
+sustainability of both the platform itself and Nixpkgs development for the
+platform.
 
-Before adding an expectation that platform non-users pay attention whether
-upgrades break a lower-tier platform, support for testing on this platform
-must be available.
+Support above Tier-2 (and expectation that platform non-users pay attention to
+the platform support on updates) requires deployment of test infrastructure
+for the platform.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -2,7 +2,7 @@
 feature: platform_support_tiers
 start-date: 2019-04-28
 author: Michael Raskin
-shepherd-team: Ryan Mulligan, Jonas Pfenniger , Graham Christensen, John Ericson
+shepherd-team: Ryan Mulligan, Jonas Pfenniger, Graham Christensen, John Ericson
 shepherd-leader: John Ericson
 co-authors: Matthew Bauer
 related-issues:
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -71,9 +71,24 @@ Same aims and tooling support, there are channel blocking jobs
 
 Fewer developers and users, less testing — significantly more broken packages
 
+If a package update breaks a previously working package on that platform,
+update author should check if some easy fix follows from the build logs or is
+mentioned in the upstream bugtracker.
+
 * `aarch64-linux`, `gcc`+`glibc`
 
+A team @NixOS/aarch64-maintainers shall be created to include people who
+understand the platform and use it.
+
+If there is a complicated problem on this platform when updating a package
+that was previously built succesfully on macOS, @NixOS/aarch64-maintainers team
+should be informed.
+
 * `x86_64-darwin`, `clang`+Darwin/macOS
+
+If there is a complicated problem on this platform when updating a package
+that was previously built succesfully on macOS, @NixOS/darwin-maintainers team
+should be informed.
 
 ### Tier 2-ε
 
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -64,7 +64,7 @@ met.
 
 ## Questions to answer for each platform
 
-* What fraction of packages is expected to work?
+* What fraction of packages are expected to work?
 * How many users there are / how much testing one can expect?
 * How much complexity is a platform-specific fix allowed to carry?
 * Is there a binary cache for the platform?
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -28,6 +28,30 @@ of the decisions of this type.
 # Detailed design
 [design]: #detailed-design
 
+## Concrete actions
+
+If this RFC is accepted, the following changes will happen:
+
+* The definition of support tiers and a list of platforms with the
+corresponding support tiers is added as an appendix to the Nixpkgs manual. The
+list of platforms is further maintained as a part of the Nixpkgs manual. The
+platform list in the manual will be initially based on the list in the appendix
+of the present RFC.
+
+* Other official documentation and marketing materials of the Nix/Nixpkgs/NixOS
+project are edited and maintained to be consistent with the platform list in
+the manual.
+
+* NixOS GitHub organisation teams @NixOS/aarch64-maintainers and
+@NixOS/exotic-platform-maintainers are created.
+
+* Platform support status at the Tier 4 and below is updated according to
+achievement of the requirements defined in this RFC.
+
+* Discussions of platform support at the Tier 3 and above start using the
+vocubulary based on this RFC and ensure that the requirements of this RFC are
+met.
+
 ## Platform elements
 
 * CPU architecture (possibly: subarchitecture, optional features)
@@ -52,62 +76,130 @@ of the decisions of this type.
 * Is Nix supported?
 * Are native bootstrap tools available?
 
-## Main dimensions
+## Mutual obligations between the platform developers and the community
 
-It is expected that for a platform in a stable situation the tiers
-corresponding to all the dimensions are the same or ar most adjacent. On the
-other hand, changes to the platform tiers usually include one dimension getting
-ahead of the other, with a different balance emerging afterwards.
+It is expected that the work necessary for supporting a platform is done by
+people using the platform, as they have both qualification and motivation for
+such work. In many cases a part or the entire build capacity will also be
+provided by the platform users. The only cost recognising platform support is
+allowed to impose on all the developers and users of Nixpkgs is the additional
+code and additional conditions for a proposed change to be considered correct.
 
-Note that Tolerance tier is never higher than Tooling tier (and normally not
-higher than Package coverage tier).
+The platform support tier defines how intrusive and widespread the changes
+needed to support the platform are allowed to be, and the amount of effort
+spent on avoiding the changes causing platform-specific regressions.
 
-### Tooling
+To qualify for a tier, the platform support should meet requirements on package
+coverage (how useful it is to users), and tooling (how easy it is to check the
+changes on the platform without access to the platform, and to start using the
+platform). Recognition of the lower tiers (Tier 5 and below) is based on
+a technical review of the platform support code quality and requirements. For
+higher tier an explicit decision has to be made based on the technical merits,
+future expectation. resource allocation and project priorities. As this is not
+a purely technical decision, this RFC only stipulates that a serious discussion
+occurs once any new platform meets technical requirements of a high tier.
 
-This dimension measures how easy it is to find out what works and what doesn't
-on the platform (in particular, without being a user of the platform).
+It is expected that a platform in a stable situation doesn't come close to
+satisfying the technical requirements of the next higher tier, neither in terms
+of coverage nor in terms of tooling. So in a stable situation the permitted
+impact, the package coverage and the tooling for a platform correspond to the
+same tier.
 
-#### Tier 1
+## Main support dimensions
 
-A lot of packages built by Hydra, full ofBorg support, some ordinary packages
-are channel-blockers on Hydra.
-
-#### Tier 2
-
-Native bootstrap tools available, cross-build toolchains in the binary cache,
-no channel-blocking jobs on Hydra.
-
-It is recommended to provide a derivation to test the software on this platform
-(e.g. a Qemu-based derivation with all the necessary scripts).
-As it is impossible to provide a legal testing setup for Tier-1 platform (macOS),
-this requirement is not strictly mandatory for Tier-2 tooling.
-
-#### Tier 3
-
-None
-
-### Tolerance/impact
+### Impact (support tier)
 
 This dimension describes the permissible impact of a platform on the Nixpkgs
 repository. This includes the necessary patches as well as the handling of the
 build failures of updates in Nixpkgs.
 
-#### Tier 1
+### Tooling requirments
 
-Problems on these platforms can block updates.
+This dimension measures how easy it is to find out what works and what doesn't
+on the platform (in particular, without being a user of the platform).
 
-#### Tier 1.5
+### Package coverage.
 
-Platform-specific patches expected to be applied as needed; updates expected
-not to break the build on these platforms, problems should be investigated
-(and reported to the platform maintenance team if no solution was found).
+This dimension measures avilability of the most important packages on the
+platform. It is intended to describe how useful is the current state of the
+platform support to the current and potential users.
 
-#### Tier 2
+## Support tiers
 
-Platform-specific fixes are expected to be rare and non-intrusive. Updates
-might break builds.
+This section defines requirements and permitted impact for different support
+tiers.
 
-#### Tier 3
+### Tier 1
+
+#### Impact
+
+Problems on these platforms can block updates for as long as necessary to
+resolve the issue.
+
+Platform-specific patches can be applied as necessary.
+
+Many ordinary packages are channel-blockers on Hydra.
+
+#### Tooling requirements
+
+Most of packages built by Hydra, full ofBorg support.
+
+#### Package coverage expectations
+
+Almost everything that is not explicitly platform specific and that is not
+abandoned (in Nixpkgs) works.
+
+### Tier 2
+
+#### Impact
+
+Updates are  expected not to break the build on these platforms, problems
+should be investigated. If no solution is easily found, the problems should be
+reported to the platform maintainers with a reasonable amount of time provided
+for fixing the issue.
+
+Platform-specific patches are applied as needed.
+
+Some ordinary packages are channel blockers on Hydra.
+
+#### Tooling requirements
+
+A lot of packages built by Hydra, full ofBorg support.
+
+#### Package coverage expectations
+
+Most packages work, credible ambition to reach Tier 1 coverage at some point.
+
+### Tier 3
+
+#### Impact
+
+Completely platform-specific fixes are expected to be rare and non-intrusive.
+Fixes to the compilation toolchains are expected.
+General cleanups of non-standard assumptions (e.g. «everything that is no x86
+is a kind of ARM» or «malloc(0) behaviour is a reliable indicator of other
+malloc features») useful for this platforms are welcome.
+
+Updates might break builds on this platform .
+
+No channel-blocking jobs on Hydra.
+
+#### Tooling requirements
+
+Native bootstrap tools available, cross-build toolchains in the binary cache.
+
+It is recommended to provide a derivation to test the software on this platform
+(e.g. a Qemu-based derivation with all the necessary scripts).
+As it is impossible to provide a legal testing setup for a Tier-2 platform
+(macOS), this requirement is not strictly mandatory for Tier-3 tooling.
+
+#### Package coverage expectations
+
+Most of the popular packages work.
+
+### Tier 4
+
+#### Impact
 
 Fixes necessary for this platforms must be either limited to compilation
 toolchains, or general cleanups of non-standard assumptions (e.g. «everything
@@ -116,69 +208,70 @@ indicator of other malloc features»). These fixes must be generic: there
 should be a reasonable expectation that other exotic platforms would equally
 benefit from the exact same fix.
 
-#### Tier 4
+#### Tooling requirements
 
-It is recommended to keep platform-specific patches to the toolchain in a
-separate package. Cleanups not necessary on any tier-3 platforms can be
-rejected.
+None.
 
-### The number of working packages
-
-This dimension measures how useful is the current state of the platform support
-to the current and potential users.
-
-#### Tier 1
-
-Almost everything that is not explicitly platform specific and that is not
-abandoned (in Nixpkgs) works.
-
-#### Tier 1.5
-
-Most packages work, credible ambition to reach Tier 1 at some point.
-
-#### Tier 2
-
-Most of the popular packages work.
-
-#### Tier 3
+#### Package coverage requirements
 
 Some packages work.
 
-#### Tier 4
+### Tier 5
+
+#### Impact
+
+It is recommended to keep platform-specific patches to the toolchain in a
+separate package. Cleanups not necessary on any Tier-4 platforms can be
+rejected if considered too intrusive.
+
+#### Tooling requirements
+
+None.
+
+#### Package coverage requirements
 
 Platform definitions present, a small number of packages might be working.
 
-## Platform list
+### Tier 6
 
-The platform list will be added as an appendix to the Nixpkgs manual, and will
-be maintained as a part of the Nixpkgs documentation.
+Work ongoing to provide/merge Tier 5 support
 
-A non-normative
-description of the platforms at the moment of discussion is provided in the 
-appendix to the RFC and will be used as a basis for a PR adding a platform list
-to the manual.
+#### Impact
 
-## Adding a new platform
+Platform definitions and separate platform-specific toolchain packages can be
+included.
 
-It is expected that Tier-4 support can be added freely, and Tier-3 support is
+### Tier 7
+
+No current support, but previous support or clear path to add support
+
+## Platform lifecycle
+
+It is expected that Tier-5 support can be added freely, and Tier-4 support is
 added once enough packages are tested and sustained development happens.
-Tier-2 support (and higher tolerance to platform-specific fixes in
+Tier-3 support (and higher tolerance to platform-specific fixes in
 non-toolchain packages) is generally linked to higher user interest and
 sustainability of both the platform itself and Nixpkgs development for the
-platform. Note that Tier-2 tooling requirements imply allocation of some amount
+platform. Note that Tier-3 tooling requirements imply allocation of some amount
 of recurring build resources (for building and testing the toolchain).
 
-Support above Tier-2 (and expectation that platform non-users pay attention to
-the platform support on updates) requires deployment of test infrastructure
-for the platform.
+Tier-2 and higher support (and expectation that platform non-users pay
+attention to the platform on updates) requires deployment of test
+infrastructure for the platform.
 
-Note that from the impact point of view Tier-3 only allows the platform to be a
+Note that from the impact point of view Tier-4 only allows the platform to be a
 motivation for generic cleanups, and further tiers require commitment of
-recurring resources. We hope that this allows Tier-4 addition and Tier-3
+recurring resources. This is the reason for Tier-5 addition and Tier-4
 promotion to happen inside the scope of normal technical review by the people
 working with similar platforms; further tiers require allocation and of
-hardware resources, and procedures for coordinating such financial decisions is
-out of scope for this RFC.
+hardware resources, and procedures for coordinating such financial decisions
+are out of scope for this RFC.
+
+When a platform starts falling out of use, its support tier (and permitted
+impact) is reduced once it becomes clear that the current tier requirements
+will stop being met in the near future. Platform-specifing patches no longer
+permissible in the context of the new support tier can be removed at will by
+the package maintainers.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -195,10 +288,12 @@ basis without a shared framework.
 Defining a scope for Nixpkgs platform support and dropping/separating support
 for some of the currently supported platforms.
 
+Defining both requirements and impact based purely on the package sets.
+
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-The list of currently supported platforms is incomplete.
+The list of currently supported platforms might still be incomplete.
 
 # Future work
 [future]: #future-work
@@ -216,6 +311,9 @@ generally encouraged could be defined. (Example: when building an FHS
 environment becomes a reasonable strategy to get something running on a NixOS
 machine?)
 
+Define the expetations of maintenance for specific packages; consider the
+notion of platform-specific maintenance.
+
 # Appendix A. Non-normative description of platforms in November 2019
 
 We currently have a relatively steady state, so the tiers for each platform do
@@ -227,9 +325,9 @@ Developer/user base: most of the Nix developers/users
 
 * `x86_64-linux`, `gcc`+`glibc`
 
-### Tier 1.5
+### Tier 2
 
-Fewer developers and users, less testing; tier-1 tooling.
+Fewer developers and users, less testing; Tier-1 tooling.
 
 * `aarch64-linux`, `gcc`+`glibc`
 
@@ -246,15 +344,14 @@ If there is a complicated problem on this platform when updating a package
 that was previously built succesfully on macOS, @NixOS/darwin-maintainers team
 should be informed.
 
-### Tier 2-ε
-
-Pure stdenv for native builds is a channel-blocking job; `wine` dependencies
-are available in the binary cache
+### Tier 3
 
 * `i686-linux`, `gcc`+`glibc` — `ofBorg` builds via `pkgsi686Linux`, binary
   cache contains `wine` dependencies
 
-### Tier 2
+As an exception, pure stdenv for native builds is a channel-blocking job;
+`wine` dependencies are available in the binary cache. These extra packages are
+maintained as a part of `x86_64-linux` Wine support.
 
 * `armv{6,7,8}*-linux`, `gcc`+`glibc`
 
@@ -266,7 +363,9 @@ are available in the binary cache
 
 * `x86_64-linux`, `gcc`+`musl`
 
-### Tier 3
+### Tier 4
+
+A special team @NixOS/exotic-platform-maintainers is created and can be consulted about issues related to these platforms
 
 * `aarch64-none`
 
@@ -298,15 +397,13 @@ are available in the binary cache
 
 * `armv{7,8}-linux` — Android
 
-A special team @NixOS/exotic-platform-maintainers is created and can be consulted about issues related to these platforms
-
-### Tier 4
+### Tier 5
 
 * `x86_64-linux`, `gcc`+`glibc` — static
 
 * `x86_64-linux`, `gcc`+`glibc` — `llvm` linker
 
-### Tier 4.5
+### Tier 6
 
 Work ongoing to provide/merge Tier 4 support
 
@@ -314,7 +411,7 @@ Work ongoing to provide/merge Tier 4 support
 
 * `powerpc64le-linux`, `gcc`+`glibc`
 
-### Tier 5
+### Tier 7
 
 No current support, but previous support or clear path to add support
 
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -39,6 +39,7 @@ such discussions more efficient.
 * How many users there are / how much testing one can expect?
 * How much complexity is a platform-specific fix allowed to carry?
 * Is there a binary cache for the platform?
+* What platform-related builds are channel update blockers?
 * Is the platform normally tested by the tools like ofBorg? Is it possible to
   get something tested with reasonable effort?
 * Is there expectation that updates do not break things for this platform?
@@ -57,13 +58,16 @@ A platform-specific fix is expected to be applied in `master`
 
 Support: good binary cache coverage, full support in tooling
 
+A platform in this tier has a set of build jobs that block a channel update in
+case of failure.
+
 Developer/user base: most of the Nix developers/users
 
 * `x86_64-linux`, `gcc`+`glibc`
 
 ### Tier 1.5
 
-Same aims and tooling support
+Same aims and tooling support, there are channel blocking jobs
 
 Fewer developers and users, less testing — significantly more broken packages
 
@@ -80,6 +84,8 @@ Platform-specific things for arbitrary packages should not be too complicated
 Support: native bootstrap tools are available, cross-build toolchains in the
 binary cache, partial tooling support
 
+Native stdenv is a channel-blocking job
+
 Package updates might break build on the platforms of this tier and lower
 
 * `i686-linux`, `gcc`+`glibc` — `ofBorg` builds via `pkgsi686Linux`, binary
@@ -91,6 +97,8 @@ Aims: most of the popular packages work
 
 Support: native bootstrap tools are available, cross-build toolchains in the
 binary cache
+
+This tier and lower doesn't affect channel updates
 
 * `armv{6,7,8}*-linux`, `gcc`+`glibc`
 
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -69,10 +69,19 @@ such discussions more efficient.
   and other compilers/build tools, or to be general cleanups that just happen
   to be optional on upper-tier platforms.
 
+* Embedded systems like `arm-embedded`, `armhf-embedded` or even `avr`.
+  Status: only cross-compilation targets, no tooling support, no expectations
+  of attention from non-users. Platform-specific fixes are expected to only
+  touch immediate compiler toolchain.
+
 ## Adding a new platform
 
-A proposal to add a new platform should justify the level of platform-specific
-fixes to be tolerated.
+It is expected that embedded-level support (platform definition, basic support
+for the target, maybe a toolchain in the binary cache) can added as long as
+the support code works and is contained in the expected places. If the
+platform users want to set higher expectations, a motivation should be
+provided. User interest, platform perspectives and development effort
+commitment are usually taken into account.
 
 Before adding an expectation that platform non-users pay attention whether
 upgrades break a lower-tier platform, support for testing on this platform
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -67,6 +67,11 @@ are channel-blockers on Hydra.
 Native bootstrap tools available, cross-build toolchains in the binary cache,
 no channel-blocking jobs on Hydra.
 
+It is recommended to provide a derivation to test the software on this platform
+(e.g. a Qemu-based derivation with all the necessary scripts).
+As it is impossible to provide a legal testing setup for Tier-1 platform (macOS),
+this requirement is not strictly mandatory for Tier-2 tooling.
+
 #### Tier 3
 
 None
@@ -165,6 +170,10 @@ are available in the binary cache
 
 * `armv{6,7,8}*-linux`, `gcc`+`glibc`
 
+* `armv{6,7,8}*-linux`, `gcc`+`glibc`, cross-compilation
+
+* `aarch64-linux`, `gcc`+`glibc`, cross-compilation
+
 * `mipsel-linux`, `gcc`+`glibc`
 
 * `x86_64-linux`, `gcc`+`musl`
@@ -200,6 +209,8 @@ are available in the binary cache
 * `aarch64-linux` — Android
 
 * `armv{7,8}-linux` — Android
+
+A special team @NixOS/exotic-platform-maintainers is created and can be consulted about issues related to these platforms
 
 ### Tier 4
 
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -21,7 +21,9 @@ When adding a new platform, be it a new C library option, a new
 cross-compilation target configuration or a new CPU architecture, there is a
 discussion of support expectations and maintenance burden. Having a documented
 vocabulary to describe the expectations and documented precedents should make
-such discussions more efficient.
+such discussions more efficient. Building a consensus about a standard list of
+restrictions for less-popular platform support can also streamline most
+of the decisions of this type.
 
 # Detailed design
 [design]: #detailed-design
@@ -52,10 +54,18 @@ such discussions more efficient.
 
 ## Main dimensions
 
+It is expected that for a platform in a stable situation the tiers
+corresponding to all the dimensions are the same or ar most adjacent. On the
+other hand, changes to the platform tiers usually include one dimension getting
+ahead of the other, with a different balance emerging afterwards.
+
 Note that Tolerance tier is never higher than Tooling tier (and normally not
 higher than Package coverage tier).
 
 ### Tooling
+
+This dimension measures how easy it is to find out what works and what doesn't
+on the platform (in particular, without being a user of the platform).
 
 #### Tier 1
 
@@ -77,6 +87,10 @@ this requirement is not strictly mandatory for Tier-2 tooling.
 None
 
 ### Tolerance/impact
+
+This dimension describes the permissible impact of a platform on the Nixpkgs
+repository. This includes the necessary patches as well as the handling of the
+build failures of updates in Nixpkgs.
 
 #### Tier 1
 
@@ -110,6 +124,9 @@ rejected.
 
 ### The number of working packages
 
+This dimension measures how useful is the current state of the platform support
+to the current and potential users.
+
 #### Tier 1
 
 Almost everything that is not explicitly platform specific and that is not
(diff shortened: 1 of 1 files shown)


Ericson2314:
Could we have an "embedded" level for and https://github.com/NixOS/nixpkgs/blob/master/lib/systems/examples.nix#L102-L113 https://github.com/NixOS/nixpkgs/issues/59285, this is where very few if any packages are expected to work, but we do provide toolchains for users' convenience. This, as the lowest level support by far, should be an auto-accept for just about anything the compiler supports.

7c6f434c:
Indeed.

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,121 @@
...
+# Detailed design
+[design]: #detailed-design
+
+## Platform elements
+
+* CPU architecture (possibly: subarchitecture, optional features)
+* OS kernel
+* C compiler
+* C library
+* NixOS/non-NixOS global layout, in case of Linux with glibc
```
shlevy: What does this mean?

7c6f434c: There are some things (OpenGL, for example) where level of support for Nix-on-non-NixOS-Linux is lower than on NixOS.

Ericson2314:
Thanks for adding embedded!

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,121 @@
...
+* Global linking options
```
zimbatm: Package sets: Gnome packages, language specific package, wayland packages, ...

7c6f434c: I would say with the caption of «platfom support» these go into the first point of «questions to answer» — «what fraction of packages works».

Indeed, my gist https://gist.github.com/7c6f434c/431c8233793571cad313e77356d2f7db does have some sketches w.r.t. package support level and some Nix-related kinds of support levels both for platforms and for Nix features, but I think there are less misunderstandings there.

(So we agree: different scope, let's start with platforms)

matthewbauer:
For what exists now, my list of tiers would be:

## Tier 1

_All packages are expected to work._

Available in binary cache + tier 2 + tier 3

- x86_64-linux
- x86_64-darwin
- aarch64-linux

## Tier 2

_Most packages are expected to work._

Bootstrap tools are available + tier 3.

- armv{6,7,8}*-linux
- mipsel-linux
- i686-linux
- powerpc64le-linux

## Tier 3

_Some packages are expected to work._

Cross compilation support is available. Toolchains are built in cross-trunk via release-cross.nix.

- aarch64-none
- avr
- arm-none
- i686-mingw32
- i686-none
- x86_64-none
- x86_64-mingw32
- wasm-wasi
- powerpc-none
- powerpcle-none
- msp430-none

## Tier 4

Broken or no support is provided, but may be possible.

_No packages are expected to work._

- aarch64-darwin
- i686-darwin

## Tier 5

Everything else.

## ABIs

ABIs are a different concern, but still worth addressing. They should map to one tier.

- gnu{,eabi,eabihf} - tier 1
- musl{,eabi,eabihf} - tier 2
- android{,eabi,eabihf} - tier 3
- mingw32 - tier 3
- apple-ios - tier 4

7c6f434c:
@matthewbauer Thanks for a much better enumeration of platforms! I will try to incorporate most of it into the RFC draft. Do you want to be a coauthor of the RFC? Do you want branch access?

I would say that an honest description of what exists now would have `x86_64-darwin` and `aarch64-linux` as tier 1.5 — we aim to have everything working, but significantly more breakages happen there than on `x86_64-linux`.

I wonder what is our static build support? It should be considered an ABI variant with just `gnu-static` and `musl-static` as tier 3 ABIs?

vcunat:
I suspect that _"all packages [with matching `meta.platforms`]"_ might be a little too hard to fulfill, though it depends on how hard we take the _"expected to work"_ part and/or whether we tweak the exact meaning of `meta.platforms`.

In any case I expect it will be better to start with this simpler style &ndash; and only eventually consider upgrading the approach, say, to (conceptual) mapping: `(package, platform) -> (level of support, maintainers)`.  I think it's quite common that (particular) maintainers only really care about some platforms, even though the package often kind-of works on more of them.  In general there can be tension when one platform "slows down" another one in some way.  There's a bit more discussion also in the thread linked by GitHub just above my post.

matthewbauer:
> @matthewbauer Thanks for a much better enumeration of platforms! I will try to incorporate most of it into the RFC draft. Do you want to be a coauthor of the RFC? Do you want branch access?

Yeah I can coauthor!

> I would say that an honest description of what exists now would have `x86_64-darwin` and `aarch64-linux` as tier 1.5 — we aim to have everything working, but significantly more breakages happen there than on `x86_64-linux`.

Yeah I agree that x86_64-linux is the gold standard. But there's no reason we can't support aarch64-linux or x86_64-darwin equally. They both are built fully by Hydra.

> I wonder what is our static build support? It should be considered an ABI variant with just `gnu-static` and `musl-static` as tier 3 ABIs?

I would consider the static builds to be an overlay - not a different platform. Support is definitely low since we don't have any Hydra jobs for it. Tier 3 for Musl probably makes the most since. My understanding is support is much better for static linking in Musl than Glibc.

FRidh:
> But there's no reason we can't support aarch64-linux or x86_64-darwin equally. They both are built fully by Hydra.

Not every developer has access to those platforms nor interest to support them.

`(package, platform) -> (level of support, maintainers)` might indeed be too fine-grained for now, however I think we should already consider something like https://nixos.wiki/wiki/User:Profpatsch#nixpkgs_support_matrix
(Was referred to in https://github.com/NixOS/nixpkgs/issues/60005)

vcunat:
For `aarch64-linux` there's [the community machine](https://github.com/nix-community/aarch64-build-box) and @grahamcOfBorg, so not too hard if one is interested;  macOS is probably hard for those who don't have it.  In any case, the worst thing is human time/effort.

7c6f434c:
@FRidh Then I look at the list of questions and list of platform dimensions in this RFC and find the linked support matrix a bit too much of a simplification…

@matthewbauer I do want to describe the current/near future state — I do not see Darwin rising from tier-1.5 to tier-1 in the next 3 months.

Overlay or whatever, but static builds are a type of compilation target in my opinion.

I tried to merge the list of build types, and I moved things down to reflect the current state. I hope I am not too wrong.

Gave you push access (apparently I can only do it for the entire fork, but whatever), included as a coauthor in the next revision.

CrystalGamma:
Would it make sense to track separate "support levels" (which describe how much is expected to work) and "acceptance levels" (which describe how much platform-specific code – be it configuration conditionals, patches or entire packages – will be accepted).

The levels should then be defined such that, given enough development activity, a "support level" corresponding to the "acceptance level" can reasonably be reached with the amount of platform-specific changes permitted by the "acceptance level".

That would allow us to differentiate between platforms that are stable (support = acceptance) or in the process of being developed (support < acceptance) without having x.5 tiers. Though perhaps shorthands like "tier 3+" for "support level 3, acceptance level 2" would still be useful.

Regardless of whether this makes sense, one thing that should be clarified is what actually constitutes a "platform-specific fix": NixOS/nixpkgs#45474 (which originally started this discussion) was about disabling the intel Mesa driver on non-x86, so that Mesa can be built on ppc64le; is that a ppc64le-specific fix? I would argue that it's actually an x86-specific change (even though it served the purpose of fixing the build on other platforms), considering that the intel driver does not work (or at least didn't at the time of that PR, haven't checked since) on other architectures.

7c6f434c:
@CrystalGamma Well, the only thing I am reasonably confident, is that we should consider _at least_ the questions listed in the general introduction. But there are too many questions to track separately… In some sense, there are three key dimensions: support, acceptance, and how many packages are used are known to work. (Even `x86_64-linux` is not perfect in that last metric).

I guess if we think in terms of per-direction tiers, acceptance is weaker than both support and coverage, and support can either come as recognition of coverage or because of coordinated effort of enthusiasts donating build cycles etc.

I tried to word that carefully: I think that disabling `intel` driver on non-x86 is a cleanup of incorrect assumptions that only ARM and x86 exist, and a problem on low-acceptance platform is enough to replace the wrong assumption with a better one.

Maybe a good test is «do we expect the exact same fix to help on currently unsupported platforms?» (be it RISC-V instead of POWER or BSD libc instead of musl). Should we add this into RFC?

Ericson2314:
> do we expect the exact same fix to help on currently unsupported platforms

That's a good question, because the answer is usually "yes"!

7c6f434c:
@Ericson2314 there is a bit of selection bias — the fixes applied to non-compiler-toolchain packages to fix minority platforms might have to be generic cleanups to stay in the mainline…

7c6f434c:
I guess another interesting question is whether we want to define a place for the support table to live, if we want the low-acceptance additions to be cheap and only increases in acceptance to require RFCs…

matthewbauer:
> I guess another interesting question is whether we want to define a place for the support table to live, if we want the low-acceptance additions to be cheap and only increases in acceptance to require RFCs…

I think it's definitely important to document what's built on Hydra + what we have bootstrap tools for.

7c6f434c:
I agree, and maybe this RFC should include the initial list, but where do we want to maintain the updates to the list?

matthewbauer:
Perhaps here: https://nixos.org/nixpkgs/

Although that's not so visible...

Ericson2314:
> @Ericson2314 there is a bit of selection bias — the fixes applied to non-compiler-toolchain packages to fix minority platforms might have to be generic cleanups to stay in the mainline…

Sure, but that we can "unblock" ourselves at all while still writing good upstreamable patches is testament to that we have a good methodology. If we were failing to support new platforms because patches were rottting or not being accepted, then I'd say more selection bias going on.

7c6f434c:
@matthewbauer maybe visibility is less of an issue, Tier 2 and below are not supposed to attract new users from scratch — to use them one needs to be comfortable and knowledgeable. Once one has read the Nixpkgs manual, Nixpkgs homepage is not so invisible.

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,218 @@
...
+* `x86_64-linux`, `clang`+`glibc`
+
+### Tier 3
+
+Aims: some packages are expected to work
+
+Platform-specific fixes limited to general cleanups of non-standard
+assumptions in the upstream code and basic toolchain fixes
+
+* `aarch64-none`
```
edolstra: What is `X-none`?

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,218 @@
...
+Support: native bootstrap tools are available, cross-build toolchains in the
+binary cache
+
+* `armv{6,7,8}*-linux`, `gcc`+`glibc`
+
+* `mipsel-linux`, `gcc`+`glibc`
+
+* `x86_64-linux`, `gcc`+`musl`
+
+* `x86_64-linux`, `clang`+`glibc`
```
edolstra: I would drop `x86_64-linux`+clang to a lesser tier, since from the end-user perspective, it doesn't really provide any value over `x86_64-linux`+gcc.

rfcs/0046-platform-support-tiers.md
7c6f434c: Embedded targets that seem to be somewhat supported by our cross-compilation tooling.

edolstra:
Currently, this RFC is vague about what "support" means, and more importantly, who is tasked with providing that support. For example, suppose that somebody submits a PR to the Firefox Nix expression, which causes Firefox to break on some Tier 3 platform. Is it the responsibility of the PR submitter, the package maintainer, or the people who care about that platform (let's call them "platform maintainers") to fix it? Or is it fine to just leave it broken? Is the package maintainer required to accept (potentially intrusive and hard-to-maintain) patches to add support for some platform?

It may also be worth defining a tier that is not only unsupported, but not even allowed in Nixpkgs. For maintainability, it's important to be able to say "no" to PRs that clutter Nixpkgs with patches and special casing for obscure platforms.

We should also define what tiers mean with respect to the binary cache and channels. For example, if we say that Tier <= 2 are built by Hydra, then we can't have `mipsel-linux` as a Tier 2 platform because we don't have the machines. We also need to define what constitutes a channel blocker. For example, if stdenv breaks on a Tier 2 platform, can the Nixpkgs channel still proceed?

rfcs/0046-platform-support-tiers.md
7c6f434c: Indeed, it looks like the largest thing depending on this is just `afl`

7c6f434c:
Thanks, I forgot about some of the points and they had not surfaced in the previous discussion.

Whether the tiers should be a linear list or a list of dimensions is also a good question…

> For example, suppose that somebody submits a PR to the Firefox Nix expression, which causes Firefox to break on some Tier 3 platform.

If there is no tooling support (more or less Tier 2 and below, maybe this should be caled «Tier-2 tooling support and below» for clarity in transitional situations), then only those who have hardware have a chance of noticing anyway — and it is them who are expected to fix or mark as broken on a platform. For Tier-3 it is completely OK to just say «Firefox currently broken here».

This is mentioned as 

>> Package updates might break build on the platforms of this tier and lower

> Is the package maintainer required to accept (potentially intrusive and hard-to-maintain) patches to add support for some platform?

In the current version, for Tier 3 the patches (for non-toolchain packages) are required to be general cleanups not specific to a single minor platform. So far it looks like cleaning up assumptions not based on any standard is enough to fix many platforms at once (well, then there are properly hardware-dependent packages that can be harder to fix). Maybe a package where such fixes are too intrusive should be declared os-specific.

> For example, if we say that Tier <= 2 are built by Hydra

Currently it is said that Tier 1 and 1.5 are the tiers natively built by Hydra (Tier 2 receives cross-toolchains from Tier 1 and maybe Tier 1.5 to these platforms, which seems to reflect current situation)

> For example, if stdenv breaks on a Tier 2 platform, can the Nixpkgs channel still proceed?

Native stdenv is not built (and not expected to be built) for Tier 2. But yes, the channel status of the builds related to the platforms will be documented. Do I understand correctly that currently only native `stdenv` is channel-blocking, including `i686-linux`?

> For maintainability, it's important to be able to say "no" to PRs that clutter Nixpkgs with patches and special casing for obscure platforms.

Well, as anything from Tier 2 down can be dropped if it creates a problem after an update, and Tier 3 gets more or less a single patchset for all platforms (if something cannot be applied uniformly, it is not a generic cleanup), all of the lower-tier platforms cannot even come close to the maintainability headache created by macOS.

shlevy:
I propose carrying over the shepherd team from the PPC64 RFC: @ryantm @zimbatm @grahamc @Ericson2314, with @Ericson2314 as leader.

shlevy:
I propose carrying over the shepherd team from the PPC64 RFC: @ryantm @zimbatm @grahamc @Ericson2314, with @Ericson2314 as leader.

shlevy:
I propose carrying over the shepherd team from the PPC64 RFC: @ryantm @zimbatm @grahamc @Ericson2314, with @Ericson2314 as leader.

ryantm:
@shlevy okay with me.

7c6f434c:
@matthewbauer in light of the discussion about various representations of support tiers, what is your opinion?

Synthetica9:
There was some discussion on this topic, starting here: https://logs.nix.samueldr.com/nixos-dev/2019-05-10#2209522;

matthewbauer:
> @matthewbauer in light of the discussion about various representations of support tiers, what is your opinion?

Seems like reasonable changes. The biggest question I see is whether a tier represents "what we think can work in theory" (@ericson2314) vs. "what's we want to advertise as working" (@edolstra). Nixpkgs portable packages principles and large contributor base means that lots of stuff *could* work, but not necessarily is guaranteed to work. I'd like to advertise what we can guarantee works, with an added caveat that we design our Nix expressions so that anything could work.

> There was some discussion related to this topic, starting here: https://logs.nix.samueldr.com/nixos-dev/2019-05-10#2209522;

Yeah there is already a "cross-trunk" jobset. I would be hesitant to incorporate it into release.nix though. Cross is, in my mind, still not the "primary" concern of Nixpkgs in the same way that x86_64-linux or x86_64-darwin support is.

Synthetica9:
> Yeah there is already a "cross-trunk" jobset. I would be hesitant to incorporate it into release.nix though. Cross is, in my mind, still not the "primary" concern of Nixpkgs in the same way that x86_64-linux or x86_64-darwin support is.

I'm not suggesting it should be in release.nix, but currently only a few packages per arch are built. I'd suggest building a larger set (probably less frequently, like once per week) on one or two selected cross-compilation vectors.

7c6f434c:
@matthewbauer but how many dimensions do we want?

rfcs/0046-platform-support-tiers.md
```diff
@@ -80,6 +84,8 @@ Platform-specific things for arbitrary packages should not be too complicated
 Support: native bootstrap tools are available, cross-build toolchains in the
 binary cache, partial tooling support
 
+Native stdenv is a channel-blocking job
```
Ericson2314: this words is a bit confusing because there is a heavily bitrotted "impure" native stdenv.

7c6f434c: Argh, indeed.

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,241 @@
...
+mentioned in the upstream bugtracker.
+
+* `aarch64-linux`, `gcc`+`glibc`
+
+A team @NixOS/aarch64-maintainers shall be created to include people who
+understand the platform and use it.
+
+If there is a complicated problem on this platform when updating a package
+that was previously built succesfully on macOS, @NixOS/aarch64-maintainers team
+should be informed.
```
vcunat: Nitpick: `s/macOS/aarch64/`, I believe.

7c6f434c: Thanks

7c6f434c:
Tried to split the tier definitions into the three groups of concerns (which might be relevant for future dynamics)

Ericson2314:
@7c6f434c and @matthewbauer the thread has quieted down. Is this ready real-time discussion with the shepherds?

lheckemann:
Do we _have_ a shepherd team?

Ericson2314:
@lheckemann maybe? :) There's https://github.com/NixOS/rfcs/pull/46#issuecomment-490887104 then https://github.com/orgs/NixOS/teams/rfc-steering-committee/discussions/41. Maybe a step was elided as the nominated team we taken from the other RFC?

oxij:
Overall, :+1:.

Typos:

-  get something tested with reasonable effort?
+  get something tested with a reasonable effort?

- Some packages work
+ Some packages work.

Then,

> ## Main dimensions

looks very reasonable, but then

> ## Current platforms

makes up another ad-hoc dimension? Confusing.

IMHO, it's better to list `config.sub`/platform strings under their respective tiers in "Main dimensions" section and then edit "Current platforms" to be a plain list with descriptions. (I would also switch the order of these sections in the document.) Especially if this RFC is expected to be edited/updated from time to time to match the current state of things.

7c6f434c:
Thanks for the feedback!

A part of the reason for the current order is that tiers 1.5 and 2-ε seem to exist and be observably different from 1 and 2 but I wanted to introduce a clean picture before going into the gory details.

shlevy:
@lheckemann Yes, sorry, that was the decision :sweat_smile:

7c6f434c:
@matthewbauer after adding you as a coauthor I would prefer some opinions from you on the structure of the explanation of the current tiers (in light of all the feedback) before I just try to react to all the feedback and make a few random judgement calls.

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,270 @@
...
+
+Work ongoing to provide/merge Tier 4 support
+
+* `wasm-wasi`
+
+* `powerpc64le-linux`, `gcc`+`glibc`
+
+### Tier 5
+
+No current support, but previous support of clear path to add support
```
danbst:
```suggestion
No current support, but previous support or clear path to add support
```
I guess a typo?

7c6f434c:
>+No current support, but previous support of clear path to add support
>
>```suggestion
>No current support, but previous support or clear path to add support
>```
>I guess a typo?

Indeed, thanks

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,272 @@
+---
+feature: platform_support_tiers
+start-date: 2019-04-28
+author: Michael Raskin
+shepherd-team: Ryan Mulligan, Jonas Pfenniger , Graham Christensen, John Ericson
```
ryantm:
```suggestion
shepherd-team: Ryan Mulligan, Jonas Pfenniger, Graham Christensen, John Ericson
```

danbst:
I propose that every platform for every tier should have a dedicated team.

For example, in case of https://github.com/NixOS/nixpkgs/issues/64471 (which I failed to resolve in 30min) I'd like to tag @nixpkgs-static team, but there is no such.

Maybe this is orthogonal to Tiers, but if we talk about "support", support team should be defined as well.

7c6f434c:
@danbst  Thanks for the suggestion; I think it is on-topic, but I am also not sure how to divide the team scopes. Should there be a team for `x86_64-linux`+`glibc` (dynamic)? How many 32-bit ARM Linux platform teams do we need? Etc.

I would like to add something like this if there is a natural-looking outline, but I am not ready to try to sketch all the distinctions.

danbst:
Yeah, Tiers do not map naturally to desired support team. 

- so, Tier-1 doesn't require special team, as we all deal with that.
- Tier-1.5 requires 2 teams - `@team-aarch` and `@team-darwin`. Those who care about package availability on these platforms.
- from Tier 2-5 I can extract need for `@team-x86_64-exotic`, which basically means everything, except `gcc+glibc`: `gcc+musl`, `gcc+musl static`, `clang+glibc`, `clang+glibc llvm linker`, `gcc+glibc static`, `gcc+glibc llvm linker`. The reasoning is that if you had ever maintained one of these configurations, you may be familiar with issues for another one (but I can be wrong here).
- from Tier2-5 I can also extract `@team-embedded`, those who care about embedded stuff. I think AVR guys won't be too distracted if pinged issue is ARM-related.
- and one more team should be `@team-exotic`, who cares about any testable platform. "Testable platform" is platform, which we can provide build support for (crosscompiling) and ability to simulate (predefined scripts to run binary in VM with correct architecture). For example, if we can run programatically run Windows in a VM and run a crosscompiled for mingw binary there, then this is on-topic platform for `@team-exotic`.

`@team-exotic` can ping other teams for questions/review.
-----

This brings to another idea on how to split Tiers.
- Tier 1 is our official, most tested platforms
- Tier 2 are emerging, well-tested platforms
- Tier 3 are platforms, for which we have instructions on how to perform build/testrun manually
- Tier 4 are platforms, which some of us care about, but no build/testrun instructions documented - you should figure out everything yourself
- Tier 5 are those which nobody cares and we have no instructions

7c6f434c:
That sounds good; `@team-exotic` might also reinforce the idea that Tier-3-tolerance systems normally get fixes useful for many Tier-3-tolerance systems at once.

Did you mean to also have some static-specific team.

danbst:
@matthewbauer , what do you think about covering -static, -musl, -clang and -llvm linker topics under single umbrella of `@team-x86_64-exotic` or even `@team-exotic-platforms`?

matthewbauer:
> @matthewbauer , what do you think about covering -static, -musl, -clang and -llvm linker topics under single umbrella of `@team-x86_64-exotic` or even `@team-exotic-platforms`?

Yeah that sounds reasonable to me.

Ericson2314:
Sheesh it has been almost a month since it was cleared up that there is a shepherd time and I am indeed the leader :blush:. I am on vacation this week, but I'll be back next week (and also lowering the variance of our timezone distribution in the process). Should we schedule a meeting for then? @7c6f434c @matthewbauer  @ryantm @ryantm @zimbatm @grahamc.

7c6f434c:
(sigh; moderately exhausted, with a medium cold, trying to juggle conflicting deadlines…)

Ericson2314:
@7c6f434c yeah I feel you. Not sick ATM but certainly juggling multiple projects and deadlines. In my view at least, the shepherds and process owes the authors a "speedy trial" but aren't obliged to rush the authors, especially as someone else can go fork champion the RFC instead.

We can either just wait for a better time, or you and @matthewbauer can decide he alone is fine for the first meeting. (The process as written prioritizes public discussion, so I think one one author for an initial meeting is fine.)

7c6f434c:
@Ericson2314 well, the sick part is not that bad per se, but obviously amplifies my default slightly negative attitude to voice calls…

@matthewbauer what is your opinion on the order of «passes» in the RFC? There are definitions of main tiers (per dimension), definitions of special tiers (2-ε, maybe 1.5), lists of platforms per tier (per-dimension and global) — what of these should be included, and in what order (some of the lists can go together, of course)?

Maybe it is a good idea to make some decision based on the feedback and apply the decision to the RFC before the synchronous call. In general I am surely OK with @matthewbauer participating in the meeting without me and posting the notes (or saying that there is no difference in impression with shepherd team notes).

7c6f434c:
Re: platform teams: maybe we should discuss a timeout such that «ofBorg fails on darwin and mentioning the darwin team doesn't lead to reaction → update can go forward, package is marked broken on macOS until someone fixes it»?

zimbatm:
@Ericson2314 any time next week except Wed should work for me

zimbatm:
I think we could reduce the complexity of this RFC drastically by changing the `meta.maintainers` attribute to map maintainers to architectures.

One issue that I have with the support tiers is that if I am only on Darwin and submit a new package, how am I going to test it for NixOS? Or the other way around. As a maintainer I should be allowed to say that I support the package but only for a given set of architectures. Otherwise it's expanding on the things that I decided I was willing to support, which becomes work.

So let's say that I submitted `nixpkgs.curl` for NixOS. It might compile for Darwin but since I am only supporting Linux I become free to break the Darwin build during upgrades. Same for any packages that depend on it. If somebody else, or a team wants to also support Darwin they can become the maintainers for that.

TODO: we also have to define what it means to be a maintainer at some point. Are there any obligations attached to it?

Once that's in place, each architecture supports the subset of packages for which they have maintainers. And if you want to assign teams then you can decide on your own which subset you really want to support. The only other attribute is whenever the packages are built on the main Hydra cluster or not.

7c6f434c:
>So let's say that I submitted `nixpkgs.curl` for NixOS. It might compile for Darwin but since I am only supporting Linux I become free to break the Darwin build during upgrades. Same for any packages that depend on it. If somebody else, or a team wants to also support Darwin they can become the maintainers for that.

That is likely to lead just to a split of a package into two versions once the two teams have different opinions on the optimal date for a version bump… I don't say it is necessarily a bad idea.

And also there are some opinions about the amount of annoying patches permitted in various cases — the «tolerance» tiers in terms of the current proposal — which are dependent on tooling and current coverage.

zimbatm:
If need be, it might not be a bad thing to split the package in two. Some less maintained architectures might depend on an old version of a library. Same with the patches, they should probably only be added on a per-architecture basis unless all maintainers agree to maintain them.

Another advantage is that once the maintainer role and responsibility is better defined, it will become possible to use a hasMaintainer meta check which is essentially a more strict version of `allowBroken = false`.

davidak:
> but where do we want to maintain the updates to the list?

The supported plattforms should also be updated here: https://nixos.org/nixos/about.html

Maybe we can have it as structured data and just import at the needed places in documentation generation? Then it will not get out of sync.

Related: https://github.com/NixOS/nixos-homepage/issues/285

danbst:
@zimbatm I'd say, that `meta.maintainers` is kinda joke currently. I even dare to say, this should not belong to `nixpkgs` repo, it should be managed centralized in some other way. It is only a _guidance_ on who may be a maintainer, not a definition.

But we use it to define people commitment. Well, people need commitment, but this different from end-user support.

I think this RFC omits the per-package platform support question, because it is yet to be defined. It's purpose is more of whether _some platform_ is supported by nixpkgs, up to which point and percentage of packages possible to build/cross-build. This information should published, yes.

oxij:
Well, regardless, adding an option to `maintainer-list.nix` that lists systems one has access to seems like a good idea to me. The only technical question there is whether `maintainer-list.nix` should import `lib` to use `lib.platforms` or not (i.e. use strings instead).

I also think membership in GitHub teams should be mentioned in `maintainer-list.nix` (if not just straight derived from it) since `nix-instantiate` is much easier to use than GitHub API.

As to what `meta.maintainer` should mean. I think, OfBorg should mention `meta.maintainer`s in PRs with 1-10 changes that change packages those maintainers maintain. Because I wish it did so for packages for which I have added myself to `meta.maintainers`.

7c6f434c:
I kind of still hope that once we define the list of relevant platform/platform groups here, the maintainer list field addition can go via a normal medium-impact PR. (As for pinging: I hope #39 gets resolved in some direction and everyone gets the pings/review requests for the maintained packages…)

zimbatm:
Maybe we can schedule a call next week to clear things up? I think we all agree that having the platforms better defined is a good thing. The question is how to achieve this.

The current version of the RFC is quite complicated, it spends a lot of effort defining and mapping all the possible states a platform can be in. This maps to the currently existing complexity.

If I put myself into a new user's shoes, all I want to know is if the platform is officially supported or not. Iinstead of creating this gradient of support, would it be possible to simplify the list somehow?

For example, promote nixpkgs-darwin and nixos-aarch64 as officially supported platforms. Ensure that both platforms have Hydra channels dedicated to them. Then all the rest, define that it's allowed to host them in the nixpkgs repo and have wiki pages to maintain their individual state?

7c6f434c:
Re: new-users: I stand my ground: https://github.com/NixOS/rfcs/pull/46#issuecomment-489210564

(Tier 1 is the flagship platform, Tier 1.5 is supported, come back for more when you can read Nix and understand the intent of default `stdenv` phases)

On the other hand, if our new users start by reading the actual RFCs, maybe we should ask them to read the three manuals first, and after the three manuals this RFC should not be too hard to understand.

I think that developers notice the distinction of Tier 2 working better than Tier 3, with the corresponding difference in the patch tolerance. As long as we don't have default NixOS installation using multiple overlays (or will these be Flakes?), I would say that purging borderline-acceptable Tier-2 patches from mainline Nixpkgs would be a loss.

CrystalGamma:
Let me make another proposal to hopefully get the ball back rolling on this:

* add a `meta` option `supportlevel` for packages, with initially supported values `core` (all packages required to build a cross-stdenv), `bootstrap` (all packages required to bootstrap a pure stdenv on the platform and build bootstrap files), `system` (all packages required to build a reasonable basic configuration of NixOS) and absent / null.
* add nixpkgs option `maxsupportlevel` with the same values, with `pkgs/stdenv/generic/check-meta.nix` disallowing packages above the specified level, to allow automatic detection of wrongly-classified packages in Hydra
* platform support is then determined by what support level can be expected to work for the given platform; specifically core-level support (targeted for things like microcontroller platforms) being build of a cross-stdenv, bootstrap-level support being cross *and* native build of bootstrap files and native pure stdenv, system-level support being building a certain reference configuration of NixOS (which needs to be bootable), and full support being all packages not marked as `broken` for the platform in `meta`.
* acceptance level is determined by the minimum of *platform-specific* patches and Nix expression conditionals required to achieve the corresponding support level. In particular this means platform-specific patches should not be accepted for packages that are of higher `supportlevel` than the given acceptance level. Patches that isolate/remove non-portability (such as assumptions that are only true on x86 or (x86 || ARM)) are generally exempt from this, i. e. they not only have to be accepted (if they are technically sound) for packages of  `supportlevel` up to acceptance level, but also should be considered on packages on a higher level, as long as their size isn't excessive (which is hopefully rare).

Examples:

* x86_64-linux-glibc has full support
* aarch64-linux-glibc has system-level support, but full acceptance
* x86_64-linux-musl and aarch64-linux-musl and have bootstrap-level support, but full acceptance
* arm-linux-glibc (and potentially arm-linux-musl?) has bootstrap-level support and system-level acceptance
* avr-none has core-level support and acceptance
* ppc64le has core-level support, but would go up to bootstrap-level if bootstrap files (could be cross-built last I checked) were provided.

Open questions, random comments:
* infrastructure: core-level should be easily CI-checkable by building cross-stdenvs. Only for bootstrap-level and up are native systems required. This should make acceptance of new core-level platforms reasonably easy, maintenance-wise.
* responsibility to fix: changes that break CI need to be backed out if not repaired by the original author. For (platform-specific) breakage not detected by CI, a maintainer team for each platform might need to be established to fix the bug.
* how to determine system-level support for platforms that can't build NixOS (i.e. macOS)? Just build all packages with `supportlevel` = `system` that are available for that platform?
* most changes that require full acceptance (and to some extent system-level acceptance) can be provided by overlays as long as that acceptance level is not reached by mainline nixpkgs.
* changes to NixOS modules (e.g. bootloaders)? these are not providable by overlays so far AFAIK, so maintaining them out-of-tree is a pain. Unsure how to compensate for that as long as there is no extension mechanism akin to overlays.

7c6f434c:
@CrystalGamma this looks like a sketch for a competing RFC (and maybe it is indeed a cleaner approach)

A question that is missing in your plan: I think there is a difference between «platform-specific patches» and «assumption-cleanup patches that just happen not to be needed for Tier 1» (my approach is basically all about trying to describe the current practices around that).

And for the current approach I would like some feedback what different people see as the clearest order of explaining the tiers and the platform list.

ehmry:
To approach the subject of adding new platforms to Nixpkgs from a different angle, perhaps its worth considering removing platform definitions from Nixpkgs? With the exception of x86_64-unknown-linux as a reference, platforms could be defined externally using a module system similar to NixOS. In this way exotic operating systems and new architectures could be built by overlaying the platform and patches over Nixpkgs, rather than forking Nixpkgs.

This is off-topic, but this discussion is a continuation of the ppc64(le) thread, and I don't see that support tiers solve the problem of stale patches for forgotten platforms or the reluctance to add exotic platforms.

7c6f434c:
> With the exception of x86_64-unknown-linux as a reference, platforms could be defined externally using a module system similar to NixOS.

Does this mean effectively demoting macOS support, because the mainline uses `x86_64-linux` and doesn't observe if an update breaks something on macOS? (Right now there is everything to see the failure, and then it is a human judgement call)

>This is off-topic, but this discussion is a continuation of the ppc64(le) thread, and I don't see that support tiers solve the problem of stale patches for forgotten platforms or the reluctance to add exotic platforms.

I do hope that having explicit expectations on how much impact can a freshly added platform have (and putting specific minimal requirements for promotion to more impact) will reduce the perceived risk of adding platforms.

Also, if we have a list of expectations, one can claim that some platform support has degraded below them. Asking whether someone is interested enough to fix might be felt more constructive than starting a removal discussion out of the blue.

ehmry:
>> With the exception of x86_64-unknown-linux as a reference, platforms could be defined externally using a module system similar to NixOS.

> Does this mean effectively demoting macOS support, because the mainline uses `x86_64-linux` and doesn't observe if an update breaks something on macOS? (Right now there is everything to see the failure, and then it is a human judgement call)

Ok, moving out actively maintained platforms is probably too drastic and destructive.

>> This is off-topic, but this discussion is a continuation of the ppc64(le) thread, and I don't see that support tiers solve the problem of stale patches for forgotten platforms or the reluctance to add exotic platforms.

> I do hope that having explicit expectations on how much impact can a freshly added platform have (and putting specific minimal requirements for promotion to more impact) will reduce the perceived risk of adding platforms. Also, if we have a list of expectations, one can claim that some platform support has degraded below them. Asking whether someone is interested enough to fix might be felt more constructive than starting a removal discussion out of the blue.

I am definitely for adding more platforms, and I have no reason to be against support tiers. I'm interested in allowing experimenters and researchers to build against exotic platforms, beyond any notion of support. I'm happy to make a complementary RFC if there is any other interest in this.

illegalprime:
I would also like to have better visibility into what cross compilation vectors work and what don't. They also seem orthogonal to the platform list (because, as I suspect everyone in this thread knows, cross compiling to aarch64 has different hashes than aarch64). Probably `x86 -> aarch64` is the most popular here and it can at least be _built_ on any x86 maintainer's machine, so maybe we should be more strict to not let cross build regressions make it into master?

Ideally i'd be able to see a % coverage for packages of a platform and separately cross compiling to that platform from another, but it's too much for now.

I also wonder if people can sponsor different platforms or cross-vectors for their professional use-case (just have packages built for them on hydra and possibly track their regressions).

7c6f434c:
>I would also like to have better visibility into what cross compilation vectors work and what don't.

Well, we can call cross-targets separate platforms if we want. In the current version some platforms are cross-only, actually, maybe we should separate that (and indeed duplicate `aarch64` and `armvX` which are also viable is direct native platforms).

> Probably `x86 -> aarch64` is the most popular here and it can at least be _built_ on any x86 maintainer's machine, so maybe we should be more strict to not let cross build regressions make it into master?

Given that impact is not supposed to be higher than support level, this probably needs some survey of current coverage status and maybe contributing a bit of cross-check sugar to ofBorg… Do we even have an official cross-compilation team?

>I also wonder if people can sponsor different platforms or cross-vectors for their professional use-case (just have packages built for them on hydra and possibly track their regressions).

Resource management for providing the tooling is orthogonal for the RFC which tries to introduce the language to describe the outcomes. Unless we introduce recognition for third-party Hydras as fully supported Hydras, we are not forced to care where the build machinces come from.

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,272 @@
...
+
+Work ongoing to provide/merge Tier 4 support
+
+* `wasm-wasi`
+
+* `powerpc64le-linux`, `gcc`+`glibc`
+
+### Tier 5
+
+No current support, but previous support or clear path to add support
```
alyssais: Illumos / Solaris (we’re inconsistent about what we call it in Nixpkgs) should be added here.

zimbatm:
I just saw that Zig is also adopting a tier support approach: https://ziglang.org/download/0.5.0/release-notes.html#Support-Table

rfcs/0046-platform-support-tiers.md
7c6f434c: What is the platform identifier you would pick here? It is not really a commitment to anything anyway.

alyssais: Let’s go with `x86_64-illumos` and `i686-solaris`, because I think that’s what we actually have targeted in the past. They should be differentiated because they’ve significantly diverged at this point.

7c6f434c:
>Let’s go with `x86_64-illumos` and `i686-solaris`, because I think that’s what we actually have targeted in the past. 

Done, thanks

shlevy:
@Ericson2314 What's your impression of current status here?

Ericson2314:
@shlevy Yeah I've been reminding myself it's time to try to schedule a meeting agin. I need to go re-read everything too, but that's no reason to wait

CrystalGamma:
As recommended by @7c6f434c I've submitted RFC #54 today, which may be of interest for people discussing here.

7c6f434c:
@Ericson2314 yes, now it is better.

Re: slow process —  please do not usurp all the blame; there were also times when I lacked energy to react in a timely manner, there were cases where I waited for feedback from my co-author on specific points despite forgetting to mention what these points _are_, etc. Then there was just a lot of useful feedback that typically makes any process slower (but hopefully more useful!)

7c6f434c:
OK, I have mentioned in #45 that next week Monday, Tuesday, Friday or Thursday afternoon should be fine with me, but of course I do not want intersecting timeslots…

7c6f434c:
In _this_ RFC discussion I can say that I think part of the reason why this RFC is more complicated and long than #54 is that I tried to collect corner cases.

I have incorporated some changes — stressing that ARM platforms are used both as cross and as native platforms and might have different issues, mentioning a team for exotic platforms, mentioning desirability to have test scripts (hardware-independent) for Tier-2 tooling.

Are there any changes suggested that I didn't respond to (or agreed and didn't apply)? Any new suggestions?

ehmry:
I volunteer, this is a topic that I've been tracking for a while.

7c6f434c:
@ehmry did you mean to volunteer to be a shepherd for #54 ?

ehmry:
Yes, #54, this is confusing.

shlevy:
@Ericson2314 Are we keeping this open? If so what's the status?

Ericson2314:
OK NixCon is done and I am not rushing around traveling. So let me try to schedule a meeting: https://www.when2meet.com/?8339971-9OoVc

Ericson2314:
OK, we finally had a meeting! We were missing people \[long story short, scheduling confusion, I take the blame for this], but we decided to hold a mini-meeting anyways.

First we compared RFCs 54 and 46.
Michael thought that 54 was currently simpler, but once it had stabilized a bit might no longer be simpler.
We agreed with this mini-meeting to just talk about RFC 46 for now.

We then talked about the non-natural-number 1.5 and 2-e tiers.
This lead to a more broad discussion on the difference between the definitional states, and wherever platforms happen to fall.
Aggregating all dimensions of a platform leads to finer-grained tiers, unless there is some rounding process.

Ryan thought it would good to move/delete description of the status quo  under "Current platforms".
I agreed, thought it would good in an "examples" section.
Michael was fine with moving it to be that or an appendix. 

We talked about the changes that have happened in practices.
In particular, we referenced when people (often me!) add some platform and Eelco and others worry about maintenance.
Michael reaffirms that the RFC says lower tier platforms should be made to work through platform-agnostic fixes.
Michael also says cross fixes (fixes for build != host, irrespective of concrete platforms) is a platform-agnostic fix but still can be invasive.

Then we talked about "statics vs dynamics" or the state each tier is defined to be vs the rules one how a platform transitions between them.
The RFC is currently clearer on what the tiers mean than how do you move between tiers.
Michael thought that lower tier plaforms can just "come into being", with a PR that meets the standards for that tier which is merged.
Michael was worried about writing down a policy for how to move to a higher tier, as that is more a social (admin time) and financial (build machines) than technical question that a code change alone cannot satisfy.

Ryan reference Eelco and others being worried about "slippery slopes" where we lower our standards over time.
Michael wasn't so concerned: if we follow the process at every step that should not happen.
Then we talked about Darwin.
Darwin didn't cause a slippery slope in adding more platforms, but it did become popular itself, and then as non-cli/daemon packages became the next thing to package, started taking more new patches than before/
In short it became popular and then more invasive.
Michael didn't want this RFC to say whether "Darwin was a mistake", that should happen separately.
Michael also brought up the Arm, which lead to Aarch64 and its lower tier.

7c6f434c:
I think I expect 54 to be still _simpler_ than 46 once it stabilises, but probably not as simple as now (which is also a bit longer than initial submission). If we accept complexity level of 46, 54 stays useful as it defines properties of packages (this could also be referenced in the «Package support» part of 46).

Moved platform list to a non-normative appendix, mentioned dynamics, stressed the limitations of fixes at Tier-3 and below, and presence of financial decisions at Tier-2 and above.

This RFC can be interpreted as written under assumption that Darwin and ARM and x86_64 were not a mistake.

ryantm:
Here are the notes from our meeting on Thursday

# NixOS RFC 0046, Nov 16, 2019 meeting

## Participants

* Graham
* Ryan
* Michael
* zimbatm

* John

## Decisions

* Explicitly say that the only immediate outcome of RFC process is an update to the Nixpkgs manual
* (well, we do need to update Nixos.org/nixpkgs to not claim i686-linux is better supported than aarch64-linux)
* Make Tiers centered around impact, the rest are requirements
* Slightly edit the procedure descriptions for changing tiers
* 1.5 → 2, 2 → 3…
* We want to invite CrystalGamma to the next meeting

## Chatlog



Zimbatm

* this is partly about maintainer responsibilities.
* complexity in number of tiers

Michael

* much of the complexity might be inherent
* platform support is not binary

Zimbatm

* can we move the platform support tier into the package and make it the package maintainer's responsibility?

Graham
* talk of adding risc-v or power9 builder or arm-v7
* "when it gets to it" support tier for builder (maybe weekly)

Zimbatm

* pull model vs push model
* push: for officially supported, ideally merges only happen when the CI is green
* pull: maintainers fix packages asychronously

* add actionable items to the RFC?

Graham

* glibc musl - willdietz argued that getting musl in would help improve glibc too


People's experience with porting to other platforms
* Zimbatm: not exotic ones
* Graham: not much
* Ryan: not much

Zimbatm

* The RFC does not appear to have much of a roadmap

Michael

* It has a bit of a procedure

Ryan
* RFC will be something to point to in PRs & have a common vocabulary


Zimbatm
* Might be useful to have a guide from the perspective of someone who is adding platform support

Ryan
* Tiers maybe should be the top-level thing

John arrived

Zimbatm
* Can we get rid of the 0.5 tiers by renumbering?

Graham
* Let's get rid of epsilon

Zimbatm
* We need to clarify what a maintainer means because maintainers cannot support all the different platforms. Maybe out of scope from this RFC.

John
* Would like to see it possible to add a platform without resistance as long as it does not pose a maintenance burden.
* Windows is difficult because of no bash, so definitely OK with requiring community agreement first
* Maybe there is less resistance to adding platforms now

Graham
* It would be nice to have intermediate goals for progressing up support tiers.

Michael
* How long should be people wait to merge as non-platform user/maintainer?

John
* Requirements for tier *n* vs maintenance burden of tier *n*
* Tolerance/Impact is the output of the tier, Tooling and number of working packages in the input

RyanTM
* seems like this could simplify it a lot by saying that the only tiers we externally communicate are the impact/tolerance with the other ones being requirements of that

Michael
* Higher impact/tolerance tiers probably require RFCs
* The outcome of the RFC is accepted is to add a section to the manual for what platforms support what tiers and definitions of tiers

Zimbatm
* We should update the website to talk about the platform

Michael
* https://nixos.org/nixpkgs/ is not up to date and indicates i686 is our main platforms

RyanTM
* of course we should get something on the website because marketing this will make it better for all of us

Michael
* nixpkgs on non-NixOS platforms (and NixOS.org homepage)


We are planning to meet at the same time next week (John might not be there.)

7c6f434c:
I ended up changing more than expected, so both feedback on the significantly rewritten content and on the general correctness of my shuffling stuff around is needed.

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,426 @@
...
+needed to support the platform are allowed to be, and the amount of effort
+spent on avoiding the changes causing platform-specific regressions.
+
+To qualify for a tier, the platform support should meet requirements on package
+coverage (how useful it is to users), and tooling (how easy it is to check the
+changes on the platform without access to the platform, and to start using the
+platform). Recognition of the lower tiers (Tier 5 and below) is based on
+a technical review of the platform support code quality and requirements. For
+higher tier an explicit decision has to be made based on the technical merits,
+future expectation. resource allocation and project priorities. As this is not
```
ryantm:
```suggestion
future expectations, resource allocations and project priorities. As this is not
```

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,426 @@
...
+
+## Main support dimensions
+
+### Impact (support tier)
+
+This dimension describes the permissible impact of a platform on the Nixpkgs
+repository. This includes the necessary patches as well as the handling of the
+build failures of updates in Nixpkgs.
+
+### Tooling requirments
```
ryantm:
```suggestion
### Tooling requirements
```

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,426 @@
...
+
+This dimension measures how easy it is to find out what works and what doesn't
+on the platform (in particular, without being a user of the platform).
+
+### Package coverage.
```
ryantm:
```suggestion
### Package coverage
```

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,426 @@
...
+
+This dimension measures avilability of the most important packages on the
```
ryantm:
```suggestion
This dimension measures availability of the most important packages on the
```

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,426 @@
...
+
+#### Impact
+
+Completely platform-specific fixes are expected to be rare and non-intrusive.
+Fixes to the compilation toolchains are expected.
+General cleanups of non-standard assumptions (e.g. «everything that is no x86
+is a kind of ARM» or «malloc(0) behaviour is a reliable indicator of other
+malloc features») useful for this platforms are welcome.
+
+Updates might break builds on this platform .
```
ryantm:
```suggestion
Updates might break builds on this platform.
```

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,426 @@
...
+
+Most packages work, credible ambition to reach Tier 1 coverage at some point.
+
+### Tier 3
+
+#### Impact
+
+Completely platform-specific fixes are expected to be rare and non-intrusive.
+Fixes to the compilation toolchains are expected.
+General cleanups of non-standard assumptions (e.g. «everything that is no x86
```
ryantm:
```suggestion
General cleanups of non-standard assumptions (e.g. «everything that is not x86
```

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,426 @@
...
+is a kind of ARM» or «malloc(0) behaviour is a reliable indicator of other
+malloc features») useful for this platforms are welcome.
```
ryantm:
```suggestion
malloc features») useful for these platforms are welcome.
```

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,426 @@
...
+
+Most of the popular packages work.
+
+### Tier 4
+
+#### Impact
+
+Fixes necessary for this platforms must be either limited to compilation
+toolchains, or general cleanups of non-standard assumptions (e.g. «everything
+that is no x86 is a kind of ARM» or «malloc(0) behaviour is a reliable
```
ryantm:
```suggestion
that is not x86 is a kind of ARM» or «malloc(0) behaviour is a reliable
```

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,426 @@
...
+motivation for generic cleanups, and further tiers require commitment of
+recurring resources. This is the reason for Tier-5 addition and Tier-4
+promotion to happen inside the scope of normal technical review by the people
+working with similar platforms; further tiers require allocation and of
+hardware resources, and procedures for coordinating such financial decisions
+are out of scope for this RFC.
+
+When a platform starts falling out of use, its support tier (and permitted
+impact) is reduced once it becomes clear that the current tier requirements
+will stop being met in the near future. Platform-specifing patches no longer
```
ryantm:
```suggestion
will stop being met in the near future. Platform-specific patches no longer
```

7c6f434c:
Thanks for the careful reading!

7c6f434c:
@CrystalGamma I would like to request a permission to use your package scope definitions as a part of package coverage guidelines in #46, and, if you are interested, invite you as a coauthor of the current RFC.

You might be interested to know that the structure of tiers has been somewhat reworked here (and also the discussion is actually moving forward now). Regardless of whether you are interested in becoming a coauthor, you are definitely welcome to participate in the Shepherd meetings on #46 (Thursdays 14:00 UTC).

If you are interested in merging more of the #54 into #46, PMG expectations for high-tier platforms would probably also make sense in #46.

ryantm:
Comments from meeting:

## Participants 2019-11-21

* Michael
* Ryan
* Zimbatm

## Decisions

* Tell John to call for RFC FCP

## Chatlog

Ryan

* the updates mostly match what we had talked about

Zimbatm

* I thought we wanted to extend the motivation a bit

Zimbatm

* Four angles of motivation
  1. Common language
  2. Clarifies what to users what will work 
  3. What should the infrastructure support?
  4. Maintainers need to know whether their package should work

Michael

* Require platform maintainer groups?
* Darwin has 4 maintainers, one is maybe inactive

RyanTM

* RFC 54 would say Darwin's support would be restricted by the maintainer number

Zimbatm

* Lets limit prescribing rules because this is open source

Michael

* Should we have a ARM{6,7,8} maintainers group?

Zimbatm

* No
* What is the form the team should take?


RyanTM

* It would be nice to know how to get access to teams

Zimbatm

* Does a team have IRC?
* Looks like the RFC is in a pretty good state

Michael

* Can I reuse stuff from 54?

RyanTM

* The license is CC-SA

Michael

* Maybe copying stuff sometimes is a passive agressive thing

Zimbatm

* Ryan what do you think about the RFC?

RyanTM

* I'm happy with it and would vote in favor of RFC entering FPC with disposition to merge


Zimbatm

* I would too

Michael

* I don't have pending edits to do

Ryan

* What did you want to take from 54?

Michael

* Maybe something about package guidelines
* Let's not do that now.
* I'd help with 54 after 46 is wrapped up (but right now I'm a little biased)

Zimbatm

* Instead of hard requirements, we should try to have tooling
* Once we have a good way of managing teams, we could make an RFC to make that more clear

Michael

* I think it is OK to have an RFC to create new guidance, if people are allowed not to change their current

ryantm:
@Ericson2314 zimbatm, 7c6f434c, and I are in favor of entering FCP with disposition to merge as it stands. So we need you to officially call for the FCP and @grahamc to sign off.

Ericson2314:
I'll go with the consensus, but should we do FCP with disposition to merge before #54 has had a meeting?

ryantm:
I just realized I'm the Shepherd leader for #54! Well, now I'm biased. 

During the meeting zimbatm, 7c6f434c, and I decided that we should try to get this one through, then work on incorporating whatever good stuff is left in #54. I think this still holds, but we should make a big deal in the FCP announcements that there is a competing RFC that people might prefer.

Ericson2314:
So that plan includes potentially merging this (if no objections during FCP, and timer goes off) and then merging the next one after?

ryantm:
Yes, and #54 would most likely need substantial reworking to remove redundant parts.

Ericson2314:
OK, well then let it be final comment period with disposition to merge.

7c6f434c:
Technically, for #54 there will be a choice of being an amendment or dropping the only impact-limiting rule it has and becoming an RFC about language for supporting specific sets of packages.

ryantm:
@Ericson2314 I'm not sure we should open the FCP until @grahamc has signed off. By the way, when the shepherd leader opens the FCP, it is their responsibility to advertise it widely (Discourse, NixOS weekly, etc.).

Ericson2314:
@ryantm OK sorry. I edited the above, and will do those things if @grahamc approvals. I guess one of my mistakes made the other less bad!

domenkozar:
ping @grahamc if we should move this to FCP.

Profpatsch:
(meta) Can we add the sheperds of RFCs to the original post? It’s hard to reconstruct from history with Github’s folding feature.

ryantm:
@Profpatsch We have the shepherds in the [RFC text](https://github.com/7c6f434c/rfcs/blob/platform-support-tiers/rfcs/0046-platform-support-tiers.md).

ryantm:
@Ericson2314 graham told me he is having difficulty getting to this RFC, and everyone else is in agreement, so I think we should move forward with FCP. He can still comment during FCP, and, of course, we can make tweaks outside of the RFC process, or major changes in a future RFC.

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,430 @@
...
+achievement of the requirements defined in this RFC.
+
+* Discussions of platform support at the Tier 3 and above start using the
+vocubulary based on this RFC and ensure that the requirements of this RFC are
+met.
+
+## Platform elements
+
+* CPU architecture (possibly: subarchitecture, optional features)
+* OS kernel
```
ehmry: As mentioned in 54, I believe the phrase "OS ABI" is more accurate than "OS kernel" in this context. It may seem like a minor nitpick now, but this could paint us into a problematic corner later. Consider that the FreeBSD kernel can execute binaries compiled for the Linux ABI, but there are Linux utilities that interact with Linux-specific kernel interfaces (networking, power-management, VMs). Giving that set of packages with specific interactions with the Linux kernel is a very small subset of packages that can be executed on Linux, I argue that the ABI platform axis is more significant than the kernel platform axis.

7c6f434c: Agreed. I keep kernel, but move it next to the more exotic considerations like NixOS/non-NixOS layout

edolstra:
@ryantm Okay, do you want to declare the FCP?

ryantm:
The Final Comment Period for RFC 0046 has started with disposition to merge and, barring any blocking issues, will be merged after 2020-01-16. Your opinions, comments, and approvals are welcome!

rfcs/0046-platform-support-tiers.md
```diff
@@ -0,0 +1,431 @@
...
+* OS ABI
+* C compiler
+* C library
+* OS kernel
+* NixOS/non-NixOS global layout, in case of Linux with glibc
+* Global linking options
+
+## Questions to answer for each platform
+
+* What fraction of packages is expected to work?
```
michaelpj:
```suggestion
* What fraction of packages are expected to work?
```

7c6f434c: (looking up) oh indeed, the modern majority position is «are», even if there are still some style guides that favour «is» here (of course I only searched five minutes ago because when writing it was obviously «a fraction is» to me).

ryantm:
The FCP has ended without any objections. @NixOS/rfc-steering-committee please merge.

erictapen:
The RFC mentions an appendix to the Nixpkgs manual:

> If this RFC is accepted, the following changes will happen:
>
> * The definition of support tiers and a list of platforms with the corresponding support tiers is added as an appendix to the Nixpkgs manual. The list of platforms is further maintained as a part of the Nixpkgs manual. The platform list in the manual will be initially based on the list in the appendix of the present RFC.

Did this already happen? If not is there a way to help with it?

Mic92:
I don't think so.

7c6f434c:
>> * The definition of support tiers and a list of platforms with the corresponding support tiers is added as an appendix to the Nixpkgs manual. The list of platforms is further maintained as a part of the Nixpkgs manual. The platform list in the manual will be initially based on the list in the appendix of the present RFC.
>
>Did this already happen? If not is there a way to help with it?

No, not yet, sorry.

This requires a bit of writing (to edit the parts of RFC into something that will look natural in the manual) and I now have a sequence of completely unrelated deadlines including writing, and also it looks like the manual is getting some structural/format changes right now which doesn't help given lack of energy to write.

Also checking the recent additions to platform list in <nixpkgs>/lib/systems/platforms.nix and <nixpkgs>/lib/systems/doubles.nix and evaluation which of the low tiers should be assigned to them would be needed.

If you are willing to write the glue text (probably by composing it together out of pieces of RFC and editing), edit/update the list itself, and PR against Nixpkgs manual (I guess Markdown chapter would be easier because of preservation of some RFC formatting? And after the conversion we will see what happens), please mention me, I will review and (unless there are objections) merge it after some time (probably a bit more than a week).

L-as:
Is there a list of the platforms other than this RFC?

sternenseemann:
> how's the powerpc64le-linux support looking currently? I can't get things to build on my Power9 system.

There has been work on this platform recently by mainly @amjoseph-nixpkgs which is available in the unstable nixpkgs/nixos channels. There is no formal support / support requirement at the moment. You can report any specific issues you are encountering over at the [nixpkgs issue tracker](https://github.com/NixOS/nixpkgs/issues).

amjoseph-nixpkgs:
> > how's the powerpc64le-linux support looking currently? I can't get things to build on my Power9 system.
> 
> There has been work on this platform recently by mainly @amjoseph-nixpkgs

By cherry-picking the PRs below (and their dependent PRs) you can cross-compile a NixOS install image for powerpc64le-linux or mips64el-linux from x86-linux:

- https://github.com/NixOS/nixpkgs/pull/194153
- https://github.com/NixOS/nixpkgs/pull/192672

The powerpc64le-linux image even boots.  I haven't tried the mip64el-linux image yet.

In the long term, it would be nice if we could add an (absolutely **not** channel-blocking in any way, shape or form) Hydra job for these install images.  That way people can try NixOS on these machines without having to trust binaries from anywhere other than Hydra.

That's the limit of my ambitions for these platforms.  I don't think it makes sense to build anything more than the installer on Hydra.  It would be nice if we had a name for this level of functionality; something like "enabled but unsupported".

rjzak:
I haven't cross-compiling from x86_64 since I have a ppc64le machine I use as my desktop, and Void Linux PPC has a package for Nix, which seems to work (except for getting things to compile due to the stack checking method).

amjoseph-nixpkgs:
You can native-compile the installer too.

asymmetric:
Is there a reason why the `@NixOS/aarch64-maintainers` was never created?

zimbatm:
Probably because it wasn't clear who should do it. Let me know who should be on the team and I will create it.

asymmetric:
Any takers?

Atemu:
Well, at least we know why it wasn't created now.

kip93:
> Probably because it wasn't clear who should do it. Let me know who should be on the team and I will create it.

Hello!

Heard about this on matrix. Don't have much experience working on nixpkgs itself but I have been using nix for many years now, mostly for aarch64 systems and doing a lot of cross compiling, so maybe I can be of some help here? If so, I'd be more than happy to join and help.

delroth:
> Probably because it wasn't clear who should do it. Let me know who should be on the team and I will create it.

I use NixOS on 3 different AArch64 servers, happy to help with aarch64-specific issues.

aforemny:
> Any takers?

I am using NixOS on aarch64 machines (laptop, chromebook) and would be interested in becoming a platform maintainer. Could you please add me to the team?

PR #48, opened 2019-06-19
"accepted" directory -> "rfcs" directory

edolstra:
This reflects the fact that there is no "accepted" directory.


zimbatm:
The PR doesn't change any process, merely corrects the doc to correspond to our existing process so I think it could be merged without too much ceremony.

zimbatm:
The PR doesn't change any process, merely corrects the doc to correspond to our existing process so I think it could be merged without too much ceremony.

PR #52, opened 2019-09-05
[RFC 0052] Away from static IDs

Infinisil:
# Summary
[summary]: #summary

A lot of NixOS modules are [assigning static uids/gids](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/misc/ids.nix#L36) to their users. This has resulted in less than 90 static ids left in the reserved range from 0 to 400.

This RFC deprecates the practice of doing that and instead suggests to
- If applicable use systemd's `DynamicUser`
- Otherwise let NixOS assign dynamic persistent ids, which happens automatically when `users.users.<name?>.uid`/`users.groups.<name?>.gid` is not set. For users, `users.users.<name?>.isSystemUser` should be set so that only uids under 1000 are used by NixOS services.

Only in special circumstances are static ids allowed anymore.

[Rendered](https://github.com/Infinisil/rfcs/blob/dynamic-ids/rfcs/0052-dynamic-ids.md)

Partial implementation PR: https://github.com/NixOS/nixpkgs/pull/65698

cc @edolstra @aanderse @ryantm @arianvp @globin @volth @arcnmx

### DIFF
--- a/rfcs/0052-dynamic-ids.md
+++ b/rfcs/0052-dynamic-ids.md
@@ -70,7 +70,7 @@ An example:
 ### Ensuring correct directory permissions for services
 
 Services must be able to access their directories.
-- The easiest way to achieve this is to use `systemd.services.<name?>.serviceConfig.StateDirectory = "myservice"`, which ensures that `/var/lib/myservice` belongs to the services user. See `man systemd.exec` for other related directives.
+- The easiest way to achieve this is to use `systemd.services.<name?>.serviceConfig.StateDirectory = "myservice"`, which ensures that `/var/lib/myservice` belongs to the services user. See `man systemd.exec` for info on this and the related directives `CacheDirectory`, `LogsDirectory` and `ConfigurationDirectory`.
 - `systemd.tmpfiles.rules = [ "Z '/var/lib/myservice' - myuser mygroup - -" ]` can also be used, with the disadvantage that it will only run at system activation and not when the service starts. It also recursively fixes the permissions every time, meaning it can lead to considerable slowdown with many files.
 - An alternative is to assign `serviceConfig.ExecStartPre = "+${pkgs.writeScript "myservice-prestart" "..."}"` with a script to fix the permissions, where the `+` makes the script run with full root permissions as documented in `man systemd.service`.
 
--- a/rfcs/0052-dynamic-ids.md
+++ b/rfcs/0052-dynamic-ids.md
@@ -76,7 +76,7 @@ Services must be able to access their directories.
 
 ## Enlarge the reserved range of system users/groups
 
-How uid/gid ranges are decided: If `isSystemUser = false`, dynamic uids are allocated in the range `UID_MIN` to `UID_MAX`, while with `isSystemUser = true`, it's `SYS_UID_MIN` to `SYS_UID_MAX`. In contrast, gids are always allocated in the range `SYS_GID_MIN` to `SYS_GID_MAX`.
+How uid/gid ranges are decided: If `isSystemUser = false`, dynamic uids are allocated in the range `UID_MIN` to `UID_MAX`, while with `isSystemUser = true`, it's `SYS_UID_MIN` to `SYS_UID_MAX`. In contrast, gids are always allocated in the range `SYS_GID_MIN` to `SYS_GID_MAX`. See [#65698](https://github.com/NixOS/nixpkgs/pull/65698) for implementation details.
 
 Since the new recommendation is to use `isSystemUser` for system users and to not set static ids, the available ids now have the range `SYS_UID_MIN` to `SYS_UID_MAX` and `SYS_GID_MIN` to `SYS_GID_MAX`, defined in [shadow.nix](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/programs/shadow.nix#L13-L21), both of which currently span from 400 to 499. This means a NixOS system could only have 100 different services using dynamically allocated ids over its lifetime. Since this is not very much, this range will be changed to span from 400 to 999, reserving an additional 500 ids for system users/groups. The range from 500 to 999 is currently not reserved for anything.
 
@@ -96,7 +96,7 @@ Are there any problems when moving between different user declarations? Note tha
 
 In addition, a transition from `isSystemUser = false` to `isSystemUser = true` can't be done automatically, the `/var/lib/nixos/uid-map` file needs to be adjusted manually for that. However without doing so, NixOS will happily continue to use the previously assigned uid without problems. This means changing this value is unproblematic.
 
-Note that changing all current NixOS services to use dynamic ids is [future work][#future].
+Note that changing all current NixOS services to use dynamic ids is [future work][#future-work].
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/rfcs/0052-dynamic-ids.md
+++ b/rfcs/0052-dynamic-ids.md
@@ -1,0 +1,112 @@
+---
+feature: dynamic-ids
+start-date: 2019-09-05
+author: Silvan Mosberger
+co-authors: (find a buddy later to help our with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: https://github.com/NixOS/nixpkgs/pull/65698
+---
+
+# Summary
+[summary]: #summary
+
+A lot of NixOS modules are [assigning static uids/gids](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/misc/ids.nix#L36) to their users. This has resulted in less than 90 static ids left in the reserved range from 0 to 400.
+
+This RFC deprecates the practice of doing that and instead suggests to
+- If applicable use systemd's `DynamicUser`
+- Otherwise let NixOS assign dynamic persistent ids, which happens automatically when `users.users.<name?>.uid`/`users.groups.<name?>.gid` is not set. For users, `users.users.<name?>.isSystemUser` should be set so that only uids under 1000 are used by NixOS services.
+
+Only in special circumstances are static ids allowed anymore.
+
+# Motivation
+[motivation]: #motivation
+
+We are running out of static ids in the reserved range from 0 to 400. If services continue to reserve ids for themselves this will run out eventually.
+
+In addition, a central list of ids is annoying to maintain and leads to merge conflicts.
+
+# Detailed design
+[design]: #detailed-design
+
+## Documentation updates
+
+New best practices for declaring users are to be documented. This includes sections like the following
+
+### Using `DynamicUser` for declaring users
+
+If the service is fit for `DynamicUser`, this is the preferred solution, as it doesn't require any persistent ids. However this only works well for services whose state is self-contained. As soon as the services data is needed somewhere else, this approach can fall flat. `DynamicUser` also enables a bunch of options restricting the service (see `man systemd.exec` for details), some of which might need to be disabled in order for it to work.
+
+An example:
+
+```nix
+{
+  systemd.services.myservice.serviceConfig.DynamicUser = true;
+}
+```
+
+### Using ids dynamically allocated by NixOS
+
+
+By not setting `users.users.<name?>.uid`/`users.groups.<name?>.gid`, the ids will be dynamically allocated by NixOS. For users, `users.users.<name?>.isSystemUser = true` should be set as well such that only ids below 1000 are to be used. These ids are persistent over the lifetime of a NixOS system, even when services are disabled and enabled again. The generated mapping from names to ids is stored in `/var/lib/nixos/uid-map`/`/var/lib/nixos/gid-map`, so if this directory is backed up, the mappings will persist too when restoring.
+
+An example:
+
+```nix
+{
+  users = {
+    users.myservice = {
+      description = "My service user";
+      group = "myservice";
+      isSystemUser = true;
+    };
+    groups.myservice = {};
+  };
+}
+```
+
+### Ensuring correct directory permissions for services
+
+Services should make permissions of their directories are usable by them.
+- The easiest way to achieve this is to use `systemd.services.<name?>.serviceConfig.StateDirectory = "myservice"`, which ensures that `/var/lib/myservice` belongs to the services user.
+- `systemd.tmpfiles.rules = [ "Z '/var/lib/myservice' - myuser mygroup - -" ]` can also be used, with the disadvantage that it will only run at system activation and not when the service starts. It also recursively fixes the permissions every time, meaning it can lead to considerable slowdown with many files.
+- An alternative is to assign `serviceConfig.ExecStartPre = "+${pkgs.writeScript "myservice-prestart" "..."}"` with a script to fix the permissions, where the `+` makes the script run with full root permissions as documented in `man systemd.service`.
+
+## Enlarge the reserved range of system users/groups
+
+Since the new recommendation is to use `isSystemUser` for users and to not set static ids, the available ids now have the range `SYS_UID_MIN` to `SYS_UID_MAX` and `SYS_GID_MIN` to `SYS_GID_MAX`, defined in [shadow.nix](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/programs/shadow.nix#L13-L21), both of which currently span from 400 to 499. This means a NixOS system could only have 100 different services using dynamically allocated ids over its lifetime. Since this is not very much, this range will be changed to span from 400 to 999, reserving an additional 500 ids for system users/groups. The range from 500 to 999 is currently not reserved for anything.
+
+Therefore the number of different services using dynamic ids that can be enabled on a single NixOS system is 600. In comparison with static ids, the number of *enabled* services on a system grows much slower than the number of total *existing* services in NixOS (each of which would need one of the 400 static ids).
+
+The implementation of this is in PR [#65698](https://github.com/NixOS/nixpkgs/pull/65698).
+
+A note on uids vs gids: If `isSystemUser = false`, dynamic uids are allocated in the range `UID_MIN` to `UID_MAX`, while with `isSystemUser = true`, it's `SYS_UID_MIN` to `SYS_UID_MAX`. In contrast, gids are always allocated in the range `SYS_GID_MIN` to `SYS_GID_MAX`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+If data is restored from a backup without restoring `/var/lib/nixos` as well and the service doesn't [ensure correct directory permissions](#ensuring-correct-directory-permissions-for-services), then the service can fail to start. When copying data from a different machine this can happen as well. This is fixable by either manually changing `/var/lib/nixos/{u,g}id-map` to map the name to the old id, by recursively `chown`ing the restored data to the new id, or by making the NixOS service fix permissions itself.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Increase the range of static ids up to 999, adding another 600 static ids. While this would solve the problem of running out of static ids temporarily, over time this range will fill up again, at which point this alternative can't be used again.
+- Use a big range above `UID_MAX`/`GID_MAX`=29999, perhaps from 30000 to 40000. This would be enough static ids for a long time, but having to maintain a static mapping in nixpkgs will still be annoying.
+
+Not doing anything is not an option as the currently used range is finite.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What is the migration path from different user declarations?
+
+| From \ To | `DynamicUser` | dynamic NixOS ids | static NixOS ids |
+| --- | --- | --- | --- |
+| `DynamicUser` | - | ??? | ??? |
+| dynamic NixOS ids | ??? | - | ??? |
+| static NixOS ids | ??? | ??? | - |
+
+# Future work
+[future]: #future-work
+
+All unnecessary static ids can over time be removed from the [central listing](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/misc/ids.nix#L36), replacing them with `DynamicUser` where applicable and/or dynamic ids allocated by NixOS.
--- a/rfcs/0052-dynamic-ids.md
+++ b/rfcs/0052-dynamic-ids.md
@@ -5,7 +5,7 @@ author: Silvan Mosberger
 co-authors:
 shepherd-team: Arian van Putten, asymmetric, Eelco Dolstra, Jörg Thalheim, Ryan Mulligan
 shepherd-leader: Ryan Mulligan
-related-issues: https://github.com/NixOS/nixpkgs/pull/65698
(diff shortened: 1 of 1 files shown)


Profpatsch:
Does `DynamicUser` magically work when migrating between machines (as compared to the mapping file)?

Infinisil:
@Profpatsch If you need a stateful directory with `DynamicUser`, you need to use `StateDirectory` for it, which ensures correct permissions, so yeah this should just work.

edolstra:
This PR is now open for shepherd nominations. Any volunteers?

ryantm:
I volunteer to be a shepherd.

Mic92:
I volunteer to be a shepherd.

arianvp:
I volunteer too!

asymmetric:
I too volunteer to be a shepherd 🐑.

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+# Summary
+[summary]: #summary
+
+A lot of NixOS modules are [assigning static uids/gids](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/misc/ids.nix#L36) to their users. This has resulted in less than 90 static ids left in the reserved range from 0 to 400.
+
+This RFC deprecates the practice of doing that and instead suggests to
+- If applicable use systemd's `DynamicUser`
+- Otherwise let NixOS assign dynamic persistent ids, which happens automatically when `users.users.<name?>.uid`/`users.groups.<name?>.gid` is not set. For users, `users.users.<name?>.isSystemUser` should be set so that only uids under 1000 are used by NixOS services.
+
+Only in special circumstances are static ids allowed anymore.
```
7c6f434c: I think it would be a good idea to list such special circumstances. I would guess that something that doesn't fit `StateDir` model and locks one out of the system on failure (at least from remote access) would be an example?

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+# Detailed design
+[design]: #detailed-design
+
+## Documentation updates
+
+New best practices for declaring users are to be documented. This includes sections like the following
+
+### Using `DynamicUser` for declaring users
+
+If the service is fit for `DynamicUser`, this is the preferred solution, as it doesn't require any persistent ids. However this only works well for services whose state is self-contained. As soon as the services data is needed somewhere else, this approach can fall flat. `DynamicUser` also enables a bunch of options restricting the service (see `man systemd.exec` for details), some of which might need to be disabled in order for it to work.
```
Mic92: I still have some hopes that at some point we will have a shiftfs in linux mainline: https://github.com/linuxkit/linuxkit/tree/master/projects/shiftfs
Then we could have `DynamicUser` for all services because it removes the overhead of adapting filesystem uids at service startup.

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+```nix
+{
+  systemd.services.myservice.serviceConfig.DynamicUser = true;
+}
+```
+
+### Using ids dynamically allocated by NixOS
+
+
+By not setting `users.users.<name?>.uid`/`users.groups.<name?>.gid`, the ids will be dynamically allocated by NixOS. For users, `users.users.<name?>.isSystemUser = true` should be set as well such that only ids below 1000 are to be used. These ids are persistent over the lifetime of a NixOS system, even when services are disabled and enabled again. The generated mapping from names to ids is stored in `/var/lib/nixos/uid-map`/`/var/lib/nixos/gid-map`, so if this directory is backed up, the mappings will persist too when restoring.
```
Mic92: It would be great if we can defer this if possible and only use dynamically allocated uids if absolute necessary. Users may also allocate on their own static uids/gids for system users to avoid the issue when migrating data between machines.

Mic92: Just to throw in an idea:
If there would be a mechanism that would allow users to put ad-hoc allocated uid/gids into the nixos configuration, than this information could be easily backed and put into version control.

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+related-issues: https://github.com/NixOS/nixpkgs/pull/65698
+---
+
+# Summary
+[summary]: #summary
+
+A lot of NixOS modules are [assigning static uids/gids](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/misc/ids.nix#L36) to their users. This has resulted in less than 90 static ids left in the reserved range from 0 to 400.
+
+This RFC deprecates the practice of doing that and instead suggests to
+- If applicable use systemd's `DynamicUser`
```
Mic92: To help the reader to decide when not to use `DynamicUser`:
Services not applicable for `DynamicUser` are services that own many files,
so that the service startup would be significantly delayed by adapting filesystem permission.

manveru: From http://0pointer.net/blog/dynamic-users-with-systemd.html
> the allocation of dynamic UIDs has been tweaked in two ways to avoid the necessity to do this expensive operation in most cases: firstly, when a dynamic UID is allocated for a service an allocation loop is employed that starts out with a UID hashed from the service's name. This means a service by the same name is likely to always use the same numeric UID. That means that a stable service name translates into a stable dynamic UID, and that means recursive file ownership adjustments can be skipped (of course, after validation). Secondly, if the configured state directory already exists, and is owned by a suitable currently unused dynamic UID, it's preferably used above everything else, thus maximizing the chance we can avoid the chown()ing.

So I'm not sure how much of an issue that is in practice without benchmark, but it sounds like it would work even for this case.

Mic92: I think this should be good enough for us.

rfcs/0052-dynamic-ids.md
Infinisil: That's always possible by assigning `users.users.<name?>.uid` in their config. I guess I haven't mentioned this, but there's nothing wrong with the end user using static ids.

However I think it would be good if we could get away from static ids completely and not object anybody to decide on these arbitrary numbers. When things like `StateDirectory` are used throughout, non-static ids won't be a problem even when migrating data.

rfcs/0052-dynamic-ids.md
Infinisil: Not sure if there's any program that would actually do that. As an actual example, I think there is NFS3 which might not be able to map ids from other machines to local ones, but NFS4 doesn't have that problem anymore I'm pretty sure.

In general the problem might be programs that transfer files between machines without adjusting the ids accordingly. And arguably since ids don't have to be the same on different machines, that's really a bug in the program then.

Another example might be programs that mention that their user should have a certain fixed uid, because it might be hardcoded in the source, or because it might not have code to handle different ones on different machines. No concrete I know, but I wouldn't be surprised to find this somewhere.

rfcs/0052-dynamic-ids.md
ryantm: https://discuss.linuxcontainers.org/t/lxd-3-12-has-been-released/4483

> This is a feature we’ve been looking forward to for years and that we are really excited to finally see coming to completion. shiftfs allows for an unprivileged container experience that doesn’t need any shifting of the filesystem, instead having the kernel do it on the fly.
> 
> This requires kernel support through the shiftfs filesystem which is currently a custom patchset that will be present in the Ubuntu 19.04 kernel.
> 
> LXD automatically detects support for this and will transparently start using it whenever possible.

Sounds like maybe Ubuntu adopted a `shiftfs` kernel patch.

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+
+Therefore the number of different services using dynamic ids that can be enabled on a single NixOS system is 600. In comparison with static ids, the number of *enabled* services on a system grows much slower than the number of total *existing* services in NixOS (each of which would need one of the 400 static ids).
+
+The implementation of this is in PR [#65698](https://github.com/NixOS/nixpkgs/pull/65698).
+
+A note on uids vs gids: If `isSystemUser = false`, dynamic uids are allocated in the range `UID_MIN` to `UID_MAX`, while with `isSystemUser = true`, it's `SYS_UID_MIN` to `SYS_UID_MAX`. In contrast, gids are always allocated in the range `SYS_GID_MIN` to `SYS_GID_MAX`.
```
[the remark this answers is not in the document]
Infinisil: Usually not, no.

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+  };
+}
+```
+
+### Ensuring correct directory permissions for services
+
+Services should make permissions of their directories are usable by them.
+- The easiest way to achieve this is to use `systemd.services.<name?>.serviceConfig.StateDirectory = "myservice"`, which ensures that `/var/lib/myservice` belongs to the services user.
+- `systemd.tmpfiles.rules = [ "Z '/var/lib/myservice' - myuser mygroup - -" ]` can also be used, with the disadvantage that it will only run at system activation and not when the service starts. It also recursively fixes the permissions every time, meaning it can lead to considerable slowdown with many files.
+- An alternative is to assign `serviceConfig.ExecStartPre = "+${pkgs.writeScript "myservice-prestart" "..."}"` with a script to fix the permissions, where the `+` makes the script run with full root permissions as documented in `man systemd.service`.
```
arianvp: Would be nice if we could have this syntax work in the `preStart` and `postStart` shorthands that we currently have in the `systemd` module

rfcs/0052-dynamic-ids.md
asymmetric: Or one could reword it to: "usage of static ids has to be explicitly justified", so as to put the burden on the code's author rather than the RFC's (assuming we could even come up with an exhaustive list of reasons where static IDs are necessary)

rfcs/0052-dynamic-ids.md
asymmetric:
```suggestion
By not setting `users.users.<name?>.uid`/`users.groups.<name?>.gid`, the ids will be dynamically allocated by NixOS. For users, `users.users.<name?>.isSystemUser = true` should be set as well such that only ids below 1000 are used. These ids are persistent over the lifetime of a NixOS system, even when services are disabled and enabled again. The generated mapping from names to ids is stored in `/var/lib/nixos/uid-map`/`/var/lib/nixos/gid-map`, so if this directory is backed up, the mappings will persist too when restoring.
```

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+      isSystemUser = true;
+    };
+    groups.myservice = {};
+  };
+}
+```
+
+### Ensuring correct directory permissions for services
+
+Services should make permissions of their directories are usable by them.
```
asymmetric:
```suggestion
Services must be able to access their directories.
```

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+- The easiest way to achieve this is to use `systemd.services.<name?>.serviceConfig.StateDirectory = "myservice"`, which ensures that `/var/lib/myservice` belongs to the services user.
```
asymmetric:
```suggestion
- The easiest way to achieve this is to use `systemd.services.<name?>.serviceConfig.StateDirectory = "myservice"`, which ensures that `/var/lib/myservice` belongs to the services user. See `man systemd.exec` for other related directives.
```

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+- An alternative is to assign `serviceConfig.ExecStartPre = "+${pkgs.writeScript "myservice-prestart" "..."}"` with a script to fix the permissions, where the `+` makes the script run with full root permissions as documented in `man systemd.service`.
+
+## Enlarge the reserved range of system users/groups
+
+Since the new recommendation is to use `isSystemUser` for users and to not set static ids, the available ids now have the range `SYS_UID_MIN` to `SYS_UID_MAX` and `SYS_GID_MIN` to `SYS_GID_MAX`, defined in [shadow.nix](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/programs/shadow.nix#L13-L21), both of which currently span from 400 to 499. This means a NixOS system could only have 100 different services using dynamically allocated ids over its lifetime. Since this is not very much, this range will be changed to span from 400 to 999, reserving an additional 500 ids for system users/groups. The range from 500 to 999 is currently not reserved for anything.
```
asymmetric:
```suggestion
Since the new recommendation is to use `isSystemUser` for system users and to not set static ids, the available ids now have the range `SYS_UID_MIN` to `SYS_UID_MAX` and `SYS_GID_MIN` to `SYS_GID_MAX`, defined in [shadow.nix](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/programs/shadow.nix#L13-L21), both of which currently span from 400 to 499. This means a NixOS system could only have 100 different services using dynamically allocated ids over its lifetime. Since this is not very much, this range will be changed to span from 400 to 999, reserving an additional 500 ids for system users/groups. The range from 500 to 999 is currently not reserved for anything.
```

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+
+Therefore the number of different services using dynamic ids that can be enabled on a single NixOS system is 600. In comparison with static ids, the number of *enabled* services on a system grows much slower than the number of total *existing* services in NixOS (each of which would need one of the 400 static ids).
```
asymmetric:
```suggestion
Therefore the number of different services using dynamic ids that can be enabled on a single NixOS system is 600. In comparison with static ids, the number of *enabled* services on a system grows much more slowly than the number of total *existing* services in NixOS (each of which would need one of the 400 static ids).
```

rfcs/0052-dynamic-ids.md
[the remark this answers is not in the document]
asymmetric: I'd move this note to the top of the section, since it explains concepts that are useful to understand the section itself.

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+| From \ To | `DynamicUser` | dynamic NixOS ids | static NixOS ids |
+| --- | --- | --- | --- |
+| `DynamicUser` | - | ??? | ??? |
+| dynamic NixOS ids | ??? | - | ??? |
+| static NixOS ids | ??? | ??? | - |
+
+# Future work
+[future]: #future-work
+
+All unnecessary static ids can over time be removed from the [central listing](https://github.com/NixOS/nixpkgs/blob/044cc701c23ede96355eb1dc997985b4dfac0372/nixos/modules/misc/ids.nix#L36), replacing them with `DynamicUser` where applicable and/or dynamic ids allocated by NixOS.
```
asymmetric: At the end of this, `SYS_UID_MIN` could be lowered to 1.

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,112 @@
...
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What is the migration path from different user declarations?
+
+| From \ To | `DynamicUser` | dynamic NixOS ids | static NixOS ids |
+| --- | --- | --- | --- |
+| `DynamicUser` | - | ??? | ??? |
+| dynamic NixOS ids | ??? | - | ??? |
+| static NixOS ids | ??? | ??? | - |
```
asymmetric: I'm not sure this RFC is about any other case than "static NixOS ids" -> `DynamicUser`, and the migration path for that is the one you describe below in "Future work".

Or do I misunderstand what the unresolved question is?

Infinisil: I think we should think (and test) how all of these behave, making sure they actually work and don't break services or how to ensure they don't:

- static NixOS id -> dynamic NixOS id: For migrating from static ids
- static NixOS id -> `DynamicUser`: For migrating from static ids
- `DynamicUser` -> dynamic NixOS id: When `DynamicUser` was found to be too restrictive
- dynamic NixOS id -> `DynamicUser`: When `DynamicUser` was found out to be feasible

rfcs/0052-dynamic-ids.md
7c6f434c: Well, «explicitly justified» is not really a clear description of _level_ of burden of proof.

ryantm:
Since the discussion seems to have died down, and there doesn't seem to be anything major outstanding, I motion for the final comment period with disposition to merge.

@arianvp, @asymmetric, @edolstra please state your agreement or disagreement with entering the FCP in a comment below.

edolstra:
@ryantm :+1: on merging.

Infinisil:
I'll be doing some changes addressing the reviews today

Infinisil:
Alright done, changes are:
- Small sentence adjustments that were suggested
- Mention that this RFC only applies to NixOS services, not to end-users
- Mention that `DynamicUser` can be used even with large amounts of files
- Move the note for uid vs gid ranges to the top of the section
- Fill out the migration table and move it to the details section
- Note how we could lower `SYS_UID_MIN` to 6 in the future

Infinisil:
And finally: Add information that changing `isSystemUser` isn't problematic either, however it doesn't have anything effect if the user existed already.

(This is good news for the couple services that have a dynamically allocated uid without using `isSystemUser`)

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,121 @@
...
+    };
+    groups.myservice = {};
+  };
+}
+```
+
+### Ensuring correct directory permissions for services
+
+Services must be able to access their directories.
+- The easiest way to achieve this is to use `systemd.services.<name?>.serviceConfig.StateDirectory = "myservice"`, which ensures that `/var/lib/myservice` belongs to the services user. See `man systemd.exec` for other related directives.
```
arianvp: Do we want to add some advice about avoiding including new `dataDir` options in newly created modules unless there's a very good usecase for it? (E.g. postgres)

aanderse: Out of scope for this RFC. I strongly agree with that opinion, but there are many who strongly disagree so we shouldn't muddy the waters here with that.

arianvp: Followup question.  Should we mention  `ConfigurationDirectory, CacheDirectory, LogsDirectory` here too? They all have the same kind of behaviour when `DynamicUser=true`  See https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=

edolstra: It's just a suggestion right? It doesn't say you *must* use `StateDirectory`.

arianvp:
Apart from the small comment I jud made, I am :+1:

rfcs/0052-dynamic-ids.md
aanderse: I definitely think we should be using `StateDirectory`, I just figured that might become a point of contention for some people. If everyone agrees then I guess there isn't a problem...

Infinisil: I added a reference to the other directives in this sentence in https://github.com/NixOS/rfcs/pull/52/commits/ff7636a43fbb69da1b75f314b94e1d88634d7bc1. I feel like they deserve at least one mention.

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,121 @@
...
+
+| From \ To | `DynamicUser` | dynamic NixOS ids | static NixOS ids |
+| --- | --- | --- | --- |
+| `DynamicUser` | - | unproblematic because `StateDirectory` was used | unproblematic because `StateDirectory` was used |
+| dynamic NixOS ids | unproblematic because `StateDirectory` needs to be used | - | needs manual `/var/lib/nixos/{g,u}id-map` change if different id |
+| static NixOS ids | unproblematic because `StateDirectory` needs to be used | unproblematic | - |
+
+In addition, a transition from `isSystemUser = false` to `isSystemUser = true` can't be done automatically, the `/var/lib/nixos/uid-map` file needs to be adjusted manually for that. However without doing so, NixOS will happily continue to use the previously assigned uid without problems. This means changing this value is unproblematic.
+
+Note that changing all current NixOS services to use dynamic ids is [future work][#future].
```
ryantm:
```suggestion
Note that changing all current NixOS services to use dynamic ids is [future work][#future-work].
```

rfcs/0052-dynamic-ids.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### Ensuring correct directory permissions for services
+
+Services must be able to access their directories.
+- The easiest way to achieve this is to use `systemd.services.<name?>.serviceConfig.StateDirectory = "myservice"`, which ensures that `/var/lib/myservice` belongs to the services user. See `man systemd.exec` for info on this and the related directives `CacheDirectory`, `LogsDirectory` and `ConfigurationDirectory`.
+- `systemd.tmpfiles.rules = [ "Z '/var/lib/myservice' - myuser mygroup - -" ]` can also be used, with the disadvantage that it will only run at system activation and not when the service starts. It also recursively fixes the permissions every time, meaning it can lead to considerable slowdown with many files.
+- An alternative is to assign `serviceConfig.ExecStartPre = "+${pkgs.writeScript "myservice-prestart" "..."}"` with a script to fix the permissions, where the `+` makes the script run with full root permissions as documented in `man systemd.service`.
+
+## Enlarge the reserved range of system users/groups
```
ryantm: In light of @asymmetric 's [comment]( https://github.com/NixOS/nixpkgs/pull/65698#issuecomment-532835429) about how systemd ignores `login.defs` this section might need to be slightly tweaked, or maybe the PR needs to be fixed to have systemd and login.defs depend on the same nix definition.

Infinisil: See https://github.com/NixOS/nixpkgs/pull/65698#issuecomment-537918789, I'll mention here that people should see that PR for implementation details

Infinisil: So here `SYS_UID_MIN` and such would be representative of whatever needs to be changed to get the change working (and really, `SYS_UID_MIN` should be the way to configure this, I think it's a bug that it doesn't work).

ryantm:
@asymmetric 's GitHub profile says he is on vacation and he hasn't replied in 6 days, and the rest of us are unanimous that we would like to enter FCP with disposition to merge, so I'm going to proceed to publicize the FCP after checking with @Infinisil.

ryantm:
The Final Comment Period for this RFC has started and, barring any blocking issues, will be merged after 2019-10-18. Your opinions, comments, and approvals are welcome!

rfcs/0052-dynamic-ids.md
asymmetric: Are we going to fix this?

rfcs/0052-dynamic-ids.md
asymmetric: Can we mark this as resolved?

Infinisil:
Ah yes, forgot to push the changes for those

Infinisil:
Lucky coincidence: @dasj just opened https://github.com/NixOS/nixpkgs/pull/71055 for making all NixOS module users use `isSystemUser = true` which I planned to do after this RFC. With https://github.com/NixOS/nixpkgs/pull/65698 the system user limit of 100 ids will be increased to non-problematic levels.

ryantm:
The FCP has ended without any objections. @NixOS/rfc-steering-committee please merge.

Infinisil:
Note to myself: This still needs the updates to the manual

PR #54, opened 2019-10-03
[RFC 0054] Support levels and tiers

CrystalGamma:
This RFC describes a scheme for expressing and managing support of packages and for platforms in nixpkgs by defining support levels for packages and support/acceptance tiers for platforms.

[Rendered version](https://github.com/CrystalGamma/rfcs/blob/rfc54/rfcs/0054-support-levels-and-tiers.md)

### DIFF
--- a/rfcs/0054-support-levels-and-tiers.md
+++ b/rfcs/0054-support-levels-and-tiers.md
@@ -1,0 +1,171 @@
+---
+feature: support_levels_and_tiers
+start-date: 2019-10-03
+author: CrystalGamma
+co-authors: –
+related-issues: –
+---
+
+# Summary
+[summary]: #summary
+
+A scheme for expressing and managing support of packages and for platforms in nixpkgs.
+
+# Motivation
+[motivation]: #motivation
+
+This RFC is an offshoot of a comment in the discussion on #46, which has been stuck in discussion for months.
+Its goals are similar, but this RFC focuses on what support expectations are reasonable for each level/tier.
+
+Providing a good base for platform discussions and defining support expectations should improve communication with users (especially those of non-mainstream platforms) and should allow establishing procedures to deal with developers wishing to add support for new architectures.
+
+# Detailed design
+[design]: #detailed-design
+
+This RFC defines support levels for packages and the corresponding support and acceptance tiers for platforms.
+
+## Support Level
+The *support level* of a package is a designator of its importance for nixpkgs.
+
+This RFC proposes an initial set of 4 support levels (in order of descending importance / expanding scope): `core`, `bootstrap`, `system` and `all`.
+(Terminology: A support level to the left of another in this list is said to be 'narrower' or 'more significant' than it. Conversely a support level to the right of another is considered 'wider'/'broader' or 'less significant'.) See [below](#relationship-between-support-levels-support-tiers-and-acceptance-tiers) for the scope of each support level.
+
+A `meta` attribute `supportlevel` which is supposed to take a value from the aforementioned enumeration is introduced for this purpose.
+A package with its `meta.supportlevel` unset is assumed to have support level `all` (in fact, setting `supportlevel` to `all` should only be done in special cases, such as for packages produced by overriding others).
+
+Scrutiny required for packages increases with its support level, as such the support level for each package should be kept as low as possible.
+
+## Platforms and support tiers
+A platform is a combination of these differentiations:
+
+* CPU architecture
+* kernel
+* libc
+* compiler
+* other system differentiations (e. g. NixOS/non-NixOS on Linux)
+
+Obviously, not all such combinations make sense and not all of these distinctions are relevant for all discussions.
+Where some of these dimensions don't make a difference, platforms may be handled as a group (particularly the term 'platform-specific' may mean 'specific to a given group of similar platforms').
+
+Each platform has a *support tier* that describes the expectations that one should have when attempting to use nixpkgs on that platform.
+
+This RFC proposes an initial set of 5 support tiers (in decreasing order): `full`, `system`, `bootstrap`, `core` and `none`.
+
+The `core`, `bootstrap` and `system` tiers are roughly characterized by the set of all packages with the corresponding support level (and more significant support level) being usable. More on this [below](#relationship-between-support-levels-support-tiers-and-acceptance-tiers). Note that higher support tiers correspond to the inclusion of less-significant support levels.
+
+Platform support tiers are advertised to users and should help them decide what platforms to use or whether the nixpkgs ecosystem is suitable for their needs, given a platform.
+
+## Acceptance tiers
+Each platform is also assigned an *acceptance tier*. It describes the amount of maintenance, infrastructure and management effort specific to that platform (or group of platforms) that is considered acceptable.
+
+Each acceptance tier corresponds to a support tier (and carries the same name): it is characterized by a reasonable estimation of effort required to reach that support tier.
+
+A platform generally has an acceptance tier that is the same (for mature platforms) or higher (for platforms that target a higher support tier) than its support tier.
+
+Acceptance tiers are used to discuss the acceptance of changes that introduce platform-specific features and workarounds: platform-specific changes that modify a package with a support level not included in the ones required for its acceptance tier should generally be rejected with reference to that acceptance tier, as should introducing platform-specific packages with less-significant support level.
+Conversely, most platform-specific changes that improve packages required for the support tier corresponding to the acceptance tier should be accepted, unless an obviously better way to achieve the same improvement is known.
+
+A complicated issue is changes that isolate platform-specific behavior, e. g. by adding conditionals in Nix expressions for flags that used to be applied unconditionally but have in fact platform-specific applicability (often this means x86-specific) or adding patches that correct unportable code in the package itself.
+As long as the added burden on maintenance of nixpkgs is not unreasonable, these should be accepted even if they only benefit platforms whose acceptance tier does not cover the package in question.
+
+## Relationship between support levels, support tiers and acceptance tiers
+
+| level/tier | code included in support level | meaning of support tier | meaning of acceptance tier |
+| --- | --- | --- | --- |
+none | N/A | There has been no testing done with packages on/for that platform, there may not even be `lib/systems` infrastructure for the platform | no code is accepted for this platform
+core | `lib/systems` infrastructure, compiler/libc packages, cross-stdenv build infrastructure | A cross-stdenv can be built for the system | code is accepted to allow building a cross-stdenv
+bootstrap | bootstrap files, packages required to build bootstrap files (stdenv bootstrap, busybox, …) | Bootstrap files (if required) can be cross- and native-built (most importantly this includes bootstrapping a stdenv on the platform itself). | changes to packages involved in bootstrapping (support level `bootstrap`) are accepted (if technically sound)
+system | OS management tools, service management systems, daemons, bootloaders, mesa, core NixOS modules and tooling, … (and their dependencies if not included in more-significant support level) | NixOS-supporting systems (i. e. Linux): a bootable mostly-default configuration of NixOS can be built. Other OSes: reasonable set of OS-specific utilities is available and works (if applicable), all inapplicable packages with support level `system` are marked as such with `meta.platforms` or `meta.badPlatforms`, all other `system` packages are expected to work. | (technically sound) changes are accepted for packages with support level `system`
+all/full | all other packages and modules | most software is expected to work, those that do not are marked as such with `meta.platforms`, `meta.badPlatforms` or `meta.broken` | Being specific to this platform is not an argument for rejection (as long as the change is technically sound)
+
+## Management of support levels
+Packages should be kept at the lowest support level possible.
+If the scope of a support level expands to include further packages (e. g. dependencies are added to a package already in scope), the support level must increase.
+
+To facilitate the detection of such cases, a nixpkgs configuration option `maxSupportLevel` (taking the same values as `meta.supportlevel` and defaulting to `all`) is added that shall be checked in `pkgs/stdenv/generic/check-meta.nix`, disabling packages with wider support level.
+Having Hydra evaluations representing the scope of each support level (i. e. stdenv for `core` and `bootstrap`, NixOS for `system`) should allow detecting packages with insufficient support level.
+Note that the evaluations should be mostly independent of platforms, since this is only about automatically detecting that packages required for a given scope are actually marked with at least the corresponding support level; the derivations needn't even be built for this purpose.
+
+Should restructuring (e. g. of stdenv bootstrap or changing the 'reference' NixOS configuration for the `system` support level/tier) exclude a package from the scope of a certain support level, the `meta.supportlevel` should be widened appropriately. There is no obvious way to check whether this is done correctly by automatic means.
+
+## Management of support/acceptance tiers
+(Note: this part in particular is a first draft. Suggestions welcome)
+
+As long as the relevant compiler supports the target, adding a platform to satisfy support tier `core` should be relatively simple. (Mainly involves adding support in `lib/platforms` and adding flags to compiler and libc if necessary)
+For that reason, acceptance of such platforms should be at `core` by default.
+
+Platform maintainer groups (PMGs) should exist for each supported platform (group). A single platform tuple may be covered by the scope of multiple PMGs (if one e. g. maintains x86 platforms and one maintains musl platforms).
+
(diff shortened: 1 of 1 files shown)


shlevy:
This RFC is now open for nominations. Any volunteers?

rfcs/0054-support-levels-and-tiers.md
```diff
@@ -0,0 +1,171 @@
...
+# Summary
+[summary]: #summary
+
+A scheme for expressing and managing support of packages and for platforms in nixpkgs.
+
+# Motivation
+[motivation]: #motivation
+
+This RFC is an offshoot of a comment in the discussion on #46, which has been stuck in discussion for months.
+Its goals are similar, but this RFC focuses on what support expectations are reasonable for each level/tier.
```
7c6f434c: I think there should be a self-contained summary of motivation that makes independent sense a year later, when this RFC is finally accepted and #46 is rejected as horribly overengineered.

rfcs/0054-support-levels-and-tiers.md
```diff
@@ -0,0 +1,171 @@
...
+
+To facilitate the detection of such cases, a nixpkgs configuration option `maxSupportLevel` (taking the same values as `meta.supportlevel` and defaulting to `all`) is added that shall be checked in `pkgs/stdenv/generic/check-meta.nix`, disabling packages with wider support level.
+Having Hydra evaluations representing the scope of each support level (i. e. stdenv for `core` and `bootstrap`, NixOS for `system`) should allow detecting packages with insufficient support level.
+Note that the evaluations should be mostly independent of platforms, since this is only about automatically detecting that packages required for a given scope are actually marked with at least the corresponding support level; the derivations needn't even be built for this purpose.
+
+Should restructuring (e. g. of stdenv bootstrap or changing the 'reference' NixOS configuration for the `system` support level/tier) exclude a package from the scope of a certain support level, the `meta.supportlevel` should be widened appropriately. There is no obvious way to check whether this is done correctly by automatic means.
```
7c6f434c: Maybe something in the direction «every package of support level X that has a source, must correspond to a depepndency of the support-level-defining expression with the same source»?

rfcs/0054-support-levels-and-tiers.md
```diff
@@ -0,0 +1,171 @@
...
+
+* `core`: 1 maintainer
+* `bootstrap`: 2 maintainers
+* `system`: 5 maintainers
+* `full`: 10 maintainers
+
+PMGs should be notified and should publically give an (informal) report on the state of the platform as part of the release process for a major nixpkgs/NixOS version.
+This can also serve as an occasion to gauge responsiveness of their individual members.
+
+If the maintainers report maturity at the current support tier, the acceptance tier may be raised one above the support tier.
```
7c6f434c: Does this require PMG to be large enough for the next support level?

rfcs/0054-support-levels-and-tiers.md
```diff
@@ -0,0 +1,171 @@
...
+If then the PMG reports adequate quality of support for the next-highest support tier, the support tier may be raised appropriately.
+
+Support tiers are dropped if the number of responsive maintainers in the PMG goes below the required number for the support level or if major breakage is reported for multiple subsequent major releases. (Note that the acceptance tier may be kept if sensible)
```
7c6f434c: Is the suport-acceptance gap allowed to become _two_ in this case?

rfcs/0054-support-levels-and-tiers.md
CrystalGamma: Makes sense. Something like "Currently there is no consensus in the nixpkgs maintainer community regarding how to deal with platform-specific code and changes, as well as few official statements and processes to represent platform support to users. As a growing distribution, the Nix community needs to establish a framework to determine platform support and decide what platform-specific packaging code is acceptable."?

rfcs/0054-support-levels-and-tiers.md
CrystalGamma: I haven't thought about that so far, but I'd say it makes more sense to only require the requisite number when raising the support tier (not level, support levels are for packages – maybe we should find a different name for one of them …).

Ericson2314:
@7c6f434c do you wish to withdraw #46? That RFC already has a shepeard team from a previous MR. @shlevy I would suggest we just carry over the same band of people again.

To be clear, the slow process is almost entirely my own fault. I think it's very good that as much RFCs have at least one meeting before NixCon, so I will try to not let the ball down with whatever I end up shepharding.

rfcs/0054-support-levels-and-tiers.md
CrystalGamma: Good question. I'd argue it shouldn't, since the purpose of having a small gap is making maintainers focus on the basics first by denying PRs for minor packages. In practice I can't imagine a platform mature enough to warrant raising the acceptance level suddenly being abandoned. I suppose it might make sense to prescribe revoking the elevated acceptance level before dropping the support level.

rfcs/0054-support-levels-and-tiers.md
CrystalGamma: Sorry, I'm having trouble understanding the suggestion … what do you mean by "source"? the `src` attribute? How would you check this?

rfcs/0054-support-levels-and-tiers.md
7c6f434c: Yes, that sounds reasonable

rfcs/0054-support-levels-and-tiers.md
7c6f434c: @CrystalGamma I can _remember_ such a thing in Nixpkgs, actually. Depending on the details of the defnitions, `i686-linux` might have eventually dropped from the flagship `all`-tier platform to a platform where part of `system`-level packages might be broken without too many people noticing.

CrystalGamma: `i686-linux` in particular might be gnarly because its main use nowadays is as a dependency for wine on `x86_64-linux`, so it doesn't need everything from `system` but does need a lot of things like sound libraries that are out of scope for `system` … we could get around that by defining support levels/tiers not as total orders, but lattices, with an additional `wine` support tier/level as a sibling of `system`, but that sounds like it could easily lead to exactly the kind of complexity I was trying to avoid :(

7c6f434c:
Due to conflict of interest, my first round of comments here assumes that «streamlined is obviously better»; outside this context I do not hold this conviction strongly enough to withdraw my alternative proposal immediately.

It does make sense to have an interesecting (or same) group of shepherds so that there can be a detailed discussion _which_ proposal to push to completion.

rfcs/0054-support-levels-and-tiers.md
7c6f434c: You have an option to say that there is an `x86_64-linux` Wine package (with multiarch support) that still depends on the same set of things, so these are supported in the interest of `x86_64-linux` by Wine package maintainers.

I do hope that the reference system configuration for the `system` package includes _some_ kind of sound support, though!

But I think that we cannot exclude that some other platform goes the same way.

rfcs/0054-support-levels-and-tiers.md
7c6f434c: We can enumerate all the derivations (say, using `release.nix`). Most derivations have `src` or `srcs`. In a single Nixpkgs checkout, «same» source usually means exact match of the source derivation output.

rfcs/0054-support-levels-and-tiers.md
CrystalGamma: Sure, `system` should probably include things like libpulse and whatever the low-level libraries for ALSA are called … I was thinking of libraries like OpenAL or cubeb (then again I don't actually know what wine uses, I was just recalling some dependencies of random video games I've seen run on Ubuntu and assumed wine might use those as well).

7c6f434c: `vkd3d` might be a good example of a thing useful for Wine and really not making sense in `system`

CrystalGamma: I suppose it makes sense to say that `i686-linux` inherits `x86_64-linux`'s acceptance tier where it makes sense for multilib.
It's probably sensible to have a provision for multilib somewhere in the RFC, in case that, as you say, it becomes relevant for another platform.

CrystalGamma:
I've been thinking … would it make sense to rename what I named "support level" (for packages) to "scope"?

There's already been confusion in the discussion between that and the "support tier" of platforms.

Renaming it "scope" would avoid that and no longer carries the assumption that the scopes are totally ordered, which would simplify the categorization of platforms mainly used for multilib (which was something I didn't think about when writing the RFC.

7c6f434c:
>I've been thinking … would it make sense to rename what I named "support level" (for packages) to "scope"?

Regardless of the structure, it does make make the two notions less similar-sounding

rfcs/0054-support-levels-and-tiers.md
CrystalGamma: So you'd check if the `src` is somewhere in the (build) closure of the defining expression? But doesn't closure-info only include the closure of the output?

7c6f434c:
>So you'd check if the `src` is somewhere in the (build) closure of the defining expression? But doesn't closure-info only include the closure of the output?

If we check the closure of the derivation file after instantiating the defining expression, it will include all the derivation files of all the build steps.

You are right, that if we check the closure of the output after building an expression, indeed only the runtime closure is included. But we do not want the check to cost a full rebuild anyway.

CrystalGamma:
That would need some code outside of Nix expressions to check, right? I was thinking of the `closureInfo` function, which AFAICT will only output the closure of the output path. But if we made a script that collected the closure of the *derivation* described by the defining expression and extracted the output paths of all the fixed-output derivations contained therein, we could add another nixpkgs option called, say, `restrictSrc` which gets the set of those paths and for each package/`stdenv.mkDerivation` checks if `src`/`srcs` contains that path.

I don't know the details of how Hydra is run, but this does sound reasonably easy to implement.

alyssais:
I think a worked example would be useful in explaining this.  Suppose I
wanted to reintroduce illumos support.  What would and would not be
expected at each point in the journey?  I'd having trouble figuring that
out from just reading the RFC.

globin:
The @NixOS/rfc-steering-committee has decided to ask for new nominations here, since #46 will remain open. Please do come forward with nominations and also feel free to nominate yourself!

Quoting RFC #36 for the responsibilities of the shepherd team:
#### Shepherd Team

A team of 3-4 community members defined unanimously by the RFC Steering Committee, responsible for accepting or rejecting a specific RFC. This team is created per RFC from community members nominated in the discussion on that RFC.

This team should be people who are very familiar with the main components touched by the RFC. The author cannot be part of the Shepherd Team. In addition, at most half of the Shepherd Team can be part of the RFC Steering Committee.

The responsibility of the team is to guide the discussion as long as it is constructive, new points are brought up and the RFC is iterated on and from time to time summarise the current state of discussion. If this is the case no longer, then the Shepherd Team shall step in with a motion for FCP.
Shepherd Leader

The person in charge of the RFC process for a specific RFC, and responsible for ensuring the process is followed in a timely fashion. The Shepherd Leader has no special responsibility with regard to moving an undecided Shepherd Team to a certain decision.

shlevy:
@ryantm @zimbatm @grahamc @Ericson2314 Do you want to carry over to this one? Also @ehmry did I understand correctly you put yourself forward for nomination?

Ericson2314:
@shlevy I could, but first me prove myself a non-shit shepherd leader by finally scheduling a meeting,

ryantm:
@shlevy yes, I'm okay with doing this one too.

globin:
The @NixOS/rfc-steering-committee has decided on @ehmry, @Ericson2314 and @ryantm as shepherds, with @ryantm as leader.

rfcs/0054-support-levels-and-tiers.md
```diff
@@ -0,0 +1,172 @@
...
+A `meta` attribute `scope` which is supposed to take a value from the aforementioned enumeration is introduced for this purpose.
+A package with its `meta.scope` unset is assumed to have scope `all` (in fact, setting `scope` to `all` should only be done in special cases, such as for packages produced by overriding others).
+
+Scrutiny required for packages increases with narrower scope, as such the scope for each package should be kept as broad as possible.
+
+## Platforms and support tiers
+A platform is a combination of these differentiations:
+
+* CPU architecture
+* kernel
```
ehmry: This is a minor nitpick, but I think the phrase "OS ABI" is more appropriate than kernel. For example, FreeBSD and Fuchsia can emulate the Linux ABI, CloudABI binaries are basically kernel-agnostic.

CrystalGamma:
I apologize for being a terrible author by disappearing without a word for 2½ months :(

Anyway, @alyssais : assuming you are aiming for `system` level support, you'd do the following:

* Add illumos to `lib/systems`, add whatever toolchain support a cross-stdenv targeting illumos needs and make a PR. Unless you are doing major things wrong (i. e. your changes aren't 'technically sound' as I called it in the RFC), this should be accepted as part of the 'accept-core-by-default' policy proposed by the RFC.
* Nominate the illumos-based platforms for `core` support, naming yourself (and possibly others) as maintainers.
* You will now be expected to check in somewhere to report on the support state of your platforms when the next release comes around.
* Assuming you are responsive enough and you can report that `core` functionality isn't broken or is at least fixed quickly, in the next release illumos will be considered `core`-supported and `bootstrap`-scoped changes will be accepted.
* You can now start upstreaming changes needed to create bootstrap tools for illumos within Nix and bootstrap a stdenv with them.
* Automated cross-builds of the bootstrap files via Hydra should be installed at some point in this phase, and, if possible, some installation (e. g. a VM) that can check that bootstrapping works on the platform.
* When you feel confident enough, you can apply for `bootstrap` support. You'll need to have found at least one other maintainer for the platform before this, as the PMG for `bootstrap`-supported platforms are required have at least two active members.
* Again, you (and the other maintainer(s)) are expected to report in on the state of the platform for the next release.
* Assuming no major breakage, illumos can be classified as a `bootstrap`-supported platform for the next release, and acceptance tier may be raised to `system`.
* Upstream `system`-scoped changes (depending on your goal this may be just a collection of relevant programs and libraries, or even a full port of NixOS to the illumos kernel – though the latter might be difficult considering the current reliance of NixOS on systemd)
* When you feel ready, you can apply for `system` support. By this point you should have:
  * at least 5 active PMG members
  * Hydra building packages on the supported illumos platforms
* As above, when the PMG reports that `system` support is stable for some release, that support tier can be awarded for that release.

It seems that during my prolonged absence, #46 has gained traction again. I'm personally not invested in which one of these wins (actually, the arguments for #46's less bureaucratic process of tier advancement is quite convincing), so if my work has inspired improvements in that RFC that helped it become ready, then my primary purpose for making this submission (providing clarity about how non-mainstream platforms are handled) is fulfilled.

ryantm:
@CrystalGamma I'm also sorry that I haven't done my duty as shepherd leader in pushing this along. It wasn't until I voted to move #46 to FCP that I realized I was the leader on this one. I think your summary of the state of #46 is accurate, and I want to encourage you to give your comments and improvement suggestions to #46, since we all would appreciate your feedback.

7c6f434c:
#54 has definitely helped to regain focus in the discussions of #46, in particular by highlighting the range of natural options in some places, and thank you for that.

I think that even if the process won't use it for limiting the packages one can touch, the definition of package scopes and reporting in terms of «90% of LiveCD packages build» is useful to have (and feels like the core part of #54, and seems pretty complete). So even if you do not want to build an argument for your versions of the conflicting normative parts, the nonconflicting ones are valuable and hopefully uncontroversial.

edolstra:
Now that #46 has been merged, should we keep this RFC open? Are there parts that are not addressed by #46?

7c6f434c:
>Now that #46 has been merged, should we keep this RFC open? Are there parts that are not addressed by #46?

The most detailed part here — classifying the packages — could be useful even without the normative part (also maybe as a part of guidelines about the more vague «most of the most popular» in #46)

Not sure if there is anyone willing to work on that restructuring in the short term, though.

ryantm:
@edolstra Yeah, I think we can close this without any formality. If @CrystalGamma requests reopening, of course we can do that no problem.

7c6f434c:
>@edolstra Yeah, I think we can close this without any formality. If @CrystalGamma requests reopening, of course we can do that no problem.

Indeed, we even have this quote from the author:

> if my work has inspired improvements in RFC #46 that helped it become ready, then my primary purpose for making this submission (providing clarity about how non-mainstream platforms are handled) is fulfilled.

PR #55, opened 2019-10-11
[RFC-0055] Retire inactive nixpkgs committers

tilpner:
Many people were given push access to the nixpkgs repository, which is kept even if these committers become inactive. This RFC proposes moving these contributors to a new team without push access.

[Rendered](https://github.com/tilpner/nixos-rfcs/blob/inactive/rfcs/0055-retired-committers.md)

### DIFF
--- a/rfcs/0055-retired-committers.md
+++ b/rfcs/0055-retired-committers.md
@@ -52,10 +52,26 @@ to be more predictable for committers and reduce the evaluations from 12 times a
 For each committer from the [Nixpkgs Committers team](https://github.com/orgs/NixOS/teams/nixpkgs-committers), the number of commits
 in that time range is checked, and the committer is considered inactive if there are none.
 
-This process occurs once after the RFC is accepted, and is then repeated at the beginning of each new year.
-
 Previous committers are moved to a new Nixpkgs Committers Emeritus team, to honor their past contributions.
 Members of this team will remain in the GitHub organisation, and may regain push access at a later time.
+
+The process consists of the following steps:
+
+- Running the [reference implementation][implementation] or a functionally equivalent application.
+- Checking the correctness of the script output by manually inspecting the recent git history of inactive committers.
+  This may be done by reviewing the activity page linked in the script output of the reference implementation.
+- Notification of the confirmed inactive committers, via a GitHub issue and an email to the committers address,
+  according to `<nixpkgs/maintainers/maintainer-list.nix>`. The text should include a notice on how to regain
+  commit permissions. The exact process of that is not part of this RFC, as the process of becoming a committer
+  in the first place is still not formalised.
+- Moving of the confirmed inactive committers from the [Nixpkgs Committers team](https://github.com/orgs/NixOS/teams/nixpkgs-committers) team
+  to the new Nixpkgs Committers Emeritus team, either directly if sufficient permissions are available,
+  or indirectly by notifying an organisation administrator.
+
+This process occurs once after the RFC is accepted, and is then repeated at the beginning of each new year.
+
+@Mic92 has committed to performing the yearly process, but any member of the NixOS organisation can execute the above
+steps if @Mic92 does not remember to do so in a reasonable amount of time after the beginning of the year.
 
 
 # Drawbacks
@@ -91,6 +107,7 @@ without being directly part of the work? -->
 - The threshold may need adjustment in the future.
 
 # Reference implementation
+[implementation] # reference-implementation
 
 ```py
 #! /usr/bin/env nix-shell
--- a/rfcs/0055-retired-committers.md
+++ b/rfcs/0055-retired-committers.md
@@ -27,6 +27,8 @@ at risk of of unauthorized modification.
 
 Because every secret with push access can be leaked, we should keep their number as low as necessary,
 here by deactivating the push access of inactive committers.
+A special case of inactive committers are those who have lost access to their GitHub account entirely,
+who would be unable to remove potentially leaked secrets from their account.
 
 As of 2019-08-18, at least 2 committers have officially stepped down, and at least 1 committer has
 not pushed to nixpkgs since 2014, but are still able to push directly to nixpkgs.
--- a/rfcs/0055-retired-committers.md
+++ b/rfcs/0055-retired-committers.md
@@ -1,0 +1,134 @@
+---
+feature: retired-committers
+start-date: 2019-08-25
+author: Till Höppner
+co-authors: Graham Christensen
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Many people were given push access to the nixpkgs repository, which is kept even if
+these committers become inactive. This RFC proposes moving these contributors to
+a new team without push access.
+
+# Motivation
+[motivation]: #motivation
+
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome? -->
+
+Each committer represents secrets and access which need to be managed carefully.
+These come in the form of passwords, SSH and GPG keys, and leaking them can put nixpkgs
+at risk of of unauthorized modification.
+
+Because every secret with push access can be leaked, we should keep their number as low as necessary,
+here by deactivating the push access of inactive committers.
+
+As of 2019-08-18, at least 2 committers have officially stepped down, and at least 1 committer has
+not pushed to nixpkgs since 2014, but are still able to push directly to nixpkgs.
+
+If implemented in this form, and assuming no further contributions, 7 contributors will be moved at the beginning of 2020.
+
+
+# Detailed design
+[design]: #detailed-design
+
+<!-- This is the bulk of the RFC. Explain the design in enough detail for somebody
+familiar with the ecosystem to understand, and implement.  This should get
+into specifics and corner-cases, and include examples of how the feature is
+used. -->
+
+Inactive committers will have their push access disabled after not committing to nixpkgs for an entire year.
+
+That year is measured from January 1 to December 31 instead of using a rolling window over the last 12 months,
+to be more predictable for committers and reduce the evaluations from 12 times a year to just once a year.
+
+For each committer from the [Nixpkgs Committers team](https://github.com/orgs/NixOS/teams/nixpkgs-committers), the number of commits
+in that time range is checked, and the committer is considered inactive if there are none.
+
+This process is repeated at the beginning of each new year.
+
+Previous committers are moved to a new Nixpkgs Committers Emeritus team, to honor their past contributions.
+Members of this team will remain in the GitHub organisation, and may regain push access at a later time.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+<!-- Why should we *not* do this? -->
+
+- It might put pressure on people because they might lose their hard-earned permissions.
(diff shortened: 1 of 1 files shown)


rfcs/0055-retired-committers.md
```diff
@@ -0,0 +1,134 @@
...
+- Committers could keep push access forever.
+- We could be even stricter, at the risk of higher contributor churn and losing low-frequency direct contributions.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+<!-- What parts of the design are still TBD or unknowns? -->
+
+- Is one year without commits a good activity threshold?
+- How are committers informed about this change, or an impending revocation?
```
Mic92: I think they should be informed some time before (maybe two months?). That their push access will be removed unless they add a commit. There might be some extreme circumstances like an illness that knock out someone for a year.

grahamc: An alternative to this alternative is make it very easy to get commit bit back. I don't think there should be any permanence to the commit bit removal. If there is nothing difficult/annoying about getting commit bit back, is there any need to let them know in order to keep it?

rfcs/0055-retired-committers.md
```diff
@@ -0,0 +1,134 @@
...
+
+<!-- Why are we doing this? What use cases does it support? What is the expected
+outcome? -->
+
+Each committer represents secrets and access which need to be managed carefully.
+These come in the form of passwords, SSH and GPG keys, and leaking them can put nixpkgs
+at risk of of unauthorized modification.
+
+Because every secret with push access can be leaked, we should keep their number as low as necessary,
+here by deactivating the push access of inactive committers.
```
Mic92: One motivation could be as well that they might have also loose access to their account.

7c6f434c:
Would we want to take into account any other kinds of projet participation? Write-access-require project participation  (Like nontrivial labelling of issues)? Commits to other NixOS organisation repos?

(I recognise that a possible answer is «would be nice, but requires too much fighting GitHub»)

rfcs/0055-retired-committers.md
timokau: But if it is very easy to get the commit bit back, doesn't that kind of negate the security motivation? At least if we assume the accounts/secrets might get compromised.

grahamc:
> want to take into account any other kinds of projet participation?

At this time, I'm not sure it is necessary. The list of people to have commit bit revoked is so low. Maybe that list could be shared, to confirm each person is actually inactive?

rfcs/0055-retired-committers.md
grahamc: Good question, I don't know.

tilpner: @Mic92 An email (only required contact information in maintainer-list) two months before the end of the year sounds good.

@grahamc I definitely want a notification, it should not feel unreasonable or be unexpected, but I'm not sure the requirement for re-activation should be subject of this RFC.

If a limit is needed, something with more friction than just dropping a line on IRC, but less than the current initial limit (?) of 50 recent PRs with non-trivial contributions should be acceptable, perhaps a tenth of that.

rfcs/0055-retired-committers.md
tilpner: Good point. Should I include that in the motivation?

tilpner:
@7c6f434c I don't know how commit access to non-nixpkgs repositories is managed. I've so far assumed the team nixpkgs-committers is only used for nixpkgs.

If possible, labelling of issues could be solved with another group, which only grants labelling permissions. That would also be useful for granting labelling rights to people without giving them commit access, but such a team is out-of-scope for this RFC.

rfcs/0055-retired-committers.md
worldofpeace: How about unresponsive maintainers and committer issues?

```
Hello, x committer has been unresponsive for x days.
They maintain:

List of things
- 
-

If anyone knows how to get in contact with them please let us know.

[Detailed section about the process with moving commit access]
```

Another issue is what do we do with packages that say they're maintained by them?
We need to be able to move them out of meta.maintainers as well.

7c6f434c:
Maybe a better version of my question would be: if the committer responds and says they are kind-of-half-active around Nixpkgs, do we cancel the revocation?

alyssais:
It's worth noting when discussing public announcements of impending
removal, etc., that the list of committers isn't currently public.  I'm
not saying it shouldn't be, but if it's going to be made semi-public
through automated(?) announcements anyway, we might as well take the
opportunity to properly make it public.

worldofpeace:
True @alyssais, and I think it's unfair to people who aren't members of the org to figure out who can commit unless they directly interact with them. This is a source of confusion because they can't even tell the team does exist (I can think back on occassion where people are like "what is nixpkgs-committers?).
I'd say if you're a committer your public membership is already prevalent, so default secrecy doesn't make sense to me.

I do believe if people desire privacy in that regard they can make their status non-public?

rfcs/0055-retired-committers.md
Mic92: Yes.

tilpner:
@7c6f434c I don't have a good answer to that question, but I would expect the specifics to matter a lot. A vague "I'm still active-ish" should be treated differently to "I didn't have time recently, but I plan to work on X during the next n months". I'm not sure it's necessary (or sensible) to write up rules for this right now. I don't expect it to occur often, and trust whoever ends up being responsible for this to be reasonable.

@alyssais The RFC currently does not propose or require public announcements. While I would welcome nixpkgs-committers becoming public, that is unrelated and I don't want that to block this RFC.

rfcs/0055-retired-committers.md
tilpner: @worldofpeace It's possible to keep maintaining their things without commit access, so I argue that removing them as a maintainer should not be part of this RFC.

Of course that's unlikely to happen, considering they haven't contributed in the past year, but expanding this RFC to affect all maintainers (as opposed to just the committers) and removing maintainers from packages (and potentially removing packages if they become unmaintained) feels very far out of scope for now.

Zimmi48:
@worldofpeace The nixpkgs-committers team could be made public to other organization members such as the members of nixpkgs-maintainers. However (and quite unfortunately), GitHub does not support making a team public to non-members of the organization. So if this is what is proposed, then it will require another means, such as maintaining a public list in a markdown file somewhere.

7c6f434c:
@tilpner Yes, with advance notification and explicit lack of attempt to automate removal the plan sounds reasonable.

Mic92:
I would like nominate myself as a shepherd for this RFC.

shlevy:
Any others interested in shepherding? We need one or two more.

domenkozar:
We need one more nominations, as there can be only 50% from SC.

Mic92:
Would you like to shepherd this one as well? @alyssais

domenkozar:
If we follow the procotol, all three shepherds are now part of RFC steering committee. So we'd need to swap one.

domenkozar:
I'd like to nominate @grahamc.

7c6f434c:
>If we follow the procotol, all three shepherds are now part of RFC steering committee. So we'd need to swap one.

… and add two to get to half-and-half

alyssais:
> Would you like to shepherd this one as well? @alyssais 

Sure.

edolstra:
We decided to go with @Mic92, @edolstra, @worldofpeace and @alyssais as the shepherd team. Strictly speaking this violates the rule that there should be two shepherds that are not on the steering committee but since @worldofpeace was not on the steering committee at the time of nomination we'll just ignore that :-)

Mic92:
To move this forward I would like to organize a short meeting for this one. I will sent out an email.

rfcs/0055-retired-committers.md
```diff
@@ -0,0 +1,136 @@
...
+A special case of inactive committers are those who have lost access to their GitHub account entirely,
+who would be unable to remove potentially leaked secrets from their account.
+
+As of 2019-08-18, at least 2 committers have officially stepped down, and at least 1 committer has
+not pushed to nixpkgs since 2014, but are still able to push directly to nixpkgs.
+
+If implemented in this form, and assuming no further contributions, 7 contributors will be moved at the beginning of 2020.
```
Mic92: Things to address in the RFC discussed in the meeting:

- set an start date
- create an issue + email
- Who is doing the removal

Mic92:
I was selected as a shepherd for this RFC.

worldofpeace:
@Mic92 Hope you don't mind I edited your commit, because I believe you meant shepherd leader.

And just a mention to @NixOS/rfc-steering-committee, we need to agree on leaders unanimously during our meetings :grin:

worldofpeace:
I'm going to be traveling next month, and my schedule actually works better if I can do things in real time. The date and time we have here will actually still be perfect when I travel, so I suggest another meeting soonish.

Mic92:
@worldofpeace Can you state your objections in this RFC issue itself? Maybe we can get those solved without requiring a meeting in person.

worldofpeace:
> @worldofpeace Can you state your objections in this RFC issue itself? Maybe we can get those solved without requiring a meeting in person.

Eventually yes, but I'm full on travel stuff Mar 6-21.

worldofpeace:
I'm going to try to find some time hopefully next week to look everything over and we can probably have motion for FCP.

Mic92:
Unless there are major objections we are merging this RFC on the 24th of May.

Mic92:
Only two days left until this is merged.

Mic92:
This is the first batch: https://github.com/NixOS/nixpkgs/issues/88867

worldofpeace:
It's been brought to my attention by @jtojnar that the team for the honor designation to past contributors has a masculine meaning.

I'll quote our discussion on IRC:
> @jtojnar
> Mic92: I only took half semester of latin before it was cancelled but committer emeritus does not sound good to me
since emeritus is masculinum singular

> @Mic92 
> Jan Tojnar: I just followed rfc55. Feel free to open a RFC against it.

> Taneb
> Jan Tojnar: it's accpeted usage as an English adjective, which doesn't change form with gender

> @worldofpeace 
> wikipedia's header for the word "Emeritus (/əˈmɛrɪtəs/; female: Emerita"
> the notes " feminine emerita or emeritus; plural emeriti (masc.) or emeritae (fem.); abbreviation emer."

> @jtojnar
> Taneb I have definitely seen emerita in English being used for female professors

> see also https://wmich.edu/writing/rules/alumni

> @worldofpeace 
> Jan Tojnar: reading that makes me not really like the word/s
> especially the usage note that like alumni is never wrong for groups that do include women, even though it's a plural male form?
> plain weird to me, but I'm not here with a better name suggestion just yet. and it's pretty easy to change team names

> @lheckemann 
> How about "former committers"?

> @jtojnar
> grammatical genders are weird, I like to think about masculinum as a gender category that also among other things refers to male people
because otherwise I get weird clashes between grammatical genders of different languages

> @worldofpeace 
> hehe, they really are. 
sphalerite I mean, former committers would be exactly that. I think this word was chosen because it was like an honor position
> like we could say "retired" or really inactive would be exact, it's just it doesn't have that honor indication
> So options would then be - Nixpkgs Former Committers Team - Nixpkgs Retired Committers Team - Nixpkgs Inactive Committers Team

@jtojnar also mentioned the existence of [Latinx](https://en.wikipedia.org/wiki/Latinx) which I responded with

> Jan Tojnar: gender-neutral in spanish is so hard. "To capture the complexity of the various terms used the term Latin* (pronounced Latin) was introduced in 2020" I see how that is related to 
sphalerite comment
I personally use latinx or X etc. for a lot things

From the logs I'm interesting in creating a slight amendment, and also to the benefit of being a more accessible english word.

PR #17, opened 2017-08-11
[RFC 0017] Intensional Store

wmertens:
This RFC builds on the implementation of RFC #62 to maximize its benefits.

- Decouple metadata (Trust DB) from the Store, keep per-user
- Move the Store to `/var/lib/nix` for better compatibility with other platforms
- Store can be shared read-write on a network share
- `nix-daemon` becomes optional
- Store paths provide dependencies without database

[Rendered](https://github.com/wmertens/rfcs/blob/master/rfcs/0017-intensional-store.md)

PR #57, opened 2019-11-13
WIP: Nix-Cas rfc

regnat:
A draft PR to gather some preliminary reviews on this RFC before (hopefully) submitting it upstream.

This ist still quite WIP and probably needs a lot of rewriting, but the bulk of the content should be there


wmertens:
Why'd you close this? It's interesting, I'm happy to see that you already have an implementation. Did you read #17 and preferably the comments? Probably we can build on your work to implement #17.

Ericson2314:
I think they meant to submit it internally and then submit it upstream?

regnat:
Yes sorry for that, this was supposed to be internal at that point (because a large part of it is still hypothetical from an organizational point of view). But hopefully it won't stay that way for long :)

Ericson2314:
Good luck! I did sneak a peak and I like what I see a lot :).

PR #58, opened 2019-11-14
[RFC 0058] Named Ellipses

deliciouslytyped:
It should be possible to bind a name to ellipses in a function definition like `{ a, ...@extra }: null`, and `{ a, extra@... }: null`. This makes intuitive sense, and could remove the need for a lot of uses of `removeAttrs` that really just want to refer to the contents of ellipses.

[Rendered](https://github.com/deliciouslytyped/rfcs/blob/patch-1/rfcs/0057-named-ellipses.md)

Edit:
Proposed discussion point:
Should scope of this be expanded to binding any function argument to new names - for consistency, even though that might be considered redundant? E.g.: `a@{b@c, d@...}: null`
Maybe it's preferable to get this simple RFC through first though.

### DIFF
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -22,6 +22,8 @@ This is pretty simple so it#s redundant with the summary:
 # Detailed design
 [design]: #detailed-design
 
+`{...@extraargs}: extraargs` should yield an attrset the extra arguments "in" `...`.
+
 Help?
 
 This is the bulk of the RFC. Explain the design in enough detail for somebody
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -17,7 +17,7 @@ It should be possible to bind ellipses to a name in a function argument definiti
 [motivation]: #motivation
 This is pretty simple so it#s redundant with the summary:
 - I think it makes sense to be able to refer to "the rest of the arguments" as a first class object
-- This should also allow nixing a lot of usages of the `removeAttrs` pattern
+- This should also allow nixing a lot of usages of the `removeAttrs` pattern TODO: substantiate this
 
 # Detailed design
 [design]: #detailed-design
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -22,7 +22,7 @@ This is pretty simple so it#s redundant with the summary:
 # Detailed design
 [design]: #detailed-design
 
-`{...@extraargs}: extraargs` should yield an attrset the extra arguments "in" `...`.
+`{...@extraargs}: extraargs` should yield as an attrset the extra arguments "in" `...`.
 
 Help?
 
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -1,7 +1,7 @@
 ---
 feature: Named Ellipses OR function argument consistency
 start-date: (fill me in with today's date, YYYY-MM-DD)
-author: @deliciouslytyped
+author: deliciouslytyped
 co-authors: (find a buddy later to help our with the RFC)
 shepherd-team: (names, to be nominated and accepted by RFC steering committee)
 shepherd-leader: (name to be appointed by RFC steering committee)
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -12,7 +12,7 @@ related-issues:
 # Summary
 [summary]: #summary
 
-It should be possible to bind ellipses to a name in a function definition like `{ a, ...@extra }: null`, and `{ a, extra@... }: null`. This makes intuitive sense, and could remove the need for a lot of uses of `removeAttrs` that really just want to refer to the contents of ellipses.
+It should be possible to bind a name to ellipses in a function definition like `{ a, ...@extra }: null`, and `{ a, extra@... }: null`. This makes intuitive sense, and could remove the need for a lot of uses of `removeAttrs` that really just want to refer to the contents of ellipses.
 
 TODO: The latter could be substantiated.
 
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -1,0 +1,53 @@
+---
+feature: Named Ellipses OR function argument consistency
+start-date: (fill me in with today's date, YYYY-MM-DD)
+author: @deliciouslytyped
+co-authors: (find a buddy later to help our with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+It should be possible to bind ellipses to a name in a function argument definitions like `{ a, ...@extra }: null`, and `{ a, extra@... }: null`. This makes intuitive sense, and would remove the need for a lot of uses of `removeAttrs` that really just want to refer to the contents of ellipses.
+
+# Motivation
+[motivation]: #motivation
+This is pretty simple so it#s redundant with the summary:
+- I think it makes sense to be able to refer to "the rest of the arguments" as a first class object
+- This should also allow nixing a lot of usages of the `removeAttrs` pattern
+
+# Detailed design
+[design]: #detailed-design
+
+Help?
+
+This is the bulk of the RFC. Explain the design in enough detail for somebody
+familiar with the ecosystem to understand, and implement.  This should get
+into specifics and corner-cases, and include examples of how the feature is
+used.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+None?
+This implements syntax that would not have worked before and so in theory shouldn't cause breakage in the Nix ecosystem. [Citation Needed]
+
+# Alternatives
+[alternatives]: #alternatives
+None?
+
+Not doing this would not have any major impact besides not making nix and nixpkgs nicer to use.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+Should scope of this be expanded to binding any function argument to new names - for consistency, even though that might be considered redundant?
+
+
+# Future work
+[future]: #future-work
+None?
+
+# Bibliography
+Named ellipses - https://github.com/NixOS/nix/issues/2998
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -19,7 +19,8 @@ Nixpkgs often gets commits like https://github.com/NixOS/nixpkgs/commit/a5065329
 # Detailed design
 [design]: #detailed-design
 
-`{...@extraargs}: extraargs` should yield as an attrset the extra arguments "in" `...`.
+The intent of this syntax is for `{...@args}:` to behave as close to the current `{...} @ args:`
+ as practical. As both `args @ {…}` and `{…} @ args` forms happen in Nixpkgs, both would be supported for named ellipses, too. As `args @ {args}: ` is an error in Nix because of duplicate argument names, so should be `{extra, ...@extra}: `. As `(a @ { ... }: a) {a=1;}` returns `{a=1;}`, so should `({ a @ ... }: a) {a=1;}`. ```
 
 TODO: consider what other languages like Haskell do
 
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -21,6 +21,8 @@ TODO: The latter could be substantiated.
 
 `{...@extraargs}: extraargs` should yield as an attrset the extra arguments "in" `...`.
 
+TODO: consider what other languages like Haskell do
+
 # Drawbacks
 [drawbacks]: #drawbacks
 None? This implements syntax that would not have worked before and so in theory shouldn't cause breakage in the Nix ecosystem. [Citation Needed]
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -3,8 +3,8 @@ feature: Named Ellipses
 start-date: 2019-11-14
 author: deliciouslytyped
 co-authors: none
-shepherd-team: (names, to be nominated and accepted by RFC steering committee)
-shepherd-leader: (name to be appointed by RFC steering committee)
+shepherd-team: Eelco Dolstra, Michael Raskin, jD91mZM2
+shepherd-leader: Michael Raskin
 related-issues:
   Named Ellipses - https://github.com/NixOS/nix/issues/2998
 ---
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -1,5 +1,5 @@
 ---
-feature: Named Ellipses OR function argument consistency
+feature: Named Ellipses
 start-date: 2019-11-14
 author: deliciouslytyped
 co-authors: none
@@ -19,8 +19,10 @@ Nixpkgs often gets commits like https://github.com/NixOS/nixpkgs/commit/a5065329
 # Detailed design
 [design]: #detailed-design
 
-The intent of this syntax is for `{...@args}:` to behave as close to the current `{...} @ args:`
- as practical. As both `args @ {…}` and `{…} @ args` forms happen in Nixpkgs, both would be supported for named ellipses, too. As `args @ {args}: ` is an error in Nix because of duplicate argument names, so should be `{extra, ...@extra}: `. As `(a @ { ... }: a) {a=1;}` returns `{a=1;}`, so should `({ a @ ... }: a) {a=1;}`. ```
+The intent of this syntax is for `{...@args}:` to behave as consistent with `{...} @ args:`
+ as practical. As both `args @ {…}` and `{…} @ args` forms happen in Nixpkgs, both would be supported for named ellipses, too. 
+ 
+ As `args @ {args}: ` is an error in Nix because of duplicate argument names, so should be `{extra, ...@extra}: `. As `(a @ { ... }: a) {a=1;}` returns `{a=1;}`, so should `({ a @ ... }: a) {a=1;}`. ```
 
 TODO: consider what other languages like Haskell do
 
@@ -30,8 +32,4 @@ This increases the amount of syntax Nix has, thus creating some maintenance cost
 
 # Alternatives
 [alternatives]: #alternatives
-Not doing this would not have any major impact besides not making nix and nixpkgs nicer to use.
-
-# Unresolved questions
-[unresolved]: #unresolved-questions
-Should scope of this be expanded to binding any function argument to new names - for consistency, even though that might be considered redundant?
+Not doing this would not have any major impact besides not making Nix and nixpkgs nicer to use.
--- a/rfcs/0057-named-ellipses.md
+++ b/rfcs/0057-named-ellipses.md
@@ -1,11 +1,12 @@
 ---
 feature: Named Ellipses OR function argument consistency
-start-date: (fill me in with today's date, YYYY-MM-DD)
+start-date: 2019-11-14
 author: deliciouslytyped
-co-authors: (find a buddy later to help our with the RFC)
+co-authors: none
 shepherd-team: (names, to be nominated and accepted by RFC steering committee)
 shepherd-leader: (name to be appointed by RFC steering committee)
-related-issues: (will contain links to implementation PRs)
+related-issues:
+  Named Ellipses - https://github.com/NixOS/nix/issues/2998
 ---
 
 # Summary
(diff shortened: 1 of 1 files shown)


deliciouslytyped:
Note, I'm not really able to implement this by myself, so someone else will have to volunteer to implement it - or wait indefinite time for me to learn how.

deliciouslytyped:
It has been proposed/pointed out to me over IRC that JavaScript uses `...somename`.
I don't particularly have a preference - I didn't know there is a language that does this, but it initially felt somewhat aesthetically ugly, having no separator between whatever `...` is, and the name. I consider my preference arbitrary though.

A more reasoned counterpoint however, would be desiring consistency with using `@` like in `a@{whatever}:`?

deliciouslytyped:
It may be relevant to this general dicussion to explicitly point out that both `a@{}: null` and `{}@a: null` are valid. Not everyone knows this.

rfcs/0057-named-ellipses.md
```diff
@@ -0,0 +1,36 @@
...
+# Detailed design
+[design]: #detailed-design
+
+`{...@extraargs}: extraargs` should yield as an attrset the extra arguments "in" `...`.
+
+TODO: consider what other languages like Haskell do
+
+# Drawbacks
+[drawbacks]: #drawbacks
+None? This implements syntax that would not have worked before and so in theory shouldn't cause breakage in the Nix ecosystem. [Citation Needed]
```
7c6f434c: I would say that Nix has a complicated enough syntax that increasing syntax complexity should be recognised as a drawback per se to be considered in comparison to the benefit. Every syntactical construct is a cost for third-party Nix-processing tooling, for example.

(I am not saying this drawback overweights the benefit here)

rfcs/0057-named-ellipses.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+Not doing this would not have any major impact besides not making nix and nixpkgs nicer to use.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+Should scope of this be expanded to binding any function argument to new names - for consistency, even though that might be considered redundant?
```
7c6f434c: I think this would increase complexity without being that much simpler than `let … in`

rfcs/0057-named-ellipses.md
```diff
@@ -0,0 +1,36 @@
...
+It should be possible to bind a name to ellipses in a function definition like `{ a, ...@extra }: null`, and `{ a, extra@... }: null`. This makes intuitive sense, and could remove the need for a lot of uses of `removeAttrs` that really just want to refer to the contents of ellipses.
+
+TODO: The latter could be substantiated.
+
+# Detailed design
+[design]: #detailed-design
+
+`{...@extraargs}: extraargs` should yield as an attrset the extra arguments "in" `...`.
+
+TODO: consider what other languages like Haskell do
```
7c6f434c: Note that you want to consider pattern matching libraries, too, not just function arguments.

Some languages use notations that wouldn't fit well with the current use of `...` in Nix: `def f(a,b,*args):` in Python, `(lambda (a b &rest args) …)` in Common Lisp.

rfcs/0057-named-ellipses.md
deliciouslytyped: I reluctantly agree and will add something later.

7c6f434c:
By the way, to avoid adding a new underspecified syntax construct in the language: is `({extra@...}: extra) {extra = 1;}` a runtime error?

rfcs/0057-named-ellipses.md
7c6f434c: Now that I think of it, the functionality (although not the syntax) could indeed be added as a library function (as a higher-order function).

rfcs/0057-named-ellipses.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Summary
+[summary]: #summary
+
+It should be possible to bind a name to ellipses in a function definition like `{ a, ...@extra }: null`, and `{ a, extra@... }: null`. This makes intuitive sense, and could remove the need for a lot of uses of `removeAttrs` that really just want to refer to the contents of ellipses.
+
+TODO: The latter could be substantiated.
+
+# Detailed design
+[design]: #detailed-design
```
asymmetric: It would be useful to have an example of how people are currently using `removeAttrs` somehwere in the RFC, for comparison.

deliciouslytyped:
> By the way, to avoid adding a new underspecified syntax construct in the language: is `({extra@...}: extra) {extra = 1;}` a runtime error?

The question is whether `extra = 1` should end up inside the `...` right?
- If yes then it should be fine though it might be a bit confusing if one happens to have a name collision?
Right now I'm leaning towards this option.
- if no, you shouldn't be allowed to do that because - how do you say this... the name is being created by the interpreter - right?
In which case I think we should fail at non-runtime if possible...

rfcs/0057-named-ellipses.md
deliciouslytyped: Yeah that's what I meant. I'll add something when I have some spare brain for it, unless someone else wants to give it a shot.

7c6f434c:
>> By the way, to avoid adding a new underspecified syntax construct in the language: is `({extra@...}: extra) {extra = 1;}` a runtime error?
>
>The question is whether `extra = 1` should end up inside the `...` right?
>- If yes then it should be fine though it might be a bit confusing if one happens to have a name collision?
>Right now I'm leaning towards this option.

Hm, given the current behaviour of `(args@{...}: args) { args = 1; }` it does sound reasonable that for `(args@{extra@...}: [extra args.extra]) { extra = 1; }` we get `[{ extra = 1; } 1]`; but this should probably be mentioned.

deliciouslytyped:
Oh yeah, that's a nice and simple way to justify that.
`(args@{...}: args) { whatever } == (args@{extra@...}: args) { whatever }`

edolstra:
This RFC is now open for shepherd nominations. Any volunteers? I'm nominating myself :-) @7c6f434c Would you be willing to shepherd?

7c6f434c:
I am close enough to being on the fence, so I accept the nomination.

I would also like to nominate @jwiegley as the lead author of a popular alternative Nix parser.

7c6f434c:
Hm. Just to make sure. @jwiegley — did the notification get lost, or do you not want to accept nomination?

Mic92:
I would like @jD91mZM2 and @matklad to shepherd this one.

jD91mZM2:
Not sure what shepherding means in this context, but sure I'm in

7c6f434c:
@jD91mZM2 shepherding is carrying out the Shepherd duties as described in https://github.com/NixOS/rfcs/blob/master/rfcs/0036-rfc-process-team-amendment.md

jD91mZM2:
Okay I'll try, let me know if I'm doing things incorrectly.

This PR looks good and although I don't know how often `removeAttrs` is used inside nixpkgs for example, I agree with the intention of the RFC.

As mentioned by @7c6f434c, expanding the syntax is not necessarily a pro. For this reason I am a little hesitant to let the user choose which side of the ellipsis to put the "binding" as it costs a bit of duplicate code while parsing. Then again, allowing both sides will keep consistency with `{}@thing` which makes sense.

In the future I'm hoping this can replace full-destruct bindings (`{}@thing`), for the sake of minimizing syntax. If so we definitely need to make sure this syntax is something that everyone agrees on. I like this, though. But currently this overall looks good, and should probably not be any trouble to parse and therefore add only a small bit of complexity. From a parser's perspective, keeping the binding to the right side of the ellipsis may make things slightly cleaner.

domenkozar:
@7c6f434c would you mind being the leader? If not, then we can move this forward.

7c6f434c:
>@7c6f434c would you mind being the leader? If not, then we can move this forward.

Let's try…

alyssais:
I don't particularly care what the syntax for this is, but I think it
would be great to have.

We need to do the builtins.removeAttrs dance every time we make a
function that is a wrapper around mkDerivation, which happens quite a
lot, and there's no way to do it with builtins.removeAttrs that isn't
extremely ugly.

Example: I recently did it to buildSkawarePackage in
https://github.com/NixOS/nixpkgs/commit/a50653295df5e2565b4a6a316923f9e939f1945b.

alyssais:
Also, I think this should be renamed RFC 0058, since I think they're
supposed to match the PR number.

asymmetric:
Is there any progress on this issue?

7c6f434c:
Sorry for my not being more active here, I am currently comprehensively failing at juggling some overlapping deadlines.

I think there are some outstanding points that @deliciouslytyped as the author generally agrees with but hasn't find the time to implement (and I am not one to throw stones here…)

I think the points are:

- `removeAttrs` use in Nixpkgs raised by you: would be nice ot have an example or two inline, and maybe some more references to use to better describe the current situation; goal: context/motivation
- in the drawbacks section: acknowledge the complexity cost; maybe remove the unresolved question for argument renaming (I see nobody who actively says this should be done)
- comparison with `@{...}`: motivation for design decisions, and an additional source of disambiguation in case implementation discovers underspecification

A point related to the last point is that maybe we should also see how `args@{...}` and `{...}@args` are used in Nixpkgs. If there is a clear supermajority, we might consider only defining a matching version for the ellipsis.

7c6f434c:
@deliciouslytyped is my understanding of your plans to address some of the in-discussion issues correct?

deliciouslytyped:
It's been a while heh. I'm not sure what you mean by "plans to address", but yeah you're correct.
Empirically, I don't plan to do a whole lot to get this through, but it seems to me like a low-effort high-return thing for someone that can actually implement it. I'd be happy to hand it off.

- https://github.com/NixOS/rfcs/pull/58#issuecomment-568281475 gives a stereotypical example of the issue.
- argument renaming can be postponed
- I didn't think of "just" doing `args@{...}`, but that once again has the problem of not separating mandatory, and "extra" arguments? An alternative approach I just thought of is `mandatory@{a,b,c}: extraargs@{...}: funcbody`, but that goes against the usual nixpkgs conventions? 
- I think it should work with both `args@{a, ...}` and `{a, ...}@args`. It would be inconsistent not to. Why would you even make them behave different? I don't know how the nix grammar defines these but I'd be a bit surprised if the code wasn't the same.

7c6f434c:
Well, as GitHub loves to hide the middle of discussion at unpredictable times, I want to make sure as much of the discussion context as makes sense is explicitly mentioned in the RFC text. 

> #58 (comment) gives a stereotypical example of the issue.

Agreed, and a mention of this code would make a nice «Motivation» part for the RFC

> I didn't think of "just" doing args@{...}

I do not say about «just» — I mean that in https://github.com/NixOS/rfcs/pull/58#issuecomment-554740713 you said that consistency of behaviour is a good disambiguation tool for the specification, so I think it woul be nice to have something about that in the RFC

> I think it should work with both args@{a, ...} and {a, ...}@args. It would be inconsistent not to. Why would you even make them behave different?

For uniformity of the code, given that there is no useful semantic difference?

Mic92:
@7c6f434c Can you set up a meeting for this RFC?

7c6f434c:
I hoped to talk @deliciouslytyped into including some more parts of the discussion into the RFC first, but yeah, maybe we should try a meeting to set the priorities instead.

https://www.when2meet.com/?8854240-aOUQn
@deliciouslytyped @edolstra @jD91mZM2 — please fill your preferences.

I support the recent discussion that it's good when people with the interest in the topic join the meeting. Please post a summary of your point of view on the proposal (and maybe your related interests) if you want to join the discussion, and fill in the time preference. I want to accomodate the interested community members, but of course having a meeting of the Shepherd Team and the author gets higher priority if there is no perfect time.

7c6f434c:
Just a reminder: @edolstra @jD91mZM2 any objections to having a meeting soon-ish? And if no, could you please enter some preferences?

jD91mZM2:
I'm okay with having a meeting soon, although I have no idea how meetings go as I'm still very new to Nix' RFC process

7c6f434c:
@jD91mZM2 there is a link for timeslot preferences. Basically, the point of the meeting is making sure everyone understands others' points of view correctly, re-summarize the situation re: RFC and discussion, and hopefully reach a consensus on further steps.

rfcs/0057-named-ellipses.md
```diff
@@ -0,0 +1,36 @@
...
+related-issues:
+  Named Ellipses - https://github.com/NixOS/nix/issues/2998
+---
+
+# Summary
+[summary]: #summary
+
+It should be possible to bind a name to ellipses in a function definition like `{ a, ...@extra }: null`, and `{ a, extra@... }: null`. This makes intuitive sense, and could remove the need for a lot of uses of `removeAttrs` that really just want to refer to the contents of ellipses.
+
+TODO: The latter could be substantiated.
```
7c6f434c:
```suggestion
Nixpkgs often gets commits like https://github.com/NixOS/nixpkgs/commit/a50653295df5e2565b4a6a316923f9e939f1945b with code that would be cleaner without the need for extra `removeAttrs`.
```

rfcs/0057-named-ellipses.md
7c6f434c:
```suggestion
This increases the amount of syntax Nix has, thus creating some maintenance cost both for Nix itself, and for tools intended to work with Nix syntax (from highlighting to hnix)
```

rfcs/0057-named-ellipses.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+None? This implements syntax that would not have worked before and so in theory shouldn't cause breakage in the Nix ecosystem. [Citation Needed]
+
+# Alternatives
+[alternatives]: #alternatives
+Not doing this would not have any major impact besides not making nix and nixpkgs nicer to use.
+
+# Unresolved questions
```
7c6f434c: Let's drop this section as no strong motivation for argument renaming has been provided.

rfcs/0057-named-ellipses.md
```diff
@@ -0,0 +1,36 @@
...
+[summary]: #summary
+
+It should be possible to bind a name to ellipses in a function definition like `{ a, ...@extra }: null`, and `{ a, extra@... }: null`. This makes intuitive sense, and could remove the need for a lot of uses of `removeAttrs` that really just want to refer to the contents of ellipses.
+
+TODO: The latter could be substantiated.
+
+# Detailed design
+[design]: #detailed-design
+
+`{...@extraargs}: extraargs` should yield as an attrset the extra arguments "in" `...`.
```
7c6f434c:
```suggestion
The intent of this syntax is for `{...@args}:` to behave as close to the current `{...} @ args:`
 as practical. As both `args @ {…}` and `{…} @ args` forms happen in Nixpkgs, both would be supported for named ellipses, too. As `args @ {args}: ` is an error in Nix because of duplicate argument names, so should be `{extra, ...@extra}: `. As `(a @ { ... }: a) {a=1;}` returns `{a=1;}`, so should `({ a @ ... }: a) {a=1;}`. ```

**Comment by deliciouslytyped** [comment]
@7c6f434c I committed your change suggestions, LGTM.

**Comment by 7c6f434c** [comment]
So, my opinion on the current state.

I tried to organise a meeting and failed to recieve a reaction from my fellow shepherds; we did discuss the RFC a bit via IRC DMs with @deliciouslytyped 

Right now I think this RFC has no clear outstanding issues. As it is, it describes a clean feature request with motivation (including context), basic outline and guidelines to resolving ambiguity. I see the reception of the proposal as positive with no negative reactions (although maybe not very active).

I think there are three reasonable outcomes from here: 

1. We accept this with minor edits or as is, as a confirmation that a implementing such a feature would be reviewed as a welcome feature (of course, with usual review of implementation trade-offs etc.)
2. Someone says they want to try implementing this, then there is a more detailed specification based on implementation experience.
3. Someone raises an objection that ends up recognised as significant enough to reject or postpone.

In my opinion cases 2 and 3 require some additional arguments coming to light, but would not be surprising. But in the current situation as it is I have a growing impression that **accepting** this RFC after a minor proofreading just in case is the most natural course of action. In the «accept as a clean feature request» case I don't even see any benefit in having a synchonous meeting. 

 @edolstra @jD91mZM2 — what is your opinion on all this? Do you agree with my summary?

To everyone — any additional input?

**Comment by zimbatm** [comment]
Extending the syntax is a serious thing. It's something we will pay for quite extensively. All the tooling needs to be updated. This includes all the nix parsers, the various nix syntax highlighting tools, which then need to trickle down into all of the platforms. And then we will keep paying for the extra complexity over the years.

In my opinion, right now, the RFC fails to motivate the change explicitly enough. The RFC should have at least a few before/after examples that show how the code base would be improved. Ideally by picking examples from nixpkgs.

A related topic is also default attributes and how they are not captured by `@attrs` in the current syntax. It would be good to also explore that area and see if it improves it.

**Comment by deliciouslytyped** [comment]
The default attribute capture situation is a good point.

**Comment by zimbatm** [comment]
I think you will see that when taking examples from nixpkgs. The desire to capture the elipsis might be related if the users re-constructs the attribute set with the default values.

```
{ a ? true, ...@attrs}: stdenv.mkDerivation (attrs // { inherit a; })
```
vs (other hypothetical notation that captures with default values)
```
{ a ? true, ... } @@attrs: stdenv.mkDerivation attrs
```

Or I might be wrong. But that's the type of analysis I am hoping to see and that would help substantiate the claim.

**Comment by deliciouslytyped** [comment]
I think I suggested doing something about that on IRC at some point and it was not well received, but I would be happy if we could do something so that people trip up on it less.

**Comment by deliciouslytyped** [comment]
Ok, I looked at it with pudb (which is a pretty nice pdb frontend actually), and I'm getting a failure with 
```
VIDIOC_REQBUFS: Inappropriate ioctl for device
```
at https://github.com/AlexEMG/DeepLabCut/blob/bcef8c961ed27a5fc8f947ef97b29a466e3f371a/deeplabcut/create_project/new.py#L156 , so it's returning an empty video_sets in the next lines.

jD91mZM2:
@7c6f434c Sorry if I missed something, I didn't see a link and assumed you'd get back to us with it. I agree with your message and don't see changing syntax as anything too troublesome.

7c6f434c:
Well, now that we do have arguments for insufficiency of motivation the situation obviously changes.

edolstra:
@7c6f434c Did you mean insufficiency or sufficiency?

I don't have major objections to accepting this RFC, though I do see @zimbatm's concerns - the language is already complicated enough, so new language features ideally should provide a big advantage.

Another issue is that this feature is potentially expensive at runtime (when used in a hot path), since it requires the input attrset to be cloned to get rid of the non-`...` arguments.

7c6f434c:
@edolstra I meant «insufficiency» but maybe my phrasing is just bad there. I pushed for a partial reflection of the same concerns, and (as an active member of discussion) I think current state is probably sufficient. But now with @zimbatm arguing that it is insufficient as it is, we have an opinion that the current state is insufficient (but probably fixable), so as a shepherd I think it is too early for FCP and these points can and should be addressed better.

 As usual, I forgot that Nix uses fully strict-on-keys attrsets so I did not ask to write down the performance point; given that people do `removeAttrs` the proposal is unlikely to make things noticeable worse. But indeed this should be reflected in the text (and later in the documentation).

edolstra:
@7c6f434c @deliciouslytyped  @jD91mZM2 This RFC has been stuck for a while, is there anything that can be done to get it unstuck? If I read the comments correctly, the main objection right now is a lack of evidence that this feature would be sufficiently useful to justify it. It could be helpful to add some examples from Nixpkgs where this feature would be useful.

7c6f434c:
>the main objection right now is a lack of evidence that this feature would be sufficiently useful to justify it. It could be helpful to add some examples from Nixpkgs where this feature would be useful.

I agree with this evaluation of the state of the RFC at the last update

deliciouslytyped:
I personally would still like to see this happen, but do not have the capacity to push for it until further notice.

Mic92:
Has somebody time else to come up with examples?

jtojnar:
Example: https://github.com/NixOS/nixpkgs/blob/67f489dae8ce99fb4a5418c87c5a91e327f3af11/nixos/tests/installed-tests/default.nix#L75-L81

Mic92:
@7c6f434c could you organize a meeting for this one?

7c6f434c:
Meeting, meeting… let's start by finding out if there is enough substance…

@deliciouslytyped are you interested in writing up more detailed motivation section for this RFC in any close future?

@jtojnar are you interested in maybe taking over / partially taking over this?

jtojnar:
@7c6f434c Unfortunately, I do not have the bandwidth at the moment.

7c6f434c:
@deliciouslytyped apparently my previous mention was just before you returned online.

deliciouslytyped:
Sorry, still distracted.

7c6f434c:
My opinion on meeting: it would be a cargo-cult as whatever is discussed will no realistically be implemented in predictable time.

My opinion on the current state: I am turning towards the idea to close with the decision like «defer». There is some level of positive interest in the community, and there is a question if enough consideration has been given to the costs and to the design trade-offs. Accepting this requires either more work on surveying the current Nixpkgs situation, potential mode of use, and consistency with other similar constructions in Nix and possibly in other languages; or maybe an implementation plan, which should both incentivise the justification work and allow to confirm the limited impact from varoius points of view.

Once someone has energy to either improve the justification part or implement it, the discussion can be continued.

[This is not yet a call for FCP, and probably not yet exactly the disposition I want to propose, just the likely direction]

lheckemann:
Shall we close this until we have some developments here?

7c6f434c:
Almost a month later, I formally motion for FCP with disposition to close until interest in further work is present. We have seen some interest in this feature, but to commit to a design as an RFC decision either more surveying/documentation work (and discussion of specific trade-offs based on more data) or some prototyping would be needed.

Continuation of discussion might be a reopening or a new RFC.

@edolstra @jD91mZM2

edolstra:
@7c6f434c I agree, let's FCP this for now. (Maybe @jD91mZM2 wants to reply as well.)

jD91mZM2:
Sure, I agree, let's close it. It's a convenient feature, but not something we have seen massive reason to push for yet.

7c6f434c:
Thanks.

Final Comment Period starts on this RFC with a disposition to close as «deferred». The proposal got positive feedback, but more work is needed for final commitment to accept details of the design.

Thanks to everyone who contributed to the discussion, hopefully the record will be useful at some future point when we revisit this topic.

edolstra:
Thanks, I've closed with label "status: deferred".

PR #59, opened 2019-11-16
[RFC 0059]: Systemd Service Secrets

d-goldin:
Hi,

This is a first draft for an RFC about a possible approach to secrets management for systemd services.

Curious to hear your thoughts.

[Rendered](https://github.com/d-goldin/rfcs/blob/master/rfcs/0059-secrets-for-services.md)


rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+# Detailed design
+[design]: #detailed-design
+
+To summarize, necessary preconditions:
+
+* Delivery of secrets to target systems is a solved problem
+* It's sufficiently secure to store the secrets or access tokens in a location
+  only accessible by root on the system
+* The secrets store locations is secure at rest, such as full-disk-encryption.
+* Interactive unlocking scenarios are should be treated separately
```
turion:
```suggestion
* Interactive unlocking scenarios should be treated separately
```

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+* Retrieved secrets are reliably cleaned up when the services stop, crash,
+  receive sigkill or the system is restarted
+
+Core concepts and terminology:
+
+* *Secrets store*: a secure file-system based location, in this document
+  `/etc/secrets`, only accessible to root
+* A *fetcher* function: a function whose task it is to resolve the secret
+  identifier, retrieve the secret and place it in the service process' private
+  namespace within /tmp name
```
turion:
```suggestion
  namespace within `/tmp` name
```

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+* Simple helper functions to *enrich* expressions defining systemd services
+  with secrets
+* "Side-car" service: A privileged systemd service running the fetcher
+  function to retrieve the fetcher function, and initially create the service
+  namespace
```
turion: "Running the fetcher function to retrieve the fetcher function"? Should one of the two mentions of "fetcher function" mean something else?

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+
+* *Secrets store*: a secure file-system based location, in this document
+  `/etc/secrets`, only accessible to root
+* A *fetcher* function: a function whose task it is to resolve the secret
+  identifier, retrieve the secret and place it in the service process' private
+  namespace within /tmp name
+* Simple helper functions to *enrich* expressions defining systemd services
+  with secrets
+* "Side-car" service: A privileged systemd service running the fetcher
+  function to retrieve the fetcher function, and initially create the service
```
turion:
```suggestion
  function to retrieve the secret, and initially create the service
```

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+More specifically, in this example a secrets scope is created - to allow for
+extensibility and differentiation a store has a type. In this case "folder"
+denotes a secrets store in the form of a root-only accessible locked down
+directory on the local filesystem. Here we want to acquire access to
+2 secrets, and 2 secrets only, which are specified in `loadSecrets`, by
+their id. How a secrets identifier is resolved, should be up to the fetcher
+function and here it's just trivially the file-name (this of course does not
+allow for file extensions).
+
+These secrets are then made acessible to the target service's unit definitions as
```
turion:
```suggestion
These secrets are then made accessible to the target service's unit definitions as
```

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+
+This is mostly functionality containing a _registry_ of existing fetchers, which
+might need to be configured by the user via their system configuration, the
+fetcher logic itself and functionality to generate side-car services and
+expose the secrets scope.
+
+## Rotating secrets
+
+Right now, secrets rotation is not done automatically. When new secrets are
+pushed, it is the responsibility of the user to restart the services affected.
```
turion: Is it possible to at least create a warning or suggestion at the end of e.g. `nixos-rebuild`, with the correct `systemctl restart foo.service` command?

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+no further harm is done aside from failing to access the resources (and possibly
+restart on its own).
+
+It would be possible to allow for automatic restarts using systemd path monitors.
+Also see _Future work_.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+I can't really think of a serious drawback right now, but hopefully the
```
turion: How does the secret store deal with system/main config upgrades? When I uninstall an old service and install a new one that asks for a secret of the same name, does the new service get access to the old secret?
When I uninstall a lot of services, are old secrets garbage collected? Do downgrades always work seamlessly?

Can you outline the possible pitfalls regarding that topic in an extra section?

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+
+Persistent confusion by new-comers and veterans alike, about what the a
+recommended way looks like and a variety of different approaches within
+nixos service modules.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* Is it sufficient to put responsibility on restarting services after key changes
+  onto the user or would an automated mechanism be better?
```
turion: I think typically an automated mechanism would be better. Most services just restart gracefully, so an additional unexpected restart is better than a confusing error message and downtime.

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+
+* Would it be better to create a side-cart per secret instead of per secret-scope+service?
+
+# Future work
+[future]: #future-work
+
+* When using a scope with multiple services, ideally only the secrets
+  referenced in the services definition should be made available to each
+  service. Right now all the secrets of the scope are blindly copied.
+* Transition of most critical services to use proposed approach
```
turion: I think it would be better to transition some non-critical services first, to gain some experience without breaking anything. If this works for 10+ regularly used non-critical e.g. webapps, then transition everything to this approach.

d-goldin: I will fix the wording on that one. Parts of those section are maybe still a bit too "note to self"-like. What I intended was not for "inclusion into upstream" but more to hypotehtically verify this is covering all the most important aspects needed for the most critical services. Your version is definitely better for an actual slow inclusion into upstream step.

rfcs/0058-secrets-for-services.md
d-goldin: In the current shape, no such management would happen. The secrets would just remain in the secrets store and if a config author decides to re-use them, then so be it. I can include that. I was initially thinking of the drawbacks of "sth that would be worse when adding this", but I guess it depends on the reference point. Definitely good points to add, even if they would remain unhandled.

rfcs/0058-secrets-for-services.md
d-goldin: Yeah, maybe I was being a bit over-specific there - my intention was to mention the indirection/dispatch by secret store type, but it reads a bit confusingly. Let's simplify that for now.

rfcs/0058-secrets-for-services.md
d-goldin: I'll resolve this for now, since I accepted the simplified suggestion below. If it turns out to be confusing, I'll rephrase. It rather obvious from the example code though.

rfcs/0058-secrets-for-services.md
d-goldin: For this we'd have to look at some metadata of the secrets, like a digest or at least mtime. Right now this is not done. This would be probably not needed if the consensus ends up at automatic restarts (unless we add a flag to disable automatic restarts for some cases).

rfcs/0058-secrets-for-services.md
d-goldin: This very much depends on the service. If I'm dealing with a very heavy service, like for instance an ElasticSearch node, where a restart can cause re-replications or similar it can be very useful to tightly coordinate restarts in a larger setup. So I'm sympathetic to the average case where automatic restart is nice and stress-less but I think if we do it, we should make it configurable to.

Ericson2314:
The detailed design is quite long. I think it's better to clean up the Nix secrets story before moving downstream, and I suspect trying to do everything at once will leave us stuck.

rfcs/0058-secrets-for-services.md
turion: Well, it's maybe a strawman, but if you put a secret in `/nix/store`, it's going to vanish upon garbage collection when I uninstall the package. If I don't care whether it's readable by root or by all users (e.g. inside a container), this is a drawback now because no such garbage collection would take place.

As in "something that would be worse when adding this", you could also say that people possibly expect that NixOS doesn't do any state management by default and that downgrading will bring their machine back into the exact state, but that's not the case anymore.

turion: > you could also say that people possibly expect that NixOS doesn't do any state management by default

The difference being that before this RFC, I'd manually handle the secret state and remember to handle it. After all, I'm deleting the line saying `service.foo.secretsFile = "/foo/bar";`, and when I delete the line, I'll delete the file. This is now not the case anymore. Instead, I need to remember how this semiautomatic secret handling works.

globin: > if you put a secret in `/nix/store`
It is exactly the purpose of this RFC not to have secrets in the nix store. This is possible without any problems and if they are referenced in the nixos config they would not be garbage collected as currently is the case. And this can still be achieved with this proposal as before.

> If I don't care whether it's readable by root or by all users (e.g. inside a container)

Actually currently the nix store of (nixos) containers is shared with the host and all other containers, so that might be an issue nonetheless.

With regard to secret state handling, a lot of services have the possibility do use a `passwordFile`, `secretFile`, etc. but there is no abstraction for handing out permissions to secrets correctly, which is something like the proposal in this RFC is needed. Currently you have to not only drop the files on the machine yourself in the correct destination (same with this RFC, as long as you are using a fetcher that requires that), but also set the correct permissions for the respective service, which might be a chicken-and-egg problem of adding users for service by enabling it and the service requiring the secret.

globin:
@Ericson2314 I think this proposal is orthogonal to the secrets-in-nix-store issue and that this is quite a nice way to move forward without getting in the way of that, as this RFC abstracts over where the secrets are fetched from. In the case of nix store secrets being implement one could easily add a fetcher to use that, but also keep the abstraction in place for remote secret management systems if for some reason you are required to use that.

But obviously I'd be very happy if #5 or similar would be continued!

globin:
> Side note: This approach does not prevent one service from reading the secrets of another, e.g. through privilege escalation or remote code execution. But I think that issue is beyond the scope of this RFC, and should better be handled with something like containers

@turion I'm not sure how you mean this? This as far as I can see is possible with this RFC as you can also use non-FS fetcher functions. And obviously always depends on how far you want to go with isolation.

turion:
> > Side note:

It was really just a comment that all secrets are still root-readable, and that this RFC is orthogonal to more isolation/virtualization.

aanderse:
From reading this over it seems once a module uses this proposal there is no opt out for the user, correct?

If say the drawback is that this adds more complexity to the system. When I was new to nix I always enjoyed being able to read a modules source and usually understand what is happening... translate things into what a regular distro does. This is an entirely new abstraction that has no equivalent in a regular distro and after reading through it twice I still feel very uncomfortable thinking about how I can explain/justify this mechanism to my co-workers who manage nixos servers with me, but have not dived into nixos enough to understand all the nuances yet. I foresee having to answer many questions to puzzled looks from people wondering why we can't continue to keep doing what we're doing, and wondering why, as non "nix people", they no longer understand module source code.

I also wonder if a product similar to vault should be put in the alternatives section. I'm under the impression that while this may not be an entirely solved problem, there are existing solutions out there that can fit.

globin:
> I also wonder if a product similar to vault should be put in the alternatives section. I'm under the impression that while this may not be an entirely solved problem, there are existing solutions out there that can fit.

I think that this rather allows us to consistently use something like vault or some otherwise created secrets folder etc. and not having to depend on every module implementing support for it. It does create a further abstraction but by that allows all modules to have a consistent way of passing secrets and not a different implementation for each of them.

aanderse:
@globin I meant a single solution like vault could be chosen by the sysadmin for every secret on a machine and then module code could continue as is, unmodified. This provides greater flexibility, though currently at the cost of upfront work for the sysadmin. That being said I concede that the open issues regarding secrets in the nix store have been open for years without resolution, and no one has provided a nice to use module for something like vault that integrates nicely into nixos yet.

I guess some sort of system/module that keeps this logic out of the module system would have been my desired outcome.

I hope my opinion has provided at least *minimal* value as a differing point of view.

globin:
The thing is that is not possibly currently.

For all modules that only allow passing secrets by string (secret content) you'd have to modify it to pass in the secret differently as it would otherwise be added to the nix store nonetheless.

This change should happen anyway even without this proposal and a lot of modules now at least allow passing by file, where one can choose to use a non-nix-store-file with more specific permissions or just a `pkgs.writeText` in one does not care. For the ability to use something like vault one would still have to create new services, similar to the proposal here. That's why I think it would be nice to have an abstraction so that we can standardise all services handling secrets to have a consistent interface.

aanderse:
@globin I entirely agree that all modules need to replace any `password` options with `passwordFile` (or the like) regardless of this RFC, and that is something I (along with many others) have slowly been working toward. I also agree that `pkgs.writeText` in conjunction with `passwordFile` is the proper justification for entirely eliminating any `password` like options, like I proposed when I removed the `dbPassword` option from the `zabbixServer` module.

d-goldin:
> From reading this over it seems once a module uses this proposal there is no opt out for the user, correct?

Do you mean for a module author or a user? If a module is switched to this, then no, without "more involved" fiddling about a user wouldn't be able to change that unless author has created multiple avenues. We could of-course add a "secret store" that is just the nix store as some sort of fallback for scenarios where people don't care and they can at least go and add their secrets to the nix store as before.

> If say the drawback is that this adds more complexity to the system. [...]. I foresee having to answer many questions to puzzled looks from people wondering why we can't continue to keep doing what we're doing, and wondering why, as non "nix people", they no longer understand module source code.

This is of course different perceptions, but for me NixOS is already quite an opinionated distribution that does a lot of things quite differently and brings a few abstractions that no other distro has.
Imho pretty much anything new of this sort will have the issue of change/surprise/added complexity.

I wonder, why does this particular code/interface seem more complex than for instance all the module evaluation and merging logic in https://github.com/NixOS/nixpkgs/blob/master/lib/modules.nix? It is of course possible to simplify and document to help people understand it more easily, once we understand what points of confusion exist.

> I also wonder if a product similar to vault should be put in the alternatives section. I'm under the impression that while this may not be an entirely solved problem, there are existing solutions out there that can fit.

Well, yes. I thought of this as a given, that people can still go and use vault directly if they want/their software supports it. If all things we are dealing with would support sth like vault, then it would be great. But as @globin said, we'd still want the ability to plug in another service to manage secrets if need be, so some abstraction would likely emerge.

I will keep the points in mind when I do the next pass.

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+  `/etc/secrets`, only accessible to root
+* A *fetcher* function: a function whose task it is to resolve the secret
+  identifier, retrieve the secret and place it in the service process' private
+  namespace within `/tmp` name
+* Simple helper functions to *enrich* expressions defining systemd services
+  with secrets
+* "Side-car" service: A privileged systemd service running the fetcher
+  function to retrieve the secret, and initially create the service
+  namespace
+* Secrets scope: provides a context in swhich secrets are accessible as
```
Lassulus:
```suggestion
* Secrets scope: provides a context in which secrets are accessible as
```

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+handle secrets more consistently while being able to integrate pre-existing solutions
+like NixOps, or a simple secrets folder.
+
+# Motivation
+[motivation]: #motivation
+
+There is currently a lack of consistent and safe mechanisms to make secrets
+available to systemd services in NixOS. Various modules implement it in various
+ways across the ecosystem. There have also been ideas like adjustments to the
+Nix Store (like issue https://github.com/NixOS/nixpkgs/issues/8), which
```
edolstra: This should be https://github.com/NixOS/nix/issues/8.

d-goldin: Whoops. Thanks!

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+* *Secrets store*: a secure file-system based location, in this document
+  `/etc/secrets`, only accessible to root
+* A *fetcher* function: a function whose task it is to resolve the secret
+  identifier, retrieve the secret and place it in the service process' private
+  namespace within `/tmp` name
+* Simple helper functions to *enrich* expressions defining systemd services
+  with secrets
+* "Side-car" service: A privileged systemd service running the fetcher
+  function to retrieve the secret, and initially create the service
+  namespace
```
edolstra: It's not clear to me why the helper unit is needed. Can't the keys be fetched using an `ExecStartPre=+...` command?

More generally, instead of creating our own mechanism for passing keys to services, maybe the kernel keyring mechanism can be used for this? Units would call `keyctl request/search/read/...` to fetch keys. These keys would either be preloaded into the keyring or produced on demand using the [`request-key` program](https://linux.die.net/man/5/request-key.conf).

d-goldin: I briefly tried `ExecStartPre`, but unfortunately it does not seem to run within the mount namespace, so it doesn't have a access to the `PrivateTmp` we want. I am not sure if this is intentional or not. 

Regarding kernel keyring mechanism - I agree, it might be a good default backend. The directory based thing was just the dumbest proof-of-concept case I came up with (given that similar approaches are used in nixops and krops/stockholm). Part of the intention is to have a somewhat agnostic interface.

rfcs/0058-secrets-for-services.md
```diff
@@ -0,0 +1,273 @@
...
+in
+   systemd.services = secretsScope ({ secret1, ... }: {
+      foo = {
+        description = "Simple test service using a secret";
+        serviceConfig = {
+          ExecStart = "${pkgs.coreutils}/bin/cat ${secret1}";
+          DynamicUser = true;
+        };
+      };
+    };
```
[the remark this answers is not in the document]
d-goldin: > This seems unnecessarily verbose. I would do something like this:
I agree, it's more compact. Why I initially decided against it and for something that modifies the resulting structure separately was to reduce changes to the existing systemd modules. At the same time I thought it would be nice to be able to get the secrets as arguments, which I thought made it nicer to deal with in code. In the `needsSecrets` case, if I understand it correctly, the user would be requesting a secret by its name, like in the scope creation, but then would have to possibly deal with paths (as strings) to pass the secret as an argument to the service, or load it into an env-var, because there would not be an automatic mapping mechanism anymore. Unless we pack it up into a shell variable or so.

> It also avoids imperative-sounding function names like mkSecretsScope which suggest that they allocate a unique new scope, but that's not the case, e.g. ...

I do not necessarily perceive `mk*` as an imperative terminology, but that depends on the reader. We have a lot of pure `mk*` functions that construct some structure. But I'm not at all attached to the naming here, so we can change it to whatever seems more suitable if they're still around further down the road.

edolstra:
This RFC is now open for shepherd nominations. Any takers? @aanderse @globin Would you be willing to shepherd this?

domenkozar:
I nominate @dhess and @fpletz

aanderse:
@edolstra thanks for the nomination. I would be able to take this on conditional to the team agreeing to a 30 minute limit on meetings.

rfcs/0058-secrets-for-services.md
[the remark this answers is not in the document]
d-goldin: @edolstra: Did this sufficiently address your question? I'm not super familiar with the inner workings of the process, so for now I just left those as discussion comments here, but I'm willing to incorporate some of the things pointed out into "alternatives" or the core section, if there is some consensus around that.

Mic92:
I would like to nominate @Lassulus, maintainer of krops.

dhess:
I'm grateful for the nomination as shepherd. However, I don't think I meet the requirements as stated by the RFC process, namely: "This team should be people who are very familiar with the main components touched by the RFC." 

I think the project would be better served by someone who has, e.g., hacked on NixOps' key distribution code, or worked with the Linux kernel's `keyrings` facility, both of which seem very relevant to the goal of this RFC.

How about this: I'll agree to shepherd the RFC, despite my lack of expertise, if a better candidate doesn't step up in a reasonable timeframe.

rfcs/0058-secrets-for-services.md
arianvp: Kernel keyring mechanism sounds overkill to me.  Its usecase is not to communicate files between userland processes, but between userland and kernel drivers.  Files are a perfectly sufficient abstraction for passing secrets around in userland.

arianvp:
I'm not sure if this RFC actually solves the problem it's trying to solve.



In the case documented in this PR, the `fetcher` is responsible for copying the file from `/run/secrets` into the the service's mount namespace.  But the only way I see this working is either have the secret world-readable (in which case the service might as well read it from `/run/secrets` directly, as the service has access to that anyway),  or to statically allocate a `uid` and `guid` for the user, and `chown` the secret to that user; but in that case `DynamicUser=` will disable itself automatically and you lose its 'benefits'

> Note that if User= is specified and the static group with the name exists, then it is required that the static user with the name already exists. Similarly, if Group= is specified and the static user with the name exists, then it is required that the static group with the name already exists.


Reason is, Because the  sidecar-service has no idea what the `uid` and `guid` of the `DynamicUser` is going to be apriori (because they're dynamically allocated on service startup), it can not `chown` the secret in the service's private `/tmp` to the right permissions to make the secret readable by the service.  To mitigate this problem you can either make it world-readable (in which the service can just directly read from `/run/secrets`)  or you must allocate the `uid` upfront (in which the `DynamicUser=` mechanism does not allocate a `uid` automatically anymore)

If a service wants to access secretse, I advice using a statically allocated user and/or group, and then use filesystem permissions to scope the secret to the specific service. This is what filesystem permissions are for.   `/run/secrets/mykey.key` can then be owned by the `nginx` group, making sure that only `nginx` can access it.

We can still keep `DynamicUser=true` as it will still default to a whole bunch of useful isolation features even if the `uid` for the provided `User=` is statically allocated; but I think both using dynamic `uid` allocation and getting the secret into the container with the right permissions at the same time is impossible.

domenkozar:
@dhess I'd like you to reconsider, not everyone is expected to know everything. We need people with security mindset around.

@Lassulus, @globin and @aanderse are accepted.

We need the leader - do you mind being one @aanderse (30min cap is perfect).

And final word from @dhess.

dhess:
@domenkozar OK, I accept!

flokli:
Possibly relevant: https://github.com/systemd/systemd/issues/14264

aanderse:
@domenkozar sounds good. I'll wait to hear back from @d-goldin, once the outstanding comments have been addressed.

d-goldin:
@aanderse: Which comments in particular would you like to have addressed?

aanderse:
Oh I wasn't sure if you wanted to address the most recent comment by @arianvp or not. I'm happy to schedule a call whenever you are ready. Just give the word.

d-goldin:
@aanderse: Alright. I'll ask for some clarification on that one and we can probably get in next week or so.

d-goldin:
@arianvp: I tried to address a few parts of your argument, but please clarify further where necessary. The core suggestion I think I'm seeing here is to not do any of this and leave things as they are with static service users and somehow set up permissions, but I don't really see how one hurts the other and why just static users is better than static users+additional solutions. This "classical" approach is by the way listed in alternatives.

> I'm not sure if this RFC actually solves the problem it's trying to solve. But please correct me if I am wrong.

There are roughly two things this RFC tries to propose:
* An API for how secrets could be described and referenced in Nix for use with systemd services
* Rough mechanics of how those secrets can be made available to services

Does it solve neither?

> In the case documented in this PR, the `fetcher` is responsible for copying the file from `/run/secrets` into the the service's mount namespace. But the only way I see this working is either have the secret world-readable (in which case the service might as well read it from `/run/secrets` directly, as the service has access to that anyway)

Right now the "store" is described like this: 
> Secrets store: a secure file-system based location, in this document /etc/secrets, only accessible to root. 

It is mostly assumed that the service itself, running as a service user or _especially_ `DynamicUser` does not have access to this store directly. 

In case of a known service user, this is possible to solve by the config author ensuring correct permissions (likely manually?), but as you correctly noticed further below, not in the case of `DynamicUser` as things stand right now (unless systemd solves this for us, like in the systemd issue linked by @flokli). 

The fetcher runs as a privileged user though and has no issues accessing the store, which is the whole purpose of the side-car service (be it a folder as a secrets store, or vault).

> [...]
> Reason is, Because the sidecar-service has no idea what the `uid` and `guid` of the `DynamicUser` is going to be apriori (because they're dynamically allocated on service startup), it can not `chown` the secret in the service's private `/tmp` to the right permissions to make the secret readable by the service. To mitigate this problem you can either make it world-readable (in which the service can just directly read from `/run/secrets`) or you must allocate the `uid` upfront (in which the `DynamicUser=` mechanism does not allocate a `uid` automatically anymore)

Maybe this aspect in the RFC doc itself is not sufficiently spelled out yet, but DynamicUser is specifically called out as something that this aims to solve, like:

> With the introduction of Systemd's DynamicUser, the more traditional approaches of manually managing permissions of some out-of-store files could become cumbersome or slow down the adoption of DynamicUser and other sandboxing features throughout the nixpkgs modules.

The simple POC implementation of the fetcher does set the secret to world-readable *within* the services private mount namespace (https://github.com/d-goldin/nix-svc-secrets/blob/master/secretslib.nix#L19) to deal with the `DynamicUser` part. And one key advantage here is that less additional out-of-config state such as permissions needs to be kept consistent with the config by the user because the fetchers can make them accessible based on nix config and nothing else.

What I don't seem to understand in this argument is why you equate world-readable within the private mount namespace with the ability to read from the secrets store directly (`/run/secrets`, as you call it here).

> If a service wants to access secretse, I advice using a statically allocated user and/or group, and then use filesystem permissions to scope the secret to the specific service. This is what filesystem permissions are for. `/run/secrets/mykey.key` can then be owned by the `nginx` group, making sure that only `nginx` can access it.

While there is not much on the case of static users + secrets accessible with that user in the RFC right now, I think it should still map well enough interface wise but would require adjustments in how the solution _resolves_ the location of the secret for such a service, which could be done rather transparently.

Which specific adjustments do you think are necessary to address your concerns, or do you generally think that there is no problem to be solved here?

I do see the need to be a bit more explicit in the document about how the dynamic UID/GUID is handled as it currently relies on the reader to actually look into the POC, but before I make further changes, I'd like to see this clarified first.

dhess:
My role here as shepherd will be to play the naïve user, asking the dumb questions.

First off, I'm unfamiliar with `DynamicUser`. I infer from reading this RFC that `DynamicUser` is important to the future of NixOS module design. Can someone briefly explain to me what benefits `DynamicUser` provides for NixOS module design, and why `DynamicUser` doesn't work well with the current best-practices approach for out-of-store secrets, as exemplified by NixOps?

Also, I infer from reading this RFC that it exists primarily because of the needs of `DynamicUser` services; but in my opinion, there are some comparatively straightforward features that a robust `systemd` secrets service should also provide, which are either not addressed by this RFC, or are mentioned only in passing. For example, many services need not be given read access to any secrets-containing files at all, if only NixOS's `systemd` facilities had a standard, automated way to set environment variables containing those secrets in the target process's environment -- consider, for example, a process that requires AWS credentials. This mechanism would work irrespective of `DynamicUser`, so long as the application can read secrets from environment variables.

That's just one example, but there are several others I can think of just off the top of my head. There are some real low-hanging fruit with respect to the handling of secrets in NixOS. It would be a shame not to address at least some of those features in this RFC, in addition to whatever is required by `DynamicUser`. Perhaps those other features are out of scope for this particular RFC. If so, that's fair, but in that case, I think that the RFC should be re-titled to indicate that it is specifically addressing the needs of `DynamicUser` modules.

d-goldin:
> First off, I'm unfamiliar with `DynamicUser`. I infer from reading this RFC that `DynamicUser` is important to the future of NixOS module design.

I did not necessarily want to evoke this impression, but it's a useful feature that currently is a bit hard to use for things that also require access to secrets, which often are likely managed using regular permissions, which then stops working. It should also not be read as "this is the only way things should be done from now on" and rather as an attempt to provide some additional, optional tooling.

> Can someone briefly explain to me what benefits `DynamicUser` provides for NixOS module design, and why `DynamicUser` doesn't work well with the current best-practices approach for out-of-store secrets, as exemplified by NixOps?

* Removes need to explicitly manage users and groups for a service. This is nice not only because it saves some typing in the module definition, but because it also avoids problems such as UID/GID re-use.
* Additional sandboxing, such as mostly read-only file-system hierarchy
* Private, cleaned up /tmp
* Cleanup of IPC objects
 
As was pointed out by @arianvp, it is possible to achieve similar isolation using static users by adding a few settings (see link below). In this case we are back to the usual things that need to happen outside of nix: ship the secrets, ensure correct permissions, reference secrets loosely as path-strings which is imho a gap that would be nice to narrow. I have also some hope that it should be possible to provide an abstraction that could allow providing secrets from various stores, such as vault, to services that do not have direct support.

Why DynamicUser is mentioned in the RFC so much is because it seems to me, that if we solve this for DynamicUser, the solution should also work for static user scenarios with the added benefit of avoiding additional out-of-config permission management and hopefully adding some representation to secrets in the config that is more than just paths-as-strings.

Here is a more exhaustive write-up by Poettering himself on the motivations behind DynamicUser; http://0pointer.net/blog/dynamic-users-with-systemd.html

> Also, I infer from reading this RFC that it exists primarily because of the needs of `DynamicUser` services; 

As mentioned above, I don't think this should be the case. It also exists because there is perpetual confusion about how to manage secrets, how to use them in modules, how to avoid them being stored in the nix-store which leaves NIxOS with a usability gap and some footguns in that regard.

> but in my opinion, there are some comparatively straightforward features that a robust `systemd` secrets service should also provide, which are either not addressed by this RFC, or are mentioned only in passing. For example, many services need not be given read access to any secrets-containing files at all, if only NixOS's `systemd` facilities had a standard, automated way to set environment variables containing those secrets in the target process's environment -- consider, for example, a process that requires AWS credentials. This mechanism would work irrespective of `DynamicUser`, so long as the application can read secrets from environment variables.

Systemd does have `EnvironmentFile` which does exactly this. It also runs with systemds privileges, so it can load files the service user does not directly have permissions to access (but again, something needs to make the secret accessible to the service user, be it systemd in this case or a side-car service in the other case). This is nice, as long as a prepared environment-file can be provided. While nice enough, as a mechanism it has a few shortcomings though, one of which is environment variable size limits, cumbersome escaping being another.

Further, If I'm not mistaken, it's right now not possible to provide multiple EnvironmentFile directives in a NixOS systemd service config, which is problematic if there is need to combine entries from user-supplied data and module settings which makes this a bit cumbersome too, but it's an issue that could be fixed independently (we have no way to pass a secret as a string in a "safe" way, so if we pass a path, we're back to the need of being able to load up that path as the service user).

The POC linked in the RFC has an [ugly example](https://github.com/d-goldin/nix-svc-secrets/blob/master/secrets-test.nix#L37) of loading a secret file into the environment, so its not entirely unmentioned. Nicer mechanics for this would be nice though, such as with some wrapper or similar. 

Why I opted to make the ugly example and not just use EnvironmentFile is because supporting X different kinda-working-for-some-cases approaches was not something I was aiming for. As soon as a file can be made accessible to the service process most ways of passings secrets such as environment variables, CLI arguments, config-file templates should be dealable with and require fewer moving parts in processing of the service definition and require a less complex API. 

Ultimately, a module author should still be able choose which mechanism is the easiest to use.

> That's just one example, but there are several others I can think of just off the top of my head. There are some real low-hanging fruit with respect to the handling of secrets in NixOS.

Which additional cases are you thinking of that are still unexplained/unaddressed? A list would be generally useful.

While I do think we should address as many low hanging fruits as possible, i don't think we should support every possible mechanism under the sun for maintainability and complexity reasons. 

Hope that clarifies a bit.

7c6f434c:
> It should also not be read as "this is the only way things should be done from now on" and rather as an attempt to provide some additional, optional tooling.

Note, though, that by the (accepted) RFC #52 DynamicUser seems to be advised as the default solution whenever feasible.

globin:
Note that a rather annoying bug in systemd concerning DynamicUser is fixed in the next release (https://github.com/systemd/systemd/pull/14532).

Mic92:
I know that encryption is not a goal yet of the current RFC but I would like to propose a concept that would unify the secret management between all deployment solutions and just make it part of NixOS itself.

1. Encrypt the secret with a public key encryption like gpg (with pass as a frontend)
2. Add it to the nix store during the normal build process
3. During activation phase the key gets decrypted by the secret service using the key stored on the machine.

This approach has the following advantages: 

- We know when we have to restart services because the encrypted secret changed
- We can encrypt secrets for multiple both the users and the servers that should read them -> this also makes it easy to manage them with a VCS (i.e. git)
- We no longer need to implement an extra step of deploying secrets in each deploy tool, which makes it likely faster and less error prone
- Rollbacks now will correctly and pick up old version of secrets from nix store

flokli:
@Mic92 this would however make secrets part of your system configuration, so rotating these secrets regularily becomes complicated or impossible, and going back to older generations will also rollback to (expired) secrets. See @zimbatm's comment https://github.com/NixOS/nixpkgs/issues/24288#issuecomment-375106708 which goes more into detail about why they should not be configuration.

Mic92:
@flokli rolling back to old secret is exactly my use case. I am not an enterprise user using vault and for my small deployment I want to store my secrets in the git. However I could also imagine to extend the concept above to support taking a key name as the `secret` that would instead of decrypting at runtime from the nix store to download it from vault.

rfcs/0058-secrets-for-services.md
flokli: I have mixed feelings about this.

Using files usually implies having to worry about ACLs and who's allowed to access them. We can cheat by mounting them in a private namespace, but it's still a bit cumbersome.
Sometimes you want to have "use once" properties and provide a new key on every read / issue tokens on access etc. We could cheat again by providing these key files by a fuse filesystem, but then it just gets more complicated.

The kernel keyring might be a good abstraction over all this, it's just not widely adapted currently and lacking real-world usage. Reading [`keyrings (7)`](http://man7.org/linux/man-pages/man7/keyrings.7.html) looks promising. In addition to thread/process/session, there's also an `upcall` feature, bouncing back to userspace to request secrets which could be a request to whatever credentials provider is used.

I'd love to experiment with that a bit, or see some real-world examples. Anybody aware of these?

poettering: kernel keyring is very much intended for userspace keys too. See the kerberos stuff that has been ported to use it for that, or systemd's cryptsetup.

I think the kernel keyring has deficiencies (upcalls, yuck! also no namespacing for containers, …), but it probably is the right approach in the long run.

d-goldin:
@Mic92: The retrieval/use part (point 3) should be possible using some fetcher implementation that can decrypt the secrets from the store (this would of course still require managing the decryption key somewhere, which itself would not be able to reside in the nix store in most cases).  

As for point 1 and 2, those are so far explicitly out of scope as originally proposed. Could you imagine a solution that integrates well but is separate? If it would have to be included, what could a concrete implementation look like that fits in with the rest?

Mic92:
> @Mic92: The retrieval/use part (point 3) should be possible using some fetcher implementation that can decrypt the secrets from the store (this would of course still require managing the decryption key somewhere, which itself would not be able to reside in the nix store in most cases).
> 
> As for point 1 and 2, those are so far explicitly out of scope as originally proposed. Could you imagine a solution that integrates well but is separate? If it would have to be included, what could a concrete implementation look like that fits in with the rest?

I suppose it can be implemented with the current concept. It would be great if the fetcher would be not hard-coded though.

Mic92:
If a service fails to start for some reason what I currently sometimes do is just executing the parts manually in a shell, maybe in combination with some debugging tools (strace etc). When there are additional services added to decrypt secrets to a tmpfs, how would my debugging workflow looks like? Can I still debug my services?

Mic92:
@aanderse Did this RFC already had a meeting?

aanderse:
@Mic92 not yet. There had been a fair bit of discussion which seemed like a reason to postpone, but that seems to have settled so scheduling something now seems like a good idea.

Does anyone have some time in the coming weeks?

d-goldin:
@aanderse: Who should be present and what is usually the content of such a call?

aanderse:
The content of the call depends on the status of the RFC. The shepherd team and yourself should be present on the call.

My feeling is that the first call for this RFC would start with assessing the status of this RFC and the discussion in this comment thread up to this point. As part of clarifying the status of this RFC I believe we would attempt to determine if the discussion this far has resolved many of the questions/concerns raised this far, and then discuss next steps for this RFC.

d-goldin:
@aanderse: Sounds good. I can surely find some time. Should we set up a doodle?

aanderse:
Sounds good. Do you mind polling everyone with that?

d-goldin:
@aanderse @globin @Lassulus @dhess: https://doodle.com/poll/n4r4i9afb824izty (CET time) - if you need additional options, let me know. I will fill it out later, but most of the suggested timeslots should work for me.

d-goldin:
@aanderse: I think @globin might be vacationing for some while. Should we wait? I could extend the doodle to include to include a few more days later in the month (seems like wednesdays 19-20:00 could be a good timeslot)

aanderse:
Has everyone else responded and we have a common time? If so I think we should go ahead. The call is mostly a catchup to figure out the status, etc... We will summarize the call anyways.

d-goldin:
@aanderse: Yep. 2020-03-18 7-8pm CET it is.

dhess:
Hi, sorry but I've never participated in a Nix RFC call before. What's the procedure? Maybe I'm missing something, but I don't see any details on the Doodle page posted above.

Lassulus:
I also never participated in one. As far as I understand it, we will try to agree on a platform/software and make a conference call where we talk about the current state of this RFC and how to continue.

dhess:
As I write this, I believe the call is scheduled for less than 3 hours from now, so we've left it a bit late to agree on a conferencing system :)

dhess:
Also, I only now just noticed that I'm listed as the shepherd leader, which was quite surprising to me! I hope this doesn't mean it was my responsibility to set up the conference call!

aanderse:
I was just going to throw up a [jitsi](https://meet.jit.si/) meeting a few minutes before the call... :man_shrugging:

aanderse:
Here is a room I setup: https://meet.jit.si/AngryApplesGlowHigh

d-goldin:
Thanks for the call. There likely will be some meeting notes, but my core takeaways for now are roughly the following two points:

  * PoC implementation of a second alternative backend (like vault or keyring or ...)
  * Injection of secrets into process environment

I'll also do an editing pass to clarify/improve on some parts of the text that came up as unclear or lacking.

And then we'll see.

dhess:
Here's a brief log of our call this evening:

- @d-goldin feels there is not yet a lot of explicit support for this RFC, and was hoping for more activity in this GitHub discussion.

- `DynamicUser` is now officially accepted best practice for NixOS (cf., https://github.com/NixOS/rfcs/pull/52), so any secrets service adopted by NixOS needs to support it.

- @dhess asked that @d-goldin demonstrate how the API proposed here could accommodate another backend (e.g., Vault, `keyring`) and/or another process-visible secrets mechanism (e.g., environment variables), so that it were more clear what value this RFC provides vs simply porting NixOps's `deployment.keys` filesystem bits to NixOS.

- @aanderse pointed out that `mkaito` in IRC [1] had mentioned a working Vault secrets implementation [2] that might be useful in prototyping something for this RFC. 

- @d-goldin would like to find an interested and motivated co-author to help drive this.

- We'll consider another call in a month to check progress.

[1] https://logs.nix.samueldr.com/nixos-dev/2020-03-04#3133786
[2] https://gist.github.com/mkaito/ce20db29ed9c7285808c935597232bd8

Mic92:
@dhess it might be good to already schedule the next meeting since it is already half-way a new month.

dhess:
@d-goldin Are you ready for another meeting in mid-April?

d-goldin:
@dhess: Yep. I did some work on the 2nd integration and am currently working on triggering restarts when I find time. Though it might be best to leave at least a week between me pushing the update and the meeting just to leave some time for potential discussion, so maybe we should try to find a slot mid april and one a week later to fallback to.

d-goldin:
I pushed some updates to the proof of [concept repo](https://github.com/d-goldin/nix-svc-secrets).
It should now be a bit more eligible and contains the following improvements:

* Vault backend
* Reloading for vault and folder based backends
* Somewhat improved configurability of the backends (system level config and service level config overrides)
* Should be somewhat more legible than before (readme, some comments, a bit tidier)
* Very basic example of an activationScript to check basic settings

I will go through and adjust the RFC doc to reflect those changes today/tomorrow.

rfcs/0058-secrets-for-services.md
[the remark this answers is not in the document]
d-goldin: So, getting back to this suggestion - if somebody could point me to the simplest way of implementing it with such an additional field on any systemd service I'm willing to try and adjust the poc to see how that works. One thing I do kinda like though is the ability to have some at least very basic checking of secrets used, which wouldn't work the same way with just strings.

dhess:
OK, here are some suggested meeting times that work for me:

https://doodle.com/poll/wip9wcz226k8r24r

dhess:
@Lassulus Can you please take a look at the Doodle calendar and reply with the ones that work for you? If none work for you, let me know here. Thanks!

dhess:
While investigating some unrelated `systemd` things this past weekend, I ran across something in the `Environment=` section of the `systemd.exec(5)` man page that gave me pause:

> Note that environment variables are not suitable for passing secrets (such as passwords, key material, …) to service processes. Environment variables set for a unit are exposed to unprivileged clients via D-Bus IPC, and generally not understood as being data that requires protection.

ref: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Environment=

That was certainly news to me! I don't use Linux on the desktop, so I don't have the pleasure of using D-Bus very often, and I don't feel qualified to verify this claim; but assuming it's true, and also that I'm interpreting this statement correctly, that would mean that we can't recommend (and shouldn't implement) an environment variable-based mechanism for this RFC.

Can someone who knows how to spelunk the depths of D-Bus IPC please try to reproduce this behavior? What I'd like to know is, can an unprivileged process owned by userid `n` inspect the environment of a `systemd` unit owned by userid `m`, for `m != n`?

dhess:
OK, the next meeting will be on Wed 6 May from 5-6pm UTC. I'll create a Jitsi meeting when the time comes. Thanks to everyone for responding.

flokli:
> Can someone who knows how to spelunk the depths of D-Bus IPC please try to reproduce this behavior? What I'd like to know is, can an unprivileged process owned by userid n inspect the environment of a systemd unit owned by userid m, for m != n?

```
root@tp:~|⇒  systemd-run --pty --property=Type=oneshot --property=DynamicUser=yes --property=User=guest systemctl cat accounts-daemon.service| grep Environment
Running as unit: run-u94833.service
Press ^] three times within 1s to disconnect TTY.
Environment=GVFS_DISABLE_FUSE=1
Environment=GIO_USE_VFS=local
Environment=GVFS_REMOTE_VOLUME_MONITOR_IGNORE=1
Environment="LOCALE_ARCHIVE=/nix/store/9b725cly2a6a61vb8bgz7cyr0xr8y2av-glibc-locales-2.30/lib/locale/locale-archive"
Environment="PATH=/nix/store/ca9mkrf8sa8md8pv61jslhcnfk9mmg4p-coreutils-8.31/bin:/nix/store/hg3albf7g05ljfqrfjhd58rblimrp6ph-findutils-4.7.0/bin:/nix/store/8pajzfyqx1v7dz1znrnrc4pqj5rmnx24-gnugrep-3.4/bin:/nix/store/jpqlmf3wqg281j8fdz50kjl525pfsxjc-gnused-4.8/bin:/nix/store/zbxfs37qjj6ddrfnzrdnxnkrvvm1ddsf-systemd-245.3/bin:/nix/store/ca9mkrf8sa8md8pv61jslhcnfk9mmg4p-coreutils-8.31/sbin:/nix/store/hg3albf7g05ljfqrfjhd58rblimrp6ph-findutils-4.7.0/sbin:/nix/store/8pajzfyqx1v7dz1znrnrc4pqj5rmnx24-gnugrep-3.4/sbin:/nix/store/jpqlmf3wqg281j8fdz50kjl525pfsxjc-gnused-4.8/sbin:/nix/store/zbxfs37qjj6ddrfnzrdnxnkrvvm1ddsf-systemd-245.3/sbin"
Environment="TZDIR=/nix/store/wmry9mqmimq8ib8ijli4g1yx92gxjli5-tzdata-2019c/share/zoneinfo"
Environment="XDG_DATA_DIRS=/nix/store/89gmv6228zfl0lhmg1qaj36i5vz9wcl8-system-path/share"
```

dhess:
@flokli Thanks for that, but is that what's meant in the `systemd.exec(5)` man page by "exposed to unprivileged clients via D-Bus IPC"? What you've demonstrated there is equivalent to reading the contents of the service's `.service` file, which anybody can do without even using `systemd` since the files are kept in the Nix store, of course.

I took the warning in the man page to mean any/all environment variables in the running service's environment, not just the ones set by the `Environment=` directive in the `.service` file. Am I misunderstanding?

lheckemann:
@dhess I believe so — "environment variables set for a unit" suggests that this is for the unit files in systemd, not the processes running as part of the unit — so setting environment variables for the services is safe, just not in the units.

dhess:
@lheckemann OK, thanks for clarifying that! 

Nothing to worry about here, then.

flokli:
I poked around in the DBUS interface, and seems I cannot read the environment variables set through `EnvironmentFile` - but I definitely can read all the properties of all units.

dhess:
The previous meeting link still works, so let's use that again:

https://meet.jit.si/AngryApplesGlowHigh

poettering:
BTW, I thought about a creds concept in systemd the past weeks. It's a big topic, but this is what I'd probably propose in systemd:

1. Introduce a new setting per service setting `LoadCredential=<name>:<path>` for loading a credential off disk. The name is some user chosen ID, the path is some file system path to load the credential from.
2. Similar, add `PassCredential=<newname>:<oldname>` which takes a credential systemd itself received (think: pid1 in a container gets some creds from the container manager, or systemd --user from its parent). 
3. Similar, add `AskCredential=<name>` for querying a credential from the user via the `systemd-ask-password` framework.
4. Then, add `CredentialMode=` for configuring how to pass credentials to the service. Can be a combination of options: `file` (pass as file in the fs), `keyring` (for kernel keyring), `fd` (for passing as fd via socket activation). We'd default to `file` and `keyring`. We should be graceful here, i.e. if `keyring` cannot work (because we are in a container where kernel keyring is blocked), use only the fs. We'd always attempt to pass all configured creds through all methods in parallel.
5. Then, let's add `CredentialDirectory=`, similar to `RuntimeDirectory=` and its friends, for defining a subdir of /run/credentials/ to pass the creds in in case `file` mode is used. There'd be $CREDENTIAL_DIRECTORY (again, similar to `$RUNTIME_DIRECTORY` and friends) we'd set for services to the resulting dir)
6. When passing via file system, we'd mount a `ramfs` (and not a `tmpfs`!) to /run/credentials/<id>/ and place the creds there. `ramfs` has the benefit of never being swapped out. needs graceful fallback logic to using the dir without mount however, to cover for containers. we'd populate the fs atomically, and keep it read-only afterwards.
7. We'd never pass creds in env vars, they suck since they propagate down the process tree, even to suid programs.

everything passed will be immutable and only accessible to the user of the service, nothing else. there'd be no API for dynamically requesting more creds.

dhess:
@poettering Any plans to implement this in systemd in the immediate future?

poettering:
well, no guarantees, but yeah, i hope to deliver this soon.

dhess:
@poettering Oh, interesting. Would you like to join the call this evening (5-6pm UTC)?

dhess:
Agenda for the meeting:

1. review last meeting take-aways

2. Discuss @d-goldin's progress.

3. Discuss @poettering's proposal -- how does it affect this RFC?

4. Next steps

flokli:
@poettering I gave your thoughts a bit of a read and would have some feedback. Do you want to discuss things here, or should this be moved elsewhere?

poettering:
ah, sorry, i only work part-time currently (parental leave), I am not as attentive to things like this as I used to be. @flokli maybe open an RFE bug on systemd github. maybe called "RFE: per-service credentials system" or so

Lassulus:
@dhess did you post a summary of the meetup somewhere else? I can't find one

dhess:
@Lassulus Sorry, I haven't posted the meeting notes yet. I'll get to it soon.

dhess:
Sorry for the delay in posting the meeting minute notes, but here they are.

## Notes from 2020-05-06 call

Attending:

- @dhess 
- @d-goldin 
- @flokli 
- @Lassulus 
- @aanderse
- @globin (I think? Sorry, I forgot to ask for this participant's GitHub username)

@d-goldin reported the following:
- Nobody has stepped up to co-author yet.
- There's now a proof-of-concept repo:  https://github.com/d-goldin/nix-svc-secrets/blob/master/lib/secrets.nix
- The Hashicorp Vault implementation in this proof-of-concept implementation isn't very robust at this point.
- @poettering's proposal needs more review, but provides some good primitives. It probably doesn't solve all the NixOS problems, e.g., if we want Vault we'd still need some way of integrating with it. It only provides a few mechanisms. How does a user of a service module describe where secrets would live

@dhess & @flokli  agreed to try out the implementation before the next meeting.

@flokli:
- Will open a discussion with @poettering. NixOS should be involved from the start. 
- How we expose any systemd hooks from the NixOS module system still needs to be solved. 
- Will ask @poettering about other mechanisms, e.g., Vault.

Given @poettering's intention to work on related things in systemd, a vote was taken on this RFC, whether to proceed with the work, pause it for a bit until systemd work is better fleshed-out, or close the proposal. The votes were as follows (@d-goldin did not vote):

- @dhess: continue
- @aanderse:  pause for a month or two
- @Lassulus: continue 
- @flokli: continue
- @globin: continue

The vote was 4 : 1 : 0 (continue : pause : close), so we'll proceed with the RFC in parallel with the systemd work. Any further work should be done with the intention of integrating with systemd's primitives (if any) when the time comes.

d-goldin:
@dhess: s/@globin/@andir

dhess:
@d-goldin Thank you. I've updated the notes. @andir I'm sorry!

rfcs/0059-secrets-for-services.md
```diff
@@ -0,0 +1,431 @@
...
+The general idea is centered around this simple process:
+
+A privileged side-car service is launched first, creates a namespace, executes
+the fetcher function which retrieves the secrets and copies them into the
+private tmpfs. The side-car service and main service are configured in such a
+way that restart or termination of either causes restart or termination of the
+other. Once both services shut down the private tmpfs disappears.
+
+The target service launches once the side-car service has been launched and
+singalled successful retrieval of secrets (via systemd-notify), the target
```
dudebout, line 118: s/singalled/signaled/

rfcs/0059-secrets-for-services.md
```diff
@@ -0,0 +1,431 @@
...
+service then joins the sida-cars namespace and is able to access the secrets
```
dudebout, line 119: s/sida-cars/side-car/

lheckemann: side-car's

d-goldin: I'll fix these up a bit later. thanks!

rfcs/0058-secrets-for-services.md
bennofs: > I briefly tried ExecStartPre, but unfortunately it does not seem to run within the mount namespace, so it doesn't have a access to the PrivateTmp we want. I am not sure if this is intentional or not.

I think that's what the `!` prefix is for:

> "!" - Similar to the "+" character discussed above this permits invoking command lines with elevated privileges. However, unlike "+" the "!" character exclusively alters the effect of User=, Group= and SupplementaryGroups=, i.e. only the stanzas that affect user and group credentials. Note that this setting may be combined with DynamicUser=, in which case a dynamic user/group pair is allocated before the command is invoked, but credential changing is left to the executed process itself.

rfcs/0059-secrets-for-services.md
d-goldin: opted for BE

Mic92:
@dhess What are next steps for this RFC. Another meeting?

dhess:
@Mic92 Yes, eventually, but not much has changed since the last one, so I don't know if it's warranted at this time. I haven't had any time to try out the proof-of-concept implementation yet.

@flokli How about you, have you tried it? Also, is there any update on the discussions with upstream?

aanderse:
Thought I would link this here for future reference. Possibly of interest to people subscribed to this thread: https://cedwards.xyz/protecting-systemd-service-secrets/

Mic92:
We have now a pass based secret [implementation](https://github.com/NixOS/nixpkgs/pull/93659) and one based on [sops](https://github.com/Mic92/sops-nix).

d-goldin:
@dhess, @flokli: Have you had a chance to give it a try, as a follow-up from the last call?

d-goldin:
As things stand, I don't think this will progress much, so I propose to close this.

d-goldin:
Thanks for everyone's feedback, ideas and other contributions. I hope they'll be useful at the very least as a reference for other efforts.

Mic92:
Systemd actually has this feature now: https://github.com/systemd/systemd/pull/16568
It just got merged and we could start using it in NixOS as well.

PR #60, opened 2019-11-17
Make the detailed design terse by separating examples

Ericson2314:
A few times I have asked people to separate examples from the core change we are deciding on, to have a terse, precise, and unambiguous normative bit that won't become more confusing when the context that makes something wordier flow is lost over time.

There is prior art for this sort of template revision:

In https://github.com/ghc-proposals/ghc-proposals/pull/251 it is pointed out that the combined section *both* be too wordy and have too few clarifying examples. This is a good point, lest this template change seem to skewed in favor of exports and against beginners: Everyone can win when we have terser detailed designs and far more examples. 

https://github.com/rust-lang/rfcs/pull/2059 is not an exact match, but I think is closer in spirit given th way things worked in practice. Before that change, the "how do we teach this" section was excessively beginner-oriented, so the detailed design would still have many intermediate-oriented details. After that change, I think it was clearer that everyone can benefit from a guide / examples, and no one should feel obligated to immediately obligated to understand the reference explanation and all its ramifications on the first pass-through. This freed that section to be more terse and precise.

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -22,10 +22,19 @@ outcome?
 # Detailed design
 [design]: #detailed-design
 
-This is the bulk of the RFC. Explain the design in enough detail for somebody
-familiar with the ecosystem to understand, and implement.  This should get
-into specifics and corner-cases, and include examples of how the feature is
-used.
+This is the core, normative part of the RFC. Explain the design in enough
+detail for somebody familiar with the ecosystem to understand, and implement.
+This should get into specifics and corner-cases. Yet, this section should also
+be terse, avoiding redundancy even at the cost of clarity.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This section illustrates the detailed design. This section should clarify all
+confusion the reader has from the previous sections. It is especially important
+to counterbalance the desired terseness of the detailed design; if you feel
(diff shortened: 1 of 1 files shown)


0000-template.md
```diff
@@ -22,10 +22,19 @@ outcome?
...
 [design]: #detailed-design
 
-This is the bulk of the RFC. Explain the design in enough detail for somebody
-familiar with the ecosystem to understand, and implement.  This should get
-into specifics and corner-cases, and include examples of how the feature is
-used.
+This is the core, normative part of the RFC. Explain the design in enough
+detail for somebody familiar with the ecosystem to understand, and implement.
+This should get into specifics and corner-cases. Yet, this section should also
+be terse, avoiding redundancy even at the cost of clarity.
```
Ericson2314: I might note that an example is by definition redundant, and this is a good thing! If examples state things not covered in the formal design, then we have to reason backwards on what missing design fragment is (difficult and subjective), and also decide whether the detailed design meant to include that design fragment and forgot to, or the didn't mean to include that design fragment and the example dragged in something off-topic.

globin:
I'll bring this up in the RFCSC meeting on Thursday!

Ericson2314:
Thank you all for the quick decision!

PR #67, opened 2020-03-18
[RFC 0067] Common override interface derivations

FRidh:
This RFC came after the discussion on https://github.com/NixOS/nixpkgs/pull/46842 and https://github.com/NixOS/nixpkgs/pull/82772.

[rendered](https://github.com/FRidh/rfcs/blob/overrideattrs/rfcs/0067-common-override-interface-derivations.md)

### DIFF
--- a/rfcs/0067-common-override-interface.md
+++ b/rfcs/0067-common-override-interface.md
@@ -1,0 +1,72 @@
+---
+feature: common-override-interface
+start-date: 2020-03-17
+author: Frederik Rietdijk
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Define commonly used attributes for overriding in Nixpkgs.
+
+# Motivation
+[motivation]: #motivation
+
+In Nixpkgs several methods exist to override functions calls. The primary ones are:
+
+1. `.override` to override, typically, the first function call.
+2. `.overrideAttrs` to override the call to `stdenv.mkDerivation`.
+3. `.overrideDerivation` to override the call to `derivation`.
+
+The first two are mainly used and are typically sufficient. However, how can we override generic package builders, such as `buildPythonPackage` and `buildGoPackage`? 
+
+For the `buildPythonPackage` function the `.overridePythonAttrs` was introduced that would override the call to `buildPythonPackage` because using `.overrideAttrs` would often not result in what the user expect would happen. In hindsight, it would have been better to attach a custom `.overrideAttrs` to `buildPythonPackage`.
+
+This RFC thus proposes to let generic builders define a custom `.overrideAttrs` that overrides the call to the generic builder.
+
+# Detailed design
+[design]: #detailed-design
+
+The method `.overrideAttrs` will be modified so that instead of
+
+- `.overrideAttrs` to override the call to `stdenv.mkDerivation`.
+
+it will be
+
+- `.overrideAttrs` to override the call to the generic builder.
+
+The generic builders such as `buildGoPackage` would thus apply the function `makeOverridable` to it.
+
+In case of Python that already has `.overridePythonAttrs`, support for
+```nix
+buildGoPackage = makeOverridable buildGoPackage;
+```
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+-
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+After invoking a generic builder it is no longer possible to override the call to `stdenv.mkDerivation`.
+
+# Alternatives
+[alternatives]: #alternatives
+
+An alternative would be to add a new method for overriding, `.overrideArgs`, thus allowing one to still call `.overrideAttrs` to override `stdenv.mkDerivation`.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+-
+
+# Future work
+[future]: #future-work
+
+Implement the custom `.overrideAttrs` for the generic builders.
--- a/rfcs/0067-common-override-interface-derivations.md
+++ b/rfcs/0067-common-override-interface-derivations.md
@@ -11,7 +11,8 @@ related-issues: (will contain links to implementation PRs)
 # Summary
 [summary]: #summary
 
-Define commonly used attributes for overriding of derivations in Nixpkgs.
+Define commonly used attributes for overriding of derivations in Nixpkgs to
+improve overriding from a UI point of view.
 
 # Motivation
 [motivation]: #motivation
--- a/rfcs/0067-common-override-interface-derivations.md
+++ b/rfcs/0067-common-override-interface-derivations.md
@@ -3,8 +3,8 @@ feature: common-override-interface-derivations
 start-date: 2020-03-17
 author: Frederik Rietdijk
 co-authors: (find a buddy later to help out with the RFC)
-shepherd-team: (names, to be nominated and accepted by RFC steering committee)
-shepherd-leader: (name to be appointed by RFC steering committee)
+shepherd-team: @Infinisil, @edolstra, @Taneb, @7c6f434c
+shepherd-leader: @Infinisil
 related-issues: (will contain links to implementation PRs)
 ---
 
--- a/rfcs/0067-common-override-interface-derivations.md
+++ b/rfcs/0067-common-override-interface-derivations.md
@@ -1,5 +1,5 @@
 ---
-feature: common-override-interface
+feature: common-override-interface-derivations
 start-date: 2020-03-17
 author: Frederik Rietdijk
 co-authors: (find a buddy later to help out with the RFC)
@@ -11,7 +11,7 @@ related-issues: (will contain links to implementation PRs)
 # Summary
 [summary]: #summary
 
-Define commonly used attributes for overriding in Nixpkgs.
+Define commonly used attributes for overriding of derivations in Nixpkgs.
 
 # Motivation
 [motivation]: #motivation
@@ -22,7 +22,13 @@ In Nixpkgs several methods exist to override functions calls. The primary ones a
 2. `.overrideAttrs` to override the call to `stdenv.mkDerivation`.
 3. `.overrideDerivation` to override the call to `derivation`.
 
-The first two are mainly used and are typically sufficient. However, how can we override generic package builders, such as `buildPythonPackage` and `buildGoPackage`? 
+Also used in several places but out of scope for this RFC because it is for
+overriding package sets is `lib.overrideScope'`.
+
+The first two are mainly used and are typically sufficient. The third one should
+typically be avoided and can be considered legacy.
+
+However, how can we override generic package builders, such as `buildPythonPackage` and `buildGoPackage`?
 
 For the `buildPythonPackage` function the `.overridePythonAttrs` was introduced that would override the call to `buildPythonPackage` because using `.overrideAttrs` would often not result in what the user expect would happen. In hindsight, it would have been better to attach a custom `.overrideAttrs` to `buildPythonPackage`.
 
--- a/rfcs/0067-common-override-interface-derivations.md
+++ b/rfcs/0067-common-override-interface-derivations.md
@@ -66,7 +66,27 @@ After invoking a generic builder it is no longer possible to override the call t
 # Alternatives
 [alternatives]: #alternatives
 
-An alternative would be to add a new method for overriding, `.overrideArgs`, thus allowing one to still call `.overrideAttrs` to override `stdenv.mkDerivation`.
+An alternative would be to let generic builders add their own methods, e.g.
+`overridePythonAtrrs`, `overrideGoAttrs`, and so on. While it may be clear which
+function it overrides, the user may not actually know what functions are called
+and thus may not know which override to use. From a UI point of view, they
+should probably not even need to know; all they should know is that there is a
+method which, when called, will apply what is expected.
+
+An alternative would be to add a new method for overriding of generic builders,
+`.overrideArgs`, thus allowing one to still call `.overrideAttrs` to override
+`stdenv.mkDerivation`. This is a new function users may need to be aware of.
+From a UI point of view, it could mean `stdenv.mkDerivation`, when not wrapped
+into a generic builder, may also need to offer `.overrideAttrs` for consistency.
+
+An alternative would be to have a generic override function that can override
+any function in the stack of function calls. E.g.
+- `.override` becomes `.overrideCall "callPackage" { ... }`
+- `.overrideAttrs becomes .overrideCall "mkDerivation" { ... }`
+- `.overridePythonAttrs becomes .overrideCall "buildPythonPackage" { ... }`
+- `.overrideDerivation becomes .overrideCall "derivation" { ... }`
+While this offers maximum precision when overriding, this does not solve the UI
+problem either.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions


rfcs/0067-common-override-interface.md
```diff
@@ -0,0 +1,72 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+After invoking a generic builder it is no longer possible to override the call to `stdenv.mkDerivation`.
+
+# Alternatives
+[alternatives]: #alternatives
+
+An alternative would be to add a new method for overriding, `.overrideArgs`, thus allowing one to still call `.overrideAttrs` to override `stdenv.mkDerivation`.
+
```
7c6f434c: Another alternative could be some mechanism to unify `override`/`overrideAttrs` into a single override mechanism supporting multiple (probably named?) levels, like `overrideCall pkg "callPackage" (x: {…})`

zimbatm:
The RFC in its current state is a bit sparse in details. Is it possible to add more details on the implementation and how it would work when builders are nested?

rfcs/0067-common-override-interface.md
Infinisil: Very much this, and it's also what https://github.com/NixOS/nixpkgs/pull/87394 seems to implement. I think this is the only way to not end up with a mess when you have arbitrarily deeply nested function calls.

With this
- `.override` becomes `.overrideCall "callPackage" { ... }`
- `.overrideAttrs` becomes `.overrideCall "mkDerivation" { ... }`
- `.overridePythonAttrs` becomes `.overrideCall "buildPythonPackage" { ... }`
- `.overrideDerivation` becomes `.overrideCall "derivation" { ... }`
- etc.

With this, there's much less confusion as to how you need to override what. And it's a unified interface for the current override functions.

Infinisil:
Nominating myself as a shepherd

FRidh:
I will clearly need to extend this but just dumping here some thoughts.

Consider e.g. the Python packages set that contains primarily Python packages (`.overridePythonAttrs`), but also bindings to non-`buildPythonPackage` packages `.overrideAttrs`. 

In the last years that we've had `.overridePythonAttrs` there has not been a single case where `.overrideAttrs` was actually needed.

7c6f434c:
> In the last years that we've had .overridePythonAttrs there has not been a single case where .overrideAttrs was actually needed

This, apparently, happens because you have these four lines of code that allow arbitrary overrides of `mkDerivation` via `buildPythonPackage` arguments.

I am not sure whether we want to commit to always having this in every package creation function, and not use generic nested overrides. I am not even sure it is better in the case #58 gets unstuck and implemented and released and usable in Nixpkgs in a couple of years.

Separately, if we are to touch the overrides at all, maybe say that the correct way to use overrides is `lib.override* x something`, not `x.override something`? I have an impression that if we start optimising the override memory cost, the former approach will leave us more freedom.

B4dM4n:
Since opening NixOS/nixpkgs#87394 and being redirected here I experimented a bit more with different overriding variants.

I'm currently favoring a layered approach over shadowing lower functions to keep the current flexibility of nixpkgs.

The difference might not be a big deal in nixpkgs because it's self contained and developers have full control of the definition and usage of a function, but projects outside nixpkgs or even the `configuration.nix` (which got me into implementing NixOS/nixpkgs#87394) might want to access any of the provided override functionality for small modifications.

The current set of overriding capabilities by all the `. override*` functions is hard to understand when using them for the first time. Some of these functions are called with a set, others take a function returning a set and some might also accept both.

Especially when working outside nixpkgs (projects that use nixpkgs) it can sometimes take two or three of these to replace a single dependency, so removing override functions might restrict users from using the nixpkgs package and simply copy paste the code into their own files:
- `.override` to replace a input package
- `.overridePythonAttrs` to switch a build step
- `.overrideAttrs` to change a `buildPythonPackage` behavior

Sure, it can all be done with one `.overrideDerivation`, but this wouldn't be very user friendly.

> In the last years that we've had `.overridePythonAttrs` there has not been a single case where `.overrideAttrs` was actually needed.

Even if `.overrideAttrs` isn't needed inside nixpkgs, changing it's behavior from `override mkDerivation` to `override the top level builder` would probably break most of the code outside of nixpkgs that relies on this.

Since NixOS/nixpkgs#87394 I have tried to implement a layered variant of `makeOverridable`.

The first attempt used the shadowing approach for `.overrideAttrs` and included a `overrideAttrsNested [ ]` function to override deeper levels.
This wan't very ergonomic, because even a simple task as `override passthru in mkDerivation` (from `pythonPackages.toPythonModule`) wasn't as trivial to implement as before, because the function had to check `how many layers has the builder of this derivation`.

The second attempt (B4dM4n/nixpkgs@2793c4a11220b3e3e5fa9cfdae06142314b1b104) uses a "named layers" approach where each named layer can be specifically overridden and also allows to override layers bottom up without naming them. Unnamed layers (produced by `makeOverridable`) can be inserted in between without influencing the layers, but providing the `.override` function.

It currently adds two new functions to provide it's features:
- `.overrideLayerName` to override a specific layer by name. This can be use do replace functions like `.overridePythonAttrs` (B4dM4n/nixpkgs@34d6cc523c4cfc42390ed86f76dcaf9c022f13c2)
- `.overrideLayers` allows overriding multiple layers at the same time either by name or as a list from bottom to top.

Theses two examples are equivalent (with the current set of named layers):
```nix
nixpkgs-fmt.overrideLayers {
  derivation = (old: {
    buildInputs = old.buildInputs ++ [ hello ];
  });
  rust = {
    version = "0.9.1";
    cargoSha256 = "1wybvm9qckx9cd656gx9zrbszmaj66ihh2kk6qqdb6maixcq5k0b";
  };
}
```

```nix
nixpkgs-fmt.overrideLayers [
  (old: {
    buildInputs = old.buildInputs ++ [ hello ];
  })
  {
    version = "0.9.1";
    cargoSha256 = "1wybvm9qckx9cd656gx9zrbszmaj66ihh2kk6qqdb6maixcq5k0b";
  }
]
```

As a side effect from moving all override handling into one place, this also allows to use attribute sets and functions interchangeably as override arguments.

With this approach `.overrideDerivation` and `.overrideAttrs` could in theory be deprecated as they can be replaced by named layers and with free functions for convenient access.

Mic92:
We still don't have enough shepherds here to move this RFC to the next phase.

Taneb:
I'd like to nominate myself to help shepherd this PR.

rfcs/0067-common-override-interface.md
FRidh: I like this approach, but I do want to clarify that from a UI point of view, end users should *typically* need one and only one function that helps them. That is my motivation for shadowing `.overrideAttrs`. Of course, if we can provide something like proposed here, *as well* as a call that is expected to override what is *typically* expected, then that would be great.

FRidh:
Clarified the scope is overriding of derivations, not package sets.

FRidh:
At this point I won't add more to the RFC. For me the technical implementation is a detail. What matters is that we agree on the requirement, that is, that there is at least one *typical* override that will help users 99% of the cases, whether they used `stdenv.mkDerivation` directly, or another generic builder such as `buildPythonPackage`. I'm all for the proposed alternative where you can more specifically override specific calls, *as long* as there is a sufficiently generic override that users can rely on.

7c6f434c:
1. Changing `overrideAttrs` to behave differently is also an UI problem, given existence of material about Nixpkgs outside the Nixpgks repository. If it is not a large problem, such a claim needs to be substantiated at least a bit.

2. Should the universal interface be accessed via an attribute or be a function in `lib`?

FRidh:
1) is about backwards compatibility. Depends on how important we deem that. Considering we do not have many custom override functions (I think its just Python, and it hasn't existed that long), I don't think it is such a big problem.
2) Whether it is a lib function or an attribute is a detail that, to me, does not matter. Attribute is slightly more convenient I find but I suppose it comes with a performance penalty.

7c6f434c:
>1) is about backwards compatibility. Depends on how important we deem that. Considering we do not have many custom override functions (I think its just Python, and it hasn't existed that long), I don't think it is such a big problem.

Well, changing the semantics of overrideAttrs (which is almost always applicable in principle) is a backwards incompatible change.

>2) Whether it is a lib function or an attribute is a detail that, to me, does not matter. Attribute is slightly more convenient I find but I suppose it comes with a performance penalty.

It will probably depend on implementation, but not creating a closure might hopefully save us a bit…

lheckemann:
IMHO: a better solution should not be sacrificed for backwards compatibility. We have pinning unlike many other distros, and stable releases where the "old way" is available for 6 months, which should allow us to make backwards-incompatible changes if they're required for improvements.

This doesn't address the use of `rec` or `let` bindings like when `src` depends on `version`, right? Fair enough if you consider it out of scope, but while we're RFCing on a quite closely related subject it'd be "nice to have" as well.

https://gist.github.com/edolstra/29ce9d8ea399b703a7023073b0dbc00d is also somewhat related (though old and not updated for a while? Maybe @edolstra can say more about its current status).

FRidh:
> This doesn't address the use of rec or let bindings like when src depends on version, right? Fair enough if you consider it out of scope, but while we're RFCing on a quite closely related subject it'd be "nice to have" as well.

Defining an interface for derivations is a different, bigger task, that is out of scope here. I think that is very important well, but will take far more time, and require massive changes. The purpose here is to improve the current state with relatively small effort but provide users with the means for performing the common overrides they need, solving an issue that has come up many times.

edolstra:
@B4dM4n @7c6f434c Are you interested in shepherding this RFC? Currently we have two shepherd nominations but it would be better to have at least one more.

7c6f434c:
I feel that I am a bit too behind on things I want to do around Nix; I guess if this is blocked on having enough shepherds, I can promise to continue rereading the proposal and keeping track of the discussion (which I do anyway so far) and allocate some synchronous time for meeting.

So I accept the nomination as a non-leader shepherd, but I also acknowledge I will probably be worse at that than someone currently willing to allocate significant energy to this topic.

Mic92:
Ok. Is there maybe one more shepherd?

edolstra:
I'd like to nominate myself.

lheckemann:
Shepherd team is @infinisil (leader), @edolstra, @Taneb, @7c6f434c

rfcs/0067-common-override-interface.md
olmokramer: I fail to see how this helps to be honest, because the difficult part (at least in my opinion) is understanding the semantics of the different `override` functions, not their actual name. But all this (I'm using @Infinisil's example here because it is concrete) does is rename `override` to `overrideCall "callPackage"` etc. but I still have to look at the code to see which of the strings (instead of functions) I have to use to achieve some specific goal.

I also don't know if shoving all these functions under one name will be very helpful documentation wise, as now I have to read more, and more importantly about a wider range of topics, to find out if this one function will solve my problem or not.

I think the problem lies more in the fact that it is difficult to grasp all the mechanics at play in nixpkgs, but I don't think that's solved by overloading a function.

Mic92:
@Infinisil could you organize a meeting for this one?

Infinisil:
Calling for a meeting from 14. to 30. November. Please enter your available times in https://www.when2meet.com/?10278623-oIIve until 13. November (end of next week) @edolstra @Taneb @7c6f434c @FRidh

7c6f434c:
The range is limited 12:00 UTC to 23:00 UTC, right, it is not me missing something?

Infinisil:
@7c6f434c Correct

Infinisil:
Thanks for entering your dates @edolstra @Taneb @7c6f434c @FRidh. We will meet on 15:00 UTC on 17 November, or `date --date=@1605625200` in https://meet.jit.si/NixRFC67 for an expected hour. I have sent out a calendar invite to your email addresses as well.

Infinisil:
## Meeting 1

Present were @FRidh, @7c6f434c, @Taneb and @Infinisil 

### Notes
Which override functions do people actually use?
* sometimes (Haskell) you need to override source in `mkDerivation` and not in the builder
* user UI: is knowledge of correct overrides needed?
  * ask for overriding the expression you see
  * but still need to understand the semantics

Stack of overridable things, e.g. for python:
* .override for callPackage
* .overridePythonAttrs for buildPythonPackage
* .overrideAttrs for mkDerivation
* .overrideDerivation for derivation

We don't want to conflate overrides of different layers with each other. Each override function should be for its specific function. Having `.overrideAttrs` override `buildPythonPackage` arguments would cross layers

To unify these layers we could give string names to each layer
* `.overrideLayer "layerName"` overrides arguments of the given layer
* This `.overrideLayer` function can make sure that all the other layer overrides get updated as well (no more complexity of this in each override function declaration)

With this in place, it might be possible to create a layer-inferencing override mechanism
* `.overrideLayerAuto` applies attributes to the layer they are inferred from
* This allows users to not have to know the implementation details anymore to know how things can be overridden
* If the layer inference turns out to be ambiguous, throw an error telling the user to use `.overrideLayer "<name>"` instead

With this, we could also add smarts around function arguments
* Each layer could add `description` values to arguments, which document their use
* For ambiguous overrides, these descriptions are displayed
* Can also declare `supersedes`-like relation, where e.g. the `buildInputs` of `mkDerivation` supersedes the `buildInputs` of `derivation`, therefore not triggering an ambiguity error

Should these overrides be a library function like `lib.overrideLayer` instead?
* Doing this might allow Nix to collect Nix values which previously were kept around
* We could also make it backwards compatible by aliasing `.overrideLayer` to `lib.overrideLayer`
 
Conclusions:
* Let's not shadow `.overrideAttrs` arguments with `.overridePythonAttrs`
* Builders may have their own override, as it mirrors the `.overrideLayer` approach, and we can introduce aliases later
* However, it's very tricky to properly implement such custom `.overrideFooAttrs` functions, as these need to be aware of any other layers and override them all

Mic92:
What is the next step in this RFC?

Infinisil:
I have started an implementation of a generic way to specify overridable layers. It doesn't work in all cases, but it's pretty neat still. Gonna explain it in short here:

We can define a stack of functions and give each layer of that stack a name so we can later override it:
```nix
{
  # An example function stack, where each layer can transform the input, passing it to the next layer
  testStack = [
    { name = "layer1"; fun = arg: arg // { foo = "foo-${arg.bar}"; }; }
    { name = "layer2"; fun = arg: { result = arg.bar + ", " + arg.foo; }; }
  ];

  # Calls the function stack with an argument, applying all the layers in order
  callStack = stack: arg: lib.foldl' (arg: el: el.fun arg) arg stack;

  result.r1 = callStack testStack { bar = "bar"; };
  # { result = "bar, foo-bar"; }
}
```

We can't override anything yet, so let's create a function that lets us redo the call with a modified stack

```nix
{
  # Modifies the layer with the given name, calling the modification function on the old element to get the new one
  modify = modName: mod: stack: /* ... */;

  # Same as callStack, but it also sets a 
  callStackOverridable = stack: arg:
    let
      result = callStack stack arg;

      # Return the result but with a modified stack
      modifyStack = mod: callStackOverridable (mod stack) arg;

      # Convenience function for modifying the arguments of a given layer
      overrideLayerArg = name: mod: modifyStack (modify name (old: old // {
        fun = arg: old.fun (arg // mod);
      }));
    in result // {
      inherit modifyStack overrideLayerArg;
    };

  result.r2 = callStackOverridable testStack { bar = "bar"; };
  # { result = "bar, foo-bar"; ... }

  result.r3 = result.r2.overrideLayerArg "layer1" { bar = "qux"; };
  # { result = "qux, foo-qux"; ... }}
```

Nice! Now let's look at an example of how this would look like for python:

```nix
{
  buildPythonPackage = callStackOverridable [
    {
      name = "python";
      fun = { dontWrapPythonPrograms ? false, ...}@args:
        removeAttrs args [ "dontWrapPythonPrograms" ] // {
          postFixup = lib.optionalString (!dontWrapPythonPrograms) ''
            wrapPythonPrograms
          '' + args.postFixup or "";
        };
    }
    {
      name = "mkDerivation";
      fun = args: args // {
        builder = pkgs.stdenv.shell;
        system = builtins.currentSystem;
      };
    }
    {
      name = "derivation";
      fun = args: derivation (removeAttrs args [ "passthru" ])
        // lib.optionalAttrs (args ? passthru) {
          inherit (args) passthru;
        };
    }
  ];
}
```

Here we defined each layer to transform the argument in some way. Notably the `python` layer doesn't call `mkDerivation` itself, but it rather just transforms its arguments for the next layer, which is the "mkDerivation" one.

This then allows us to call the function and override any layer of it later:
```nix
{
  result.r4 = buildPythonPackage { name = "test"; };

  result.r5 = result.r4.overrideLayerArg "python" { dontWrapPythonPrograms = true; };

  result.r6 = result.r4.overrideLayerArg "mkDerivation" { passthru.foo = "foo"; };

  result.r7 = result.r4.overrideLayerArg "derivation" { builder = "/bin/sh"; };
}
```

The full source is here:
<details>

```nix
let
  pkgs = import <nixpkgs> {};
  inherit (pkgs) lib;

  # An example function stack, where each layer can transform the input, passing it to the next layer
  testStack = [
    { name = "layer1"; fun = arg: arg // { foo = "foo-${arg.bar}"; }; }
    { name = "layer2"; fun = arg: { result = arg.bar + ", " + arg.foo; }; }
  ];

  # Calls the function stack with an argument, applying all the layers in order
  callStack = stack: arg: lib.foldl' (arg: el: el.fun arg) arg stack;

  result.r1 = callStack testStack { bar = "bar"; };
  # { result = "bar, foo-bar"; }


  # Modifies the layer with the given name, calling the modification function on the old element to get the new one
  modify = modName: mod: stack:
    let index = findIndex (el: el ? name && el.name == modName) stack;
    in lib.take index stack ++ [ (mod (lib.elemAt stack index)) ] ++ lib.drop (index + 1) stack;

  # Same as callStack, but it also sets a 
  callStackOverridable = stack: arg:
    let
      result = callStack stack arg;

      # Return the result but with a modified stack
      modifyStack = mod: callStackOverridable (mod stack) arg;

      # Convenience function for modifying the arguments of a given layer
      overrideLayerArg = name: mod: modifyStack (modify name (old: old // {
        fun = arg: old.fun (arg // mod);
      }));
    in result // {
      inherit modifyStack overrideLayerArg;

      # Backwards compatible .overrideAttrs
      overrideAttrs = overrideLayerArg "mkDerivation";
    };

  result.r2 = callStackOverridable testStack { bar = "bar"; };
  # { result = "bar, foo-bar"; ... }

  result.r3 = result.r2.overrideLayerArg "layer1" { bar = "qux"; };
  # { result = "qux, foo-qux"; ... }


  buildPythonPackage = callStackOverridable [
    {
      name = "python";
      fun = { dontWrapPythonPrograms ? false, ...}@args:
        removeAttrs args [ "dontWrapPythonPrograms" ] // {
          postFixup = lib.optionalString (!dontWrapPythonPrograms) ''
            wrapPythonPrograms
          '' + args.postFixup or "";
        };
    }
    {
      name = "mkDerivation";
      fun = args: args // {
        builder = pkgs.stdenv.shell;
        system = builtins.currentSystem;
      };
    }
    {
      name = "derivation";
      fun = args: derivation (removeAttrs args [ "passthru" ])
        // lib.optionalAttrs (args ? passthru) {
          inherit (args) passthru;
        };
    }
  ];

  result.r4 = buildPythonPackage { name = "test"; };

  result.r5 = result.r4.overrideLayerArg "python" { dontWrapPythonPrograms = true; };

  result.r6 = result.r4.overrideLayerArg "mkDerivation" { passthru.foo = "foo"; };

  result.r7 = result.r4.overrideLayerArg "derivation" { builder = "/bin/sh"; };



  #### Extra functions and utilities for modifying a function stack

  # Finds the index of a list element or -1 otherwise
  findIndex = pred: list:
    let
      start = { index = 0; found = -1; };
      fun = { index, found }: el: { index = index + 1; found = if found == -1 && pred el then index else found; };
      result = lib.foldl' fun start list;
    in result.found;

  # Inserts an element into a list at a specific index
  insert = index: el: list:
    lib.take index list ++ [ el ] ++ lib.drop index list;

  # Inserts a new layer into a function stack just before the layer with the given name
  insertBefore = beforeName: el: stack:
    let index = findIndex (el: el ? name && el.name == beforeName) stack;
    in insert index el stack;

  # Inserts a new layer into a function stack just after the layer with the given name
  insertAfter = afterName: el: stack:
    let index = findIndex (el: el ? name && el.name == afterName) stack;
    in insert (index + 1) el stack;

  # Prepends a layer to the function stack, applying it at the very start
  prepend = name: fun: stack:
    [ { inherit name fun; } ] ++ stack;

  # Appends a layer to the function stack, applying it at the very end
  append = name: fun: stack:
    stack ++ [ { inherit name fun; } ];

in result
```
</details>

---

Now there's a problem with this approach, which can however be fixed: In nixpkgs, functions can depend on the arguments itself, e.g.
```nix
{ stdenv }: stdenv.mkDerivation { /* ... */ }
```

And a function stack as shown before can't represent this, because later layers can't depend on the arguments of earlier ones. I believe this can be fixed by using a linked stack representation like
```nix
{
  testStackLinked = {
    name = "layer1";
    fun = arg: {
      result = arg // { foo = "foo-${arg.bar}"; };
      next = {
        name = "layer2";
        fun = arg: {
          result = { result = arg.bar + ", " + arg.foo; };
        };
      };
    };
  };
}
```

Another problem is that such a function stack needs to be extracted from the current nixpkgs conventions, such that no packages need adjustments. This could be very easy if the linked stack is implemented, and probably only needs to be used by `callPackage` and all builders. This might be pretty tricky to get right though.

B4dM4n:
### `overrideAuto`

Based on my layer overriding approach from earlier, I also put together a `overrideAuto` implementation on top of that ([B4dM4n:override-attrs](https://github.com/NixOS/nixpkgs/compare/master...B4dM4n:override-attrs)), with a very basic layer-inferencing mechanism. 

Given an attribute set of arguments to override, it works roughly like this:
- filter out the args for the current layer with `lib.functionArgs`
- update the current function args with the new filtered set
- apply the new set of arguments
- take all remaining arguments from the override set and apply the same algorithm to the next layer down
- if any arguments are left after the lowest layer is done, raise an error

Using `lib.functionArgs` for "inference" has the downside that all all function arguments that are used inside a layer must be explicitly named in the argument set. For example `buildPythonPackage` currently uses the arguments `pname`, `version`, `src`, `postFixup` and `checkPhase` inside its body via the `@ attrs` set and not by naming them directly. This prohibits `lib.functionArgs` from seeing this arguments and therefore would not allow `overrideAuto` to modify them. The solution to this issue is that every layer function needs to name all its arguments explicitly. It is still fine to capture arguments in `@ attrs` sets, but they should not be referenced in the function itself, only passed to lower layers.

In most cases, this behavior should match what users try to achieve with overrides. In cases where users want to skip certain layers and explicitly modify a lower argument (the Haskell source example), they probably know why and can use `overrideLayerName` to achieve their goal.

#### `overrideAuto` example
```nix
with import ./lib;
let
  f1 = makeOverridableLayer "f1" ({ name, src } @ args: f2 {
    name = "f1-${name}";
    src = "f1-${src}";
    test = false;
  });
  f2 = makeOverridableLayer "f2" ({ name, src, test } @ args: {
    name = "f2-${name}";
    src = "f2-${src}";
    inherit test;
  });
  res = f1 {
    name = "test1";
    src = "src1";
  };
in
[
  res
  (res.overrideAuto {
    name = "override";
    test = true;
  })
]
# [ 
#   {
#     name = "f2-f1-test1";
#     src = "f2-f1-src1";
#     test = false;
#     override… = …;
#   } {
#     name = "f2-f1-override";
#     src = "f2-f1-src1";
#     test = true;
#     override… = …;
#   }
# ]
```

### `overrideAuto` with functions

During the implementation I wanted to keep the possibility to use override functions with `overrideAuto`. This turned out to be complicated and/or inconvenient for the user. The passed function would either need to be passed the arguments from all layers to be able to access the old state for each argument or it would be needed to be invoked for each layer separately, which puts the burden of checking the available arguments or which layer it would currently override to the user.

#### Problematic example
```nix
res.overrideAuto (prev: { # How to build prev?
  name = "override-${prev.name}";
  test = !prev.test;
})
```

The solution I came up with is a special "override function" pattern which could be recognized:
```nix
res.overrideAuto {
  name = { prev }: "override-${prev}";
  test = { prev }: !prev;
}
```

Because override arguments can also be functions, I used the named argument syntax to check for a function with exactly one `prev` argument. When such a function is detected, it is called with the previous value of the argument.

#### Usage in python-packages.nix
```nix
…
scipy_1_3 = self.scipy.overrideAuto rec {
  version = "1.3.3";
  src = { prev }: prev.override {
    inherit version;
    sha256 = "02iqb7ws7fw5fd1a83hx705pzrw1imj7z0bphjsl4bfvw254xgv4";
  };
  doCheck = false;
  disabled = !isPy3k;
};
…
```

FRidh:
Recently several new PR's were opened changing how to do overrides with package sets. Its really important we converge on this point:
- PHP https://github.com/NixOS/nixpkgs/pull/107044
-

FRidh:
I like what is proposed here. For @Infinisil 's part, it is not necessary to explicitly state the parameters, but I think we should in this RFC recommend builders do so, especially if that makes `overrideAuto` available thereby reducing (typically) the need to know what layer(s) to override. Longer term, I can imagine that by splitting into layers that transform the data before passing to a builder, we get closer to an approach where we can have Nix support for builders inheriting from other builders.

Shall we meet again in the first half of January?

rfcs/0067-common-override-interface.md
FRidh: Yes, we need to be better at documenting the parameters and how they transform the data. In the past there were ideas for attaching docstrings to functions https://github.com/NixOS/nixpkgs/pull/23505.

shlevy:
I haven't reviewed the whole discussion here, just the RFC itself. I like the idea, though I think it would be good to spend some time thinking about if there's a good way to precisely state where in the stack of generic builders you want to override.

My current approach to working around this issue is to almost never use overrideAttrs. Instead I use override, and override the argument corresponding to the set-specific `mkDerivation` (e.g. how `overrideCabal` works). This works fine but is somewhat boilerplate-ridden and assumes there's no relevant scope above the function being overridden that such things are being pulled from.

edolstra:
@FRidh Yes sounds good, do you want to create a doodle for a meeting?

Mic92:
I think someone else could also provide doodle link.

FRidh:
Hooks often rely on environment variables. When a builder builds upon that hook, it means the environment variables used by the hook need to be declared by the builder as well.

Mic92:
@FRidh can you organize a meeting? https://www.when2meet.com/

spacekookie:
@Infinisil, @edolstra, @Taneb any plans to schedule a meeting for this RFC?

Mic92:
Is this RFC stalled? @Infinisil, @edolstra, @Taneb, @7c6f434c
It looks like no one is interested in organizing a meeting.

7c6f434c:
Hm. I am not sure, @Infinisil posted a pretty nice proposal as intermediate and I am not sure if there was any later changes to it not shared or should we start from there and try to move forward (I still hold the opinion expressed in the shpeherd meeting notes that the approach needs some changes, indeed in the direction @Infinisil has tried)

spacekookie:
@Infinisil, @edolstra, @Taneb, @7c6f434c since this RFC has stalled we will close it for the time being. Please feel free to re-open it again once work on this continues

roberth:
https://github.com/NixOS/nixpkgs/pull/119942 (which, I should add, is fully backwards compatible and has merit in its own right) changes the landscape a bit. It could [probably simplify](https://github.com/NixOS/nixpkgs/pull/119942#issuecomment-1014718769) the python expressions by making `overridePythonAttrs` unnecessary, or even an alias for `overrideAttrs`, defined via `passthru`.

roberth:
I think it's worth noting that if you can separate your overriding pipeline into separate functions each with a name, you'r basically one step away from putting those attrsets in a module together, with the additional benefit of having access to all layers without even having to think about layers.

I suppose either a module-like solution or the more constrained interface proposed here can solve the issue of `callPackage` having to be aware of `mkDerivation`'s returned overriding functions.

As a final thought, having to serialize the chain of overriding methods doesn't generalize too well. Perhaps we should first have a pattern where each overridable thing has a function as its parameter that specifies how its _creator_ can be overridden. That may solve aforementioned architectural problem when, say, a package defined an extra overriding methods using `passthru.extraOverrideAttrs = args: finalAttrs.finalPackage.overrideAttrs (foo args)`. I don't see how the pipeline of named functions idea would allow the equivalent of `pkg.extraOverrideAttrs.override` to work.
Probably the definition should look somewhat like `passthru.extraOverrideAttrs = args: (finalAttrs.finalPackage.overrideAttrs (foo args)) // callerOverrides` to support that.

PR #69, opened 2020-06-02
Expand the RFC acronym at least once

andir:
While it is probably know to most it should also be documented what we expand RFC to.


README.md
```diff
@@ -1,4 +1,4 @@
-# Nix RFCs
+# Nix RFCs (request for comments)
```
zimbatm, line 1:
```suggestion
# Nix RFCs (Request For Comments)
```
it's a custom that the acronym letters be upper-cased

PR #70, opened 2020-06-20
[RFC 0070] Merge nixos-hardware in to nixpkgs (#70)

Mic92:
We are currently having a separate repository to providing NixOS profiles for hardware specific extensions at [nixos-hardware](https://github.com/NixOS/nixos-hardware). This RFC proposes to merge this repository (4000 LOCs/350 commits) and add support for detecting hardware in `nixos-generate-config`.

[Rendered](https://github.com/Mic92/rfcs/blob/nixos-hardware/rfcs/0070-nixos-hardware.md)

### DIFF
--- a/rfcs/0070-nixos-hardware.md
+++ b/rfcs/0070-nixos-hardware.md
@@ -1,0 +1,103 @@
+---
+feature: nixos_hardware
+start-date: 2020-06-20
+author: Jörg Thalheim (@Mic92)
+co-authors: Profpatch (@Profpatsch)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+We are currently having a separate repository to providing NixOS profiles for hardware specific extensions at [nixos-hardware](https://github.com/NixOS/nixos-hardware). This RFC proposes to merge this repository (4000 LOCs/350 commits) and add support for detecting hardware in `nixos-config-generate`.
+
+
+# Motivation
+[motivation]: #motivation
+
+[Nixos-hardware](https://github.com/NixOS/nixos-hardware) relies on specific packages/kernel versions being present in nixpkgs. There is currently no way of preventing nixos-hardware breakages when doing changes in nixpkgs itself. The current best practice is to make sure it does not break with nixpkgs master, which makes less usable for stable users/newcomers. Since it is in an external repository we also cannot integrate  it in [nixos-config-generate](https://github.com/NixOS/nixos-hardware/issues/49) like we do for some other environments like vm nixos profiles because this would require information from the very same repository.
+
+Additionally, some support for hardware is already in nixpkgs (for example [for raspberrypi 3 and 4 images](https://github.com/NixOS/nixpkgs/issues/63720)), but we cannot move them nixos-hardware to not break existing users.
+
+Proper integration in nixpkgs would increase discoverability even just through grep, and enable us to have much more tight tooling around building different kinds of NixOS artifacts, like VMs and images, which would benefit tools like [`nixos-generators`](https://github.com/nix-community/nixos-generators).
+
+# Detailed design
+[design]: #detailed-design
+
+In [PR #91167](https://github.com/NixOS/nixpkgs/pull/91167) there is on-going work on how this will be integrated into nixpkgs. In a nutshell it will be a merge of the [nixos-hardware](https://github.com/NixOS/nixos-hardware)'s git history into nixpkgs.
+
+In a follow up PR we would add `nixos-generate-config` support
+to add profiles based on the result of dmidecode.
+
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+The profiles will be imported like existing profiles:
+
+```nix
+{
+  imports = [
+     <nixpkgs/nixos/hardware/lenovo/thinkpad/t440s>
+  ];
+}
+```
+
+
+# Licensing
+
+The authors believe that [the `nixos-hardware` license](https://github.com/NixOS/nixos-hardware/blob/master/COPYING)
+is CC0, and thus changing it to nixpkgs’ MIT should be possible.
+
+In case it is not legally possible, we will ping all contributors
+(which is feasible looking at the github statistics) and remove
+their contributions if they don’t agree with the license change.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+* nixpkgs increases by some 4000 lines of code and 350 commits.
+* The hardware definitions might be expected to work flawlessly.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Alternatively we can keep nixos-hardware as a separate repository.
+
+There have been only a handful of contributors in a few years,
+which has a few possible explanations:
+
+
+* People write all their hardware definitions manually
+* People use nixos-hardware but don’t contribute back
+* People just don’t know about nixos-hardware (because it’s not discoverable by grepping through nixpkgs)
+* It’s too complicated or too much effort to use/contribute to nixos-hardware, because it’s not in nixpkgs
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+1. How to include the hardware configurations without using `<nixpkgs>`?
+
+   Using`pkgs.path` in module `import`s leads to an infinite recursion.
+   We could have a mechanism similar to `modulesPath` to work around that. 
+   In the beginning, we can just tell people to use `<nixpkgs/nixos/hardware>`. The latter is how nixos-hardware
+   imports profiles right now.
+
+2. How to integrate it into hydra/ofborg
+   
+   We cannot test the workings of the hardware of course,
+   but we can add the image builds (or VM builds) to our hydra
+   to have better integration than before (before there was no integration).
+   
+   Some images can only be built on `aarch64`, but that’s what hydra is already set up for.
+   As a basic minimum we can ensure evaluation of all nixos profiles.
+
+
+# Future work
+[future]: #future-work
+
+* Make nixos/nixpkgs best-in-class for generating images and isos for embedded systems and cloud platforms.
+* Enable full cross-compilation support for generating artifacts for hardware with different architecture, which is common with embedded systems.
+* Provide a more solid basis for third-party efforts like [Mobile NixOS](https://github.com/NixOS/mobile-nixos).
--- a/rfcs/0070-nixos-hardware.md
+++ b/rfcs/0070-nixos-hardware.md
@@ -3,8 +3,8 @@ feature: nixos_hardware
 start-date: 2020-06-20
 author: Jörg Thalheim (@Mic92)
 co-authors: Profpatch (@Profpatsch)
-shepherd-team: (names, to be nominated and accepted by RFC steering committee)
-shepherd-leader: (name to be appointed by RFC steering committee)
+shepherd-team: @samueldr, @garbas, @edolstra, @Kloenk
+shepherd-leader: @Kloenk
 related-issues: (will contain links to implementation PRs)
 ---
 
--- a/rfcs/0070-nixos-hardware.md
+++ b/rfcs/0070-nixos-hardware.md
@@ -11,13 +11,13 @@ related-issues: (will contain links to implementation PRs)
 # Summary
 [summary]: #summary
 
-We are currently having a separate repository to providing NixOS profiles for hardware specific extensions at [nixos-hardware](https://github.com/NixOS/nixos-hardware). This RFC proposes to merge this repository (4000 LOCs/350 commits) and add support for detecting hardware in `nixos-config-generate`.
+We are currently having a separate repository to providing NixOS profiles for hardware specific extensions at [nixos-hardware](https://github.com/NixOS/nixos-hardware). This RFC proposes to merge this repository (4000 LOCs/350 commits) and add support for detecting hardware in `nixos-generate-config`.
 
 
 # Motivation
 [motivation]: #motivation
 
-[Nixos-hardware](https://github.com/NixOS/nixos-hardware) relies on specific packages/kernel versions being present in nixpkgs. There is currently no way of preventing nixos-hardware breakages when doing changes in nixpkgs itself. The current best practice is to make sure it does not break with nixpkgs master, which makes less usable for stable users/newcomers. Since it is in an external repository we also cannot integrate  it in [nixos-config-generate](https://github.com/NixOS/nixos-hardware/issues/49) like we do for some other environments like vm nixos profiles because this would require information from the very same repository.
+[Nixos-hardware](https://github.com/NixOS/nixos-hardware) relies on specific packages/kernel versions being present in nixpkgs. There is currently no way of preventing nixos-hardware breakages when doing changes in nixpkgs itself. The current best practice is to make sure it does not break with nixpkgs master, which makes less usable for stable users/newcomers. Since it is in an external repository we also cannot integrate  it in [nixos-generate-config](https://github.com/NixOS/nixos-hardware/issues/49) like we do for some other environments like vm nixos profiles because this would require information from the very same repository.
 
 Additionally, some support for hardware is already in nixpkgs (for example [for raspberrypi 3 and 4 images](https://github.com/NixOS/nixpkgs/issues/63720)), but we cannot move them nixos-hardware to not break existing users.
 


nh2:
> People just don’t know about nixos-hardware (because it’s not discoverable by grepping through nixpkgs)

Yes, I'm for making it part of nixpkgs, and mentioning this functionality right up in the NixOS manual in the installation section.

Ericson2314:
Yes since day 1 nixos-hardware has been hampered by this: especially given that the hardware compatibility deals with versioning/interface *violations*---subtle variations in software and hardware that have a far more disastrous consequences than intended---I don't think it's feasible to evolve or version nixos and nixos-hardware separately. A means to synchronize everything (as a single repo provides) is absolutely required.

lukegb:
Yes, please: this would make adopting nixos-module much easier. Some of the hardware I use already has modules in nixos-hardware, but since I didn't know about it I ended up reinventing the wheel myself at first.

rfcs/0070-nixos-hardware.md
```diff
@@ -0,0 +1,103 @@
...
+1. How to include the hardware configurations without using `<nixpkgs>`?
+
+   Using`pkgs.path` in module `import`s leads to an infinite recursion.
+   We could have a mechanism similar to `modulesPath` to work around that. 
+   In the beginning, we can just tell people to use `<nixpkgs/nixos/hardware>`. The latter is how nixos-hardware
+   imports profiles right now.
+
+2. How to integrate it into hydra/ofborg
+   
+   We cannot test the workings of the hardware of course,
```
grahamc, line 90: I wonder about validating that they at least evaluate coherently? Understandably we can't actually assert they're sensible.

Profpatsch:
> One problem with the built-in profiles is I find they're good for a while, and then usually I want to override them in some way, sometimes in ways which are difficult to do without copying their code in to my repo.

They would slowly evolve into more composable modules definitions, as with everything.

rfcs/0070-nixos-hardware.md
Profpatsch: Evaluation and probably also build of the images.

grahamc:
I wonder if this would also be a good place to put expressions like NixOS on:

- Packet
- AWS
- Hetzner
- DigitalOcean
- Vultr
- Oracle Cloud

etc.?

grahamc:
This turns out to be a bit of a loaded question,  and I didn't mean for it to be. The reason it is a bit loaded is because to generate the expression for Packet, you have to print out a good bit of information unique to each machine. If moving the packet expr into nixos-hardware is appropriate, then it opens a question for me if the generation should be moved in to nixos-generate-config, too.

Ericson2314:
@grahamc Yeah I don't think we need to figure out every architectural decision up front. Combing the repos to me is just signalling "we can't meaningfully version these separately". Exactly how to integrate the two I rather figure out via experiment----we don't even have to integrate anything at all, and we still get the benefits of synchronous development. (For the record, I do think integrating, somehow, with nixos-generate-config would be good.)

Ericson2314:
Also

> One problem with the built-in profiles is I find they're good for a while, and then usually I want to override them in some way, sometimes in ways which are difficult to do without copying their code in to my repo.

I have the same problem with `nixos-generated-config`, already. On one hand, "Thou shalt not commit generated code", on the other hand wantonly re-running `nixos-generated-config` terrifies me.

Many things we could try. The most complex might be stratifying included/generated code by both "how sure we are about the auto-detection we are making our decision on", and "how essential vs merely recommended is the config item is".

edolstra:
I'm strongly opposed to this. `nixos-hardware` is intended as a low-barrier, little reviewed dumping ground for configurations. It should not be merged into `nixpkgs` which is already way too bloated, is a PR bottleneck, and therefore shouldn't accrue even more functions.

Ma27:
I think that this is actually a good idea, also because it would be possible to generate even more hardware-specific configuration using `nixos-generate-config` with `nixos-hardware` expressions as mentioned in the RFC. @edolstra in case we don't accept this, what would you suggest to do for such a feature? (I'm not opposed to e.g. flakes, but from what I've seen it will take some time until those are stabilized).

> is a PR bottleneck

Hmm, `nixos-hardware` doesn't seem to be a high-traffic repository, so I'm not even sure if we would notice an actual increase of PRs/issues.

> should not be merged into nixpkgs which is already way too bloated

I know that I'll probably be known as the guy who initially hijacked the discussion from now on, but whenever I read something like this when people want to add e.g. a new feature, I wonder if we should reach at least some basic consensus about the direction `nixpkgs` should go in the future first.

bbigras:
Would we have some QA for those hardware profiles?

For example, I made one profile for my laptop and I'm afraid to open a PR since I have no idea if I choose the right stuff. tlp vs thermald, sssd sutff like trim or not.

Ericson2314:
> little reviewed dumping ground for configurations.

I think it's fine to merge alone as-is, but yes we should treat it as initially dubious, and carefully vet afresh any bit that gets included in `nixos-generate-config`.

> I wonder if we should reach at least some basic consensus about the direction `nixpkgs` should go in the future first.

Yes, we really need to have that discussion! (but in a dedicated thread not here.)

> flakes

I don't think flakes will work because the vagaries of comparability are inherently fast moving. A kernel update can change an interface breaking an obscure out-of-tree driver at any time. The version nightmare is a social problem caused mainly by sloppy hardware manufactures that no technical solution like flakes can fix.

samueldr:
This may need considerations for backporting hardware profile changes to "current stable".

The separate repository is good because it generally means the same expressions are used for both stable and unstable branches of Nixpkgs.

Folding this into Nixpkgs, I'm calling it, will make nixos-hardware as good as stale every six months.

While it is useful for older hardware too, I think many users are using it to get around issues with fresh new hardware. If it's getting stale every six months on stable, then it's going to lose most of its attraction.

samueldr:
Another issue, comes with NixOS on ARM which I wanted to write a not-RFC about for the nixos-hardware project.

Right now, [NixOS on ARM](https://nixos.wiki/wiki/NixOS_on_ARM) are guidelines on a wiki, sometimes without actual expressions that are easy to use.

I was interested into folding in SBC expressions into NixOS hardware.

That in itself is not an issue. It's only a couple options. But there is one option that *is* an issue, and is an issue even outside of ARM SBCs.

**Do we want Nixpkgs to start hoarding more and more arbitrarily-sourced forks of the Linux kernel?**

Many ARM SBCs end up needing patches on top of mainline, or an outright fork of the kernel to be usable.

This is not an SBC-specific issue, sometimes other class of all-purpose hardware are also better served with patches or forks of the kernel.

Let's say we add support for a Chromebook range that requires patches for something to work right, until it's part of mainline, or the pinebook pro, which requires some patches or fork of mainline. They're not SBCs, might not even be ARM, but add additional kernels that would be part of Nixpkgs.

In my opinion *even the raspberry pi kernel fork* shouldn't be in Nixpkgs proper. Other single-purpose (as in hardware support) kernel forks shouldn't either.

Though there is a good reason to get them in Nixpkgs, even if I don't like the idea: cached builds! That chromebook user, or even me and my Pinebook would be much better served if the single-purpose forks are built and cached!

  * Is the added Hydra load okay? (probably)
  * Is the added package maintenance workload okay? (probably not)
  * Who updates those kernels? (nobody?)
  * What happens when they get stuck on "bad" versions?

samueldr:
> One problem with the built-in profiles is I find they're good for a while, and then usually I want to override them in some way, sometimes in ways which are difficult to do without copying their code in to my repo. Is there a style guide or best practice we could come up with which makes this easier for users?


I hope there is a way to automatically mark anything part of `<nixpkgs/nixos/hardware/>` as a new default level for options, e.g. `mkOpinionatedDefault` or something of the like, so setting conflicting options in a user's configuration works out of the box, without having to pepper `mkOpinionatedDefault` all over the `<nixpkgs/nixos/hardware/>` expressions.

Ericson2314:
@samueldr So just to understand your argument a bit better, you think that back porting changes to `nixos/hardware` (let's call it) from unstable like we do for other things will not be feasible? Is that because of the shear amount of stuff to backport / rate of change, or because some of them are breaking changes?

Profpatsch:
> Do we want Nixpkgs to start hoarding more and more arbitrarily-sourced forks of the Linux kernel?

We should make a distinction between “common hardware that people want to use and/or only needs a few defaults” and “support every crazy device under the sun including their kernel blobs”.

For reference, this RFC is about the first kind.

Mic92:
> I'm strongly opposed to this. `nixos-hardware` is intended as a low-barrier, little reviewed dumping ground for configurations. It should not be merged into `nixpkgs` which is already way too bloated, is a PR bottleneck, and therefore shouldn't accrue even more functions.

Actually our review standards for nixos-hardware are higher than for nixpkgs. We have branch protection and require at least one review before changes can get merged to it. I don't think nixpkgs is a PR bottleneck either. Only 3 % of all PRs are still open, for Nix it is 15 % and hydra for 16 %. I would say the average time a nixpkgs PR is open is quite significant lower than most open source projects.

Mic92:
> > Do we want Nixpkgs to start hoarding more and more arbitrarily-sourced forks of the Linux kernel?
> 
> We should make a distinction between “common hardware that people want to use and/or only needs a few defaults” and “support every crazy device under the sun including their kernel blobs”.
> 
> For reference, this RFC is about the first kind.

Yes the cost of maintaining a non-mainline kernel is quite high we should not do it.

Mic92:
> I wonder if this would also be a good place to put expressions like NixOS on:
> 
>     * Packet
> 
>     * AWS
> 
>     * Hetzner
> 
>     * DigitalOcean
> 
>     * Vultr
> 
>     * Oracle Cloud
> 
> 
> etc.?

We already have Hypervisor specific profiles in nixpkgs (KVM/Xen etc). Most often they are not vendor specific.

Mic92:
> Would we have some QA for those hardware profiles?
> 
> For example, I made one profile for my laptop and I'm afraid to open a PR since I have no idea if I choose the right stuff. tlp vs thermald, sssd sutff like trim or not.

In nixos-hardware we often rely on other sources like the archlinux wiki. We also got in contact with the [TLP maintainer](https://github.com/NixOS/nixos-hardware/pull/154#issuecomment-617181517) if necessary.

bbigras:
> In nixos-hardware we often rely on other sources like the archlinux wiki. We also got in contact with the TLP maintainer if necessary.

Thank you very much @Mic92. I'll check the arch wiki and create a PR soon for my config.

Mic92:
> One problem with the built-in profiles is I find they're good for a while, and then usually I want to override them in some way, sometimes in ways which are difficult to do without copying their code in to my repo. Is there a style guide or best practice we could come up with which makes this easier for users?

If you have a look at the current code, it is actually quite modularized. Most laptop model profiles only consist of imports of other modules like `common/pc/laptop` or `common/cpu/intel`. Some profile even purely consists of imports. Still we encourage those files to be added so that people have an anchor point to get started with. Things like adding intel microcode updates is not something one would think about adding but is important for security updates.

samueldr:
> @samueldr So just to understand your argument a bit better, you think that back porting changes to `nixos/hardware` (let's call it) from unstable like we do for other things will not be feasible? Is that because of the shear amount of stuff to backport / rate of change, or because some of them are breaking changes?

Feasible? Yes.

Will it be done? Hopefully, but hopefully is not the answer I'd like to hear.

I don't know that there would be breaking changes, or anything like that, but doesn't it scream "out of repo" if you end up implementing the same changes across two-to-three branches in parallel all the time?

timokau:
> Do we want Nixpkgs to start hoarding more and more arbitrarily-sourced forks of the Linux kernel?

I think package-support levels (possibly induced by maintainer support levels) could solve this. That's another RFC to be written (which I won't do though). It doesn't solve the bloat part, but that would be part of the other general-direction discussion.

samueldr:
> (about "complex" cases)

What are the plans for all the "crazy device under the sun" then? I think that arbitrarily limiting this to "simple" cases is going to be harmful in the end, or worst, ignored and some non-simple cases merged in.

This is going to come back to creating two distinct classes of support. This is counter-productive considering a reason to fold `nixos-hardware` into Nixpkgs is for first-class support of the `nixos-hardware` repository.

And then, what's determining that those "crazy" devices are not worth the support, compared to other hardware? Wouldn't an install base of hundreds of "crazy" warrant better first-class support than one one-off device from one user? (Sure, this is disregarding the maintenance cost, but here I assume the maintenance cost is split between all the user of one profile, assuming it comes out basically equal.)

And before ARM SBCs end up shunned for no good reason (yes they are messy), this is not only about ARM SBCs.

Mic92:
> > One problem with the built-in profiles is I find they're good for a while, and then usually I want to override them in some way, sometimes in ways which are difficult to do without copying their code in to my repo. Is there a style guide or best practice we could come up with which makes this easier for users?
> 
> I hope there is a way to automatically mark anything part of `<nixpkgs/nixos/hardware/>` as a new default level for options, e.g. `mkOpinionatedDefault` or something of the like, so setting conflicting options in a user's configuration works out of the box, without having to pepper `mkOpinionatedDefault` all over the `<nixpkgs/nixos/hardware/>` expressions.

Nixos-hardware is pretty strict about avoid opinionated options. Those are usually kept as part of the documentation for the reader to apply if applicable. We use `mkDefault` everywhere so that overriding can be done by users without any hassle.

Mic92:
> This may need considerations for backporting hardware profile changes to "current stable".
> 
> The separate repository is good because it generally means the same expressions are used for both stable and unstable branches of Nixpkgs.
> 
> Folding this into Nixpkgs, I'm calling it, will make nixos-hardware as good as stale every six months.
> 
> While it is useful for older hardware too, I think many users are using it to get around issues with fresh new hardware. If it's getting stale every six months on stable, then it's going to lose most of its attraction.

Testing is only done against unstable. There has been cases where master has broken nixos-hardware because they changed options in nixpkgs without fixing nixos-hardware or nixos-hardware broken a stable release because they adapted those changes without checking backwards compatibility to stable.
Merging this repository in hopefully improves the situation because of better evaluation checks.
Regarding backports the current situation is that we backport all kernels to stable to improve hardware support for staging. We could establish a similar policy for hardware profiles as well. Most configuration there are bugfixes/quirks anyway rather than features.

Mic92:
> > (about "complex" cases)
> 
> What are the plans for all the "crazy device under the sun" then? I think that arbitrarily limiting this to "simple" cases is going to be harmful in the end, or worst, ignored and some non-simple cases merged in.
> 
> This is going to come back to creating two distinct classes of support. This is counter-productive considering a reason to fold `nixos-hardware` into Nixpkgs is for first-class support of the `nixos-hardware` repository.
> 
> And then, what's determining that those "crazy" devices are not worth the support, compared to other hardware? Wouldn't an install base of hundreds of "crazy" warrant better first-class support than one one-off device from one user? (Sure, this is disregarding the maintenance cost, but here I assume the maintenance cost is split between all the user of one profile, assuming it comes out basically equal.)
> 
> And before ARM SBCs end up shunned for no good reason (yes they are messy), this is not only about ARM SBCs.

We should only maintain what want and can maintain. That being said we could instead of merging the whole nixos-hardware repository only migrate profiles where we have a clear active maintainer.

wmertens:
I would love to have this as part of nixpkgs but I don't think it's good enough yet. 

It's all pretty haphazard and disjointed, which makes sense due to the varying degrees of skill and device ownership.

IMHO, we should first grow this repo by advertising it when you install NixOS, as well as mentioning it in the nixpkgs readme and manual, several times.

samueldr:
> We should only maintain what want and can maintain. That might include some hardware that is too hard to support (i.e. this lovely microsoft hardware: NixOS/nixos-hardware#170).

I'm not sure if you're saying this is an example of a "complex" hardware that would be accepted, or not. But this is a good example of a non-SBC that is complex and brings in a bunch of maintenance work.

7c6f434c:
>I don't know that there would be breaking changes, or anything like that, but doesn't it scream "out of repo" if you end up implementing the same changes across two-to-three branches in parallel all the time?

Just like with Firefox and I guess OpenSSL etc.?

On the other hand, doesn't it scream «single repo» when NixOS changes in anything about hardware should be reflected in-sync in hardware definitions?

samueldr:
> That might include some hardware that is too hard to support (i.e. this lovely microsoft hardware: [NixOS/nixos-hardware#170](https://github.com/NixOS/nixos-hardware/pull/170)).

I'm not sure whether this is an example of a "complex" hardware definition with a custom kernel that you would say should or shouldn't be part of Nixpkgs.

But the same points apply, this specific example adds churn and kernels that could get outdated quickly in Nixpkgs.

Mic92:
> I would love to have this as part of nixpkgs but I don't think it's good enough yet.
> 
> It's all pretty haphazard and disjointed, which makes sense due to the varying degrees of skill and device ownership.

I am not quite sure what you mean, can you give example of profiles in a bad shape? Maybe we should remove those than before merging it in.

Mic92:
> > We should only maintain what want and can maintain. That might include some hardware that is too hard to support (i.e. this lovely microsoft hardware: [NixOS/nixos-hardware#170](https://github.com/NixOS/nixos-hardware/pull/170)).
> 
> I'm not sure if you're saying this is an example of a "complex" hardware that would be accepted, or not. But this is a good example of a non-SBC that is complex and brings in a bunch of maintenance work.

We should not maintain something like this unless we have maintainer that put serious commitment to support this. Custom kernel can become easily a maintenance burden when maintaining out-of-tree kernel modules.

FRidh:
> On the other hand, doesn't it scream «single repo» when NixOS changes in anything about hardware should be reflected in-sync in hardware definitions?

And if say a project of mine is tightly connected to a Nixpkgs library, and that library changes (e.g. a major version bump), should my project then also be merged into Nixpkgs because the sync needs to be happening in sync?

@Ma27 is right that we need to define the scope of Nixpkgs. That will be a large discussion and I fear there will not be any agreement. However, by postponing that discussion and decision we will only frustrate each other more and more.

@samueldr already brought up ARM profiles and heck, even the entire of Mobile NixOS. If we go this direction, where do we stop? Should Home-manager go in then as well? And musnix?

Now, I think `nixos-hardware` is small, won't be much of a maintenance burden, and having it readily available for users adds value, but including this opens the gate for other projects to be merged as well. And that I do not want.

Mic92:
> IMHO, we should first grow this repo by advertising it when you install NixOS, as well as mentioning it in the nixpkgs readme and manual, several times.

This is already the case. It's part of the installation manual since September 2019.

wmertens:
> This is already the case. It's part of the installation manual since September 2019.

Yes but my point is that it should be in more places - the GitHub readme, the welcome of the installer image, and several times in the manual.

Mic92:
> > This is already the case. It's part of the installation manual since September 2019.
> 
> Yes but my point is that it should be in more places - the GitHub readme, the welcome of the installer image, and several times in the manual.

But what places except the installation part make sense to include this? I don't want to clutter the manual with inappropriate references. I don't think we should link in the tty to nixos-hardware. I will add it to the README.md of nixpkgs though.

ehmry:
As I recall @garbas and I created this repository to drain the wiki of hardware support articles, and was actually intended as a dumping ground of mediocrity. That's not a slight pointed to anyone of us, I wouldn't claim anything I've committed to this repository was better informed than "seems to work at the moment". I think our hardware support will generally be low assurance, and that's not our fault, that's just the Linux status quo. I say we keep the repositories separate, and exchange expressions between the two if we can establish some quality control.

Mic92:
> As I recall @garbas and I created this repository to drain the wiki of hardware support articles, and was actually intended as a dumping ground of mediocrity. That's not a slight pointed to anyone of us, I wouldn't claim anything I've committed to this repository was better informed than "seems to work at the moment". I think our hardware support will generally be low assurance, and that's not our fault, that's just the Linux status quo. I say we keep the repositories separate, and exchange expressions between the two if we can establish some quality control.

Does this mean we should upstream profiles that are well-tested and have a maintainer?

ehmry:
> Does this mean we should upstream profiles that are well-tested and have a maintainer?
Yes I think so, and we should also have a space where not-well tested stuff can collect.

Mic92:
> > Does this mean we should upstream profiles that are well-tested and have a maintainer?
> 
> Yes I think so, and we should also have a space where not-well tested stuff can collect.

Ok than we are on the same page.

wmertens:
> I don't think we should link in the tty to nixos-hardware. I will add it to the README.md of nixpkgs though.

Actually, I think that when people are running the installer, that is the best time to remind them that there is a repo full of expressions they might be able to use.

Mic92:
> > I don't think we should link in the tty to nixos-hardware. I will add it to the README.md of nixpkgs though.
> 
> Actually, I think that when people are running the installer, that is the best time to remind them that there is a repo full of expressions they might be able to use.

`nixos-install` is largely non-interactive. How would this look like?

Mic92:
There is now a PR for the [README](https://github.com/NixOS/nixpkgs/pull/91215) and [nixos-generate-config](https://github.com/NixOS/nixpkgs/pull/91216)

7c6f434c:
>And if say a project of mine is tightly connected to a Nixpkgs library, and that library changes (e.g. a major version bump), should my project then also be merged into Nixpkgs because the sync needs to be happening in sync?

There should be an expression for this project in Nixpkgs, then we see that we need to pass the previous branch of the library to the project until the upstream bump lands in Nixpkgs.

wmertens:
> `nixos-install` is largely non-interactive. How would this look like?

I meant in the message you see when you run the installer CD and you get a terminal. There, you get instructions about ssh and I think a short blurb like "for many systems there are configurations available at nixos-hardware" would be ok.

Shou:
To me it seems we're trying to move away from managing every package under the sun in the nixpkgs repository through Flakes, so it would make sense that we focus more on the system part of NixOS and including `nixos-hardware` is an obvious win then.

Hypothetically though if we decide we don't want to include this in the nixpkgs codebase, could we not add something like a first-class fetcher for `nixos-hardware` to make it really easy and obvious to include in your system, and then add specific nixos-hardware manual pages (like we have pages specific to nixpkgs, and to all the options provided by NixOS) and also include results in https://nixos.org/nixos/options.html?

Kloenk:
I also would like to keep nixos-hardware outside of nixpkgs. Maybe it would be a good way to add nixos-hardware as a default installed channel? So <nixos-hardware/...> can resolve to something?

garbas:
While [nixos-hardware](https://github.com/NixOS/nixos-hardware) was created to have a one place for people to share and work together on hardware specific configs. I found hard to test any of the hardware specific bugs and you only have to trust that others really tested it. The best you can do is check if things get evaluated.

I do think some hardware specific configuration should be present in nixpkgs and already is (amazon images/azure/rpi/...), but I would limit it (at least for now) for things that are well defined and that we can even test. Maybe this bits can be better defined in how we structure nixpkgs and maybe even document why some "hardware" is documented here and point to nixos-hardware repo.

I do think nixos-hardware *can* make our installation story incredible. I hope one of the first steps could be to just mention its existence on nixos.org in the tutorials/guides/manuals. 

I would really like this RFC to make a smaller step and not trying to do too much. Only an integration of nixos-hardware repository with `nixos-generate-config` might be a big enough step to take and will raise many questions we have to answer. This - only in my opinion - will bring the most value to the end user.

edolstra:
A few more thoughts:

* A lot of nixos-hardware modules are quirks/workarounds for issues that should be handled automatically in NixOS. For example, picking [this random module](https://github.com/NixOS/nixos-hardware/blob/master/apple/macbook-pro/11-5/default.nix), it does a couple of things: 1) It enables the `ati` X11 driver. This can be done by `nixos-generate-config` (and probably already is). 2) It disables 32-bit DRI support because of an issue with `ati_unfree`. This doesn't sound like an Apple-specific issue, so it can be included in `nixos/modules/hardware/video/ati.nix`. 3) It adds a udev rule. This udev rule can probably be included in the default udev ruleset. 4) It enables some SSD-related options like `fstrim`. This can also be done by `nixos-generate-config` if it detects SSDs. In fact, it could even be the default nowadays. 5) It enables Intel CPU related stuff and Broadcom drivers. Again, `nixos-generate-config` can take care of this.

* We don't have a way to test these modules automatically. It was mentioned that they tend to bitrot outside of nixpkgs, but I'd rather have them bitrot outside of nixpkgs than *in* nixpkgs.

Ma27:
> A lot of nixos-hardware modules are quirks/workarounds for issues that should be handled automatically in NixOS

Wouldn't it be a first step then to import nixos-hardware to nixpkgs and then try to refactor "generic fixes" into the distro itself? Having nixos-hardware in nixpkgs may make it easier to perform those changes, after that we'd have a few more hardware-specific profiles without those "hacks".

> We don't have a way to test these modules automatically. It was mentioned that they tend to bitrot outside of nixpkgs, but I'd rather have them bitrot outside of nixpkgs than in nixpkgs.

This is currently in the "Unresolved questions" -section, however it's true that we should discuss about this. What about adding maintainers (with e.g. `meta.maintainers` in the hardware-modules) listing people with the needed hardware (i.e. people who can actually test changes to the module)?

Ofborg should be able to ensure that PRs touching those profiles are reviewed by some qualified testers then.

edolstra:
> Wouldn't it be a first step then to import nixos-hardware to nixpkgs and then try to refactor "generic fixes" into the distro itself?

I don't see that. There is nothing preventing improvements to NixOS without having `nixos-hardware` in tree.

Profpatsch:
At this point I see 25 upvotes and 4 hearts.

It’s a nearly pure bikeshedding topic, with the energy invested in the comments here we could have done the merge (which I already did [here](https://github.com/NixOS/nixpkgs/pull/91160) and which was reverted for licensing reasons) and improved the existing files to fit into nixpkgs. We could have a good basis on which to incrementally improve integration to reach the goals outlined in Further Work.

In my opinion this is a good indication of the state of the RFC process in general.

I had intended this to be a quick win on the weekend, but I’m likely not gonna invest any time into similar ideas in my free time in the future.

Profpatsch:
To be clear: the licensing was never the problem, since nixos-hardware is CC0.

Ma27:
> It’s a nearly pure bikeshedding topic, with the energy invested in the comments

I disagree (don't get me wrong, I'm totally in favor of this RFC as stated above): we still have people (also in the core-team) who are pretty much against it.

> We could have a good basis on which to incrementally improve integration to reach the goals outlined in Further Work.

It's also been mentioned that it's unsure what to do if someone wants us to pull a change in those modules (since it's hard to test hardware-specific configurations) and this is something we should talk about before integrating `nixos-hardware`.

alyssais:
>> It’s a nearly pure bikeshedding topic, with the energy invested in the comments
>
> I disagree (don't get me wrong, I'm totally in favor of this RFC as
> stated above): we still have people (also in the core-team) who are
> pretty much against it.

There is no such thing as a Nix/Nixpkgs "core-team".  There once was,
and it was disbanded (with good reason).

Ma27:
I'm sorry, my wording was wrong. What I wanted to say is that core-contributors (i.e. active ones like @edolstra) have concerns about this.

As I mentioned earlier, I think that we should sooner or later try to reach a consensus about what nixpkgs should look like in the future (i.e. whether it should remain a monorepo). Otherwise we'll have those disagreements over and over again whenever something like this is proposed and that's why I don't consider this discussion a "nearly pure bikeshedding topic".

samueldr:
> At this point I see 25 upvotes and 4 hearts.

So reactions on GitHub can now bypass the RFC process? Good to know.

> It’s a nearly pure bikeshedding topic, with the energy invested in the comments here we could have done the merge [...] and improved the existing files to fit into nixpkgs. We could have a good basis on which to incrementally improve integration to reach the goals outlined in Further Work.

Whilst completely overlooking the concerns.

> In my opinion this is a good indication of the state of the RFC process in general.

[assuming this is about github reactions] I don't think so. GitHub reactions are drive-by feel-good shoot and forget. 

[assuming this is about bikeshedding] Sure, RFCs can end up into purely bikeshedding discussions, but is this really now a purely bikeshedding discssion?

There is a process that has been decided by the community to follow when we [a substantial change to the Nix ecosystem is proposed](https://github.com/NixOS/rfcs#nix-rfcs-request-for-comments). Should this be skipped whenever convenient for you? Maybe we'd rather enshrine that into the RFC process.

> I had intended this to be a quick win on the weekend, but I’m likely not gonna invest any time into similar ideas in my free time in the future.

Sorry that you feel that way, at the very least you made your intents clear in that you believe this should be a part of a mono-repo all-encompassing Nixpkgs. Though it happens that contributors are of the opinion that this is a substantial change that should go through the process.

But please, don't feel discouraged into trying and proposing things. I'm sure it sucks, I'm sure it stings, and the process does make for huge speed bumps and road blocks, but ideas sure are welcome. I don't know what your other "similar ideas" are, hopefully this hasn't squashed interesting changes.

matthewbauer:
It sounds like everyone agrees that we need to improve nixos-generate-config to cover more of the hardware in nixos-hardware. I do think a detection-based approach is going to be much more robust than enumerating every potential hardware configuration (often, even the same laptop will have multiple hardware configurations). But the truth is nixos-generate-config gets very little attention. I think archiving nixos-hardware and forcing everyone to contribute directly to nixos-generate-config would be a really good idea!

Some improvements we could make to nixos-generate-config:

- Rewriting in Python or something else might help get more contributors
- Tweak ACPI configs for acpi_backlight, mem_sleep_default when Linux defaults are wrong. Some rationale for why Linux does it wrong would be nice, instead of just linking to Arch wiki :)
- Occasionally blacklistedModules are needed. This should be accompanied with a Linux bug report explaining for the misbehaving module.
- HiDPI detection - x11 still has poor autodetection so setting xserver.dpi is necessary.
- Initrd/DRM module detection - we need to specify which modules are needed before we can do autodetection
- Ask the user if they would like to enable unfree drivers on hardware that lacks good open source drivers (nvidia)
- Adding more udev rules for ALL configs

There are a few things that I think we want to avoid in NixOS, since they are inherently opinionated, like:

- setting vm.swappiness (users need to choose this for themselves)
- enabling Linux features explicitly marked experimental. Linux detects hardware too, let it choose the default for things like i915.enable_rc6, i915.enable_fbc, i915.enable_psr, and i915.fastboot
- choose a driver when multiple working drivers exist

So I like this PR, but only in the sense that we would merge it directly into nixos-generate-config, instead of just dumping it to Nixpkgs. Note that even in Nixpkgs, we will still struggle to deal with out of date configs given the user can configure which kernel they want.

matthewbauer:
I've also been thinking about a testing strategy for nixos-generate-config. Maybe we could create a semi-automated ISO that:

- boots NixOS like the installer does
- runs nixos-generate-config automatically
- switches to new configuration
- reboots

NixOS users could sign up to spend 30 minutes or so and submit issues. It would be non-destructive so users can just switch back to their previous config afterwards.

7c6f434c:
>NixOS users could sign up to spend 30 minutes or so and submit issues. It would be non-destructive so users can just switch back to their previous config afterwards.

For most of the issues, it does not even need to touch the built-in storage, if we assume USB drive as the boot device…

Of course, storage questions are a separate can of worms: tautologically requiring touching the storage of the main system to test, and there are indeed a few problems (which are even hard to test quickly!) about storage hardware outlight lying with naive interpretation of its lies leading to problems with reliability or wear.

Shados:
@matthewbauer 
> * enabling Linux features explicitly marked experimental. Linux detects hardware too, let it choose the default for things like i915.enable_rc6, i915.enable_fbc, i915.enable_psr, and i915.fastboot
> 
> * choose a driver when multiple working drivers exist

Oddly enough, those are two of the main areas I inevitably end up needing to carefully experiment with to get *cleanly* working graphics on Intel laptops. There may be multiple "working" drivers, but if one or more are demonstrably *problematic* (e.g. screen tearing issues, hardware cursor corruption, random stalls), you'd still prefer not selecting a known-good one as part of nixos-generate-config...? At what point is a driver considered "not working"?

Mic92:
> It sounds like everyone agrees that we need to improve nixos-generate-config to cover more of the hardware in nixos-hardware. I do think a detection-based approach is going to be much more robust than enumerating every potential hardware configuration (often, even the same laptop will have multiple hardware configurations). But the truth is nixos-generate-config gets very little attention. I think archiving nixos-hardware and forcing everyone to contribute directly to nixos-generate-config would be a really good idea!
> 
> Some improvements we could make to nixos-generate-config:
> 
>     * Rewriting in Python or something else might help get more contributors
> 
>     * Tweak ACPI configs for acpi_backlight, mem_sleep_default when Linux defaults are wrong. Some rationale for why Linux does it wrong would be nice, instead of just linking to Arch wiki :)
> 
>     * Occasionally blacklistedModules are needed. This should be accompanied with a Linux bug report explaining for the misbehaving module.
> 
>     * HiDPI detection - x11 still has poor autodetection so setting xserver.dpi is necessary.
> 
>     * Initrd/DRM module detection - we need to specify which modules are needed before we can do autodetection
> 
>     * Ask the user if they would like to enable unfree drivers on hardware that lacks good open source drivers (nvidia)
> 
>     * Adding more udev rules for ALL configs
> 
> 
> There are a few things that I think we want to avoid in NixOS, since they are inherently opinionated, like:
> 
>     * setting vm.swappiness (users need to choose this for themselves)
> 
>     * enabling Linux features explicitly marked experimental. Linux detects hardware too, let it choose the default for things like i915.enable_rc6, i915.enable_fbc, i915.enable_psr, and i915.fastboot
> 
>     * choose a driver when multiple working drivers exist
> 
> 
> So I like this PR, but only in the sense that we would merge it directly into nixos-generate-config, instead of just dumping it to Nixpkgs. Note that even in Nixpkgs, we will still struggle with version variations given the user is free to choose which kernel they want.

I largely agree and made a draft PR to implement this: https://github.com/NixOS/nixpkgs/pull/91353

Mic92:
Wo wants to shepherd this RFC?

samueldr:
I would like to nominate myself for shepherding.

garbas:
I would also like to nominate myself for shepherding.

Mic92:
Ok. We would need two more.

Mic92:
Tomorrow is the next rfc steering committee meeting. The next one is in two weeks. Only around those events it would be opened up for discussion.

Mic92:
Before proceeding with my Rust prototype I also would like some feedback before investing too much time into it.

Kloenk:
@Mic92 maybe we can change `nixos-generate-config`, `nixos-rebuild`, `nixos-install`, `nixos-enter`, `nixos-options`, `nixos-build-vms`, `nixos-container` and `nixos-version` to be inside a package. This also would make testing of your new `nixos-generate-config` much easier.

Ma27:
@Kloenk we had this discussion in the past: https://github.com/NixOS/nixpkgs/pull/48122#issuecomment-430344548

Mic92:
> @Mic92 maybe we can change `nixos-generate-config`, `nixos-rebuild`, `nixos-install`, `nixos-enter`, `nixos-options`, `nixos-build-vms`, `nixos-container` and `nixos-version` to be inside a package. This also would make testing of your new `nixos-generate-config` much easier.

That might be an future perspective, but nothing I would do right now. It's too much work for one person.

mexisme:
*Very* late to this party...

I'm generally a huge advocate for monorepo-style development, but one thing `nixos-hardware` helps is faster-moving updates to what can be a very tricky area for Linux:  hardware that requires a lot of effort to reverse engineer for Linux.

I've found it supremely helpful to have the two repos running at rather different release cadences (even though `nixos-hardware` doesn't actually "release" per se...)
Are there docs for easy ways to overlay a faster-moving set of Nix expressions into the main channel, that doesn't involve a lot of clever Nix and/or creating your own local repo that manually merges the two together?

(I have experience with dealing with stuff like this in larger monorepos, but it may not quite apply cleanly to Nix...)

Mic92:
I made a pull request to include nixos-hardware in the flake registry: https://github.com/NixOS/flake-registry/pull/3 and convert nixos-hardware to a flake: https://github.com/NixOS/nixos-hardware/pull/176

Mic92:
We still need two shepherds until this Thursday to move this RFC to the next phase. If you want to know what the responsibilities of a shepherd is, ping Mic92 in #nixos on IRC (freenode)

edolstra:
I'd like to nominate myself as shepherd.

Kloenk:
I'd like to nominate myself as shepherd too

mexisme:
> I made a pull request to include nixos-hardware in the flake registry

What an awesome idea to do that, @Mic92 !!
Has that not solved many of the issues that started this RFC?

Mic92:
> > I made a pull request to include nixos-hardware in the flake registry
> 
> What an awesome idea to do that, @Mic92 !!
> Has that not solved many of the issues that started this RFC?

1. nixpkgs updates can still break nixos-hardware as there are no evaluation tests. There are different solutions to that. One would be to include nixos-hardware into nixpkgs and the other is to integrate nixos-hardware into the CI.
2. nixos-generate-config is still not very powerfull in apply hardware fixes. I am working on that however currently there is another project in the pipeline that has priority over this. So I will get back to it after it.

flokli:
IIRC, these are NixOS system configuration snippets, which also interact a lot with the rest of the NixOS system. There seems to be the need to keep this tightly coupled. Moving it into a flake doesn't change the fact the code lives in a separate repo, and that pointers need to be manually synchronized. Also, refactorings with `nixpkgs` NixOS code gets much more complicated, as well as testing.

Having to use an [experimental feature](https://github.com/NixOS/rfcs/pull/49#issuecomment-657527216) just to make some configuration snippets available seems a bit risky and overkill, which is also why I'm in favor of this RFC.

In any case, can the RFC Steering Committee move this RFC forward?

Mic92:
> IIRC, these are NixOS system configuration snippets, which also interact a lot with the rest of the NixOS system. There seems to be the need to keep this tightly coupled. Moving it into a flake doesn't change the fact the code lives in a separate repo, and that pointers need to be manually synchronized. Also, refactorings with `nixpkgs` NixOS code gets much more complicated, as well as testing.
> 
> Having to use an [experimental feature](https://github.com/NixOS/rfcs/pull/49#issuecomment-657527216) just to make some configuration snippets available seems a bit risky and overkill, which is also why I'm in favor of this RFC.
> 
> In any case, can the RFC Steering Committee move this RFC forward?

We will tomorrow. Last week there were not enough committee members present to set the shepherd team.

Kloenk:
@samueldr, @garbas, @edolstra, @Kloenk, I created a dudle poll, hope one of the time slots are fitting.
doodle: https://dudle.crypto.koeln/poll.php?poll=22afe196
jitsi: https://meet.pbb.lc/nixos-rfc-70

garbas:
@Kloenk I can only make it 21:00CEST or later. Is it possible to also add this option to the list?

Kloenk:
> @Kloenk I can only make it 21:00CEST or later. Is it possible to also add this option to the list?

In that time I only have to possible dates in the next 2 weeks. I created a new poll: https://dudle.crypto.koeln/poll.php?poll=11447229

edolstra:
@Kloenk Unfortunately it turns out I cannot make it on Tuesday after all :-(

Mic92:
@Kloenk Could we get new dates?

Kloenk:
> @Kloenk Could we get new dates?

Yes, we'll see this evening. School just started for me, so not as much free time anymore, but will find some free dates.

Kloenk:
@samueldr, @garbas, @edolstra, @Kloenk, @Mic92, I created a new poll, hope it fits. made nearly everything after 21:00 CEST. 
also the jitsi link has changed: https://meet.petabyte.dev/nixos-rfc-70
poll: https://dudle.crypto.koeln/poll.php?poll=870f4c6b

zimbatm:
we should also invite the RFC author :-D

Mic92:
Ok. So we meet `Tuesday, 2020-08-18 19:00 UTC`

rfcs/0070-nixos-hardware.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+1. How to include the hardware configurations without using `<nixpkgs>`?
+
+   Using`pkgs.path` in module `import`s leads to an infinite recursion.
+   We could have a mechanism similar to `modulesPath` to work around that. 
+   In the beginning, we can just tell people to use `<nixpkgs/nixos/hardware>`. The latter is how nixos-hardware
+   imports profiles right now.
```
Infinisil, line 86: You can put it under `nixpkgs/nixos/modules/profiles/hardware` instead, then `modulesPath + "/profiles/hardware/..."` can be used to import them. This also makes sense for it to be under `profiles` because nixos-hardware modules are essentially profiles (modules that aren't imported by default).

Mic92:
I hope I can do a writeup about the last meeting soonish. However I am in the process of moving places while having a paper deadline.

Mic92:
It's still in the back of mind and eventually I will get back to it.

roberth:
For module quality we can already use `meta.maintainers` in modules, just like we do with packages. This way hardware owners can review each other's changes.

Mic92:
As cooperation between Numtide and Clan, we are making a new attempt to build a nixos hardware configuration tool. If we have fleshed out enough details, I will re-open the RFC.
The idea is that we make the tool generate more raw hardware information facts and that NixOS modules than can use to pull in the right configuration. This is also roughly what we discussed in the first and only meeting of this RFC.

infinisil:
https://discourse.nixos.org/t/better-hardware-detection-with-nixos-facter/49813

mexisme:
> As cooperation between Numtide and Clan, we are making a new attempt to build a nixos hardware configuration tool. If we have fleshed out enough details, I will re-open the RFC.
> The idea is that we make the tool generate more raw hardware information facts and that NixOS modules than can use to pull in the right configuration. This is also roughly what we discussed in the first and only meeting of this RFC. 

Very cool!

Side note:

Would be especially keen to understand how "exceptions" are handled.
Things that come to mind incl. the multiple extra patches and config. needed for `linux-surface` or some of the unusual interim/temporary tricks required to help enable certain drivers.

Mic92:
> > As cooperation between Numtide and Clan, we are making a new attempt to build a nixos hardware configuration tool. If we have fleshed out enough details, I will re-open the RFC.
> > The idea is that we make the tool generate more raw hardware information facts and that NixOS modules than can use to pull in the right configuration. This is also roughly what we discussed in the first and only meeting of this RFC.
> 
> Very cool!
> 
> Side note:
> 
> Would be especially keen to understand how "exceptions" are handled. Things that come to mind incl. the multiple extra patches and config. needed for `linux-surface` or some of the unusual interim/temporary tricks required to help enable certain drivers.

We don't want to deprecate nixos-hardware but simplify it. We can use nixos-facter to import modules from nixos-hardware as needed i.e. using sku names.

roberth:
I like that a profile-like structure implements a [minimal modules list](https://github.com/NixOS/rfcs/pull/22) for this area of configuration. `nixos-hardware` is a bit to hierarchical and arbitrary, but adding imports based on hardware is good for evaluation performance.

PR #64, opened 2020-01-05
[RFC 0064] New Documentation Format for nixpkgs and NixOS

Infinisil:
Many people from the Nix community are interested in evaluating alternatives to DocBook as the documentation format for nixpkgs/NixOS.
Through the process of this RFC, a potentially new doc format will be decided.

In short, with this RFC a set of requirements for a doc format is decided, after which candidates that fulfil them are collected and evaluated. A public poll is then held allowing everybody to vote which formats they'd prefer. The result of this poll is then to be used by the shepherds to come to a final decision.

[Rendered](https://github.com/Infinisil/rfcs/blob/docformat/rfcs/0064-documentation-format.md)

Pinging some people who have shown interest in this topic and/or are knowledgeable in certain formats: @edolstra @grahamc @domenkozar @zimbatm @alyssais @peterhoeg @danbst @chreekat

PR #72, opened 2020-07-13
[RFC 0072] Switch to CommonMark for documentation

mboes:
This RFC proposes to convert the documentation for the Nix, Nixpkgs and NixOS projects from Docbook to CommonMark Markdown. It proposes requirements for the appearance of web facing documentation generated from Markdown. It does not mandate a choice of toolchain to generate a website with documentation. The choice is left to the implementers and maintainers.

[rendered](https://github.com/mboes/rfcs/blob/markdown-docs/rfcs/0000-commonmark-docs.md)

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/rfcs/72)

### DIFF
--- a/rfcs/0000-commonmark-docs.md
+++ b/rfcs/0000-commonmark-docs.md
@@ -136,7 +136,7 @@ demo below, it was easy to fix 5 bugs already.
 We propose to keep the man pages as Docbook for now. They are
 self-contained documents, whose form is constrained by convention and
 the limits of the man page format. When the time comes to convert the
-man pages as well, we can turn here again to prior art. The Kurnetes
+man pages as well, we can turn here again to prior art. The Kubernetes
 project uses [md2man][md2man] to generate man pages from CommonMark.
 This is a small Go command with a 4.1MB closure size (including 2MB
 for `tzdata`).
--- a/rfcs/0000-commonmark-docs.md
+++ b/rfcs/0000-commonmark-docs.md
@@ -166,7 +166,7 @@ https://nixos-docs-mockup.netlify.app
 It features the following sections of the Nix manual:
 * introduction
 * quickstart and installation
-* the Nix experssions chapter
+* the Nix expressions chapter
 * advanced topics
 
 The content is the unedited output of running the Pandoc conversion,
--- a/rfcs/0000-commonmark-docs.md
+++ b/rfcs/0000-commonmark-docs.md
@@ -144,6 +144,7 @@ for `tzdata`).
 [gatsby]: https://gatsbyjs.org
 [sphinx]: https://www.sphinx-doc.org
 [pandoc]: https://pandoc.org/
+[md2man]: https://github.com/cpuguy83/go-md2man
 
 # Examples and Interactions
 [examples-and-interactions]: #examples-and-interactions
--- a/rfcs/0000-commonmark-docs.md
+++ b/rfcs/0000-commonmark-docs.md
@@ -81,8 +81,8 @@ humans and their toolchains is Markdown. It's all the more easy to
 create good looking documentation with Markdown that the tools
 available to process it are plentiful and flexible (JavaScript
 converters with plugin support like [Remark][remark], [static site
-generators][staticgen] by the dozen, [MDX][mdx] to extend CommonMark with
-arbitrary React components, etc).
+generators][staticgen] by the dozen, [MDX][mdx] to extend CommonMark
+with arbitrary React components, etc).
 
 [gatsby-docs]: https://www.gatsbyjs.org/docs/
 [gitstar-rankings]: https://gitstar-ranking.com/repositories
@@ -98,9 +98,11 @@ arbitrary React components, etc).
 # Detailed design
 [design]: #detailed-design
 
+## Documentation website requirements
+
 AsciiDoc, CommonMark and RST are all formats that support basic
 markup: emphasis, bold, (nested) (un)numbered lists, headings, inline
-and display code, tables (as HTML in the case of CommonMark), etc. The
+and display code, tables (using a CommonMark extension), etc. The
 requirements we set below pertain to the appearance of the
 documentation available on the website.
 
@@ -121,12 +123,36 @@ The key requirements we work towards are:
    a statement to newcomers that the community takes documentation
    seriously, with both good form and good content.
 
+## Choice of format
+
 Satisfying all requirements above is possible with a number of
 toolchains, including Asciidoc-specific or RST-specific toolchains
-(not just Markdown). We propose CommonMark as the documentation
-format. The choice of toolchain is left at the discretion of the
-implementers. We feature two demos below (one uses [Gatsby][gatsby]
-and another uses [Sphinx][sphinx]).
+(not just Markdown). We propose CommonMark plus a small number of
+extensions as the documentation format. The choice of toolchain is
+left at the discretion of the implementers. We feature two demos below
+(one uses [Gatsby][gatsby] and another uses [Sphinx][sphinx]).
+
+The choice of CommonMark extensions is also left to the implementors.
+However, this RFC stipulates the following guidelines:
+
+* The overall number of extensions should be kept small, to facilitate
+  interoperability. The goal is not perfect compliance with a standard
+  (i.e. pure CommonMark), but it should nonetheless remain easy to
+  switch from one toolchain to another for generating HTML from the
+  CommonMark source, with minimal manual work. For example,
+  * enabling an extension for tables is acceptable because few tables
+  appear in the documentation and converting them by hand to a new
+  format is not labour intensive;
+  * YAML frontmatter for metadata is also acceptable, because nearly
+    all toolchains support this.
+* It is acceptable to introduce a new extension only if this extension
+  is supported by three or more popular toolchains. Toolchain-specific
+  extensions should not be used.
+* CommonMark allows HTML span and block elements. These should be
+  avoided in documentation source, because this complicates targeting
+  multiple output formats (e.g. man pages, epub, etc).
+
+## Transition to CommonMark
 
 The one-time conversion from Docbook to CommonMark is lossy, because
 CommonMark has far less expressive markup. It is done using
@@ -135,13 +161,15 @@ a high-quality writer available for CommonMark. The Pandoc Docbook
 reader isn't perfect, but in the process of putting together the quick
 demo below, it was easy to fix 5 bugs already.
 
-We propose to keep the man pages as Docbook for now. They are
-self-contained documents, whose form is constrained by convention and
-the limits of the man page format. When the time comes to convert the
-man pages as well, we can turn here again to prior art. The Kubernetes
-project uses [md2man][md2man] to generate man pages from CommonMark.
-This is a small Go command with a 4.1MB closure size (including 2MB
-for `tzdata`).
+We propose to convert man pages to CommonMark as well. However, this
+transition need not happen concurrently to the format transition for
+the rest of the documentation. They are self-contained documents,
+whose form is constrained by convention and the limits of the man page
+format. When the time comes to convert the man pages as well, we can
+turn here again to prior art. The Kubernetes project uses
+[md2man][md2man] to generate man pages from CommonMark. This is
+a small Go command with a 4.1MB closure size (including 2MB for
+`tzdata`).
 
 [gatsby]: https://gatsbyjs.org
 [sphinx]: https://www.sphinx-doc.org
@@ -193,12 +221,12 @@ CommonMark as the input format doesn't force unreasonable compromises.
 The current DocBook format is semantically richer. There are specific
 tags for definitions, environment variables, user accounts, various
 types of callouts, etc. CommonMark's data model isn't nearly as rich,
-so in converting to CommonMark, some information is lost. However,
-experience in other very large ecosystems with many users tells us
-that authors are happy to make do with an inexpressive but familiar
-format, which in any case *can* be extended with the wise use of
-`<span>`-like HTML tags and custom tags that expand to HTML. That
-appears to be seldom necessary in practice (see e.g. the five
+so in converting to CommonMark, even with extensions, some information
+is lost. However, experience in other very large ecosystems with many
+users tells us that authors are happy to make do with an inexpressive
+but familiar format, which in any case *can* be extended with the wise
+use of `<span>`-like HTML tags and custom tags that expand to HTML.
+That appears to be seldom necessary in practice (see e.g. the five
 documentation examples in [Motivation][motivation]).
 
 # Alternatives
--- a/rfcs/0000-commonmark-docs.md
+++ b/rfcs/0000-commonmark-docs.md
@@ -2,7 +2,7 @@
 feature: commonmark-docs
 start-date: 2020-07-05
 author: mboes
-shepherd-team: Rok Garbas, zimbatm, Infinisil
+shepherd-team: garbas, zimbatm, Kloenk
 shepherd-leader: Infinisil
 ---
 
--- a/rfcs/0000-commonmark-docs.md
+++ b/rfcs/0000-commonmark-docs.md
@@ -236,7 +236,7 @@ Pros:
 
 - reStructuredText (RST) has wider adoption than Asciidoc.
 - Most Python projects use it. The Python ecosystem is deep and wide.
-- Good and mature toolchains like Sphynx exist to process RST files.
+- Good and mature toolchains like Sphinx exist to process RST files.
 
 Cons:
 
--- a/rfcs/0000-commonmark-docs.md
+++ b/rfcs/0000-commonmark-docs.md
@@ -78,8 +78,8 @@ contributors. The *lingua franca* across subcommunities for both
 humans and their toolchains is Markdown. It's all the more easy to
 create good looking documentation with Markdown that the tools
 available to process it are plentiful and flexible (JavaScript
-converters with plugin support like [Remark][remark], static site
-generators by the dozen, [MDX][mdx] to extend CommonMark with
+converters with plugin support like [Remark][remark], [static site
+generators][staticgen] by the dozen, [MDX][mdx] to extend CommonMark with
 arbitrary React components, etc).
 
 [gatsby-docs]: https://www.gatsbyjs.org/docs/
@@ -91,6 +91,7 @@ arbitrary React components, etc).
 [vscode-docs]: https://code.visualstudio.com/docs
 [mdx]: https://github.com/mdx-js/mdx
 [remark]: https://remark.js.org/
+[staticgen]: https://www.staticgen.com/
 
 # Detailed design
 [design]: #detailed-design
--- a/rfcs/0000-commonmark-docs.md
+++ b/rfcs/0000-commonmark-docs.md
@@ -1,0 +1,302 @@
+---
+feature: commonmark-docs
+start-date: 2020-07-05
+author: mboes
+---
+
+# Summary
+[summary]: #summary
+
+Nix, Nixpkgs and NixOS documentation is currently in Docbook format.
+We propose to migrate all existing content to CommonMark, a flavour of
+Markdown.
+
+# Motivation
+[motivation]: #motivation
+
+Documentation ought to be easy to write, easy to maintain, easy to
+make pretty, and its format be so boring as to seldom be the object of
+future RFC's. But beyond that, the significance of the documentation
+format is in its very real impact on the perception of the project.
+Documentation is the storefront of any project online in addition to
+its front page. Modern looking documentation that can be edited at the
+click of a button sends a signal: that the Nix* projects are forward
+looking, are welcoming to novice contributors and feel familiar.
+
+The motivation to switch away from Docbook is that it's unfamiliar.
+None of the [top 100 projects on Github][gitstar-rankings] use
+Docbook. The motivation to switch to CommonMark is that among these
+projects, since roughly 10 times more projects use Markdown (or some
+variation thereof) than all other documentation formats combined, it
+can be considered the default. It's worth considering alternatives,
+but the burden ought to lie with the proponents of other format/tool
+combinations to convince this community that following the precedent
+set by the overwhelming majority of communities larger than ours won't
+work well for us.
+
+Increasing the number of contributors to the documentation, increasing
+its coverage and improving its quality are social goals, not technical
+requirements. To recruit more writers and users, should we make
+support for callouts, admonitions and a precise taxonomy (like Docbook
+has) requirements? Should literate programming and the ability to do
+transclusions also be requirements? Does reducing documentation build
+times from seconds to milliseconds matter? Maybe meeting these
+requirements helps us reach our social goals. Maybe they are
+immaterial. What we can go by is the following piece of evidence:
+nearly all of the most beautifully laid out, high-quality and
+easy-to-navigate documentation in the GitHub Top 100 are in
+CommonMark. The only notable exception is the [Tensorflow
+documentation][tensorflow-docs], which uses Jupyter notebooks for
+everything.
+
+Consider the following 5 examples of great Markdown documentation in
+terms of breadth, presentation and richness (cross references,
+definition lists, tables, callouts, integration with playgrounds,
+etc):
+
+- [GatsbyJS documentation][gatsby-docs] (using Gatsby with [MDX][mdx])
+- Facebook's [React documentation][react-docs] (using Gatsby with
+  plain CommonMark)
+- Microsoft's [VS Code documentation][vscode-docs] (using Jekyll)
+- [Kubernetes documentation][kubernetes-docs] (using Hugo)
+- The [Rust book][rust-docs] (using mdBook)
+
+The goal of this RFC is to change the *form* of the current
+documentation for Nix, Nixpkgs and NixOS to look similar to any one of
+the above 5 projects (see requirements in the next section). We submit
+that the least effort route to do so is to use the same toolchain as
+they do (CommonMark or MDX input and one of Gatsby, Jekyll or Hugo to
+generate a static site).
+
+What all of the projects cited above have in common is a large
+contributor base with heterogeneous skillsets and multiple
+subcommunities. Just like in Nixpkgs, they cannot leverage the
+hegemony of RST in Python, because not all their users are Python
+programmers. They prefer not to count on writers learning Docbook or
+Asciidoc, because many documentation patches are from casual
+contributors. The *lingua franca* across subcommunities for both
+humans and their toolchains is Markdown. It's all the more easy to
+create good looking documentation with Markdown that the tools
+available to process it are plentiful and flexible (JavaScript
+converters with plugin support like [Remark][remark], static site
+generators by the dozen, [MDX][mdx] to extend CommonMark with
+arbitrary React components, etc).
+
+[gatsby-docs]: https://www.gatsbyjs.org/docs/
+[gitstar-rankings]: https://gitstar-ranking.com/repositories
+[kubernetes-docs]: https://kubernetes.io/docs/home/
+[react-docs]: https://reactjs.org/docs/getting-started.html
+[rust-docs]: https://doc.rust-lang.org/book/
+[tensorflow-docs]: https://github.com/tensorflow/docs
+[vscode-docs]: https://code.visualstudio.com/docs
+[mdx]: https://github.com/mdx-js/mdx
+[remark]: https://remark.js.org/
+
+# Detailed design
+[design]: #detailed-design
+
+AsciiDoc, CommonMark and RST are all formats that support basic
+markup: emphasis, bold, (nested) (un)numbered lists, headings, inline
+and display code, tables (as HTML in the case of CommonMark), etc. The
+requirements we set below pertain to the appearance of the
+documentation available on the website.
+
+The key requirements we work towards are:
+
+1. easy-to-find "Edit on GitHub" button to increase drive-by
+   contributions,
+1. good quality documentation search engine,
+1. syntax highlighting of all code,
+1. separate page per chapter, instead of the current monolithic page
+   for each manual,
+1. table of content for each chapter available in a side bar to easily
+   jump through long content,
+1. make warning and info boxes stand out from the rest of the content
+   with colour,
+1. easy to customize HTML and CSS, since tired old templates we've all
+   seen too many times are best avoided. The objective is to make
+   a statement to newcomers that the community takes documentation
+   seriously, with both good form and good content.
+
+Satisfying all requirements above is possible with a number of
+toolchains, including Asciidoc-specific or RST-specific toolchains
+(not just Markdown). We propose CommonMark as the documentation
+format. The choice of toolchain is left at the discretion of the
+implementers. We feature two demos below (one uses [Gatsby][gatsby]
+and another uses [Sphinx][sphinx]).
+
+The one-time conversion from Docbook to CommonMark is lossy, because
+CommonMark has far less expressive markup. It is done using
+[Pandoc][pandoc], which has a reader available for Docbook and
+a high-quality writer available for CommonMark. The Pandoc Docbook
+reader isn't perfect, but in the process of putting together the quick
+demo below, it was easy to fix 5 bugs already.
+
+We propose to keep the man pages as Docbook for now. They are
+self-contained documents, whose form is constrained by convention and
+the limits of the man page format. When the time comes to convert the
+man pages as well, we can turn here again to prior art. The Kurnetes
+project uses [md2man][md2man] to generate man pages from CommonMark.
+This is a small Go command with a 4.1MB closure size (including 2MB
+for `tzdata`).
+
+[gatsby]: https://gatsbyjs.org
+[sphinx]: https://www.sphinx-doc.org
+[pandoc]: https://pandoc.org/
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This RFC is *not* about vetting a single implementation choice. The
+five large projects cited above use one of Hugo, Gatsby, Jekyll, or
+mdBook. We can add Sphinx to the set of available implementation
+choices, which likewise has excellent support for Markdown. In this
+section we show how two of these choices could work in the context of
+the Nix* projects. *Selecting a toolchain is left to the discretion of
+the implementors. Discussing implementation choices that are
+transparent to the user is out of scope for an RFC process*.
+
+The following demo shows how to satisfy all requirements above using
+CommonMark as the input format and a Gatsby-based [starter kit from
+Hasura][hasura-docs-starter] to generate a ready-to-deploy static
+website:
+
+https://nixos-docs-mockup.netlify.app
+
+It features the following sections of the Nix manual:
+* introduction
+* quickstart and installation
+* the Nix experssions chapter
+* advanced topics
+
+The content is the unedited output of running the Pandoc conversion,
+which still has bugs (in particular the handling of cross-references).
+
+This other demo shows how to satisfy many of the same requirements
+using Sphinx, again with CommonMark as the input:
+
+https://nixpkgs-manual-sphinx-markedown-example.netlify.app
+
+The look and feel is customizable, and indeed could be made the same
+in both cases. Both are examples meant to demonstrate that choosing
+CommonMark as the input format doesn't force unreasonable compromises.
+
+[hasura-docs-starter]: https://github.com/hasura/gatsby-gitbook-starter
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The current DocBook format is semantically richer. There are specific
+tags for definitions, environment variables, user accounts, various
+types of callouts, etc. CommonMark's data model isn't nearly as rich,
+so in converting to CommonMark, some information is lost. However,
+experience in other very large ecosystems with many users tells us
+that authors are happy to make do with an inexpressive but familiar
+format, which in any case *can* be extended with the wise use of
+`<span>`-like HTML tags and custom tags that expand to HTML. That
+appears to be seldom necessary in practice (see e.g. the five
+documentation examples in [Motivation][motivation]).
+
+# Alternatives
+[alternatives]: #alternatives
+
+Other formats share the desirable properties of the various flavours
+of Markdown (of which CommonMark is but one):
+
+- textual format that is easy to diff,
+- reuses familiar conventions from decades of plain text emails,
+- terseness of the markup and consistent levels of indentation.
+
+The main two other contenders are AsciiDoc and RST. There are
+technical reasons to prefer either of them to CommonMark, despite
+social factors in favour of CommonMark.
+
+## AsciiDoc
+
+AsciiDoc has the same data model as Docbook - just a different syntax.
+This makes the two formats interchangeable in principle. The markup
+language is arguably better designed than CommonMark and is more
+expressive. However, AsciiDoc is not as precisely specified as
+CommonMark. It has only two extant toolchains ([Asciidoc][asciidoc]
+and [Asciidoctor][asciidoctor]) for transforming into HTML, one of
+which is infrequently maintained (3 releases in 7 years). In the top
+100 projects on GitHub, only one project uses Asciidoc for their
+documentation format.
+
+Asciidoc is also effectively a format transformer dead-end:
+round-tripping via Docbook using `asciidoc` and `docbookrx` doesn't
+work and Pandoc does not support Asciidoc as an input.
+
+[asciidoc]: https://asciidoc.org/
+[asciidoctor]: https://asciidoctor.org/
+
+## reStructuredText
+
+Pros:
+
+- reStructuredText (RST) has wider adoption than Asciidoc.
+- Most Python projects use it. The Python ecosystem is deep and wide.
+- Good and mature toolchains like Sphynx exist to process RST files.
+
+Cons:
+
+- the syntax for simple things like links or inline code is
+  non-standard. Inline code requires double backticks, but single
+  backticks is legal syntax and used for links, so it's easy to get
+  things wrong that don't lead to build errors.
+- RST constructs do not compose nearly as well as they should. What is
+  trivial in CommonMark is impossible in RST. For example, links with
+  inline code in the title are not expressible. Links with italics in
+  the title are not expressible either. In general, inline markup does
+  not concatenate well. Backslashes are required in common constructs
+  like the following:
+  ```
+  Python ``list``\s use square bracket syntax.
+  This is a long\ *ish* paragraph
+  ```
+- nesting block elements suffers from gotchas (like [this
+  one][rst-nested-lists]).
+
+[rst-nested-lists]: https://stackoverflow.com/questions/44557957/unexpected-indentation-with-restructuredtext-list
+
+## Are they popular?
+
+The extra expressiveness of Asciidoc or RST over CommonMark was not
+deemed to be a crucial requirement by many other large projects. In
+the GitHub Top 100 (projects ranked by number of stars), only one
+project (Spring Boot) uses Asciidoc, and only two projects (Ansible
+and Linux) use RST other than the predominantly Python projects. Swift
+uses a mix of RST and Markdown, what with being in the middle of
+a transition to full Markdown (grep for `[Gardening] De-RST` in [the
+history][swift-docs-history]).
+
+[swift-docs-history]: https://github.com/apple/swift/tree/master/docs
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+We propose to take CommonMark as a baseline. This is sufficient to
+support all the markup we need. It's also a stepping stone towards
+[MDX][mdx], which enables embedding custom widgets if so desired. For
+example, Gatsby uses MDX to include newsletter signup forms, embed
+videos from third-party training websites, or to generate
+documentation sitemaps within a section (e.g. the bottom of [this
+page][gatsby-graphql]). It's unclear that custom widgets are worth the
+trouble at this stage.
+
+The choice of toolchain to publish the documentation on the website
+and as man pages is left open.
+
+[gatsby-graphql]: https://www.gatsbyjs.org/docs/graphql/
+
+# Future work
+[future]: #future-work
+
+- Build out the documentation demo above into a full replacement for
+  each of the Nix, Nixpks and NixOS user manuals online.
+- Fix all Pandoc bugs encountered along the way, in particular to get
+  working cross-references.
+- Customize the CSS and layout of the demo to be more in line with Nix
+  branding.
+- Integrate the result into the Nix and Nixpkgs repositories and hook
+  into their respective CI/CD pipelines.
--- a/rfcs/0000-commonmark-docs.md
+++ b/rfcs/0000-commonmark-docs.md
@@ -2,6 +2,8 @@
 feature: commonmark-docs
 start-date: 2020-07-05
 author: mboes
+shepherd-team: Rok Garbas, zimbatm, Infinisil
+shepherd-leader: Infinisil
 ---
 
 # Summary
--- a/rfcs/0072-commonmark-docs.md
+++ b/rfcs/0072-commonmark-docs.md
@@ -9,7 +9,7 @@ shepherd-leader: Infinisil
 # Summary
 [summary]: #summary
 
-Nix, Nixpkgs and NixOS documentation is currently in Docbook format.
+Nixpkgs and NixOS documentation is currently in Docbook format.
 We propose to migrate all existing content to CommonMark, a flavour of
 Markdown.
 
@@ -28,7 +28,7 @@ looking, are welcoming to novice contributors and feel familiar.
 The motivation to switch away from Docbook is that it's unfamiliar.
 None of the [top 100 projects on Github][gitstar-rankings] use
 Docbook. The motivation to switch to CommonMark is that among these
-projects, since roughly 10 times more projects use Markdown (or some
+projects, roughly 10 times more projects use Markdown (or some
 variation thereof) than all other documentation formats combined, it
 can be considered the default. It's worth considering alternatives,
 but the burden ought to lie with the proponents of other format/tool
@@ -64,7 +64,7 @@ etc):
 - The [Rust book][rust-docs] (using mdBook)
 
 The goal of this RFC is to change the *form* of the current
-documentation for Nix, Nixpkgs and NixOS to look similar to any one of
+documentation for Nixpkgs and NixOS to look similar to any one of
 the above 5 projects (see requirements in the next section). We submit
 that the least effort route to do so is to use the same toolchain as
 they do (CommonMark or MDX input and one of Gatsby, Jekyll or Hugo to
@@ -145,9 +145,9 @@ However, this RFC stipulates the following guidelines:
   format is not labour intensive;
   * YAML frontmatter for metadata is also acceptable, because nearly
     all toolchains support this.
-* It is acceptable to introduce a new extension only if this extension
-  is supported by three or more popular toolchains. Toolchain-specific
-  extensions should not be used.
+* An extension for defining references between section should be
+  supported, since this is widely used in the current manuals and
+  essential for navigating around.
 * CommonMark allows HTML span and block elements. These should be
   avoided in documentation source, because this complicates targeting
   multiple output formats (e.g. man pages, epub, etc).


rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,302 @@
...
+CommonMark as the input format and a Gatsby-based [starter kit from
+Hasura][hasura-docs-starter] to generate a ready-to-deploy static
+website:
+
+https://nixos-docs-mockup.netlify.app
+
+It features the following sections of the Nix manual:
+* introduction
+* quickstart and installation
+* the Nix experssions chapter
```
Profpatsch, line 169:
```suggestion
* the Nix expressions chapter
```

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,302 @@
...
+The key requirements we work towards are:
+
+1. easy-to-find "Edit on GitHub" button to increase drive-by
+   contributions,
+1. good quality documentation search engine,
+1. syntax highlighting of all code,
+1. separate page per chapter, instead of the current monolithic page
+   for each manual,
+1. table of content for each chapter available in a side bar to easily
+   jump through long content,
```
Profpatsch, line 113: The “good quality search engine” part becomes less of a problem when you put everything on one page, because the browser’s search bar is a tool that works everywhere, without internet connection.

Do we have too much documentation to fit everything on one page? Or what is the reasoning behind splitting it up?

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,302 @@
...
+they do (CommonMark or MDX input and one of Gatsby, Jekyll or Hugo to
+generate a static site).
+
+What all of the projects cited above have in common is a large
+contributor base with heterogeneous skillsets and multiple
+subcommunities. Just like in Nixpkgs, they cannot leverage the
+hegemony of RST in Python, because not all their users are Python
+programmers. They prefer not to count on writers learning Docbook or
+Asciidoc, because many documentation patches are from casual
+contributors. The *lingua franca* across subcommunities for both
```
domenkozar, line 77: That's exactly where the marketing and reality of Markdown differs. There are many standards of Markdown and most tooling implements their own format, meaning you're not using Markdown but a flavor.

For that reason it's important to talk about the tooling used, as each documentation tool will use a different flavor and thus define the "vendor lock-in".

mboes: That's why the title of this RFC has CommonMark in it, not Markdown. CommonMark is a [well specified](https://spec.commonmark.org/current/) dialect of Markdown that is also a de facto standard (understood by GitHub, Sphinx, Gatsby, Pandoc and many others). We propose to use CommonMark for now: nothing added, nothing less.

domenkozar: That's quite nice in theory, but in practice the need for extensions quickly arises. My bet is we'll use one even as we port the existing documentation. 

For example there are no footnotes, tables in CommonMark, what do you propose to do with the existing ones?

rfcs/0000-commonmark-docs.md
mboes: That's a good point. The reasoning for the split-up is that this is what other projects with extensive documentation do as well. Presumably to make the documentation less intimidating (imagine coming to a project for the first time and realizing you're going to need to wade through 40k+ words of documentation...), but I'm not a tech writer so I'm not sure whether there are strong reasons.

Navigating from section to section currently isn't easy. But that can be solved with a sidebar TOC whether we're in single-page or multi-page.

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,302 @@
...
+The current DocBook format is semantically richer. There are specific
+tags for definitions, environment variables, user accounts, various
+types of callouts, etc. CommonMark's data model isn't nearly as rich,
+so in converting to CommonMark, some information is lost. However,
+experience in other very large ecosystems with many users tells us
+that authors are happy to make do with an inexpressive but familiar
+format, which in any case *can* be extended with the wise use of
+`<span>`-like HTML tags and custom tags that expand to HTML. That
+appears to be seldom necessary in practice (see e.g. the five
+documentation examples in [Motivation][motivation]).
```
domenkozar, line 198: Drawback: no support for PDF, epub or manual pages.

rfcs/0000-commonmark-docs.md
asymmetric: I do think that the wall of text that are the current manuals is intimidating for first-time users. A collapsible TOC on the side would greatly help, plus it also degrades gracefully to the current experience when one has no JS.

I do think a one-page-man is more annoying (and resource-intentive) on mobile devices, though.

Emacsy projects usually have both split and unified manpages; that could be another option.

rfcs/0000-commonmark-docs.md
domenkozar: TOC is another example of such feature.

domenkozar: Only looking at the rust markdown, it's not just CommonMark as it uses extensions for includes: https://github.com/rust-lang/book/blob/master/src/ch18-03-pattern-syntax.md

FRidh: Right, same with in-document links and generation of anchors.

mboes: > but in practice the need for extensions quickly arises.

I don't believe that's a foregone conclusion. Don't take my word for it: for example, all of React's documentation is in plain CommonMark. This is why I cite 5 large and incredibly popular projects (according to GitHub) with (IMHO) great documentation in the motivation section: they all use plain CommonMark except for one (Gatsby), which uses MDX. I'm very familiar with the Gatsby documentation since I use it a lot in my day job. They use MDX to embed custom components like newsletter sign-up forms and embedding videos to third-party training websites. No footnotes. Tables are supported the way Pandoc translates them: using HTML tags. See [here](https://nixos-docs-mockup.netlify.app/advanced-topics) for an example in one of the demos (ignore the headings, which have broken cross-references due to a Pandoc bug).

domenkozar: I've just pointed out that your examples don't use only CommonMark. I can go through the rest of them but I'll be surprised if most of them don't use extensions. 

As soon as you start using HTML, you lose portability of any kind. 

https://www.ericholscher.com/blog/2016/mar/15/dont-use-markdown-for-technical-docs/ well summarized these issues with markdown.

rfcs/0000-commonmark-docs.md
grahamc: Our single page manual is no longer indexed by Google.

domenkozar: Reasoning of split-up is that google with rank one page SEO farms quite low.

Additionally, URLs have to be search engine friendly so that random first-time Nix contributor blog is not ranked higher than the official documentation.

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,302 @@
...
+[asciidoc]: https://asciidoc.org/
+[asciidoctor]: https://asciidoctor.org/
+
+## reStructuredText
+
+Pros:
+
+- reStructuredText (RST) has wider adoption than Asciidoc.
+- Most Python projects use it. The Python ecosystem is deep and wide.
+- Good and mature toolchains like Sphynx exist to process RST files.
```
FRidh, line 239:
```suggestion
- Good and mature toolchains like Sphinx exist to process RST files.
```

rfcs/0000-commonmark-docs.md
mboes: > Only looking at the rust book, it's not just CommonMark as it uses extensions for includes:

That's true. They use a preprocessor on top of Markdown. So do the projects I mention who use Hugo or Jekyll as their static site generator. These toolchains add special macros for includes. The Gatsby and React projects don't use that, and I believe we can get by just fine without a preprocessor given the current content.

Macros for textual includes could conceivably become useful down the road, but it will have little impact on users, who can still just write the CommonMark they know.

mboes: > TOC is another example of such feature.

TOC's are a must in my book. I think we're in alignment there. Whether they need to be programmable from within the document markup is a different matter. The ReactJS documentation (and our demos) have TOC's auto-generated from the document structure. They are prominently displayed in the sidebar. Gatsby's documentation also has that. Although this is one other place where they use an MDX component: they go the extra mile and generate section specific "site maps" at the bottom of a few of their documentation pages.

FRidh: I think CommonMark allows us to get most of the documentation writing done without many issues. In case one does touch upon its limits, things get harder, and one has to fallback to another language like e.g. HTML.

In my opinion this RFC should define what we fall back to. E.g., fallback to HTML, or, fallback to say reStructuredText? See e.g. how one can create a table with Sphinx + CommonMark:

```
    ```eval_rst
    +------------+------------+-----------+ 
    | Header 1   | Header 2   | Header 3  | 
    +============+============+===========+ 
    | body row 1 | column 2   | column 3  | 
    +------------+------------+-----------+ 
    | body row 2 | Cells may span columns.| 
    +------------+------------+-----------+ 
    ```
````
Clearly this now becomes toolchain-dependent. In this case falling back to reStructuredText instead of HTML still allows the creation of pdf's.

Comment (mboes) [reply rfcs/0000-commonmark-docs.md:77]: > As soon as you start using HTML, you lose portability of any kind (which may not be a requirement, but then we need to pick the tooling and commit to it).

That's true, but in practice:

1. HTML is seldom needed except for a few key things (tables and callouts),
2. different parts of the documentation are written for different media types.

The man pages are written for `man`. Tables in man pages are not a must have (and even if they were they are still supported by a number of man generation tools even if they were, like go-md2man, Pandoc and possibly Sphinx). Some users may be interested in PDF output rather than HTML output, but that's a matter of having decent CSS for the `print` media type. Likewise, I don't think EPUB support should be the deciding factor either, and we likely don't want to support that officially. But for those users who really want it Pandoc understands CommonMark (including HTML tables) as an input very well.

File: rfcs/0000-commonmark-docs.md (hunk 2 above)
Comment (edolstra) [reply rfcs/0000-commonmark-docs.md:113]: Fun fact: the manuals used to be split into multiple pages, and everybody complained about it and demanded a single-page manual.

File: rfcs/0000-commonmark-docs.md (hunk 4 above)
Comment (edolstra) [reply rfcs/0000-commonmark-docs.md:198]: We don't need PDF or epub. Manpages are apparently supported by Sphinx.

File: rfcs/0000-commonmark-docs.md (hunk 3 above)
Comment (mboes) [reply rfcs/0000-commonmark-docs.md:77]: > Clearly this now becomes toolchain-dependent. In this case falling back to reStructuredText instead of HTML still allows the creation of pdf's.

@fridh what you're saying is that there exist corner cases, but those corner cases don't mean we're SOL - just that what happens then is not specified by any spec. I agree. AFAICS there are two or more tool choices available for all such corner cases (in particular tables) and all media types (in particular man pages). So long as this holds true, we're in a good place: it means we don't need to commit as part of this process to a specific implementation - just agree on the requirements for any such implementation, safe in the knowledge that satisfying the requirements won't be onerous.

There was enough bikeshedding in #64 that I'd rather make progress by agreeing on the format and the requirements for any toolchain, so that implementers don't have to spend heaps of time working towards a dead end because the community can't agree on the format (or indeed the requirements). (cc @domenkozar)

File: rfcs/0000-commonmark-docs.md (hunk 2 above)
Comment (mboes) [reply rfcs/0000-commonmark-docs.md:113]: @domenkozar very convincing reasoning re page split-up, thanks.

Comment (asymmetric) [reply rfcs/0000-commonmark-docs.md:113]: Just to understand: if I google "nixpkgs manual", the first results is nixos.org/nixpkgs/manual.

In which instances are the manuals not shown?

Comment (abathur) [reply rfcs/0000-commonmark-docs.md:113]: The monolith manual is useful at times and a daunting impediment at others. I think it's still desirable to publish both versions.

Comment (Ma27) [reply rfcs/0000-commonmark-docs.md:113]: > In which instances are the manuals not shown?

I assume if you search for content that exists in the manual, the results won't be found by google (or rated very low).

Comment (Ma27) [reply rfcs/0000-commonmark-docs.md:113]: > The “good quality search engine” part becomes less of a problem when you put everything on one page

I disagree. When I was new to Nix(OS) and I didn't know the terminology too well, I often had trouble finding things in the manual because I used too general terms (e.g. "build derivation") which resulted in a lot of unrelated results.

Also, on a search engine you can just combine terms that are relevant for the search, when performing a content-search in the manual using a browser you can't search e.g. for something like "nix rust derivation".

Don't get me wrong though, I really love the offline manual (including the PDF variants). Ideally, I'd love to see a split-up documentation on nixos.org and keep a single-page manual available offline (e.g. via `nixos-help` as it is currently the case).

File: rfcs/0000-commonmark-docs.md (hunk 4 above)
Comment (domenkozar) [reply rfcs/0000-commonmark-docs.md:198]: So the tooling matters here again, otherwise we will have to ditch `services.nixosManual.showManual` option.

Comment (edolstra) [reply rfcs/0000-commonmark-docs.md:198]: Why? That option only requires an HTML rendering of the manual.

Comment (domenkozar) [reply rfcs/0000-commonmark-docs.md:198]: Ah right, I forgot that uses HTML instead of the manual page.

**Comment by abathur** [comment]
I worry a little (both in #64 and here) that the natural order to make these decisions directs a lot of attention away from something that seems like one of the most-useful razors: 

Across the ecosystem, what high-ROI tasks/work would people like to automate that will require access to some information in/from/about the docs? Will any given decision enable or impede doing so?

File: rfcs/0000-commonmark-docs.md [hunk 6]
```diff
@@ -0,0 +1,302 @@
...
+[gatsby-docs]: https://www.gatsbyjs.org/docs/
+[gitstar-rankings]: https://gitstar-ranking.com/repositories
+[kubernetes-docs]: https://kubernetes.io/docs/home/
+[react-docs]: https://reactjs.org/docs/getting-started.html
+[rust-docs]: https://doc.rust-lang.org/book/
+[tensorflow-docs]: https://github.com/tensorflow/docs
+[vscode-docs]: https://code.visualstudio.com/docs
+[mdx]: https://github.com/mdx-js/mdx
+[remark]: https://remark.js.org/
+
```
davidak, line 94:
```suggestion
converters with plugin support like [Remark][remark], [static site
generators][staticgen] by the dozen, [MDX][mdx] to extend CommonMark with
arbitrary React components, etc).

[gatsby-docs]: https://www.gatsbyjs.org/docs/
[gitstar-rankings]: https://gitstar-ranking.com/repositories
[kubernetes-docs]: https://kubernetes.io/docs/home/
[react-docs]: https://reactjs.org/docs/getting-started.html
[rust-docs]: https://doc.rust-lang.org/book/
[tensorflow-docs]: https://github.com/tensorflow/docs
[vscode-docs]: https://code.visualstudio.com/docs
[mdx]: https://github.com/mdx-js/mdx
[remark]: https://remark.js.org/
[staticgen]: https://www.staticgen.com/

```

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,302 @@
...
+the limits of the man page format. When the time comes to convert the
+man pages as well, we can turn here again to prior art. The Kurnetes
+project uses [md2man][md2man] to generate man pages from CommonMark.
+This is a small Go command with a 4.1MB closure size (including 2MB
+for `tzdata`).
+
+[gatsby]: https://gatsbyjs.org
+[sphinx]: https://www.sphinx-doc.org
+[pandoc]: https://pandoc.org/
+
```
davidak, line 146:
```suggestion
[md2man]: https://github.com/cpuguy83/go-md2man

```
@mboes i guess you mean this

mboes: @davidak thanks!

rfcs/0000-commonmark-docs.md
asymmetric: > > In which instances are the manuals not shown?
> 
> I assume if you search for content that exists in the manual, the results won't be found by google (or rated very low).

I googled "declarative package management" and the nixpkgs manual is the 3rd result.

@grahamc @domenkozar is this a case of the ranking being too low? It seems like the page is definitely indexed though.

mweinelt: One problem with the one-page manual is that Google won't do proper deeplinking. Search for `makeWrapper` for example and it will send you to https://nixos.org/nixpkgs/manual/, no anchor applied.

samueldr: The manual pages are also becoming more and more difficult to render on lower powered devices. It seems that throwing text on the screen using modern browsers is a hard task.

This, thus, reduces the accessibility of the manuals from users that don't have access to top tier machines like I figure a good chunk of Nixpkgs developers have.

While this is becoming less and less true in the developed world, bad internet that slows to a crawl is still a reality that exists. Splitting in distinct pages will also help accessibility in these instances.

samueldr:
Is the *form* of the output in scope for this RFC? I am talking about the *design*, looks-wise, and also some of the *function*, which could be a magnet for bikeshedding.

If it isn't, it should be clearly stating it to prevent bikeshedding. I also propose that it states *how* the actual final form of the docs will be selected. (Another RFC? Collaboration between website team and docs team? Other proposal?)

If it is part of the discussion for this RFC, I think it will needlessly bog it down.

Now, the form of the output *only lightly* depends on the selected toolchain, in my experience. This RFC could (should?) still decide on a specific toolchain without ruling on the final looks and behaviour of the output.

I see that in *Future work* there is

> * Customize the CSS and layout of the demo to be more in line with Nix branding.

So I believe I am understanding right that the output *looks and behaviour* is not in scope of this RFC.

Infinisil:
In #64 we deemed some things as essential for a documentation format, one point of which was the ability to create references and being able to link to them from anywhere. This is very useful for option definitions being referenced in the manual, or by other options. Also very useful for link anchors.

As far as I can see, the CommonMark standard does not have any such thing. There's [link references](https://spec.commonmark.org/0.29/#link-reference-definitions), but those don't work outside of the current document. So, if this RFC were accepted, we would not be able to ever have inter-document references. Is that correct?

If the answer is "Yes, but we can use this extension in $toolchain which can do this", then I reply with the fact that this then isn't CommonMark anymore, and we're then tied to $toolchain. So unless we can be sure that CommonMark has absolutely everything we will ever need, this RFC should instead be about choosing the tooling that can provide everything we need.

You also mention that

> The motivation to switch to CommonMark is that among these projects, since roughly 10 times more projects use Markdown (or some variation thereof) than all other documentation formats combined, it can be considered the default.

How many of these projects are actually using spec-compliant CommonMark? Since you mentioned Markdown in that sentence, this let's me believe that most projects aren't using CommonMark, but rather CommonMark/Markdown + tooling-specific extensions, suggesting that pure CommonMark wasn't good enough for all of these projects.

So I really think we should include tooling in this decision, with all the extensions it provides that we might want to rely on.

michaelpj:
I have to agree with @Infinisil that the lack of proper inter-document cross-references is a deal-breaker.

I have also never worked on a project that used pure Markdown, there were *always* extensions because Markdown is just too minimal for any serious documentation project.

mboes:
@samueldr 
> Is the form of the output in scope for this RFC? I am talking about the design, looks-wise, and also some of the function, which could be a magnet for bikeshedding.

Design: no. Function: yes. Quoting from the RFC:

> The goal of this RFC is to change the *form* of the current documentation for Nix, Nixpkgs and NixOS to look similar to any one of the above 5 projects (see requirements in the next section).

In the "Detailed design" section, we list a number of requirements, which all have to do with what you call *function* above (but not design). While I really really want to avoid discussing implementation to avoid further bikeshedding, we should make sure that the choice of format doesn't paint us into a corner where important functionality would be impossible or hard to achieve. I listed a few requirements, but I may have overlooked some other important ones.

I'm adamant that implementation should *not* be discussed as part of this RFC, for the following reasons:

* There are a lot of factors to consider when choosing an implementation, including operational concerns and ease of maintenance. Choosing the implementation is an ill-formed question and bakes in endless potential for bikeshedding.
* RFC's are a heavyweight process involving the entire community and require dozens of hours of work from the authors and the participants. Because of this, they have a strong normative power. We shouldn't have to go through the process all over again if the need arises to make big changes to the implementation (say as part of a revamp of the website).
* Agreeing on requirements is one thing. But why should we involve the entire community on implementation choices that are mainly going to affect the livelihood of the authors (who will do the brunt of the initial work) and the maintainers (who will be saddled with the result for years to come)? I think we should defer implementation to a regular issue + PR process involving only authors and maintainers. This is both more efficient and more fair.
* The choice of toolchain to generate the website does not affect users, or even documentation contributors. Beyond the choice of format, I submit that nothing here constitutes "a substantial change to the Nix ecosystem", as defined in the [README for the RFC process](https://github.com/NixOS/rfcs/blob/master/README.md). No changes to Nix itself. No restructuring of Nixpkgs. No new syntax in Nix-the-language.

> I also propose that it states how the actual final form of the docs will be selected. (Another RFC? Collaboration between website team and docs team? Other proposal?)

Fair point. This is currently left implicit in the RFC as: whatever is the usual process for getting a change landed on the website (create an issue, open a PR, get it reviewed, merge). A collaboration between website and docs team (with external contributions) sounds like the right approach to me.

FRidh:
Agreeing with the above I propose we extend the RFC stating we use Sphinx as tool, using the recommonmark plugin as CommonMark backend.
- recommonmark has an ``````eval_rst ...``` for falling back to rst, allowing you to create tables and all other "more specific" things. Won't be needed for 95% of the docs but still adds significant values when it is needed;
- recommonmark allows you to create relative links;
- the community has many Python developers that have experience with Sphinx as well;
- its plugin system and support for domains will be very valuable for us, e.g. so we can implement a Nix domain for describing functions.

(The gatsby mockup looks way nicer though, but I don't see any reason we could not achieve the same with Sphinx)

mboes:
@michaelpj @Infinisil

Folks keep on saying that there is no such thing as plain CommonMark, but fall short of stating what additional markup they believe is absolutely essential.

Furthermore, it would be good to back up any such claim with evidence. The RFC does include evidence to the contrary. As highlighted again in a [previous comment](https://github.com/NixOS/rfcs/pull/72#discussion_r453612518), all 5 large projects cited in the RFC use plain CommonMark for markup except Kubernetes that [uses](https://github.com/kubernetes/website/blob/master/config.toml#L38-L41) one extension for tables and one for definition lists. Gatsby uses MDX to include newsletter signup forms, embed videos from third-party training websites, and build partial documentation sitemaps.

Lastly, while this RFC does not propose to do so, what exactly would be so bad if we end up using one or two extensions here and there? Some extensions, like tables and definition lists, are already well supported by many documentation processing tools, as well as translation tools like Pandoc, and even if that weren't the case, translating the odd table or definition list by hand should we move to NewMarkupLangPro in the future takes far less time than this RFC discussion will likely take.

We could amend this RFC to require non-HTML tables. That would still not force a single implementation. Tables are supported by all of Gatsby, Hugo, Jekyll and mdBook, the toolchains used by the 5 projects cited in the RFC. Same goes for footnotes.

mboes:
@FRidh thanks for pointing out that Sphinx supports these too, via `eval_rst`. So I guess that expands the set of possible implementations still further, even if the requirements were to be tightened to include a small number of CommonMark extensions. As I said above, I'm wary of discussing implementation options. Sphinx on the face of it looks like a perfectly fine contender, and indeed it is the basis of one of the demos cited in the RFC. I just had a comb through the GitHub Top 100 again. I may well have missed others, but I could find only one project that uses the Sphinx + recommonmark combo: Apple Swift (see [here](https://github.com/apple/swift/tree/master/docs)), who uses a mix of RST and Markdown files (newer files appear to be in Markdown). This is not a strong point either way - it just suggests that the pros and cons would have to be investigated further.

mboes:
@Infinisil 

> > The motivation to switch to CommonMark is that among these projects, since roughly 10 times more projects use Markdown (or some variation thereof) than all other documentation formats combined, it can be considered the default.
>
> How many of these projects are actually using spec-compliant CommonMark? Since you mentioned Markdown in that sentence, this let's me believe that most projects aren't using CommonMark, but rather CommonMark/Markdown + tooling-specific extensions, suggesting that pure CommonMark wasn't good enough for all of these projects.

I asked myself the same question and would love to know a good answer, but such an investigation ended up taking far too long. Focusing on just the 5 projects I cite in the RFC that are both in the Top 100 and also large and diverse projects (quite apart from their popularity), it appears that some (like React) use no extensions and no textual inclusion macros, Kubernetes uses 2 extensions (tables, definition lists) and macros, VSCode uses the table syntax from GFM but is otherwise plain CommonMark, and Gatsby uses a handful of custom MDX components, mostly to embed content.

mboes:
@Infinisil 

> As far as I can see, the CommonMark standard does not have any such thing. There's [link references](https://spec.commonmark.org/0.29/#link-reference-definitions), but those don't work outside of the current document. So, if this RFC were accepted, we would not be able to ever have inter-document references. Is that correct?

I'm not sure what requirements you have in mind (if you write them out I can include them in the RFC), but what happens in practice is that cross-references are done with (root-)relative links and anchors. An example is the Gatsby project, which has extensive cross-referencing. Look at e.g. this page: https://www.gatsbyjs.org/docs/write-pages/. The sources (accessible using a button at the bottom) use a lot of links with anchors to entities defined elsewhere. It's up to the toolchain to check for dead links.

zimbatm:
Since we did a few rounds of this documentation discussion already, we all know that the set that includes everybody's preferences is empty. At the same time, I think that we can all agree that any of the proposed formats are better than DocBook. In order to move forward, it will require some of you to let go of some of the technical details that are dear to you.

I support and would like to nominate myself as a shepherd if possible.

7c6f434c:
>I think that we can all agree that any of the proposed formats are better than DocBook. 

Somehow I do not get the same impression from the discussion. It looks more like some of the people who wrote the most text for the manual have doubts whether any of the other formats are good enough, but consider Docbook acceptable. They might be willing to see whether the marketing team actually can organise a significant improvement of documentation if that specific thing changes, but this is not the same as agreeing that some format is definitely better.

(As for me, last time I tried to write a bit of documentation, I failed to understand where in the current documentation it should even fit and gave up; I do not have any real opinion on these formats, although of course I like the clear escaping rules brought with XML, and dislike some of the error messages about wrong element nesting)

FRidh:
> Somehow I do not get the same impression from the discussion. It looks more like some of the people who wrote the most text for the manual have doubts whether any of the other formats are good enough, but consider Docbook acceptable. 

Interesting, it is my impression that we generally agreed that, while technically all of the proposed formats are a step backwards, it is worth it for the overall project. And that at that point it also doesn't matter much which language/tool we pick considering the status quo.

It is therefore also my opinion that we should not investigate much further and just go ahead with any language/toolchain combo. Indeed CommonMark is a good fit, and as pointed out by @mboes there are different tools/extensions available for the other IMO important features such as tables and references. Then let's pick one and move on.

> (As for me, last time I tried to write a bit of documentation, I failed to understand where in the current documentation it should even fit and gave up;

This is a *far* more interesting discussion that we ought to spend our time on, contents!

Let's be pragmatic here. Thus, to continue my previous post, would anyone be opposed to the Sphinx + CommonMark combo? If so, why?

jtojnar:
> I'm not sure what requirements you have in mind (if you write them out I can include them in the RFC), but what happens in practice is that cross-references are done with (root-)relative links and anchors.

Well, the problem is CommonMark spec does not have the notion of anchors. Some implementations add anchors to headings automatically, but implicit heading anchors
* are limited to headings
  * often, it makes sense to link to other elements, like a bullet point (example: https://nixos.org/nixpkgs/manual/#ssec-gnome-common-issues-unwrappable-package-gsettings-vala)
* typically make the section address depend on the heading content
  * so even minor changes break the anchor
    * and while you can check the references in-document, you will not be aware you are breaking external references
      * explicit anchors make it clearer there is a kind of a contract
      * maybe even better solution would be some kind DOI-like permalink database but until we have that, explicit anchors are reasonable compromise

So the requirement would be to be able to specify explicit anchors, both to sections (through headings) and to other types of content ([paragraphs](https://nixos.org/nixos/manual/#para-module-syn), [list items](https://nixos.org/nixpkgs/manual/#ssec-gnome-common-issues-unwrappable-package-gsettings), [tables](https://nixos.org/nix/manual/#table-operators), [code listings](https://nixos.org/nixpkgs/manual/#ex-bowerJson)…).

There appears to be a [CommonMark extension](https://github.com/commonmark/commonmark-spec/wiki/Deployed-Extensions#attributes) for that but, again, compostability is an issue. MDX might fare better.

davidak:
>Let's be pragmatic here. Thus, to continue my previous post, would anyone be opposed to the Sphinx + CommonMark combo? If so, why?

I think the syntax of Sphinx is not very user friendly. (but haven't used in a long time) So combining it with CommonMark seems not the best fit. That it is not popular in the top100 may be due to that.

So i would rather use one of the never, popular tools. Hugo is extremely popular in the space of static site generators and is actively developed. We can also use it for the website. We would be very flexible with it's huge ecosystem. (It supports interlinks https://gohugo.io/content-management/cross-references/ but has no native PDF export) (haven't used Gatsby, Jekyll or mdBook)

Mixing RST with Markdown might be confusing since they are very similar.

But i guess Sphinx is better suited for documentation and "Nix domain for describing functions" sounds great! So i would not opposed.

Infinisil:
To throw another possibility into the mix I'm very intrigued by: [MyST](https://myst-parser.readthedocs.io/en/latest/)

> **A fully-functional markdown flavor and parser for Sphinx.**
> MyST allows you to write Sphinx documentation entirely in markdown. MyST markdown provides a markdown equivalent of the reStructuredText syntax, meaning that you can do anything in MyST that you can do with reStructuredText. It is an attempt to have the best of both worlds: the flexibility and extensibility of Sphinx with the simplicity and readability of Markdown.

davidak:
>So the requirement would be to be able to specify explicit anchors, both to sections (through headings) and to other types of content (paragraphs, list items, tables, code listings…).

Looks like a small heading would also work in that cases.

jtojnar:
> > So the requirement would be to be able to specify explicit anchors, both to sections (through headings) and to other types of content (paragraphs, list items, tables, code listings…).
> 
> Looks like a small heading would also work in that cases.

Sure, but then we are trading for readability.

Some issues appear repeatedly in the issue tracker due to incompatibility of Nix model with some common global state nexi. When that happens, I like to link to specific passage in the manual explaining how to deal with the situation. For that reason, I liberally sprinkle the documentation I write with anchors.

If we added heading to every place where there is an anchor, it would be even less readable.

<details>
<summary>Example of such use of anchors in the manual</summary>

Try importing the following CSS into your browser and visit https://nixos.org/nixpkgs/manual/#ssec-gnome-common-issues-unwrappable-package-gsettings-vala.


```css
[id]::before {
    content: attr(id);
    color: red;
    border: 1px solid;
    padding: 0.1em 0.2em;
    font-weight: bold;
}
```

many anchors in the GNOME docs

And now imagine these are headings.

</details>

samueldr:
@7c6f434c wrote:

> [...] They might be willing to see *whether the marketing team actually can organise a significant improvement of documentation if that specific thing changes*, but this is not the same as agreeing that some format is definitely better. 

Speaking personally, the *authoring format* (e.g. docbook, markdown) is not important here. I had worked on an experimental one-page-per-chapter revision with search and other things with graham, basically what this RFC calls "Modern looking documentation", entirely based on the Docbook input.

The effort was hampered by the **tooling choice**, not by the input format. The first (and only) iteration was probably accidentally quadratic in some ways and caused docs build to be absurdly long (minutes!). The current tooling maybe can be used to achieve what we desired, the effort was stopped once it was obvious that the documentation format is going to be affected by an RFC.

What I want to say with this is that:

 * The tooling to produce the format will matter if we want some amount of control
 * The input format (e.g. docbook, markdown) is an implementation detail for "Modern looking documentation".

Since the final tooling choice is going to be selected *outside of this RFC*, then I don't see much I can say as part of the Marketing team / Website team.

samueldr:
@mboes wrote: 

> The choice of toolchain to generate the website does not affect users, or even documentation contributors. [...]

The tooling choice affects documentation contributors greatly. They need to build the documentation. If the documentation takes minutes to generate, this is going to be harmful to their experience.

But you're right that it does not affect *authors* during *authoring* if we assume markdown is markdown and the author does not need to build until they are done authoring.

The tooling choice can affect users if the generated documentation is opinionated and makes bad decisions that are hard to work around.

Though, as this RFC does not select the tooling to be used, this comment is mostly off-topic.

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,304 @@
...
+[unresolved]: #unresolved-questions
+
+We propose to take CommonMark as a baseline. This is sufficient to
+support all the markup we need. It's also a stepping stone towards
+[MDX][mdx], which enables embedding custom widgets if so desired. For
+example, Gatsby uses MDX to include newsletter signup forms, embed
+videos from third-party training websites, or to generate
+documentation sitemaps within a section (e.g. the bottom of [this
+page][gatsby-graphql]). It's unclear that custom widgets are worth the
+trouble at this stage.
```
jtojnar, line 287: It might be nicer to use MDX from get-go for admonitions rather than using [`<div>` HTML element](https://github.com/mboes/nixos-docs-mockup/blob/34355ccd73262f273c6b3412c878aa6008614dda/content/advanced-topics.md#L33-L41).

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,304 @@
...
+* the Nix expressions chapter
+* advanced topics
+
+The content is the unedited output of running the Pandoc conversion,
+which still has bugs (in particular the handling of cross-references).
+
+This other demo shows how to satisfy many of the same requirements
+using Sphinx, again with CommonMark as the input:
+
+https://nixpkgs-manual-sphinx-markedown-example.netlify.app
```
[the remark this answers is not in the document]
mboes: Hm. I should include a lot more detail in the RFC. The source for that site is here: https://github.com/garbas/nixpkgs/blob/manual-markdown-sphinx/nixos/doc/manual-rst/. The README explains the history of that branch. It's based on @nlewo's earlier RST prototype prepared as part of #64, but with one file converted to Markdown. See https://github.com/garbas/nixpkgs/blob/manual-markdown-sphinx/nixos/doc/manual-rst/source/installation/installing.xml.md.

rfcs/0000-commonmark-docs.md
mboes: I'm all for using MDX, but that may be a more controversial choice than plain CommonMark, or even well supported slight extensions of it, like GFM. I agree that MDX would be nicer for admonitions, because inside a `<div>` CommonMark markup is not interpreted. React (ab)uses blockquotes for admonitions (see e.g. [here](https://raw.githubusercontent.com/reactjs/reactjs.org/master/content/docs/strict-mode.md)). It's a good enough solution, but yes, MDX would be even nicer if the community can stomach it.

mboes:
@jtojnar 
> Well, the problem is CommonMark spec does not have the notion of anchors. Some implementations add anchors to headings automatically

That's right. Heading anchors are automatic. Additional anchors can be added using `<span id="..."/>`.

Automatic anchor names are not ideal, for the reasons you mention. But an outdated anchor name in a URL is less bad than an outdated path (the latter will give you a 404), especially in setup where each chapter is its own page. Gatsby and React are examples of projects that accept this.

> MDX might fare better.

Interesting. How so? You mean that one could create a special `<LinkTarget>` component or somesuch?

mboes:
@samueldr 
> The tooling choice affects documentation contributors greatly. They need to build the documentation. If the documentation takes minutes to generate, this is going to be harmful to their experience.

Agreed, but how is this different from any other implementation concern that impacts performance? e.g. poor implementation choices in `nix` subcommands could cause a similarly bad user experience. Would you be happy addressing performance problems if any during PR review submitted to the nixos website? At least by then, we'll hopefully have gotten the format discussion out of the way. If you have specific performance targets in mind (I know the existing Docbook toolchain isn't fast), then we could add those to the list of requirements in this RFC.

samueldr:
@mboes 

I don't have specific targets, but this is a requirement *when the tooling is going to be selected*. The tooling should process the documentation, including the generated sections, in a satisfying length of time.

In my opinion it's more of a comparative weight in the available tooling than an actual metric. A trade-off to help make the choice in due time.

jtojnar:
> > MDX might fare better.
> 
> Interesting. How so? You mean that one could create a special `<LinkTarget>` component or somesuch?

Exactly.

andir:
> @jtojnar
> 
> > Well, the problem is CommonMark spec does not have the notion of anchors. Some implementations add anchors to headings automatically
> 
> That's right. Heading anchors are automatic. Additional anchors can be added using `<span id="..."/>`.

That only works if we restrict ourselves to HTML. I for one regularly use the man pages as they are faster to navigate.

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,304 @@
...
+CommonMark has far less expressive markup. It is done using
+[Pandoc][pandoc], which has a reader available for Docbook and
+a high-quality writer available for CommonMark. The Pandoc Docbook
+reader isn't perfect, but in the process of putting together the quick
+demo below, it was easy to fix 5 bugs already.
+
+We propose to keep the man pages as Docbook for now. They are
+self-contained documents, whose form is constrained by convention and
+the limits of the man page format. When the time comes to convert the
+man pages as well, we can turn here again to prior art. The Kurnetes
```
ryantm, line 139:
```suggestion
man pages as well, we can turn here again to prior art. The Kubernetes
```

worldofpeace:
This RFC is now open for nominations for the shepherd team.

Profpatsch:
I nominate

- @garbas because he did a phenomenal job with the new website
- @zimbatm (who already nominated himself earlier)
- @FRidh since it looks to me like they want to bring the doc horror story to a conclusion :)

davidak:
FYI: this theme looks very modern and clean https://squidfunk.github.io/mkdocs-material/getting-started/

Infinisil:
As the author of the previous RFC (#64) I nominate myself as a shepherd for this one

garbas:
@Profpatsch thank you for the nomination. I gladly accept it.

edolstra:
I've done a conversion of the Nix manual to Markdown. It uses `mdbook` to render the HTML manual and `lowdown` for the manpages. Both have a small closure size (41 MiB for `mdbook` and 31 MiB for `lowdown` - most of that is Glibc).

You can see the HTML result here: https://hydra.nixos.org/job/nix/markdown/manual/latest/download-by-type/doc/manual. Note that there are still a bunch of things to be fixed (e.g. internal references, footnotes) and in particular the manpages are not converted properly. A manually converted manpage is here: https://hydra.nixos.org/build/124287980/download/1/manual/command-ref/nix-copy-closure.html

zimbatm:
Let's meet-up next week to talk about the RFC: https://framadate.org/6fWhtIpILlKrQgFf

garbas:
@zimbatm Would it be possible to have a meeting at 21:00CEST (or later)?

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,304 @@
...
+a transition to full Markdown (grep for `[Gardening] De-RST` in [the
+history][swift-docs-history]).
+
+[swift-docs-history]: https://github.com/apple/swift/tree/master/docs
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+We propose to take CommonMark as a baseline. This is sufficient to
+support all the markup we need. It's also a stepping stone towards
```
8573, line 281: I think other reviewer comments rebut the claim that "CommonMark [...] is sufficient to support all the markup we need".

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,304 @@
...
+1. good quality documentation search engine,
+1. syntax highlighting of all code,
+1. separate page per chapter, instead of the current monolithic page
+   for each manual,
+1. table of content for each chapter available in a side bar to easily
+   jump through long content,
+1. make warning and info boxes stand out from the rest of the content
+   with colour,
+1. easy to customize HTML and CSS, since tired old templates we've all
+   seen too many times are best avoided. The objective is to make
```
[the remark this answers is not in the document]
samueldr: Here's my opinion about that:

First, it's a branding thing, *mdbook* AFAIK was made for the Rust documentation, so it's already their branding for their docs. 

Then, this simple objective is a way to avoid *locking ourselves* into what the tool provides. It may be that we need such capabilities not to brand the docs, but to implement missing features according to our needs. If extensibility is hard, or impossible, it's like painting us in a corner.

Finally, a throwback to the first point, is that we don't want to *lose* features. The current documentation toolchain allows a tight integration of the docs into the website. If the chosen solution doesn't allow us to integrate the docs in the website "as native as it can be", it's a loss of functionality.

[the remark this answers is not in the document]
edolstra: `mdbook` is customizable, see https://rust-lang.github.io/mdBook/format/theme/index.html. However, I don't think the current tight integration (e.g. that the docs have the same navbar as the rest of the site) is essential. I think it would be enough to have a visual style that looks similar to the rest of the site.

Or we could just upload the docs to readthedocs.io...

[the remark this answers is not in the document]
8573: > _mdbook_ AFAIK was made for the Rust documentation, so it's already their branding for their docs.

Ah, I didn't know that.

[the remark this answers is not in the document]
Mic92: We should have a `highlight.js` that can provide syntax highlighting for nix and bash.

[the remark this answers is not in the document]
evanjs: Started a mdBook docset for a different project lately, and found it's extremely easy (like most things with mdBook so far) to add additional languages for highlight.js.

I downloaded a script with the languages I needed from https://highlightjs.org/download/ and dropped it in `src/highlight.js`.
The available syntax highlighting for Nix also works great, of course.

One thing that irritates me is the default code-block width, but we could style this as needed.

Without said custom styles, I had to add a few line breaks to large expressions, such as <a href="https://github.com/Infinisil/system/blob/eb1046aa55802989ac8baffc8f476a5e3a682451/config/new-modules/default.nix">this one</a>.
Otherwise, it only displayed ~50% of the code block.

<details>
<summary>Small code block width</summary>

image
</details>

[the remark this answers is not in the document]
samueldr: (In case it wasn't obvious, I wasn't talking badly about mdbook, I haven't looked into any of the tooling options since anyway this RFC is not about tooling.)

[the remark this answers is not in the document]
edolstra: @evanjs Thanks, I've added a `highlight.js` with Nix support to the manual. Example: https://hydra.nixos.org/build/124655688/download/1/manual/expressions/language-constructs.html

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,304 @@
...
+[future]: #future-work
+
+- Build out the documentation demo above into a full replacement for
+  each of the Nix, Nixpks and NixOS user manuals online.
+- Fix all Pandoc bugs encountered along the way, in particular to get
+  working cross-references.
+- Customize the CSS and layout of the demo to be more in line with Nix
+  branding.
+- Integrate the result into the Nix and Nixpkgs repositories and hook
+  into their respective CI/CD pipelines.
```
blaggacao, line 304: I like the way https://cert-manager.io/ does give you a binary choice on their landing page: Docs or Repo? The githup repos are integral part of the experience and most often your primary (google) search intent pivots around exactly this binary choice.

I feel this is out of scope for this RFC, but might result as a collateral of the concrete implementation.

blaggacao:
When the day of choosing the tooling arrives, the organ making the call shall consider the upsides of using a tool that can query and render arbitrary auxiliary data sources. Since the nix project is so vibrant and interactions usually go by discourse and GitHub, it makes sense to preserve the option to consume those APIs in order to enrich documentation with specific widgets. No examples provided, but one idea: render context dependent open PRs/Issues from specific Doc pages. Bad example. Just consider keeping this door open.

lheckemann:
Shepherd team is @Infinisil (leader), @garbas, @zimbatm

domenkozar:
That violates rfc process as Tweag has majority of the shepherd team.

Mic92:
@domenkozar So we could basically fix that by adding another shepherd right?

domenkozar:
That seems best :)

Mic92:
> That seems best :)

How about you?

Infinisil:
As far as I know, neither @zimbatm nor @garbas is from Tweag (and I'm not from Tweag either (yet)). But also, the rfcs readme only specifies
> at most half of the Shepherd Team can be part of the RFC Steering Committee.

So even if the majority would be from a single company, that wouldn't be disallowed (though not ideal)

Mic92:
> As far as I know, neither @zimbatm nor @garbas is from Tweag (and I'm not from Tweag either (yet)). But also, the rfcs readme only specifies
> 
> > at most half of the Shepherd Team can be part of the RFC Steering Committee.
> 
> So even if the majority would be from a single company, that wouldn't be disallowed (though not ideal)

@garbas works for Tweag according https://twitter.com/garbas and I know @zimbatm also has a work relationship with Tweag. In the past we always tried to avoid these constellations when setting up a shepherd team in the RFC shepherd committee.

domenkozar:
> > That seems best :)
> 
> How about you?

Sorry, I have to pass this one. Maybe someone else from the previous team?

Kloenk:
I probably would have time to be part of yet another Shepard team, but then it is enough ;-)

mboes:
For full disclosure, in case this matters, I'm the author of this RFC and I'm also at Tweag (as stated in my GitHub profile).

It's for the community to decide what norms we want to set on shepherd teams, and to decide whether they apply in the same way to all RFC's (as @samueldr points out above, the scope of this particular RFC is small). But as the founder of Tweag, let me just register that it would be Tweag's preference to *not* have a majority on the shepherds team.

This RFC has been stuck at the "putting together the shepherds team" stage for 1 month now. So it would be great if other folks could nominate themselves as well. :)

Mic92:
I talked with the rest of the RFC steering committee team, we will just add @Kloenk to the Shepherd team.

So it is @Infinisil (leader), @garbas, @zimbatm and @Kloenk 
@mboes could you add this to the metadata of this file.
@Infinisil could you arrange meeting dates via some doodle clone or use the existing one that zimbatm has made?

mboes:
> @mboes could you add this to the metadata of this file.

Sure thing.

Infinisil:
Alright, I created https://www.when2meet.com/?9536530-Vwk23 to find a time to meet for next week (when2meet is very nice for entering many entries, though not open source), please fill it out until the end of the weekend @mboes, @garbas, @zimbatm and @Kloenk.

Infinisil:
I guess the weekend isn't optimal for trying to find a time. Let's wait until Monday evening for @mboes and @garbas, otherwise we'll try to find a time in the following week instead.

Mic92:
> I guess the weekend isn't optimal for trying to find a time. Let's wait until Monday evening for @mboes and @garbas, otherwise we'll try to find a time in the following week instead.

You can also try to write them an email directly in case they do not check their github notifications frequently.

mboes:
Done. Best availabilities are for today and tomorrow.

zimbatm:
Let's meet at 17:00 UTC today. Rok is on holiday today and tomorrow so a full majority is not possible for this round anyways.

Infinisil:
### Meeting 1

Present were @mboes, @Kloenk, @zimbatm and me
Meeting notes: https://hackmd.io/IJYrTWhMRL6wJD1rjpdpdw
General discussion result: We need extensions! Looking good otherwise

The next meeting is scheduled for next Monday same time (2020-08-24 17:00 UTC or `date -d@1598295600`). Waiting for @garbas to confirm this time works for him when he returns from vacation, otherwise we'll find a new date.

Kloenk:
I'm sorry, something came up. Unsure if I can make it today at 17:00 UTC

Infinisil:
What about you @garbas?

zimbatm (approved):
Ok, this looks good enough to me for FCP.

garbas:
Hey I'm back from vacation. I generally can not make it between 13:00-19:00UTC (15:00-21:00CEST).

blaggacao:
I'd like to give another loose input as collateral / derivation of this RFC:

It would be just plain less fastidious if for example
- [This file](https://github.com/NixOS/nixpkgs/blob/master/doc/builders/images/dockertools.xml)
- would be a Markdown named `Readme.md` here: https://github.com/NixOS/nixpkgs/tree/master/pkgs/build-support/docker
- and rendered thereof in the docs

(similar for other documentation)

davidak:
@blaggacao yes, but restructuring the documentation is a different topic. let's decide on a format first :)

blaggacao:
Another such loose feedback: (since some have risen the issue of having the possibility of doing those references)

image

vs

https://skaffold.dev/docs/references/yaml/


The latter is better - you can do: https://skaffold.dev/docs/references/yaml/#build to point people arround

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,334 @@
...
+# Motivation
+[motivation]: #motivation
+
+Documentation ought to be easy to write, easy to maintain, easy to
+make pretty, and its format be so boring as to seldom be the object of
+future RFC's. But beyond that, the significance of the documentation
+format is in its very real impact on the perception of the project.
+Documentation is the storefront of any project online in addition to
+its front page. Modern looking documentation that can be edited at the
+click of a button sends a signal: that the Nix* projects are forward
```
blaggacao, line 25: Excellent proof why this is _urgently_ necessary: https://github.com/NixOS/nixpkgs/issues?q=is%3Aissue+is%3Aopen+%22Docs%3A+dockerTools%22

blaggacao: And ventilating a little further my **_frustration_** of not having been able to get a doc improvement merged myself within due effort: https://discourse.nixos.org/t/dockertools-imprecisions/8770/3

jtojnar:
> Another such loose (offtopic) feedback: (since some have risen the issue of having the possibility of doing those references)
> 
> image
> 
> vs
> 
> [skaffold.dev/docs/references/yaml](https://skaffold.dev/docs/references/yaml/)
> 
> The latter is not only visually more pleasent - you can do: [skaffold.dev/docs/references/yaml/#build](https://skaffold.dev/docs/references/yaml/#build) to point people around. Now that I verifyed this claim, I only figured out you can do - somewhat - the same thing with the current pointers. Still not as good as the skaffold example.

This is largely presentational difference. If your format supports some form of callouts, you can just render them on the side. The only reason we do not do  Or if the whole page is actually generated from structured data like your example seems to be:

https://github.com/GoogleContainerTools/skaffold/blob/9cd8bdd8e50434370b68fa4927efbb27d3646cb9/docs/content/en/docs/references/yaml/_index.html
https://github.com/GoogleContainerTools/skaffold/blob/f16bafaa55b8ad981911d72dbebb4aa04a11d0dc/docs/content/en/schemas/v2beta1.json

But I agree

jtojnar:
> Another such loose (offtopic) feedback: (since some have risen the issue of having the possibility of doing those references)
> 
> image
> 
> vs
> 
> [skaffold.dev/docs/references/yaml](https://skaffold.dev/docs/references/yaml/)
> 
> The latter is not only visually more pleasent - you can do: [skaffold.dev/docs/references/yaml/#build](https://skaffold.dev/docs/references/yaml/#build) to point people around. Now that I verifyed this claim, I only figured out you can do - somewhat - the same thing with the current pointers. Still not as good as the skaffold example.

This is largely presentational difference. If your format supports some form of callouts, you can just render them on the side. If your tooling is not PITA to modify like XSLT is. Or if the whole page is actually generated from structured data like your example seems to be:

https://github.com/GoogleContainerTools/skaffold/blob/9cd8bdd8e50434370b68fa4927efbb27d3646cb9/docs/content/en/docs/references/yaml/_index.html
https://github.com/GoogleContainerTools/skaffold/blob/f16bafaa55b8ad981911d72dbebb4aa04a11d0dc/docs/content/en/schemas/v2beta1.json

But I agree that the ability to be able to annotate code and link to its parts is important. Unfortunately, it is at odds with simplicity of “readable” markup languages. Annotating code listings using plain text comments inlined in the code itself (and possibly linking numbered lines) will have to suffice :crying_cat_face:

blaggacao:
> Annotating code listings using plain text comments inlined in the code itself (and possibly linking numbered lines) will have to suffice

Yeah! For the moment being. I'm still eyeing on MDX or similar pre/post processing stuff once the first move to md has been done.

samueldr:
Callouts do offer one advantage, which is important in a lazy language: you can order the callouts in a different order than the code reads.

blaggacao:
Yet another loose idea (might we record them in a brainstorming section in this RFC - maybe?):

- Problem: The docs are far from context free.
- Solution: To provide additional context to the reader https://rstudio.github.io/tufte/ has had good results with his handout style.

Such can be made more easy to implement using markdown.

davidak:
>might we record them in a brainstorming section in this RFC - maybe?

keep the RFC simple. that are implementation details

>https://rstudio.github.io/tufte/ has had good results with his handout style

isn't that also solved with admotions like the note in sphinx? https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html?highlight=warning#paragraph-level-markup

blaggacao:
> isn't that also solved with admotions like the note in sphinx?

Sort of - the added value of tufte is that those live on the right margin and do not disturb the main reading flow. It is a very strcutured way of layering the discussion. And maybe one would put stuff there that wouldn't fit well into in-text admonitions. The idea of a tufte handout is that it's roughly about the same amount of ink in both parts (a little less on the margin - but still rather dense in "further info").

edolstra:
Regarding callouts: FWIW, for the Nix manual, I replaced callouts by Unicode circled digits (①, ②, ...). See for example https://github.com/NixOS/nix/blob/c3efef9275ce6db3d65a2eb97aee507e79b626a6/doc/manual/src/expressions/language-constructs.md#assertions.

davidak:
@edolstra this breaks the code. it should at least in the rendered manual not be included when you select and copy it

samueldr:
Alternatively, making them comments like `# (1)` or `# ①` would make the code be correct again.

blaggacao:
To ease future refactoring of all existing callouts (see tufte handout) during the transformation:
```md
<callout>

1. First point ...
2. First point ...

</callout>
```
<callout>

1. First point ...
2. First point ...

</callout>

That should be enough to restore the association programmatically later on.

rfcs/0000-commonmark-docs.md
```diff
@@ -0,0 +1,334 @@
...
+  switch from one toolchain to another for generating HTML from the
+  CommonMark source, with minimal manual work. For example,
+  * enabling an extension for tables is acceptable because few tables
+  appear in the documentation and converting them by hand to a new
+  format is not labour intensive;
+  * YAML frontmatter for metadata is also acceptable, because nearly
+    all toolchains support this.
+* It is acceptable to introduce a new extension only if this extension
+  is supported by three or more popular toolchains. Toolchain-specific
+  extensions should not be used.
```
Infinisil, line 150: This should be loosened. I think it should be up to the implementers to decide whether an extension is standard enough to be used, instead of requiring a specific number of toolchains.

Infinisil (commented):
I haven't posted this yet, but the second meeting was held last Monday (August 24) with @mboes, @zimbatm and me. I admittedly did a pretty bad job at trying to take notes (I'll try to improve!), but I'm going to leave a comment about the part I think needed improving. Other than that, I think this RFC is ready for FCP!

rfcs/0000-commonmark-docs.md
samueldr: Following this to the letter, this would also forbid us from using a home-grown extension unless added to three or more popular toolchains.

Profpatsch:
> Another such loose (offtopic) feedback: (since some have risen the issue of having the possibility of doing those references)

Can we please keep this RFC on-topic? How callouts are linked does not have anything to do with the format at all.

Profpatsch:
Were there any new developments over the last two weeks? I don’t see any blockers for this to go ahead/get merged, apart from “we are gonna have to use some extensions, in a restrained manner and in good taste”.

zimbatm:
Did we already enter the FCP period yet? If yes then this is good to merge!

ryantm:
@Infinisil has not called for FCP, nor widely announced FCP.

7c6f434c:
>Were there any new developments over the last two weeks? I don’t see any blockers for this to go ahead/get merged, apart from “we are gonna have to use some extensions, in a restrained manner and in good taste”.

(I think it is worth saying that the point about extensions is well-described in the RFC, so it is _not_ a blocker in the sense «agreement is reached but needs writing down»)

Profpatsch:
The [comment by @Infinisil](https://github.com/NixOS/rfcs/pull/72#discussion_r481357053) has not yet been included it looks like.

ryantm:
I'm not sure what the proper protocol is here, and I don't mean any offense to the shepherd leader, but I request that @NixOS/rfc-steering-committee step in, because this seems to be blocked by shepherd leader inactivity.

Infinisil:
I think this RFC looks good, but the thing mentioned in https://github.com/NixOS/rfcs/pull/72#pullrequestreview-480022705 is still not addressed, and I don't think it makes sense to accept this RFC without at least changing that.

Since I can't change the RFC contents myself, because I don't have any commit permissions in this repo, I propose we either ask somebody with commit permissions to push such a change, or we close this RFC and create a new one with the required changes.

As a shepherd leader, it isn't really my task to adjust the RFC, but since it might be in limbo otherwise, I created a commit with all the small fixups I found necessary in https://github.com/Infinisil/rfcs/commit/27874c565bdeb26b8957af93cd97f9e1e81038d7, this includes:
- Nix now already uses markdown
- Fix grammar mistake
- Remove requirement of extensions having to be supported by at least 3 toolchains
- Add requirement for an extension for references
- Rename the file to its designated name

This isn't really in the RFC protocol, but if that's fine I'd like the @NixOS/rfc-steering-committee to push the above commit of mine to this PR, after which I can initiate FCP.

ryantm:
pinging @mboes for good measure

lheckemann:
I'm fine with doing that. @mic92, @edolstra, @spacekookie?

Mic92:
@Infinisil pushed!

zimbatm (approved):
can we enter the FCP now?

mboes:
Forgot about @Infinisil's comment above. Looks like that's been resolved without me, which I'm happy about. Apologies if I ended up adding any latency to the process. From my perspective as the original author I don't see any blockers for FCP.

Infinisil:
Alright, this looks good to me, I'm calling FCP. Unless any blockers are brought up, this should get accepted and merged in 10 days from now, on October 30!

Infinisil (approved):

Kloenk (approved):

ryantm (approved):

jonringer (approved):

bhipple (approved):

DamienCassou:
I would appreciate epub and info exports if possible.

DamienCassou (approved):
Great RFC, thank you!

hloeffler (approved):

dasJ (approved):

davidak (approved):

worldofpeace (approved):
Hell yes

ryantm:
The FCP is over. I'm not aware of any objections.

nh2:
> [rendered](https://github.com/mboes/rfcs/blob/markdown-docs/rfcs/0000-commonmark-docs.md)

@mboes The link in the issue description no longer works.

cole-h:
Fixed rendered link: https://github.com/mboes/rfcs/blob/markdown-docs/rfcs/0072-commonmark-docs.md

AndersonTorres:
My only concern here is:

Is there an Emacs mode for it? :)

FRidh:
There is no link to the source for the example https://nixpkgs-manual-sphinx-markedown-example.netlify.app. `View page source` shows the intermediate reStructuredText. For https://github.com/NixOS/nixpkgs/issues/102246#issuecomment-719968363 we would like to see the source.

zimbatm:
Merging. Thanks all for your participation!

Mic92:
Since Nix uses mdbook now for documentation generation is this the plan for the other manuals as well?

worldofpeace:
We made the decision during 20.09 management that the release process shouldn't be in the nixos manual. There should still be a section "NixOS Releases", but information that's useful to the user. Having the process there was really weird. It has since been split out to https://github.com/NixOS/release-wiki and I used mdbook with GitHub actions and pages.

samueldr:
> Since Nix uses mdbook now for documentation generation is this the plan for the other manuals as well?

It's a possibility. As part of next steps, we need to get the different interested parties talking about the tooling.

Part of interested parties are:

 - documentation people
 - website people

As previously stated, this is only about the input format; nothing the tooling.

asymmetric:
Could someone with admin powers (or @mboes) fix the link in the PR description? :)

FRidh:
> Since Nix uses mdbook now for documentation generation is this the plan for the other manuals as well?

As far as I could discover mdbook lacks the possibility for the amount of nesting that is needed for Nixpkgs, which manual is significantly larger than the Nix one.

garbas:
> > Since Nix uses mdbook now for documentation generation is this the plan for the other manuals as well?
> 
> As far as I could discover mdbook lacks the possibility for the amount of nesting that is needed for Nixpkgs, which manual is significantly larger than the Nix one.

Nesting (or folding as mdbook calls it) is configurable with `[output.html.fold]` -> https://rust-lang.github.io/mdBook/format/config.html

doronbehar:
> Nesting (or folding as mdbook calls it) is configurable with `[output.html.fold]` -> https://rust-lang.github.io/mdBook/format/config.html. You can actually adjust the HTML output as you want, either by config, by template
> or by writing extensions.

Thanks for the link, I couldn't find it myself before. But it seems like it's related to the way the content is presented and it doesn't seem like there is or there isn't a restriction on the amount of levels of sections.

MikiVanousek:
> So i would rather use one of the never, popular tools. Hugo is extremely popular in the space of static site generators and is actively developed. We can also use it for the website. We would be very flexible with it's huge ecosystem. (It supports interlinks https://gohugo.io/content-management/cross-references/ but has no native PDF export) (haven't used Gatsby, Jekyll or mdBook)

I have personal experience with Hugo and Jekyll. Jekyll is a pain to get working on NixOS and generally way less intuitive. Hugo is great, simple and extremely fast. I have no knowledge of Go, still, Hugo is an amazing tool.

roberth:
Here's a migration plan for the module system `description`s, etc. https://github.com/NixOS/nixpkgs/issues/175586

PR #75, opened 2020-08-16
[RFC 0075]: Declarative wrappers

doronbehar:
[Rendered](https://github.com/doronbehar/rfcs/blob/declarative-wrappers/rfcs/0075-declarative-wrappers.md).


rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,353 @@
...
+
+A contributor using `wrapGeneric` shouldn't _care_ what type of wrapping needs
+to be performed on his derivation's executables - whether these are Qt related
+wrappings or a Gtk / gobject related. `wrapGeneric` should know all there is to
+know about environment variables every library / input may need during runtime,
+and with this information at hand, construct the necessary wrapper.
+
+In order for `wrapGenric` to know all of this information about our packaged
+libraries - the information about runtime env, we need to write in the
+`passthru`s of these libraries, what env vars they need. Such information was
```
FRidh, line 224: Limiting ourselves to environment variables is at times too limiting. Also, environment variables should be the implementation detail, although the variable names could be used as identifier.

##### Python example.
If you have a closure of packages, and there happens to be a Python application with Python libraries as dependencies, then that group of packages needs to be composed. If there is another Python application depending on other libraries, then that needs to be composed separately. At least, they are not allowed to share the same `*PYTHONPATH`. (Note one also should not use `PYTHONPATH` but that's a different discussion.)

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,353 @@
...
+separate derivation so that mere changes to the environment will not trigger a
+rebuild. Make it easier to debug why env vars are added to an executable, by
+using Nix as the language to evaluate what env vars are needed, instead of not
+documented good enough and not easily debug-able shell hooks.
+
+# Motivation
+[motivation]: #motivation
+
+We have numerous issues regarding wrappers and our wrapper shell hooks. Here's
+a list of them, sorted to categories.
```
FRidh, line 25: I'd suggest using bullet points here to describe what issues this RFC resolves, before going into the details.

rfcs/0075-declarative-wrappers.md
doronbehar: > If there is another Python application depending on other libraries, then that needs to be composed separately. At least, they are not allowed to share the same *PYTHONPATH

I see this is all about ABI compatibility, and I understand the issue while I remember you mentioned this in https://github.com/NixOS/nixpkgs/pull/85103 . I'm still unsure whether it's possible it'd happen, assuming our libraries are composed OK. @FRidh could you perhaps give a situation (with perhaps some rough code examples) to a situation where say Python2's and Python3's libraries will share the same `*PYTHONPATH`? It's important to note that the current design counts upon `buildInputs` and `propagatedBuildInputs` only.

FRidh: Different Python version is one case, the other is simply applications requiring different versions of a dependency.

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,353 @@
...
+
+@ttuegel TBH I can't tell if declarative wrappers might help, but I'm linking
+this issue here because @worldofpeace wrote it might be related to wrappers?
+Feel free to suggest removing this in the RFC review.
+
+- [issue 49132](https://github.com/NixOS/nixpkgs/issues/49132)
+- [issue 54278](https://github.com/NixOS/nixpkgs/issues/54278)
+- [issue 39493](https://github.com/NixOS/nixpkgs/issues/39493)
+
+`GDK_PIXBUF` compatibilities? I haven't investigated them to the details, so feel
```
jtojnar, line 137: Both of the issues related to gdk-pixbuf only happened for non-wrapped derivations and current `wrapGAppsHook` would already fix them.

The issue was, IIRC, that gdk-pixbuf NixOS module sets `GDK_PIXBUF_MODULE_FILE` environment variable, which was enabling binary gdk-pixbuf modules with ABI incompatible with the gdk-pixbuf the programs linked against. In [one case](https://github.com/NixOS/nixpkgs/issues/54278) it was caused by a bug that slipped through the ABI stability guarantees, in the [other](https://github.com/NixOS/nixpkgs/pull/60254) by running a program for `i686` on `x86_64`.

Either way, it is somewhat orthogonal to wrappers. Only that we might want to always wrap graphical applications with librsvg gdk-pixbuf module.

jtojnar, line 137: Though there is one concern around `GDK_PIXBUF_MODULE_FILE` – the environment variable supports only a single path so we need to build the cache containing all the modules using `gdk-pixbuf-query-loaders` – the [current hook actually does not actually do that](https://github.com/NixOS/nixpkgs/blob/15affbfbd590275e9d123644b706436d285fd358/pkgs/development/libraries/gdk-pixbuf/setup-hook.sh#L3-L13)! but it is required for some image formats to work in e.g. `eog`. We will need to take inspiration in [the module](https://github.com/NixOS/nixpkgs/blob/15affbfbd590275e9d123644b706436d285fd358/nixos/modules/services/x11/gdk-pixbuf.nix#L11-L25).

michaelpj:
> In my opinion we need some kind of passthru "tag" system (where the tags can have values) that is used in conjuction with a buildEnv that takes a list of functions that use these tags to decide how to perform the composition.

This reminds me of @nbp's [old proposal](https://github.com/NixOS/rfcs/pull/3) which took this to the extreme of delaying *any* construction of derivations until the last minute. This has the advantage of giving you access to all the information that goes into a derivation when composing things, like shoving everything in `passthru`.

Obviously this is a radical change (and the motivation in the previous proposal was quite different), but I think it would help here too.

rfcs/0075-declarative-wrappers.md
timokau: This can happen for example when you have a python3 application that shells out to an python2 application.

timokau:
> This reminds me of @nbp's old proposal which took this to the extreme of delaying any construction of derivations until the last minute.

Which in turn sounds a lot like the [Nix-lang 2.0](https://gist.github.com/edolstra/29ce9d8ea399b703a7023073b0dbc00d) proposal. We shouldn't extend the scope of this RFC too much, but its at least somewhat related.

rfcs/0075-declarative-wrappers.md
FRidh: Currently we prevent some of these effects by unsetting `NIX_PYTHONPATH` when a Python program, thereby avoid leakage. This problem with wrappers and environment variables should also be noted clearly in the RFC.

FRidh:
Many derivations nowadays have multiple outputs. Will this RFC consider multiple outputs? If so, let's say we have a multiple-output derivation, and we only needed to wrap `bin/` which happens to be in the `"bin"` output, but there are also other outputs, and one of those outputs is in `extraOutputsToInstall`. Do we expect that, through our wrapper derivation, that output is picked up and included as expected?

doronbehar:
> Many derivations nowadays have multiple outputs. Will this RFC consider multiple outputs? If so, let's say we have a multiple-output derivation, and we only needed to wrap bin/ which happens to be in the "bin" output, but there are also other outputs, and one of those outputs is in extraOutputsToInstall. Do we expect that, through our wrapper derivation, that output is picked up and included as expected?

Since `extraOutputsToInstall` is a NixOS option, wrapped derivations should be handled by a NixOS module. Here's the vision I had for the far future regarding this:

Once declarative wrappers will get mature, we should make `environment.systemPackages` detect derivations that are wrapped with `wrapGeneric`. Then, if needed, make the module call `wrapGeneric` differently then how it is called in `all-packages.nix` while considering `extraOutputsToInstall`.

In general, `wrapGeneric` should symlink to a single derivation both `meta.outputsToInstall` and what's in `extraOutputsToInstall`.

rfcs/0075-declarative-wrappers.md
doronbehar: OK I think I start to understand better the concerns. But still, this is not a "Drawback" of the idea, I mean, we still need to workaround such edge cases specifically in our current methods right? I agree though that it might be more complicated if someone is obliged to use `wrapGeneric` and not a wrap hook.

As an implementation detail, it may be worth noting that it'd be nice to make `wrapGeneric` able to "catch" such incompatible values in an environment variable.

doronbehar:
Today I explored [`@nmattia`'s idea](https://www.nmattia.com/posts/2019-10-08-runtime-dependencies.html) for getting the build and runtime inputs of a derivation, distinctively. I haven't yet fully implemented it. But, the ending notes in that blog post suggests that using this implementation for our purposes would result in all wrapper derivations need to IFD. Does that mean these derivations won't evaluate on Hydra? Are there other implications?

lheckemann:
Yes, IFD is unfortunately no good for anything built by hydra and is thus generally not allowed for any "standard" use of nixpkgs. Separation of evaluation and build is important not just because hydra can't deal with them being intermingled, but because it has a massive performance impact by forcing builds to happen _completely_ serially. There are a couple of niche features in nixpkgs that use IFD, like [exposing kernel config to nix evaluation](https://github.com/NixOS/nixpkgs/blob/master/pkgs/os-specific/linux/kernel/manual-config.nix#L38-L39), yarn2nix's "easy mode" (term invented by me ;) ), and [replaceDependency](https://github.com/NixOS/nixpkgs/blob/92a047a6c4d46a222e9c323ea85882d0a7a13af8/pkgs/build-support/replace-dependency.nix); none of them can be used in a mainstream manner because of their reliance on IFD.

doronbehar:
Maybe we can make an exception for declarative wrappers as well? If the goal is only to create wrappers, doing it shouldn't require too much resources, so it can happen on the users' machine, were the unwrapped derivations _will_ be fetched from hydra.

lheckemann:
The trouble is that other packages may depend on wrapped packages, at which point it does become a major problem — consider that all python packages containing scripts will be wrapped, and that many a python package containing scripts will be depended on by other packages. Additionally, most of the desktop-related tests (which we definitely want to run on hydra!) will probably depend on GNOME apps, which are wrapped…

nmattia:
> But, the ending notes in that blog post suggests that using this implementation for our purposes would result in all wrapper derivations need to IFD.

Hey guys! I have no context whatsoever, so this is just drive by. Quoting the ending notes mentioned above:

> By far the most frightening piece of code today was:
>
> `unique_by(.path) | map(select(. as $obj | $runtime | any(.[] | . == $obj.path))| .report)| .[]`
> [...]
> Given that Nix has readFile and fromJSON, we could have lifted cinfo’s output to a Nix value, then performed all the matching on that value, and generated the report as a Nix string directly! Added bonus, the “buildtime-only” reports would never have been generated, thanks to laziness!
> 
> The problem from readFile-ing cinfo is … import from derivation. I stay away from it, but YMMV, try it! [...]

If I'm reading this correctly, the solution implemented in the article does not use IFD (we use it on Hydra). There _is_ a simpler way to implement it, using IFD, but this particular solution is _not_ using IFD.

doronbehar:
Hello @nmattia and thanks for joining in. I didn't intent to mention you but it was unavoidable in the commit message. I'm glad you are willing to give your opinion on the matter :).

To summarize what's relevant to you, we (me) would like to get a list of all inputs and the inputs' inputs and so on, of a given derivation. Additionally, we'd like to get all references and their references (recursively as well), distinctively from the unreferenced inputs. Both lists should ideally be available to us in a Nix code as a 1 big list of derivations. If I understood correctly your blog post, you are / were capable of printing those lists, but not making Nix evaluate them, unless IFD was used. Is my understanding correct?

Regards.

nmattia:
>  If I understood correctly your blog post, you are / were capable of printing those lists, but not making Nix evaluate them, unless IFD was used. Is my understanding correct?

No, I don't think that's correct, but I wrote this what feels like a lifetime ago! I believe you do have access to the evaluated derivations, without IFD.

The note about the IFD at the end is that it _can_ make the code easier, ... at the cost of IFD. So yes, it works without IFD (you get the Nix drv objects at eval time) _but_ there's an ugly bit of `jq` code. You can get rid of the ugly bit of `jq` code by using IFD.

_I think._

Mic92:
Anyone who wants shepherd this one?

Profpatsch:
If you can pull this off @doronbehar, I salute you.

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,407 @@
...
+
+- [issue 95027](https://github.com/NixOS/nixpkgs/issues/95027)
+- [issue 23018](https://github.com/NixOS/nixpkgs/issues/23018)
+- [issue 11133](https://github.com/NixOS/nixpkgs/issues/11133)
+- [pull 95569](https://github.com/NixOS/nixpkgs/pull/95569)
+
+Since our wrappers are shell scripts, `gdb` can't run them. What if we had
+written a C based wrapper, that perhaps would read what environment it needs to
+set from a JSON file, and it will call the unwrapped original executable? I
+need feedback regarding whether `gdb` will play nice with this.
```
Mic92, line 220: An alternative to this is the libtool approach, which still allow you to use strace,gdb etc: https://www.gnu.org/software/libtool/manual/html_node/Invoking-libtool.html
If a binary wrapper is used than it should be compiled for every target as this increases package size/build time. Instead having the wrapper reading its actual configuration from the same directory might work.

lheckemann:
I nominate @Profpatsch, @jtojnar, @timokau, @FRidh as shepherds for this RFC. Are any of you available for it?

I also nominate myself and accept my self-nomination ;)

timokau:
Thanks for the nomination! Unfortunately I have to decline.

FRidh:
I accept the nomination, though I wonder how it is when actively contributing to larger parts of the text in the RFC. I suppose that should be fine?

Mic92:
> I accept the nomination, though I wonder how it is when actively contributing to larger parts of the text in the RFC. I suppose that should be fine?

I don't think there is anything wrong with that.

Mic92:
We need two more shepherds for this RFC. @doronbehar do you know someone who would be willing to do this?

doronbehar:
> We need two more shepherds for this RFC. @doronbehar do you know someone who would be willing to do this?

Not really. IIRC `@ttuegel` has expressed interest in this idea, but I guess he's busy[1].

I am myself going to get less available in the next 3 months, for my 3rd semester has started yesterday. I've also had a few ideas regarding the implementation that could help this but I've yet settled down to write them.[2]

[1]: Just my impression as we have some "urgent" Nixpkgs' issues he won't reply.
[2]: In short: i) Limit the depth of the deps-of-deps recursion, to make evaluation a bit lighter and ii) use different json files in `$out/bin/` or where ever needed to express dependencies on environment (to later be read by a binary wrapper).

edolstra:
Nominating myself as a shepherd.

edolstra:
So this RFC has the following shepherd: @FRidh, @lheckemann, @edolstra. Thanks!

spacekookie:
@FRidh do you want to be the shepherd leader on this?

wmertens:
This is great, some thoughts:

* I agree that Nix should be used to define the wrapper, and we'll figure out how to simplify it along the way
* Issue https://github.com/NixOS/nixpkgs/issues/11586 is a problem that will be solved by this, on macOS (and some others) scripts can't be in shebang line
* This could also be a stepping stone for late binding, which would allow better resource use in combination with an [intensional store](#17). For example, a custom binary loader (one each for ELF, MACH, ...) would read wrapper data from the store, and resolve libraries based on that only
* Some relevant thoughts on self-references, taken from https://github.com/ebkalderon/merkle-tree-nix-store-thing/issues/3#issue-748616467:
  > ### Self-references
  > The subject of self-references is a tricky one. In particular, because the package object's hash is derived from its contents on disk, self-references are impossible to support. Computing package hashes modulo self-references and rewriting them after the fact, as suggested in [NixOS/rfcs#17](https://github.com/NixOS/rfcs/pull/17), would not work in our Merkle tree design because it creates a chicken-and-egg problem where the on-disk content no longer exactly matches its content hash.
  > 
  > I believe this can be remedied using a combination of relative paths together with the `$ORIGIN` (Linux) and `@loader_path` (macOS) ([reference](https://wincent.com/wiki/@executable_path,_@load_path_and_@rpath)) special prefixes, patched in after the fact, similar to how Spack does things ([source code](https://github.com/spack/spack/blob/a80d221bfa1c9be4b2b9eff9f057edf62c34e50b/lib/spack/spack/relocate.py)). This is tricky to pull off reliably, as described in [Strategies for Binary Relocation in Functional Build Systems](https://maxmcd.com/posts/strategies-for-binary-relocation/) by Max McDonnell (this article will be useful for implementing binary relocation later), but [the author of Spack responded to the post](https://lobste.rs/s/2lnncd/strategies_for_binary_relocation#c_btkgc0) clarifying a few things, and I believe it should work well enough for eliminating the problem of self-references. In summary, I believe we should:
  > 
  > 1. Use absolute paths (like Nix) for everything whenever possible.
  > 2. Use relative paths and `$ORIGIN`/`@loader_path` for self-references. Rewriting hashes ex post facto just won't cut it.

FRidh:
> @FRidh do you want to be the shepherd leader on this?

Yes, I can do that.

Let's have a first meeting in January. Please enter what works for you at https://www.when2meet.com/?10552518-F87eU @lheckemann @doronbehar @edolstra

doronbehar:
I will be available on the weekends only I'm afraid, marked it on the web app. I can also (and slightly prefer) do it in a weekend sooner then what's available there :pray: .

edolstra:
@FRidh I've filled in the doodle, sorry for not noticing it earlier!

FRidh:
Thanks. Good you remind me as well!

Let's meet on January 15th at 12:00 UTC in https://meet.jit.si/NixRFC75.

FRidh:
Reminder we have the meeting in less than 2 hours from now.

edolstra:
We absolutely should not move computations that can be done in bash (at build time) to Nix (at evaluation time). Evaluation is already very slow and memory-hungry. Code such as https://github.com/NixOS/nixpkgs/blob/2aacb90748d2cfa1d64a8faf492d0ae4fc889394/pkgs/build-support/neowrap.nix will likely exacerbate this problem massively. (I haven't read the code in detail, but I see things like "recursive function that goes deep through the dependency graph of a given list of packages and creates a list of all buildInputs they all depend on" and `replaceAllOutputs` that seem very expensive. Reminder that Nix is not a general purpose language... It's not convenient or efficient for writing that kind of code.)

If we ever switch to structured attrs, we could have a nicer syntax for specifying the input to `makeWrapper`. But the work should still be done at build time.

I also don't know if it's a good idea in general to have separate wrapped and unwrapped derivations by default. It makes sense for cases like Firefox where you want to be able to update the plugin path without rebuilding all of Firefox. But for most packages, having a separate derivation is just a nuisance.

FRidh:
Thank you for the meeting.

# RFC 75 meeting 1

Attendants:
- @doronbehar
- @edolstra
- @FRidh
- @lheckemann

## Scope

RFC as is covers:

1. Declarative wrappers
2. Wrapping in separate derivation

Different use cases:
1. Composition (Python env) / configuration (firefox, plugins) using new
   derivation
2. Adding required dependencies of a derivation

Standard interface. Both are covered in RFC to consider and avoid
double-wrapping.

Avoid double wrapping through smarter `makeWrapper`? `makeWrapper` with
language-specific hooks to inject header with configuration (e.g. `sys.path` in
case of Python).

Structured attributes are not yet available in `stdenv.mkDerivation` in Nixpkgs.
There are work-arounds, by passing a JSON file with your configuration to your
derivation.

During build time the (structured) attributes of dependencies are not available.

## Recursion

How to compose? We need to recurse into dependencies to find what to put
together. We typically use setup hooks. When to stop recursing? For example,
with Python packages we do not want to mix different Python environments that
exist in the transitive closure.

During eval-time ([as done for
Python](https://github.com/NixOS/nixpkgs/pull/102613) with
`requiredPythonModules`)  is expensive. How about writing "evaluated
dependencies" as JSON to `$out/nix-support`?

## Binary wrappers

There is also the idea of binary wrappers that solves the issue of nested
shebangs on Darwin. They could also use a declarative interface. [Possible
implementation](https://github.com/NixOS/nixpkgs/pull/95569). Maybe the same
JSON instruction. Maybe not create a new binary wrapper for every executable but
reuse it like is done with setuid?

## Mixing declaration and implementation detail

Current proposal mixes the declaring of what to compose with how the composition
is often implemented. That is, environment variables are used as attributes.
While technically this works, often you want something higher-level, "this is a
Qt" package, "this is a Python package".


## Next step?

Think of a declarative interface to `makeWrapper`.

Take several packages, add to it the `envInputs` and the variables they need and
write it to a JSON file. Then write a script that during build-time will wrap
the target using these JSON files.

Update the RFC to describe this approach.

edolstra:
@doronbehar Have you done any work on this lately (like the next steps in the summary above)? Just wondering what the status is :-)

doronbehar:
My last exam will be on Tuesday the 23 and then I'll be on it!

lheckemann:
Thanks @doronbehar for the work! @FRidh @edolstra shall we have a look over it and arrange a new meeting?

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,366 @@
+---
+feature: Declarative Wrappers
+start-date: 2020-08-16
+author: Doron Behar <me@doronbehar.com>
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: @FRidh, @lheckemann, @edolstra
+shepherd-leader: (name to be appointed by RFC steering committee)
```
Mic92, line 7:
```suggestion
shepherd-leader: @FRidh
```

Mic92:
Ping @FRidh (https://www.when2meet.com/)

doronbehar:
Should I try to prepare something some code for the next release (I'm not sure how to do the first steps)? Or is it more of a talk about the design and affirming that it is good?

Mic92:
> Should I try to prepare something some code for the next release (I'm not sure how to do the first steps)? Or is it more of a talk about the design and affirming that it is good?

Yes. Prototypes are a good way to guide the discussion.

abathur:
This is tangential, but I have a few ~implementation-detail requests related to something I've been working on.

I've been picking at a rudimentary way to scan package executables (for now, primarily with YARA) and categorize them based on how likely they are to exec their arguments. [resholve](https://github.com/abathur/resholve) will use this information to single out invocations of these executables for extra scrutiny. 

At least as implemented now, the analysis needs to be able to distinguish wrappers from other executables and associate compiled wrappers with the wrapped executable. At the Nix level, I need to be able to readily map a list of packages to produce this analysis for arbitrary inputs.

At least for my use, I don't think this has to be *in* the wrapper--I'd see it as a win if this information was stored in an easy-to-use format at a stable location in $out.

rfcs/0075-declarative-wrappers.md
lheckemann: @Mic92 do you mean it should _not_ be compiled for each package?

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,367 @@
...
+   This should be done with `makeWrapperAuto` as well, see (2).
+2. Design the `makeWrapperAuto` shell hook:
+  - It should introduce a shell function (to be called `wrappersInfo`) that
+    will allow piping a JSON string from `builtins.toJSON` and spit a
+    `wrappers.json` that will include both what was piped into it, and the
+    content from the package's various inputs' `wrappers.json` files.
+  - It should make the executables in `$out/bin/` get wrapped according to
+    what's currently in this package's `wrappers.json`, during `fixupPhase`.
+  - The above should be also possible to do manually for executables outside
+    `$out/bin/` with say adding to a derivation a Nix variable:
```
lheckemann, line 285:
```suggestion
    `$out/bin/` by setting `wrapExtraPrograms` on the derivation:
```
Maybe we should also not wrap everything in `$out/bin` indiscriminately though? Maybe we should have a variable specifying _exactly_ which files to wrap, not just which files to wrap in addition to `$out/bin`.

lheckemann (commented):
I've made it about halfway through, will add more soon :)

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,367 @@
...
+Most of the work to do will be:
+
+1. Gather information about what environment variables are "linked" to each
+   package, and edit these derivations to include a `wrappers.json` in them.
+   This should be done with `makeWrapperAuto` as well, see (2).
+2. Design the `makeWrapperAuto` shell hook:
+  - It should introduce a shell function (to be called `wrappersInfo`) that
+    will allow piping a JSON string from `builtins.toJSON` and spit a
+    `wrappers.json` that will include both what was piped into it, and the
+    content from the package's various inputs' `wrappers.json` files.
```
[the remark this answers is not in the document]
doronbehar: Yes, perhaps the name `wrappersInfo` isn't very explanatory, and `combineWrappers` or alike is better. I don't think though this function should be split to two. By piping I meant (as explained further on):

```nix
  postInstall = ''
    echo "${builtins.toJSON {
      GST_PLUGIN_SYSTEM_PATH_1_0 = [
        # @out@ should be expanded by `combineWrappersInfo` to what's in `$out`, see:
        # https://github.com/NixOS/nixpkgs/pull/85103#issuecomment-613071343
        "@out@/lib/gstreamer-1.0"
      ];
    }}" | combineWrappersInfo
  '';
```

`passAsFile` seems **almost** appropriate to me :). If `stdenv.mkDerivation` wouldn't have mind having attributes it can't coerce into a string, but pass them as json files. To explain, this example doesn't evaluate:

```nix
  extraWrappersInfo = {
    GST_PLUGIN_SYSTEM_PATH_1_0 = [
      "@out@/lib/gstreamer-1.0"
    ];
  };
  passAsFile = [ "extraWrappersInfo" ];
  preBuild = ''
    cat $extraWrappersInfoPath
  '';
```

rfcs/0075-declarative-wrappers.md
doronbehar: > I'd suggest using bullet points here to describe what issues this RFC resolves, before going into the details.

Done.

doronbehar:
Thanks @lheckemann for the review. I've made small progress thanks to your suggestion. I will have time to write some code after the 1st week of August as it seems.

rfcs/0075-declarative-wrappers.md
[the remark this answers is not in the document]
lheckemann: We can modify stdenv to accept `extraWrappersInfo` as a "special" attribute, like `buildInputs` or `pname`. `stdenv.mkDerivation` can then handle converting `extraWrappersInfo` to a JSON document which is what is then passed to the `derivation` builtin and added to `passAsFile`.

Conceptually, this would look like

```nix
mkDerivation = {
…
wrapperInfo ? null
…
}:
derivation ({
  …
} // optionalAttrs (wrapperInfo != null) {
  wrapperInfo = builtins.toJSON wrapperInfo;
  passAsFile = ["wrapperInfo"];
})
…
```

Plus a `writeWrapperInfoPhase` or similar which scans `buildInputs` for wrappers and combines them with `wrapperInfo`.

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,372 @@
...
+
+1. Gather information about what environment variables are "linked" to each
+   package, and edit these derivations to include a `wrappers.json` in them.
+   This should be done with `makeWrapperAuto` as well, see (2).
+2. Design the `makeWrapperAuto` shell hook:
+  - It should introduce a shell function (to be called `combineWrappersInfo`) that
+    will allow piping a JSON string from `builtins.toJSON` and spit a
+    `wrappers.json` that will include both what was piped into it, and the
+    content from the package's various inputs' `wrappers.json` files.
+  - It should wrap the executables in `$out/bin/` according to
```
lheckemann, line 287: Maybe we should also not wrap everything in `$out/bin` indiscriminately. A variable could specify which programs to wrap (as a list of glob patterns?), defaulting to `$out/bin/*` but permitting an override to _not_ wrap everything in `$out/bin` and/or _also_ wrap things outside `$out/bin`.

Also, should the wrapping apply to all outputs in multiple-output derivations? My guess (though not well-thought-out) is that `$out/bin/*`, `$bin/bin/*`, and `$out/libexec/**/*` should be wrapped by default.

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,372 @@
...
+
+Currently I imagine the usage of `combineWrappersInfo` (the name can be debated) as
+so:
+
+```nix
+  # Propagate GST plugins' path
+  postInstall = ''
+    echo "${builtins.toJSON {
+      GST_PLUGIN_SYSTEM_PATH_1_0 = [
+        # @out@ should be expanded by `combineWrappersInfo` to what's in `$out`, see:
```
lheckemann, line 322: We should probably use [`${placeholder "out"}`](https://nixos.org/manual/nix/stable/#builtin-placeholder) here? Or am I missing something?

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,372 @@
...
+        # https://github.com/NixOS/nixpkgs/pull/85103#issuecomment-613071343
+        "@out@/lib/gstreamer-1.0"
+      ];
+    }}" | combineWrappersInfo
+  '';
+```
+
+`wrapQtAppsHook` and `wrapGAppsHook` should be replaced with `makeWrapperAuto`
+while enable derivations to get rid of well known workarounds such as:
```
lheckemann, line 331:
```suggestion
which enables removing frequent workarounds such as:
```

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,372 @@
...
+```nix
+  preFixup = ''
+    makeWrapperArgs+=("''${qtWrapperArgs[@]}")
+  '';
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Using `wrapProgram` will be simpler then using `combineWrappersInfo` and it might be
```
lheckemann, line 350:
```suggestion
Using `wrapProgram` is simpler than using `combineWrappersInfo` and it might be
```

rfcs/0075-declarative-wrappers.md
doronbehar: A list of glob patterns is a good idea.

rfcs/0075-declarative-wrappers.md
doronbehar: > We should probably use [`${placeholder "out"}`](https://nixos.org/manual/nix/stable/#builtin-placeholder) here? Or am I missing something?

A lot of people ask that, see https://github.com/NixOS/nixpkgs/pull/85103#issuecomment-613071343 - I linked that explanation in the comment in the code snippet.

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,367 @@
...
+    "/nix/...gobject-introspection.../...",
+    "/nix/...librsvg.../...",
+  ],
+  "GIO_EXTRA_MODULES": [
+    "/nix/...dconf.../lib/gio/modules"
+  ],
+  "XDG_DATA_DIRS": [
+    "/nix/...gtk+3.../...",
+    "/nix/...gsettings-desktop-schemas.../..."
+  ],
```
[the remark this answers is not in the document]
doronbehar: I thought about answers to some of these questions back when I implemented https://github.com/NixOS/nixpkgs/pull/85103/files#diff-aad93ba22adfe208c7b970a7b9d0f73dba9306b8a4fc4eaacfb2ac331aed9ea1R51-R74 . I prepared some answers in the RFC to be pushed, but I don't know:

- Why order would matter?
- I never encountered a case where paths should be escaped, or when other values should be escaped. In case of environment variables values that need escaping, they probably will not be originated from packages, and that can be handled in the `postInstall` snippet example below, by manually quoting values in the json file..

> We can also go further. Instead of setting `$PATH` to a massive list does it make more sense to build a folder of symlinks and set `$PATH` to a single element? If so will we special-case some variables or should we use a more detailed schema (as proposed below)?
> 

I thought about this exactly even back when I implemented https://github.com/NixOS/nixpkgs/pull/85103 so it will be possible to avoid what https://github.com/NixOS/nixpkgs/pull/84689 fixed. I addressed this idea in the changes I made to the RFC that I haven't pushed yet.

I'm not sure I understand:

> Furthermore does this replace or prepend or append? Replace seems the purest but commands such as `parallel -- command-from-my-path` are not uncommon. In that case there seems to be no good option.

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,367 @@
...
+
+The end goal is to make the experience of getting a derivation wrapped as
+automatic as possible. A derivation that needs some environment variables in
+order to work will get these environment variables set in the wrapper by
+`mkDerivation`'s standard `fixupPhase`.  As a start, we'll (me probably) will
+introduce a new `makeWrapperAuto` setup hook that will take care of this in the
+way described as follows.
+
+As a start, we'll need to think about all packages in Nixpkgs that using them
+requires some environment variables to be set. Every such package will put in
```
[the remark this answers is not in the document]
doronbehar: > The previous wording seemed a bit funny. Does this sound better?

Yes this is better. Haven't pushed this rewording yet.

rfcs/0075-declarative-wrappers.md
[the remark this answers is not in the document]
kevincox: > Why order would matter?

If two packages have a `bin/foo` then the order will decide which one is used. For most of these `PATH`-like variables the first match is used.

> I never encountered a case where paths should be escaped, or when other values should be escaped.

It is rare but does happen. For things like `PATH` I don't think there is even a way to quote elements. You simply can't have a path that contains `:` in your `PATH`. At the very least it would be nice to raise an error. It would be even better if we can "alias" the path to a safe name or escape the invalid characters (if supported by the variable, which is rare)

> I'm not sure I understand:

My point is that the argument to a parallel is a command to run, but that command should be pulled from the PATH of the calling shell, not a PATH that may be set but a wrapper in the parallel derivation.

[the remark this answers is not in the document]
doronbehar: > > Why order would matter?
> 
> If two packages have a `bin/foo` then the order will decide which one is used. For most of these `PATH`-like variables the first match is used.

I think it's super rare to have two packages providing the same binary like that, and finding them used together in the wrapping while wishing only one package's `out/bin` to rule..

Also paths escaping is too rare, so I think it'll be best to leave these topics out of scope for the RFC.

As for `parallel` commands. Do you mean `parallel` invocations in the builds of derivations? something like:

```nix
  preInstall = ''
    parallel -- commands that take a-long time
  '';
```

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,367 @@
...
+
+```nix
+  # Propagate GST plugins' path
+  postInstall = ''
+    echo "${builtins.toJSON {
+      GST_PLUGIN_SYSTEM_PATH_1_0 = [
+        # @out@ should be expanded by `wrappersInfo` to what's in `$out`, see:
+        # https://github.com/NixOS/nixpkgs/pull/85103#issuecomment-613071343
+        "@out@/lib/gstreamer-1.0"
+      ];
```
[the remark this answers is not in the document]
doronbehar: > Wouldn't `echo "${builtins.toJSON { foo = "bar"; }}"` result in `echo "{"foo":"bar"}"`? Or am I missing something here?

Oh I understand now.. Perhaps we should also use `lib.strings.escapeShellArg`. I have changed this composition into something much nicer.

rfcs/0075-declarative-wrappers.md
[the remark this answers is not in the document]
kevincox: I don't think collisions are super rare. After all nix has a priority system for some reason. I think we should at least define how this is handled. 

As for the escaping I think that it is fine to ignore this for now. But I think it should be mentioned explicitly. 

For the "parallel problem" I mean in general. Any time you have a wrapper that both needs commands itself as well as forwards a user's command we have this problem. I think we should discuss how this is handled. Currently in nixpkgs it is very inconsistent. 

But the most important point still stands. We need *some* configuration. If only to specify the separator. So I think that needs to be included.

tomberek:
While thinking about `nix shell`/`nix-shell`/shebangs, a similar [comment](https://github.com/NixOS/nix/pull/5162#issuecomment-908255178) about structured env vars for flakes was brought up. The potential for redesign of language-specific shell hooks may impact those tools too.

doronbehar:
Back when I conceived this RFC I just started using NixOS, and at that time I encountered issues with wrappers frequently. Nowadays it's not like that, possibly because many issues got fixed, and perhaps also because I switched from AwesomeWM to Gnome which adds many environment variables to my default environment.

The point of this comment is to say that I lost a bit the motivation to implement this RFC, and I also got a lot busier. I thought I'd find time for this during the summer break but I didn't, and I'm not sure what is the state of this RFC - do we need to see a POC implementation? (to be written by me?) Or are there any details still missing?

I still think this RFC has a lot of value, but I don't like to keep edges open, if I have a responsibility.. just sharing my thoughts.

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,408 @@
...
+should be able to switch to `combineWrappersInfo` and declare any nontrivial
+environment variables with it to get propagated to reverse dependencies and to
+it's executables automatically.
+
+Currently I imagine the usage of `combineWrappersInfo` (the name can be debated) as
+so:
+
+```nix
+  # Propagate GST plugins' path
+  postInstall = makeWrapperAuto.combineWrappersInfo {
```
doronbehar, line 336: I wonder whether this is a good design... `combineWrappersInfo`, defined as a Nix function or as a shell function needs to iterate `buildInputs` and `propagatedBuildInputs`. It means that as a Nix function using it will require the user to:

```nix
inherit buildInputs propagatedBuildInputs;
```

In addition to the other arguments. I worry this might make the experience of using this function a bit clumsy.

OTOH, defining `combineWrappersInfo` as a shell function, means that it would need to iterate the inputs via shell which is less clumsy. But, how do I get the dev output of an input? I'm asking because I need to get to the `nix-support/wrappers.json` file which might reside In the `dev` output. Is there a bash equivalent to `lib.getDev`?

doronbehar:
I made some progress at:

https://github.com/NixOS/nixpkgs/pull/144722

With `combineWrappersInfo` as a Nix function. Thinking more about it persuades me that it's the right choice.

rfcs/0075-declarative-wrappers.md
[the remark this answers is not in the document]
SuperSandro2000: PATH order is support important. Tools like trash, safe-rm and molly-guard heavily rely on it. Also I commonly use a higher priority entry in PATH to create wrappers for none interactive shells. For environment variables which are similar to PATH like NIX_PATH order is also super important and ordering the entries for example alphabetically breaks things.

What is safe to do is deduplication when an entry is encountered the second time and I think this should be done by default to prevent to long environment variables.

rfcs/0075-declarative-wrappers.md
SuperSandro2000: I don't think there is a function like this in bash.

rfcs/0075-declarative-wrappers.md
```diff
@@ -0,0 +1,408 @@
...
+different inputs list?):
+ 
+```
+$ nix why-depends -f. beets gobject-introspection.dev
+/nix/store/93lfrhm8vp17m8ziqi8vp6v4cff67wkb-beets-1.4.9
+╚═══bin/beet: …-expat-2.2.8-dev/bin:/nix/store/y3ym76wrak3300vsjyf3klr52cnzmxwd-gobject-introspection-1.64.1-de…
+    => /nix/store/y3ym76wrak3300vsjyf3klr52cnzmxwd-gobject-introspection-1.64.1-dev
+```
+
+## Other issues
```
LunNova, line 180: More related issues with wrappers https://github.com/NixOS/nixpkgs/issues/160923 and the ones linked at https://github.com/NixOS/nixpkgs/issues/160923#issuecomment-1046067348

rfcs/0075-declarative-wrappers.md
rycee: Worked a bit on a wrapper tool some time ago: [svep](https://git.sr.ht/~rycee/svep). Maybe there are some useful ideas there?

doronbehar: > Worked a bit on a wrapper tool some time ago: [svep](https://git.sr.ht/~rycee/svep). Maybe there are some useful ideas there?

That's a nice tool :), but I'm not sure it helps a lot, it seems it only supplies an alternative to `makeCWrapper`, but that depends on rust, and not pure C99.

We won't be able for example to use it to save the wrapping information of a package, as we want to access that information independently, from `${pkg}/nix-support/`, and not from the binaries that the package may distribute.

PR #42, opened 2019-03-23
[RFC 0042] NixOS settings options

Infinisil:
#### Part 1: Structural `settings` instead of stringly `extraConfig`

NixOS modules often use stringly-typed options like `extraConfig` to allow specifying extra settings in addition to the default ones. This has multiple disadvantages: The defaults can't be changed, multiple values might not get merged properly, inspection of it is almost impossible because it's an opaque string and more. The first part of this RFC aims to discourage such options and proposes generic `settings` options instead, which can encode the modules configuration file as a structural Nix value. Here is an example showcasing some advantages:

```nix
{
  # Old way
  services.foo.extraConfig = ''
    # Can't be set in multiple files because string concatenation doesn't merge such lists
    listen-ports = 456, 457, 458 
    
    # Can't override this setting because the module hardcodes it
    # bootstrap-ips = 172.22.68.74

    enable-ipv6 = 0
    ${optionalString isServer "check-interval = 3600"}
  '';
  
  # New way
  services.foo.settings = {
    listen-ports = [ 456 457 458 ];
    bootstrap-ips = [ "172.22.68.74" ];
    enable-ipv6 = false;
    check-interval = mkIf isServer 3600;
  };
}
```

Jump to the [detailed design](https://github.com/Infinisil/rfcs/blob/config-option/rfcs/0042-config-option.md#part-1-1) of part 1

#### Part 2: Balancing module option count

Since with this approach there will be no more hardcoded defaults and composability is not a problem anymore, there is not a big need to have NixOS options for every setting anymore. Traditionally this has lead to huge modules with dozens of options, each of them only for a single field in the configuration. Such modules are problematic because they're hard to write, review and maintain, are generally of lower quality, fill the option listing with noise and more. Additional options aren't without advantages however: They are presented in the NixOS manual and can have better type checking than the equivalent with `settings`.

The second part of this RFC aims to encourage module authors to strike a balance for the number of additional options such as to not make the module too big, but still provide the most commonly used settings as separate options. Quality is encouraged over quantity: Authors should spend more time on writing documentation, NixOS tests or useful high-level abstractions. This is in contrast to the fiddly labor of copying dozens of options from upstream to NixOS. With a `settings` option, it's also very easy to add additional options over time if the need arises. In contrast, removing options has always been nigh impossible.

Jump to the [detailed design](https://github.com/Infinisil/rfcs/blob/config-option/rfcs/0042-config-option.md#part-2-1) of part 2

[...]

PR #78, opened 2020-10-18
[RFC 0078] System-agnostic configuration file generators

7c6f434c:
Extracting configuration file generation out of NixOS service modules in a more reusable way. Based on a discussion during NxCon 2020 Hackday.

### DIFF
--- a/rfcs/0078-configuration-file-generators.md
+++ b/rfcs/0078-configuration-file-generators.md
@@ -1,0 +1,147 @@
+---
+feature: configuration_file_generators
+start-date: 2020-10-18
+author: Michael Raskin @7c6f434c
+co-authors: (in-progress)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Provide the configuration file generation functionality separately from NixOS
+as a whole, and using the module system without the global namespace.
+
+# Motivation
+[motivation]: #motivation
+
+There is currently a lot of code duplication between
+* NixOS
+* home-manager
+* nix-darwin
+
+and obviously growing code duplication with any effort running on below-Tier-2
+platforms. That leads both to effort duplication, and to confusion among users
+using multiple Nix ecosystem tools for service generation,
+
+While fully resolving the problem is likely not fully compatible with the core
+preferences of maintainers on each specific platform, we believe that config
+file generation could be a shared resource just like Nixpkgs based on the
+following properties:
+* unlike, say, upstream `systemd` unit files we typically do not
+  reuse-and-override upstream configs, but generate our own (both in NixOS and
+  other service collections)
+* the configs for a piece of software often have few dependencies on the
+  general environment
+* it can be defined as a purely functional library, simplifying integration
+
+We aim to reduce duplication of effort and code while also having a high level
+of inspection and tweaking possibilities without needing to patch module
+source.
+
+# Detailed design
+[design]: #detailed-design
+
+This RFC proposal builds upon RFCs#42 defining a configuration file abstraction
+approach.
+
+A top-level directory is added to the Nixpkgs repository, and a library of
+functions implementing program configuration file generation is created within
+this directory.  Each such generator takes as an input a NixOS module system
+based attribute set «subtree», e.g. the attribute set typically bound to `cfg`
+variable in the current NixOS service modules.
+
+The output of each such function is an attribute set. The keys are relative
+file names of configuration files, and the values are attribute sets of RFCs#42
+settings abstraction and serialiser.
+
+The function is provided as a member of an attribute set, which also contains
+the corresponding type specifications for input and output modules are defined.
+
+In some cases we only provide low-level overridable default configuration. In
+this case the input may have the type that is always an empty attribute set,
+`{}`.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+A minimalistic silly example is:
+
+Input:
+```
+  {
+     port = 1234;
+     nodeName = "localhost";
+     content = "Hello";
+  }
+```
+
+Output:
+```
+  {
+    "subdir/listen.conf" = {
+      serialiser = toJSON;
+      settings = {
+        listenPort = "1234";
+        serverGreet = "I am localhost";
+      };
+      type = …;
+    };
+    "subdir/content.conf" = {
+      serialiser = toINI;
+      settings = {
+        greeting = {
+          greeter = "localhost";
+        };
+        payload = {
+          value = "I say to you: ‘Hello’";
+        };
+      };
+      type = …;
+    };
+  }
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The proposed approach increases the spreading of the related code, from package
+and service module to package and configuration generator and module.
+
+The proposed refactoring incurs a cost in terms of implementation effort across
+NixOS, and likely to create a medium-term situation of partial migration.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Do nothing, continuing with the current code duplication.
+
+Same, but put configuration generators closer to packages. This would mean
+widespread use of the module system inside `pkgs/`. There is also no guarantee
+that all the configuration files describing interaction of multiple software
+packages will have a clear choice of reference package.
+
+There also have been many solutions proposed based on a significant rework of the module system.
+
+Abstract generation of configuration files with package-like flat arguments and
+plain text file outputs. This approach will need less code as long as it we do
+not want type checks, or ease of overriding values.
+
+Implement a complete service abstraction not tied to global system-wide
+assumptions.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Currently none.
+
+# Future work
+[future]: #future-work
+
+Define generic service abstraction for non-`systemd` systems, possibly with a
+convertor from `systemd` units.
+
+Consider some  way of referring from package to related configuration file
+generators via `meta` or `passthru`, in a way similar to `passthru.tests`.
--- a/rfcs/0078-configuration-file-generators.md
+++ b/rfcs/0078-configuration-file-generators.md
@@ -3,8 +3,8 @@ feature: configuration_file_generators
 start-date: 2020-10-18
 author: Michael Raskin @7c6f434c
 co-authors: (in-progress)
-shepherd-team: (names, to be nominated and accepted by RFC steering committee)
-shepherd-leader: (name to be appointed by RFC steering committee)
+shepherd-team: @ehmry, @edolstra, @roberth
+shepherd-leader: @edolstra
 related-issues: (will contain links to implementation PRs)
 ---
 
--- a/rfcs/0078-configuration-file-generators.md
+++ b/rfcs/0078-configuration-file-generators.md
@@ -39,8 +39,10 @@ following properties:
 
 We aim to reduce duplication of effort and code while also having a high level
 of inspection and tweaking possibilities without needing to patch module
-source. At the same time we aim to minimise the amount of code in the NixOS
-subtree that cannot be tested on NixOS.
+source. An additional goal is to move the more package-like configuration file
+generation code closer to the Nixpkgs model with scope/visibility rules. At
+the same time we aim to avoid or minimise code in the NixOS subtree that
+cannot be tested on NixOS.
 
 # Detailed design
 [design]: #detailed-design
@@ -125,6 +127,13 @@ Same, but put configuration generators closer to packages. This would mean
 widespread use of the module system inside `pkgs/`. There is also no guarantee
 that all the configuration files describing interaction of multiple software
 packages will have a clear choice of reference package.
+
+Introduce a guideline on making configuration files available to the module
+system after generation (i.e. avoidance of only putting them in a `let` to add
+a path reference inside some string; maybe with best practices on naming the
+corresponding options). Does not help with scoping, still needs a non-trivial
+amount of refactoring. Might make other desirable refactorings such as
+multiple instances of a module marginally harder.
 
 A complete or partial merge of the module collections of the major currently
 existing module system based code bases, with internal options for platform
--- a/rfcs/0078-configuration-file-generators.md
+++ b/rfcs/0078-configuration-file-generators.md
@@ -39,7 +39,8 @@ following properties:
 
 We aim to reduce duplication of effort and code while also having a high level
 of inspection and tweaking possibilities without needing to patch module
-source.
+source. At the same time we aim to minimise the amount of code in the NixOS
+subtree that cannot be tested on NixOS.
 
 # Detailed design
 [design]: #detailed-design
@@ -59,6 +60,8 @@ settings abstraction and serialiser.
 
 The function is provided as a member of an attribute set, which also contains
 the corresponding type specifications for input and output modules are defined.
+The specifications for the input should be suitable for importing as a part of
+the NixOS service module options.
 
 In some cases we only provide low-level overridable default configuration. In
 this case the input may have the type that is always an empty attribute set,
@@ -123,7 +126,15 @@ widespread use of the module system inside `pkgs/`. There is also no guarantee
 that all the configuration files describing interaction of multiple software
 packages will have a clear choice of reference package.
 
-There also have been many solutions proposed based on a significant rework of the module system.
+A complete or partial merge of the module collections of the major currently
+existing module system based code bases, with internal options for platform
+specific implementation details. Unfortunately, this would force more coupling
+of the development cycle for the platform-specific parts. It also does not
+create a clean inspection/override point, and mixes the code with different
+platform requirements for testing.
+
+There also have been many solutions proposed based on a significant rework of
+the module system.
 
 Abstract generation of configuration files with package-like flat arguments and
 plain text file outputs. This approach will need less code as long as it we do
@@ -135,7 +146,7 @@ assumptions.
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-Currently none.
+Full proof-of-concept implementation for some module.
 
 # Future work
 [future]: #future-work


rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,142 @@
...
+NixOS, and likely to create a medium-term situation of partial migration.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Do nothing, continuing with the current code duplication.
+
+Same, but put configuration generators closer to packages.
+
+Many proposed solutions based on a siggnificant rework of the module system.
```
mweinelt, line 122:
```suggestion
Many proposed solutions based on a significant rework of the module system.
```

7c6f434c: Thanks

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,142 @@
+--- feature: configuration_file_generators start-date: 2020-10-18 author:
```
lheckemann, line 1: Looks like auto-formatting broke this header?

7c6f434c: Thanks, fixed

7c6f434c:
This is based on brainstorming during NixCon mainly with @svanderburg . Sander, do you accept to co-sign this as a co-author?

(Also thanks to aanderse jtojnar edolstra 0x4A6F LnL7 for discussion)

Mic92:
Anyone who wants to shepherd this RFC?

DamienCassou:
I like this RFC, thank you!

svanderburg:
@7c6f434c sure! RFC looks great so far. I'm going to read it in detail.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+Provide the configuration file generation functionality separately from NixOS
+as a whole, and with scoped use of the module system.
+
+# Motivation
+[motivation]: #motivation
+
+There is currently a lot of code duplication between
+* NixOS
+* home-manager
+* nix-darwin
```
Mic92, line 23: Another target that would benefit from this are containers that need configuration independent from NixOS modules since they don't support systemd etc.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+[examples-and-interactions]: #examples-and-interactions
+
+A minimalistic silly example is:
+
+Input:
+```
+  {
+     port = 1234;
+     nodeName = "localhost";
+     content = "Hello";
```
Mic92, line 80: Is this typed?

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+  }
+```
+
+Output:
+```
+  {
+    "subdir/listen.conf" = {
```
Mic92, line 87: An example how this would be integrated into NixOS modules might be helpful.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+      serialiser = toJSON;
+      settings = {
+        listenPort = "1234";
+        serverGreet = "I am localhost";
+      };
+      type = …;
```
Mic92, line 93: What is this type for?

rfcs/0078-configuration-file-generators.md
7c6f434c: True; I am not sure what publically available codebase should be referenced as an example. My real use case is not even in the list and it is closer to what you say about containers; on the other hand, highly-visible existing parallel efforts are more convenient to discuss.

rfcs/0078-configuration-file-generators.md
7c6f434c: Yes, the type is provided next to the generator function.

rfcs/0078-configuration-file-generators.md
7c6f434c: Well, the output is intended for use with an RFCs#42 implementing NixOS module, to put into (in the simplest case) `services.<service>.settings`. In the simplest cases one can just have a static type for the settings entry, but I guess it doesn't hurt to repeat it here and to allow, when necessary, some variation (depending on configuration format input option, I guess)

rfcs/0078-configuration-file-generators.md
7c6f434c: Hmm, I am not sure it can be done easily without integrating an entire example from #42 and then extending it…

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+source.
+
+# Detailed design
+[design]: #detailed-design
+
+This RFC proposal builds upon RFCs#42 defining a configuration file abstraction
+approach.
+
+A subtree is established in the Nixpkgs repository serving as a library of
+functions implementing program configuration file generation. .As the design
```
blaggacao, line 51:
```suggestion
functions implementing program configuration file generation. Alongside these functions their respective input and output types are specified. As the design
```

blaggacao (changes requested):
A great motion to factor config management, define abstract interfaces and reduce the current bundling! :heart:  I hope some of my comments prompt you to do a clarifying fixup commit.

I'd also like to see a broader perspective on config management incorporated (my stated use case is just a top-of-my-head thought, nothing duly researched)

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+NixOS, and likely to create a medium-term situation of partial migration.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Do nothing, continuing with the current code duplication.
+
+Same, but put configuration generators closer to packages.
+
+Many proposed solutions based on a significant rework of the module system.
```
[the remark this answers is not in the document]
8573: @blaggacao, I'm pretty sure it means...
```suggestion
There also have been many solutions proposed based on a significant rework of the module system.
```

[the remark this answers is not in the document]
blaggacao: So it would be probably something around:

> There also have been many proposals that factor out configuration generation based on a significant rework of the module system, such as ...

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+co-authors: (in-progress)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Provide the configuration file generation functionality separately from NixOS
```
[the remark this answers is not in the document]
7c6f434c: I am not even sure it is in scope for future work! 

Sure, that's a good idea in general, but this RFC is about sharing something that is already written now, and in a format that has seen some use by now. Inventing on-start configuration for things without persistent configuration is a good direction of work, but it is somewhat different type of work than described here.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+[design]: #detailed-design
+
+This RFC proposal builds upon RFCs#42 defining a configuration file abstraction
+approach.
+
+A subtree is established in the Nixpkgs repository serving as a library of
+functions implementing program configuration file generation. .As the design
+uses a module system, putting configuration generators next to packages is also
+not a perfect solution; also a configuration generator might end up generating
+a set of config for a combination of packages.
```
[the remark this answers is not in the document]
7c6f434c: Good point.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Provide the configuration file generation functionality separately from NixOS
+as a whole, and with scoped use of the module system.
```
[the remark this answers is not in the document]
7c6f434c: Scope in the sense of visibility… reworded

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
7c6f434c: It does. It is a non-normative section, I confirm that I have not ignored this direction, and describe it in a way that emphasises that a lot of work would be needed, taking a lot of time. Describing any of the options well would probably require a full slide deck. Like that talk about modules at NixCon.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+[alternatives]: #alternatives
+
+Do nothing, continuing with the current code duplication.
+
+Same, but put configuration generators closer to packages.
+
+Many proposed solutions based on a significant rework of the module system.
+
+Abstract generation of configuration files with package-like flat arguments and
+plain text file outputs.
```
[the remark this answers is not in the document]
7c6f434c: What you say is a reasonable implementation of this abstract interface.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+
+Implement a complete service abstraction not tied to global system-wide
+assumptions.
```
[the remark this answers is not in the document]
7c6f434c: A single statement would not be enough to do much good. A good description will overtake the rest of the RFC.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+of inspection and tweaking possibilities without needing to patch module
+source.
+
+# Detailed design
+[design]: #detailed-design
+
+This RFC proposal builds upon RFCs#42 defining a configuration file abstraction
+approach.
+
+A subtree is established in the Nixpkgs repository serving as a library of
```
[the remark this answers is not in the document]
7c6f434c: I do not think `configs` is a perfect name, but maybe it is good enough, so let's see if there are other opinions. In principle, that could be left to implementation step, should only lead to a little bit of bikeshedding

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+functions implementing program configuration file generation. .As the design
+uses a module system, putting configuration generators next to packages is also
+not a perfect solution; also a configuration generator might end up generating
+a set of config for a combination of packages.
+
+Each such generator takes as an input a NixOS module system based «subtree»,
+e.g. the values typically bound to `cfg` variable in typical modules.
```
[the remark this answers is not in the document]
7c6f434c: Well, the value of the subtree? A possibly nested attribute set, yes. But not to entire `config`, just to the service.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+
+The output of each such function is an attribute set. The keys are relative
+file names of configuration files, and the values are attribute sets of RFCs#42
+settings abstraction and serialiser.
```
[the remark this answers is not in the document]
7c6f434c: «well-known» is factually wrong here for some of the more interesting configuration file formats.

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+
+Along with the function the corresponding type specifications for input and
+output modules are provided.
```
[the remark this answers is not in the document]
7c6f434c: It is still detailed-design section. We define the function, and what else is provided. In a sense, it _is_ an afterthought, or supporting scaffolding.

ehmry:
I nominate myself as a shepherd. I don't have practical experience using Nix with other inits, but I am currently working on this.

Profpatsch:
I’d be interested maybe helping out when it’s in progress. Maybe a flakes-based setup will help realizing this.

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
blaggacao: Well, maybe you want to define at least the system boundaries so it becomes clear what "system-wide" refers to?

If you can rephrase I can re-try if it's &mdash; while applying utmost care while reading &mdash; intelligible (to me).

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
blaggacao: Agree, maybe you can change it to "folder subtree" to differenciate the token from "attribute subtree" below.

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
blaggacao: So here you could use "attribute subtree" to differenciate from "folder subtree" above. That would help to not get confused.

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
blaggacao: Oh, that was my interpretation of `logSettings` question on RFC42.

I changed the suggestion and eliminated well-known. It should be still a little clearer that way.

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
blaggacao: You are right maybe "future work" is too normative. Would you like to add a "broader context" (or a better naming!) section at the end (as an after thought)?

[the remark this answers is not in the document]
blaggacao: You are right maybe "future work" is too normative. Would you like to add a "broader context" (or a better naming!) section at the end (as an after thought)?

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
blaggacao: With the removed non-terse interlude this notion is now close enough to it's object (the functions). I reformulated for conciseness sake.

"Provided along with the function can mean many things."

"specified along with the function definition" is much less open, though still not perfect.

rfcs/0078-configuration-file-generators.md
blaggacao: That deserves introduction and propper reasoning in the detailed section, then?

rfcs/0078-configuration-file-generators.md
```diff
@@ -0,0 +1,146 @@
...
+
+Currently none.
+
+# Future work
+[future]: #future-work
+
+Define generic service abstraction for non-`systemd` systems, possibly with a
+convertor from `systemd` units.
+
+Consider some  way of referring from package to related configuration file
```
blaggacao, line 145: Why? (what for) &rarr; you might want to clarify to improve intelligibility here.

rfcs/0078-configuration-file-generators.md
7c6f434c: Well, we promise to provide type specification and serialiser.

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
7c6f434c: Hmmm. Indeed an identifier conflict, thanks. Reworded a bit.

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
7c6f434c: Hm, maybe indeed worth expanding.

rfcs/0078-configuration-file-generators.md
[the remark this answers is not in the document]
7c6f434c: Somewhat reshuffled everything in the paragraph.

rfcs/0078-configuration-file-generators.md
7c6f434c: Anyone interested in this future work item _now_ is likely to be aware of the context like `passthru.tests`; those who are not familiar with that will wait anyway until some detailed work is started

rfcs/0078-configuration-file-generators.md
aaronjanse: Option generation is also useful for `override`, such as `pkgs.rofi.override = { theme = <file>; }`

7c6f434c: True. But the packages are shared just fine already anyway, so not a strong motivation. Wrapper packages can already be done easily enough. If implementation turns out to be a huge success we might discuss this on a case by case basis whether to move stuff to the configuration file generator library, but let's get there first.

rfcs/0078-configuration-file-generators.md
blaggacao: Your response is a little bit enigmatic with regard to rhe matter of the question, but let's make some assumptions at the risk of requiring further clarification:

Do you consider here the possibility of moving service configuration still closer to the packaging itself?

If that is the case, the (new) proposed module system for packaging might come into play. And (very) early coordination with this RFC might benefit future development.

7c6f434c: We'll see from feedback, but currently the RFC argues against moving configuration generators close to packages. However, we might consider developing a way of having some kind of cross-references available at evaluation time.

06kellyjac:
https://github.com/svanderburg/nix-processmgmt/ is my favourite prototype so far for abstracting services

I know the example is listed as "a minimalistic silly example" but looking at it I have no idea how the input results in the output

7c6f434c:
@06kellyjac 

agree on `nix-processmgmt`, but it looks like NixOS is headed towards more reuse of upstream units, and in an RFC aiming to factor something out of NixOS-only I did not want to go too far into things inapplicable to current direction of NixOS.

I think how the input gets converted into the output is an implementation detail? I expect whatever is described before implementation will get even the interface slightly changed once implemented; I do not think specifying internals of the function implementation is a good idea. I guess at first it should be similar to how #42 is prototyped now.

06kellyjac:
@7c6f434c 
I wouldn't really want to switch to a new way of defining services, only to then adopt a nix-processmgmt approach and need to do it all over again.

I'm unclear as to whether this RFC is focusing on just configuration files or also the definiton of services, both of which having different implementations across nixos, home-manager, nix-darwin as you've listed.
There are a lot of mentions of systemd in the RFC but I think the problem of configuration files and services having different implementations in different projects are two separate but related issues.
If the RFC's goal is "Extracting configuration file generation out of NixOS service modules in a more reusable way." then I think the RFC should avoid changing the actual systemd service much/at all and leave that to another RFC.

---

I get not wanting to define implementation details but in the output there are file names like `listen.conf` and `content.conf`. where are they defined? what determines that `listen.conf` is JSON and `content.conf` is INI? where is it defined that the input `port` then becomes `listenPort` for `listen.conf`?

With #42 I can see how things were before, a proposed structure, and a full example module.

7c6f434c:
>I wouldn't really want to switch to a new way of defining services, only to then adopt a nix-processmgmt approach and need to do it all over again.

Nah, would be out of character for NixOS to adopt init configurability any time soon. On the other hand, #42 is a good idea anyway, and this RFC is more about making the code that would be written anyway available separately from the rest of the things, os not too much of a change of how things are written, just about at what interfaces they are split.

>I'm unclear as to whether this RFC is focusing on just configuration files or also the definiton of services, both of which having different implementations across nixos, home-manager, nix-darwin as you've listed.

As summary and detailed design say, this is about configuration files, as there the difference in desired functionality for different targets is pretty small.

>There are a lot of mentions of systemd in the RFC but I think the problem of configuration files and services having different implementations in different projects are two separate but related issues.

The mentions about RFC are: 
* explanation why the scope does not include complete service generation, only configuration files (systemd unit reuse)
* future work on possible service generation for non-NixOS

>If the RFC's goal is "Extracting configuration file generation out of NixOS service modules in a more reusable way." then I think the RFC should avoid changing the actual systemd service much/at all and leave that to another RFC.

â¦ which is why the second mention is future work and not unresolved questions.

>I get not wanting to define implementation details but in the output there are file names like `listen.conf` and `content.conf`. where are they defined? what determines that `listen.conf` is JSON and `content.conf` is INI? where is it defined that the input `port` then becomes `listenPort` for `listen.conf`?

Well, the implementation needs to provide all these facts, which are generally determined by inspecting the program being configured.

>With #42 I can see how things were before, a proposed structure, and a full example module.

I am still not sure whether there will be feedback leading to a slight course correction, so I am trying to delay making any kind of implementation prototypes.

deliciouslytyped:
As one data point for usecases; I recently started wanting to be able - in nix-on-droid - to start web servers for apps from f-droid that are just frontend clients. My understanding is Termux doesn't support systemd, or something,
I imagine standard sysv init scripts will work fine since they are just shell scripts.

...and then I ended up going doing a slight bit more research:
There are differences to linux: https://wiki.termux.com/wiki/Differences_from_Linux , probably most importantly no user separation, I'm not sure if you can get fake differring UIDs. Nix-on-droid also uses the proot functionality, which maybe does fake uid0: https://wiki.termux.com/wiki/PRoot

There appears to be a bit of infra for launching service scripts:
https://old.reddit.com/r/termux/comments/cs014w/termuxservices_new_package_to_control_daemons/
https://wiki.termux.com/wiki/Termux-services
https://wiki.termux.com/wiki/Termux:Boot

A bit of a todo/note to self:
- [ ] if we ever get sufficient functionality to generate services, it would probably be nice to tell termux upstream about it / add it to their wiki

You can observe me naively running into idiosyncrasies of the proot environment in https://github.com/t184256/nix-on-droid/issues/75#issuecomment-682234708 . IIRC the main issues were debugging difficulties and x11 needing patches from termux to handle setuid/setgid exiting with failure because proot refuses to emulate them - so nothing immediately relevant to service generation I think.

blaggacao:
Potential use case: https://github.com/ttacon/glorious/issues/49 / https://github.com/numtide/devshell/issues/47

rfcs/0072-commonmark-docs.md
```diff
@@ -0,0 +1,334 @@
+---
```
edolstra, line 1: For some reason `0072-commonmark-docs.md` is part of this RFC, could you remove it?

spacekookie:
@Profpatsch would you be interested in being a shepherd on this RFC?

Generally this RFC is in need for more shepherds before being able to move onwards.

rfcs/0072-commonmark-docs.md
7c6f434c: Squashed the edits until now. Seems to have helped.

edolstra:
I nominate myself as a shepherd.

Mic92:
Are you interested in doing shepherd for this RFC, since it might be interested for your projects: @LnL7 @rycee

lheckemann:
Nominating @roberth as auther of Arion :)

roberth:
> Nominating @roberth as auther of Arion :)

I should probably mention that Arion doesn't need this,* but may benefit from it, just like other `dockerTools` users.

Before I'll shepherd anything, I'd like to contribute to the discussion.

As I understand it, this RFC seeks to propose a common interface/convention/abstraction that sits somewhere between `pkgs.formats` and a "NixOS" module with its platform-specific bits moved to separate module.
I'm not convinced that we have enough room between these to two existing concepts to justify another "abstraction level" that we'll all(?) have to maintain.

Have we tried refactoring a module into a common module and platform-specific modules? That can be done today. It's just a matter of "coding to interfaces". That is, importing modules that only declare `options`. As a real-world example, here's [hercules-ci-agent/common.nix](https://github.com/NixOS/nixpkgs/blob/4841b307841f6ac6a7f26e0c28f4173e1f979756/nixos/modules/services/continuous-integration/hercules-ci-agent/common.nix), which only depends on `nix.*` and itself. It works in both NixOS and nix-darwin. Internal options that define the interface between common and platform-specific code are marked as `internal`, so they don't appear in the documentation.

Alternatively, you can [avoid](https://github.com/NixOS/nixpkgs/pull/82802) hard `config` dependencies by checking `options ? foo.bar`. Interfaces are to be preferred, certainly for the goal of this RFC.

All the integrations listed earlier use the module system anyway, so duplicating all those options as function arguments in a separate file seems like wasted effort. And if an integration isn't based on the module system, it can still invoke the common module because it doesn't need to pull in anything it shouldn't.

A solution like this resolves both drawbacks (although obviously some effort is always required) and I don't see any new drawbacks yet. I'll contrast this solution against the related alternatives that were mentioned:

> There also have been many solutions proposed based on a significant rework of the module system.

No rework of the module system required. The nature of the work to be done is similar to the current proposal, but supported by the (existing) module system rather than outside of it.

> Implement a complete service abstraction not tied to global system-wide
assumptions.

Similar refactoring technique, but we avoid the need for such an abstraction.

If it does come to be, the transition towards such an abstraction is less demanding, because the switch simply increases the scope of the common module. The platform-specific tweaks were already separated out when the common module was introduced, so whatever isn't captured by the service abstraction can simply be kept.

*: Arion supports full systemd-based containers, running NixOS inside the container, so it doesn't have the problem. Nonetheless, init systems (entrypoints) are pluggable in Arion, so a convenient way to define more space-efficient containers/images is a great addition.

rfcs/0078-configuration-file-generators.md
```diff
@@ -123,7 +126,15 @@ widespread use of the module system inside `pkgs/`. There is also no guarantee
 that all the configuration files describing interaction of multiple software
 packages will have a clear choice of reference package.
 
-There also have been many solutions proposed based on a significant rework of the module system.
+A complete or partial merge of the module collections of the major currently
+existing module system based code bases, with internal options for platform
+specific implementation details. Unfortunately, this would force more coupling
+of the development cycle for the platform-specific parts. It also does not
+create a clean inspection/override point, and mixes the code with different
```
roberth, line 142: Just a matter of calling `lib.evalModules`, `nix repl '<nixpkgs/nixos>'`, `arion repl` or whatnot, where the latter variations have the advantage of not having to decipher the logic that produces the inputs. Using options for this is an inspectability improvement over the usual `let` bindings and such that will be used with the function approach.

rfcs/0078-configuration-file-generators.md
```diff
@@ -123,7 +126,15 @@ widespread use of the module system inside `pkgs/`. There is also no guarantee
...
+platform requirements for testing.
```
roberth, line 143: Again, the common module isn't platform specific and can stand alone.
It's really just like a function, except you get to reuse the `option` declarations for its inputs.

The two approaches are isomorphic if it wasn't for the fact that going back to a simple function loses the option metadata and checking.

inputs <-> normal options
outputs <-> internal options
multiple invocations <-> submodules
function without free variables <-> module that only defines self-declared options
function with small closure <-> module that doesn't import much

roberth (changes requested):
A module-based approach is exceedingly similar to a function-based approach. Perhaps I didn't make that clear.

rfcs/0078-configuration-file-generators.md
```diff
@@ -39,8 +39,10 @@ following properties:
 
 We aim to reduce duplication of effort and code while also having a high level
 of inspection and tweaking possibilities without needing to patch module
-source. At the same time we aim to minimise the amount of code in the NixOS
-subtree that cannot be tested on NixOS.
+source. An additional goal is to move the more package-like configuration file
+generation code closer to the Nixpkgs model with scope/visibility rules. At
+the same time we aim to avoid or minimise code in the NixOS subtree that
+cannot be tested on NixOS.
```
roberth, line 45: This is not a goal but an implementation choice.

7c6f434c:
>> Nominating @roberth as auther of Arion :)
>I should probably mention that Arion doesn't need this,* but may benefit from it, just like other `dockerTools` users.

I think this is true for most of the use cases. There is the solution of just standartising how the configuration files are handled and then using the partial evaluation of NixOS, which only has the drawbacks of less clear separation (and, also, it avoids the step in the apparently desirable direction of declaring which parts need to see which part of the config).

>Before I'll shepherd anything, I'd like to contribute to the discussion.
>
>As I understand it, this RFC seeks to propose a common interface/convention/abstraction that sits somewhere between `pkgs.formats` and a "NixOS" module with its platform-specific bits moved to separate module.
>I'm not convinced that we have enough room between these to two existing concepts to justify another "abstraction level" that we'll all(?) have to maintain.

It's not just _platform_ specific bits, it is also all the weird global-scope specific bits.

>Have we tried refactoring a module into a common module and platform-specific modules? That can be done today. It's just a matter of "coding to interfaces". That is, importing modules that only declare `options`. As a real-world example, here's [hercules-ci-agent/common.nix](https://github.com/NixOS/nixpkgs/blob/4841b307841f6ac6a7f26e0c28f4173e1f979756/nixos/modules/services/continuous-integration/hercules-ci-agent/common.nix), which only depends on `nix.*` and itself. It works in both NixOS and nix-darwin. Internal options that define the interface between common and platform-specific code are marked as `internal`, so they don't appear in the documentation.

By now I think that common use of the same modules in NixOS, nix-darwin and home-manager is something that could also happen a few years ago, and it had not, and I assume there are reasons. (Well, there are some reasons like different development cycles and different testing procedures etc.) So I look at what all the solutions _do_ share (Nixpkgs), and I hope to make more code Nixpkgs-like and thus more shareable.

One thing I like in Nixpkgs is that things actually have scoping and so most things only have access to what they need. NixOS modules work on a global level, which is good for some use cases, but I believe it is a good idea to create a path to gradually decouple the code that does not need the global scope from the lobal scope.

Separately, I want a solution that allows organic growth of reuse by the niche parts of the ecosystem, a single clean up helping many exotic platforms (in terms of Platform Tiers RFC).

>Alternatively, you can [avoid]( https://github.com/NixOS/nixpkgs/pull/82802 ) hard `config` dependencies by checking `options ? foo.bar`. Interfaces are to be preferred, certainly for the goal of this RFC.

This is not really a solution that makes handling a move of a neighbour to multi-instance modules easier, for example.

>All the integrations listed earlier use the module system anyway, so duplicating all those options as function arguments in a separate file seems like wasted effort. And if an integration isn't based on the module system, it can still invoke the common module because it doesn't need to pull in anything it shouldn't.

Module system is a thing that has many uses. The good one is that it is the best typing approach we actually have. The sometimes necessary one is that sometimes you do need to apply a global change to the config based on a policy decision. The annoying one is increasing the coupling. I definitely want the universal parts of NixOS (configuration generation) to be available to setups that only use the module system as a solution for typing, but without global module-based scope.

>A solution like this resolves both drawbacks (although obviously some effort is always required) and I don't see any new drawbacks yet. I'll contrast this solution against the related alternatives that were mentioned:

Judging from what happens with Darwin build problems on package bumps, meaningfully Darwin-specific code inside NixOS subtree sounds like a drawback. Sure Hercules-CI has people who use macOS laptops and NixOS servers at once among the core developers, but once the macOS and NixOS user bases are more disjoint, it is a smaller step to have noarch part separate and meaningful platform code separate per platform.

It is true that alternative solutions shoule be expanded (pushed a change).

>> Implement a complete service abstraction not tied to global system-wide
>assumptions.
>
>Similar refactoring technique, but we avoid the need for such an abstraction.

Well, we keep many of the global-system properties of a module system in that case.

>If it does come to be, the transition towards such an abstraction is less demanding, because the switch simply increases the scope of the common module. The platform-specific tweaks were already separated out when the common module was introduced, so whatever isn't captured by the service abstraction can simply be kept.

7c6f434c:
>> with internal options for platform
>￼specific implementation details
>
>This misses the point entirely. The dependencies are inverted, with the internal options serving the role of the function output, so the internal options are platform agnostic.

Then I failed to understand the flow completely…

>+specific implementation details. Unfortunately, this would force more coupling
>+of the development cycle for the platform-specific parts. It also does not
>
>So does the function approach.

Not as much, because non-trivial platform-specific environment setup stuff is what gets most of the changes, and hopefully with #42 (which this RFC needs anyway) configuration files per se get to stay closer to the underlying packages in the sense of forcing changes.

>+of the development cycle for the platform-specific parts. It also does not
>+create a clean inspection/override point, and mixes the code with different
>
>Just a matter of calling `lib.evalModules`, `nix repl '<nixpkgs/nixos>'`, `arion repl` or whatnot, where the latter variations have the advantage of not having to decipher the logic that produces the inputs. Using options for this is an inspectability improvement over the usual `let` bindings and such that will be used with the function approach.

There are things that I want to insepect and tweak programmatically. Introducing a functional interface at that point kind of forces that they exist somewhere. Also, then this cut will probably stay in place. Banning `let` completely in favour of internal options would also solve the problem, but with less chances to introduce stable-ish cut points, and my estimation that such a change has a smaller chance of being adopted.

>> @@ -123,7 +126,15 @@ widespread use of the module system inside `pkgs/`. There is also no guarantee
> that all the configuration files describing interaction of multiple software
> packages will have a clear choice of reference package.
> 
>-There also have been many solutions proposed based on a significant rework of the module system.
>+A complete or partial merge of the module collections of the major currently
>+existing module system based code bases, with internal options for platform
>+specific implementation details. Unfortunately, this would force more coupling
>+of the development cycle for the platform-specific parts. It also does not
>+create a clean inspection/override point, and mixes the code with different
>+platform requirements for testing.
>
>Again, the common module isn't platform specific and can stand alone.
>It's really just like a function, except you get to reuse the `option` declarations for its inputs.
>
>The two approaches are isomorphic if it wasn't for the fact that going back to a simple function loses the option metadata and checking.

The function _is_ already specified to come with the option metadata, though, that can be then imported into the options specification.

It is forced not to care where in the namespace it will get imported, though.

>inputs <-> normal options
>outputs <-> internal options
>multiple invocations <-> submodules
>function without free variables <-> module that only defines self-declared options
>function with small closure <-> module that doesn't import much

7c6f434c:
>+generation code closer to the Nixpkgs model with scope/visibility rules. At
>+the same time we aim to avoid or minimise code in the NixOS subtree that
>+cannot be tested on NixOS.
>
>This is not a goal but an implementation choice.

Judging from observations of Nixpkgs workflows, it is a useful goal to have…

blaggacao:
[Reading & asking](https://discourse.nixos.org/t/difference-between-nix-profiles-and-home-manager/9539/5), I wonder if on a high level I'm correctly connecting the dots:

Is there a treasure trove in configuration know-how to be lifted from home-manager out of its silo by mediation if the principles proposed in this RFC?

bqv:
Remember that home-manager is implemented entirely in nix. Nothing formally prevents any part of it from being absorbed into nixpkgs besides politics and sentiment.

7c6f434c:
I believe the previous discussion to merge home-manager into nixpkgs was without any deduplication plans, so all of the coordination overhead, none of the technical benefits. If the configuration database behaves closer to package database and doesn't care what's the structure above it — well, home-manager seems to work fine with upstream Nixpkgs.

7c6f434c:
@roberth hmmm, actually, do you know of any support for module import as a subtree, i.e. importing a standalone module with options like «enable» and «port» to be at path «services.newssh», i.e. with options «services.newssh.enable» and «services.newssh.port»?

I agree that it would be cleaner than a function with a module signature for type checking on the side.

roberth:
@7c6f434c 

> @roberth hmmm, actually, do you know of any support for module import as a subtree, i.e. importing a standalone module with options like «enable» and «port» to be at path «services.newssh», i.e. with options «services.newssh.enable» and «services.newssh.port»?

Yes.

```nix
let

  inherit (import <nixpkgs/lib>) evalModules types mkOption optionalString;

  pureModule = {
    options = {
      port = mkOption { default = 22; /* etcetera */ };
    };
  };

  liftedModule = {
    options.services.ssh = mkOption {
      type = types.submodule pureModule;
    };
    config.services.ssh = {};
  };

  nixosBase = {
    # Let's keep it simple and pretend :)
    options.system.build.activationScript = mkOption { type = types.lines; };
  };

  nixosSshModule = { config, ... }: {
    options.services.ssh = mkOption {
      type = types.submodule {
        options.socketActivated = mkOption { default = true; /* etcetera */ };
      };
    };
    config.system.build.activationScript = ''
      Start openssh on port ${toString config.services.ssh.port} ${optionalString config.services.ssh.socketActivated "via socket"}
    '';
  };

  userConfiguration = {
    services.ssh.port = 2222;
  };

in
  evalModules { modules = [liftedModule nixosBase nixosSshModule userConfiguration]; }
```

```
$ nix repl subtree.nix # file above
Welcome to Nix version 2.4pre20201201_5a6ddb3. Type :? for help.

Loading 'subtree.nix'...
Added 3 variables.

nix-repl> config.system.build.activationScript
"Start openssh on port 2222 via socket\n"

nix-repl> 
```

We might benefit frome syntactic sugar for `liftedModule`, so that `nixosSshModule` can import it all by itself, but let's not jump to conclusions until we know how the usage of submodule merging develops.

roberth:
> There are things that I want to insepect and tweak programmatically. Introducing a functional interface at that point kind of forces that they exist somewhere. Also, then this cut will probably stay in place.

I've updated the example in my previous comment to include the pure function idea. `pureModule` can be called independently using `evalModules`.

In this example I had to modify the file, but normally those modules would be separate files, so you could just reference the module you need.

```nix
# let ... in
  evalModules { modules = [ pureModule ]; }
```

```
$ nix repl subtree.nix 
Welcome to Nix version 2.4pre20201201_5a6ddb3. Type :? for help.

Loading 'subtree.nix'...
Added 3 variables.

nix-repl> config.configText
"Port 22\n"
```

> Banning `let` completely in favour of internal options would also solve the problem, but with less chances to introduce stable-ish cut points, and my estimation that such a change has a smaller chance of being adopted.

I agree. Banning `let` entirely doesn't make sense. I should have clarified which let binding: the one with the configuration text or config file derivation. (OT: text is preferable when dealing with secrets)

> [If?] the function _is_ already specified to come with the option metadata, though, that can be then imported into the options specification. It is forced not to care where in the namespace it will get imported, though.

Exactly. If the module file sits next to the Nixpkgs package definition, I think expectations are managed pretty well. We can also have a unit test for the config file generation, which will enforce the purity aka "not caring where in the namespace it will get imported".
The unit test can be added to `passthru.tests`, but I expect that if you try to pass the settings module via `passthru`, you run into a strictness issue. We can still reference the file by path though.

7c6f434c:
Interesting, thanks a lot! Indeed that lifting sounds nice and with a minimal test could guarantee the same pure-function properties with smoother integration than separate module definitions alongside a function.

I agree that settins / configText / configFile should be visible/overridable simultaneously (in the submodule approach it makes more sense to specify that).

liftedModule does sound like something I kind of hope to have autogenerated inside the system-level module… Not sure if it is worth trying to add liftingImports into the modules system or there is a better way.

spacekookie:
@ehmry @edolstra @roberth Who of you would want to be the shepherd leader?

ehmry:
I've not acted as sheperd before, so I would not be first choice.

Mic92:
@edolstra said, he can do the shepherd leader for this RFC.

pasqui23:
A quick comparison of the options offered by NixOS,@LnL7 's nix-darwin and @rycee 's home-manager:
* `documentation.*` appears both on NixOS and nix-darwin, but not on home-manager
* Same for `environment.etc`, naturally.The closest equivalent on home-manager is `home.file` but is not a drop-in replacement(for starter its name often start with a dot)
* `environment.systemPackages` is another shared by both Nixos and nix-darwin and it has a drop-in replacement in home.packages, in fact many options that Nixos and nix-darwin define in `environment` have some equivalent in `home` on home-manager, often with a different name.
* home-manager has `fonts.fontconfig.enable`, while nix-darwin `fonts.enablefontdir` and `fonts.fonts`
* `nix.*` is shared by Nixos,nix-darwin and home-manager, while `nixpkgs.*` is shared by all 3
* `programs.*` and `services.*` often have the same structure for the same program on all 3 of them.
* `users.users` and `users.group` are shared between nixos and nix-darwin
* *Annoyingly `systemd.*` has a different structure on nixos and home-manager, and in a way that's difficult to upgrade
* We should also get a look implementing `xdg.*Files`, there are syste wide defaults for them https://wiki.archlinux.org/index.php/XDG_Base_Directory#Specification
What I would suggest for improving the sharing between those projects:
* Create in nixpkgs a `cfg/api` directory, containing module definitions for most of the options in `environment.*`, without their implementation
* For every program with a NixOS module that can be portable, split their more generic implementation (both the `options.*` and `config.*` sections) in a new `cfg/<appname>` directory while keeping their nixos-specific stuff under `nixos/modules`. As a bonus we can split the systemd stuff for each program in a different file under that directory

edolstra:
@7c6f434c Have you done any work on this lately? Do you think it would be useful to have a call with the shepherds sometime soon? (Also pinging @ehmry and @roberth.)

7c6f434c:
>@7c6f434c Have you done any work on this lately? Do you think it would be useful to have a call with the shepherds sometime soon? (Also pinging @ehmry and @roberth.)

I haven't done anything on this lately, sorry.

I am convinced by the arguments that importing a stand-alone module with renaming it into a subtree is better than what I have originally thought. I have not yet looked into the question how much the module system needs to be changed to make this scenario a first-class usecase (and in particular whether such changes would be too intrusive to be welcome — but hopefully not, as intermediate-module workaround seems to be very straightforward). If anyone who is up to date with what is going with the module system right now is willing to help me and others get up to date, I guess that could be a productive conversation (be it a call, a chat or a long-form conversation in the comments here)

Mic92:
@7c6f434c can you organize a meeting for example on https://www.when2meet.com?

7c6f434c:
Maybe it is indeed a good idea to review what we know now

https://www.when2meet.com/?11458739-kF6rB

@ehmry @edolstra @roberth @svanderburg and anyone who wants to bring some module system expertise to make the integration as smooth as possible

Mic92:
@7c6f434c mhm. Looks like the shepherds are not following this issue close enough. Would it be maybe more efficient to contact them via email?

7c6f434c:
***@***.*** mhm. Looks like the shepherds are not following this issue close enough. Would it be maybe more efficient to contact them via email?

Well, I do not actually have anything new to say at this time anyway… so dunno

7c6f434c:
The module system itself, as a language level construct, doesn't care about Linux, systemd, or anything. It is about managing dependent computations and mergeable assignments, or something like that. nix-darwin seems to use it, so it has to be system-agnostic! (It has no way to be system-dependent anyway)

What could be added at the module system level is a nicer way of importing a module as a submodule at as specified deep position. So NixOS/RFCs#42 like configuration file generator could be internally defined as a top-level thing, but imported into services.serviceName subtree. Than this RFC would encourage converting configuration file handling into this approach.

Presumably, systemd NixOS hopes to eventually mostly reuse the upstream unit files. I guess if import-as-submodule for configuration file is a success, then services that have anything else interesting, could be implemented the same way to enable cheaper multi-instance support without a ton of special support. In a sense, like Nix supports installing multiple versions of the same package built with different overrides without a need for separate per-package support.

nbp:
> What could be added at the module system level is a nicer way of importing a module as a submodule at as specified deep position. So [NixOS/rfcs#42](https://github.com/NixOS/rfcs/pull/42) like configuration file generator could be internally defined as a top-level thing, but imported into services.serviceName subtree. Than this RFC would encourage converting configuration file handling into this approach.

I tried this approach in the past for implementing web applications (php, wikimedia, …).

Converting namespace to submodules adds extra complexity, as submodules do not have the ability to reference the parent modules, and parent modules has to deal with merging the content provided by the submodules, which adds a lot of complexity.

One way to deal with this issue would be to add a new concept of aliases which are working across module boundaries. However this is not always desirable as this breaks the laziness at the submodule boundary. This can be a problem, as if any submodule is expected to set a parent option, then all submodule have to be evaluated to check if they do or do not set the parent module option. Aliases across module would have to be crafted carefully to avoid this issue.

7c6f434c:
>Converting namespace to submodules adds extra complexity, as submodules do not have the ability to reference the parent modules, and parent modules has to deal with merging the content provided by the submodules, which adds a lot of complexity.

Well, at least for config file generation isolation / portability is a useful property (and the intended use is just taking the configuration file obtained from the service configuration).

roberth:
> Converting namespace to submodules adds extra complexity, as submodules do not have the ability to reference the parent modules

By default they don't, but the parent module can merge its own definitions into the submodule.

> and parent modules has to deal with merging the content provided by the submodules, which adds a lot of complexity.

This complexity seems to be the complexity that is inherent to premise of the RFC; splitting config file generation from platform specifics.

> One way to deal with this issue would be to add a new concept of aliases which are working across module boundaries. However this is not always desirable as this breaks the laziness at the submodule boundary. This can be a problem, as if any submodule is expected to set a parent option, then all submodule have to be evaluated to check if they do or do not set the parent module option. Aliases across module would have to be crafted carefully to avoid this issue.

This alias feature seems unnecessary and it would make modules harder to understand. It's probably a good thing for modules to state explicitly what they are setting. I that light, I also doubt whether adding syntax sugar for importing in a submodule is a good thing or not.

I think I've overcomplicated my earlier example a bit. The "import at namespace" feature was intended to replace `liftedModule`, but the `imports` line below should already do the trick, if I'm not mistaken.

```nix
  nixosSshModule = { config, ... }: {
    options.services.ssh = mkOption {
      type = types.submodule {
        imports = [ pureModule ];  # I guess we don't need liftedModule if we have this line?
        options.socketActivated = mkOption { default = true; /* etcetera */ };
      };
    };
```

7c6f434c:
>I think I've overcomplicated my earlier example a bit. The "import at namespace" feature was intended to replace `liftedModule`, but the `imports` line below should already do the trick, if I'm not mistaken.
>
>```nix
>  nixosSshModule = { config, ... }: {
>    options.services.ssh = mkOption {
>      type = types.submodule {
>        imports = [ pureModule ];  # I guess we don't need liftedModule if we have this line?
>        options.socketActivated = mkOption { default = true; /* etcetera */ };
>      };
>    };
>```

Does that mean that the option set needs to be defined twice, once in the submodule and once in the parent module?

roberth:
> Does that mean that the option set needs to be defined twice, once in the submodule and once in the parent module?

Most options are declared in the pure module. Definitions can come from either the user configuration, platform module or pure module.
When the platform module imports the pure module in its submodule, it reuses those options.

7c6f434c:
>Most options are declared in the pure module. 

What still needs to be declared in the platform module then?

roberth:
> Most options are declared in the pure module.
> What still needs to be declared in the platform module then?

Anything that isn't in scope for the pure module, like platform-specific options or perhaps some cross-service integration options. We'll know after a couple of refactored modules.

7c6f434c:
>Anything that isn't in scope for the pure module

Ah right, so there is indeed zero code duplication. I like that solution, indeed.

nbp:
> > and parent modules has to deal with merging the content provided by the submodules, which adds a lot of complexity.
> 
> This complexity seems to be the complexity that is inherent to premise of the RFC; splitting config file generation from platform specifics.

You are right, I guess the complexity came from the fact that I was always attempting that on collections of submodules.

edolstra:
Sorry @7c6f434c, I must have missed your attempt to set up a shepherd meeting. Should we try again? (CC @ehmry @edolstra @roberth @svanderburg)

7c6f434c:
Not only you, to be fair… (I am also still in doubt whether there is something to discuss that benefits from synchronicity)

edolstra:
@7c6f434c Well, a meeting can be a good way to get the discussion unstuck and decide on steps forward. So I think it would be a good idea.

7c6f434c:
***@***.*** Well, a meeting can be a good way to get the discussion unstuck and decide on steps forward. So I think it would be a good idea.

Well, it is not that discussion is stuck as a discussion, it is that I am mostly convinced that some code should be tried out to see how it would look like / work like with a real module, but then I need to experiment a bit around the parts of the module system I never touched before to gain the confidence in what the hell I would be doing. Which I haven't got around to yet,

I guess I could quickly _spec_ stuff based on these import-as-subtree without checking if it works, but that sounds wrong.

Should this be converted to draft then? Closed with a plan to reopen once a import-as-subtree implementation has been tried by someone?

edolstra:
@7c6f434c Okay, thanks. I've marked the RFC as "on hold". It can be moved back to under discussion at any time, of course, when you're ready.

cdmistman:
👋 it's nearly been 2 years - anybody willing to pick this RFC back up? I'm relatively new to nix but RFC 42 seems to be the "default" now (at least, in home-manager and nix-darwin), and I'd love to see something like this happen!

ShamrockLee:
Maybe we could start by providing information about the system (e.g. `moduleSystemName`) or the properties of the system (e.g. `isPrivileged`) through `config`. That would make it possible to share modules between NixOS and Home Manager, as lot of HM modules are basically NixOS modules without privileged configuration.

kevincox:
We don't use the `On Hold` label anymore, marking as draft.

PR #50, opened 2019-08-17
[RFC 0050] Merge bot for maintainers

FRidh:
An RFC for having a bot that allows maintainers to merge their changes to Nixpkgs. I think this can help a lot to reduce the load on nixpkgs-committers by letting maintainers of leaf packages merge their own changes, however, there is also a big security risk. 

I think the latter outweighs the former. Even so, I hope this RFC can help us further in our quest on how to scale Nixpkgs either by directly offering a way to reduce load on the nixpkgs-committers, or by agreeing to not pursue a certain option.

PR #79, opened 2020-10-20
[RFC 0079] No more direct pushes to master and release branches

dasJ:
[Rendered](https://github.com/dasJ/rfcs/blob/enforce-prs/rfcs/0079-enforce-prs.md)

cc @matthiasbeyer @lschuermann

### DIFF
--- a/rfcs/0079-enforce-prs.md
+++ b/rfcs/0079-enforce-prs.md
@@ -18,9 +18,9 @@ This implements the four-eyes principle and allows easier change discussion, bot
 # Motivation
 [motivation]: #motivation
 
-In its current state every [Nixpkgs committer](https://github.com/orgs/NixOS/teams/nixpkgs-committers/members) is able to push arbitrary code to every branch, including master and all release branches.
+In its current state every [Nixpkgs committer](https://github.com/orgs/NixOS/teams/nixpkgs-committers/members) (139 people as of writing) is able to push arbitrary code to every branch, including master and all release branches.
 
-Pushing to critical branches branches of a software development project is not a good practice in more-than-one-contributor environments.
+Pushing to critical branches of a software development project is not a good practice in more-than-one-contributor environments.
 Security vulnerabilities, breaking changes which are not tested, and regressions are easily introduced this way.
 Those issues can been prevented by a proper workflow and tooling that comes with the workflow.
 
@@ -40,11 +40,9 @@ By changing our branching workflow to a no-push-to-master workflow, we can achie
 # Detailed design
 [design]: #detailed-design
 
-In GitHubs [branch protection](https://github.com/NixOS/nixpkgs/settings/branch_protection_rules) rules, branch protection rules which require pull request reviews, include administrators, forbid force pushes and branch deletions must be created.
+In GitHub's [branch protection](https://github.com/NixOS/nixpkgs/settings/branch_protection_rules) rules, branch protection rules which require pull request reviews, include administrators, forbid force pushes and branch deletions must be created.
 There must be rules for:
 - master
-- nixos-*
-- nixpkgs-*
 - release-*
 
 # Examples and Interactions
@@ -58,13 +56,13 @@ When pushing to a protected branch directly, they get the same message as everyo
 
 It might break the workflow of some committers which are only a small portion of the community.
 
-Also, Pull Requests might take a bit of time before they are approved by somebody else, which should't matter too much since the trust in committers is already very high and their Pull Requests are likely to be merged fast.
+Also, Pull Requests might take a bit of time before they are approved by somebody else, which shouldn't matter too much since the trust in committers is already very high and their Pull Requests are likely to be merged fast.
 
 # Alternatives
 [alternatives]: #alternatives
 
 Do nothing.
-This has the downsides mentioned in the motivation and weakens the percieved trust in the project.
+This has the downsides mentioned in the motivation and weakens the perceived trust in the project.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/rfcs/0079-enforce-prs.md
+++ b/rfcs/0079-enforce-prs.md
@@ -1,0 +1,80 @@
+---
+feature: No more direct pushes to master and release branches
+start-date: 2020-09-25
+author: Janne Heß <janne@hess.ooo>
+co-authors: Matthias Beyer <mail@beyermatthias.de>
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: None
+---
+
+# Summary
+[summary]: #summary
+
+Enforce usage of Pull Requests for **all** contributions to nixpkgs master and release branches, which implies that these branches do not allow direct pushes anymore.
+
+This implements the four-eyes principle and allows easier change discussion, both before and after the merge, and improves overall security since each Pull Request has to be approved by at least one other person prior to merging.
+
+# Motivation
+[motivation]: #motivation
+
+In its current state every [Nixpkgs committer](https://github.com/orgs/NixOS/teams/nixpkgs-committers/members) is able to push arbitrary code to every branch, including master and all release branches.
+
+Pushing to critical branches branches of a software development project is not a good practice in more-than-one-contributor environments.
+Security vulnerabilities, breaking changes which are not tested, and regressions are easily introduced this way.
+Those issues can been prevented by a proper workflow and tooling that comes with the workflow.
+
+In the last year (as of 2020-09-25), we had
+- 19471 commits on the master branch in nixpkgs
+- of which 10347 (53.14%) were merges
+- of which 9124 (46.85%) were non-merge commits
+
+<small>
+(`git log --oneline --since="1 year ago" --first-parent [--[no-]merges] | wc -l`)
+</small>
+
+This is an improvement over the previous statistic from 2019-04-13 (where 51.85% commits went directly to master) and proves the point that committers are able to work properly without pushing directly.
+
+By changing our branching workflow to a no-push-to-master workflow, we can achieve more security, stability and even more important: better scalability.
+
+# Detailed design
+[design]: #detailed-design
+
+In GitHubs [branch protection](https://github.com/NixOS/nixpkgs/settings/branch_protection_rules) rules, branch protection rules which require pull request reviews, include administrators, forbid force pushes and branch deletions must be created.
+There must be rules for:
+- master
+- nixos-*
+- nixpkgs-*
+- release-*
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This only affects nixpkgs committers.
+When pushing to a protected branch directly, they get the same message as everyone else and they have to push to an unprotected branch and create a Pull Request from there.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It might break the workflow of some committers which are only a small portion of the community.
+
+Also, Pull Requests might take a bit of time before they are approved by somebody else, which should't matter too much since the trust in committers is already very high and their Pull Requests are likely to be merged fast.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Do nothing.
+This has the downsides mentioned in the motivation and weakens the percieved trust in the project.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Whose workflows will break?
+Why are they not adoptable to a Pull Request workflow?
+
+# Future work
+[future]: #future-work
+
+Handle complaints from people who are used to direct pushing.
+
+Maybe increase the number of required approvals in the future. This however requires more active committers or Pull Requests are stalled for long times.


rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+[summary]: #summary
+
+Enforce usage of Pull Requests for **all** contributions to nixpkgs master and release branches, which implies that these branches do not allow direct pushes anymore.
+
+This implements the four-eyes principle and allows easier change discussion, both before and after the merge, and improves overall security since each Pull Request has to be approved by at least one other person prior to merging.
+
+# Motivation
+[motivation]: #motivation
+
+In its current state every [Nixpkgs committer](https://github.com/orgs/NixOS/teams/nixpkgs-committers/members) is able to push arbitrary code to every branch, including master and all release branches.
```
mweinelt, line 21: Sadly that link doesn't mean much to outsiders, as it's not publicly viewable.

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+- of which 10347 (53.14%) were merges
+- of which 9124 (46.85%) were non-merge commits
+
+<small>
+(`git log --oneline --since="1 year ago" --first-parent [--[no-]merges] | wc -l`)
+</small>
+
+This is an improvement over the previous statistic from 2019-04-13 (where 51.85% commits went directly to master) and proves the point that committers are able to work properly without pushing directly.
+
+By changing our branching workflow to a no-push-to-master workflow, we can achieve more security, stability and even more important: better scalability.
```
mweinelt, line 38: How does this improve scalability (of what?) when this obviously introduces more steps?

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+
+It might break the workflow of some committers which are only a small portion of the community.
+
+Also, Pull Requests might take a bit of time before they are approved by somebody else, which should't matter too much since the trust in committers is already very high and their Pull Requests are likely to be merged fast.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Do nothing.
+This has the downsides mentioned in the motivation and weakens the percieved trust in the project.
```
mweinelt, line 67:
```suggestion
This has the downsides mentioned in the motivation and weakens the perceived trust in the project.
```

samueldr:
Quickly noting (before reading) that this has been discussed lightly in previous instances on IRC and it would cause considerable friction in the workflow of contributors, including contributors that use the GitHub platform.

One issue (IIRC) was with signing merge commits.

Another was from automated updates like @NeQuissimus [does for Linux](https://github.com/NixOS/nixpkgs/commits/master/pkgs/os-specific/linux/kernel). It would require significant changes in the workflow, and probably cause updates to not be merged as promptly as they are now.

All that is also not considering those users that (rightly so) want to engage with GitHub as little as possible.

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+</small>
+
+This is an improvement over the previous statistic from 2019-04-13 (where 51.85% commits went directly to master) and proves the point that committers are able to work properly without pushing directly.
+
+By changing our branching workflow to a no-push-to-master workflow, we can achieve more security, stability and even more important: better scalability.
+
+# Detailed design
+[design]: #detailed-design
+
+In GitHubs [branch protection](https://github.com/NixOS/nixpkgs/settings/branch_protection_rules) rules, branch protection rules which require pull request reviews, include administrators, forbid force pushes and branch deletions must be created.
```
ryantm, line 43:
```suggestion
In GitHub's [branch protection](https://github.com/NixOS/nixpkgs/settings/branch_protection_rules) rules, branch protection rules which require pull request reviews, include administrators, forbid force pushes and branch deletions must be created.
```

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+
+This only affects nixpkgs committers.
+When pushing to a protected branch directly, they get the same message as everyone else and they have to push to an unprotected branch and create a Pull Request from there.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It might break the workflow of some committers which are only a small portion of the community.
+
+Also, Pull Requests might take a bit of time before they are approved by somebody else, which should't matter too much since the trust in committers is already very high and their Pull Requests are likely to be merged fast.
```
ryantm, line 61:
```suggestion
Also, Pull Requests might take a bit of time before they are approved by somebody else, which shouldn't matter too much since the trust in committers is already very high and their Pull Requests are likely to be merged fast.
```

rfcs/0079-enforce-prs.md
matthiasbeyer: It gives us the ability to have an evergreen master, which in turn makes the whole "Well, the broken CI is unrelated and I think we can merge!"-thing go away, because CI suddenly _means_ something. Green means good, red means a maintainer does not even have to care!

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+# Detailed design
+[design]: #detailed-design
+
+In GitHubs [branch protection](https://github.com/NixOS/nixpkgs/settings/branch_protection_rules) rules, branch protection rules which require pull request reviews, include administrators, forbid force pushes and branch deletions must be created.
+There must be rules for:
+- master
+- nixos-*
+- nixpkgs-*
+- release-*
+
```
ryantm, line 49: There should at least be some mention of what happens with other critical branches like "staging" and "staging-next". I don't know how these branches are updated currently, but I'd expect that workflows surrounding them will have to change.

ryantm (changes requested):

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: None
+---
+
+# Summary
+[summary]: #summary
+
+Enforce usage of Pull Requests for **all** contributions to nixpkgs master and release branches, which implies that these branches do not allow direct pushes anymore.
+
+This implements the four-eyes principle and allows easier change discussion, both before and after the merge, and improves overall security since each Pull Request has to be approved by at least one other person prior to merging.
```
FRidh, line 16: > Pull Request has to be approved by at least one other person prior to merging.

This was discussed on the merge bot RFC as well. The group of people with push rights is too small. There we proposed introducing a new group, trusted reviewers, that could approve and then merge via the bot.

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+By changing our branching workflow to a no-push-to-master workflow, we can achieve more security, stability and even more important: better scalability.
+
+# Detailed design
+[design]: #detailed-design
+
+In GitHubs [branch protection](https://github.com/NixOS/nixpkgs/settings/branch_protection_rules) rules, branch protection rules which require pull request reviews, include administrators, forbid force pushes and branch deletions must be created.
+There must be rules for:
+- master
+- nixos-*
+- nixpkgs-*
```
samueldr, line 47: `nixos-*` and `nixpkgs-*` (channel mirrors) are already protected by rules.

rfcs/0079-enforce-prs.md
cole-h, line 43: Do required reviews include those of us that are only maintainers and not committers? If not, I don't think requiring reviews is a good idea -- there is already a shortage of people merging Nixpkgs PRs; requiring them to approve before merging means they should spend the time testing it themselves and cannot just rely on e.g. the package maintainer's approval.

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+In the last year (as of 2020-09-25), we had
+- 19471 commits on the master branch in nixpkgs
+- of which 10347 (53.14%) were merges
+- of which 9124 (46.85%) were non-merge commits
+
+<small>
+(`git log --oneline --since="1 year ago" --first-parent [--[no-]merges] | wc -l`)
+</small>
+
+This is an improvement over the previous statistic from 2019-04-13 (where 51.85% commits went directly to master) and proves the point that committers are able to work properly without pushing directly.
```
FRidh, line 36: I typically don't merge PR's but rebase the commits onto the target branch to avoid the noise of merge commits. I don't think these metrics can be used to distinguish between rebase through GitHub versus direct push.

rfcs/0079-enforce-prs.md
FRidh: The times that something is broken due to a push is very small. Not to say we should not avoid it, but with the level of direct pushes we have nowadays it hardly every happens and tends to be corrected soon. Thus, at this point I disagree with it improving scalability.

rfcs/0079-enforce-prs.md
matthiasbeyer: No. It is not "In Github is branch protection rules", but "In Githubs branch protection rules"!

cole-h: That `'s` is actually possessive, not a contraction.

FRidh (changes requested):
I think it is a good idea to protect the branches against direct pushes. There is however a use case which is important to me. I often merge the staging(-next) branches into master and vice-versa. Depending on how involved I am this happens daily. Pushing this to GitHub, and then waiting for CI just takes too much time. Thus, unless we let a bot do the bulk of these merges, this should remain possible manually otherwise it simply won't happen.

Requiring approval of Pull Requests is orthogonal to the title of this RFC and should in my opinion be considered in a separate RFC, not here.

matthiasbeyer:
> Requiring approval of Pull Requests is orthogonal to the title of this RFC and should in my opinion be considered in a separate RFC, not here.

I see your point and agree.

cole-h:
I wish GitHub had more fine-tuned permissions for this, because I think direct-pushes to release branches should be allowed -- but only by release managers.

rfcs/0079-enforce-prs.md
lheckemann: @matthiasbeyer it only works like that for "it", unfortunately…

samueldr: For the possessive, IIRC a trick is to think it as if it was `'s own` for `'s` when it's not for "it".

> "In GitHub's own branch protection rules..."

jtojnar:
Personally, I do not care about stable releases but I cherry-pick fixes to stable branches when reasonably certain it will not cause issues and someone asks me for it. If this RFC required me to open a PR, wait for CI to pass, and only then merge it, I would never backport anything. The same would happen for other small things like typo fixes, I just could not be bothered. Or the master→staging-next→staging merges Frederik mentions. From the previous discussion of this policy, I am not the only one who considers this too much overhead to be worth it for small changes.

Really, we need an automerge bot before this can be considered.

FRidh:
@jtojnar I completely agree with you. I entirely forgot about the backporting case.

samueldr:
@jtojnar I knew I was missing one important part of those previous discussions in my initial comment. And you're right, for backports multiple contributors said outright they just wouldn't bother.

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,80 @@
...
+Enforce usage of Pull Requests for **all** contributions to nixpkgs master and release branches, which implies that these branches do not allow direct pushes anymore.
+
+This implements the four-eyes principle and allows easier change discussion, both before and after the merge, and improves overall security since each Pull Request has to be approved by at least one other person prior to merging.
+
+# Motivation
+[motivation]: #motivation
+
+In its current state every [Nixpkgs committer](https://github.com/orgs/NixOS/teams/nixpkgs-committers/members) is able to push arbitrary code to every branch, including master and all release branches.
+
+Pushing to critical branches branches of a software development project is not a good practice in more-than-one-contributor environments.
```
andir, line 23:
```suggestion
Pushing to critical branches of a software development project is not a good practice in more-than-one-contributor environments.
```

rfcs/0079-enforce-prs.md
andir: I don't think requiring 4-eye-review for stable branches will in fact improve the situation there at all. From subjective observation there appear to be only at most a hand full of people caring for a released version of NixOS besides some random back ports and high visibility issues. During the 19.09 & 20.03 release cycles there were multiple cases were PRs to fix security related stuff had been open for days. If I wouldn't be able to merge my own PRs (even without CI as they might be rebuilding everything due to touching OpenSSL) I'd probably not have done that work (and they might still be open until today).

We can probably discuss how useful protecting the branch against (fast-forward) pushes is but the review requirement should probably get a dedicated RFC (and discussion).

andir:
I pretty much agree with what @samueldr said above. I unfortunately think that no kind of GitHub provided configurable policy is really helping us here. It will only come with downsides that we can't (easily) mitigate otherwise.

Our primary lack of resource is human attention and by requiring more of that we might end up with less people willing to put up with the work. While I agree that every commit should go through CI and a PR I can also see cases where that, and the review requirement, can be counter productive. That being said I think we should focus on the branch protection in this PR and not the review requirements. Both of these should have their own discussion.

NeQuissimus:
A few things off the top of my head:

Have we considered something like Mergify or a [GitHub action](https://github.com/marketplace/actions/merge-pull-requests) (or teach @ofborg about this)?

I would imagine the workflow similar to this:
- Submit PR
- Tests pass OR maintainer/GitHub code owner approved PR
- PR merged

The real issue (apart from the special cases with `staging` and friends) is when a sole maintainer for a package submits a PR. Who is meant to approve?
This is why I would suggest having a mechanism that helps with merging. If the package defines NixOS tests, those must pass and can replace approval from another human.

I'll make it specific in my case: The kernel updates
Generally, I run a script, it checks for updates and pushed those up. The commits are cherry-picked onto the last two release branches as long as they are applicable.
If I were to open a PR for each of the commits, that would be a significant amount of work for other maintainers to review. And the cherry-picks to the release branches would be additional PRs? Or can cherry-picks be done without the PR workflow?
There definitely needs to be an "efficiency escape hatch" of some sort. Otherwise the velocity of nixpkgs updates will plummet.

Maybe a test-and-automatic-merge workflow would also lead to more/better tests, something we can definitely use :)

8573:
@FRidh wrote—
> Thus, unless we let a bot do the bulk of these merges, this should remain possible manually otherwise it simply won't happen.

@jtojnar wrote—
> Really, we need an automerge bot before this can be considered.

When I saw the title of this RFC, I was expecting it to propose the use of such a bot as [Rust](https://github.com/rust-lang/rust)'s Bors (which remains the state of the art as far as I'm aware), but, as far as I see, the nearest it comes to suggesting such a thing is a single, unelaborated mention of "tooling" in a non-normative section, namely—
> Those issues can been prevented by a proper workflow and tooling that comes with the workflow.

I was disappointed to see that this RFC does not discuss "prior art", i.e., at least one other large repository that (successfully) operates with this requirement, such as Rust.

I suggest that this RFC would do better to support itself with a discussion of "prior art" and to propose, normatively, the use of tooling such as an "automerge bot".

blaggacao:
Further discussion/input about merge bots can be found here:

- https://discourse.nixos.org/t/new-merge-policy-for-an-always-green-hydra/8889
- https://discourse.nixos.org/t/label-query-based-automerge/8537

/cc @kevincox probably should be involved (as co-author) either in this RFC or in a sister RFC

rfcs/0079-enforce-prs.md
matthiasbeyer: Okay, thanks for clearing that up for me! :smile:

rfcs/0079-enforce-prs.md
matthiasbeyer: I agree. As far as I know, CI only ensures the `staging` branch builds before merging it, but not whether the actual merge builds, does it?
Because that would be (my) major concern here. Other than that, from what I know, I don't think the workflow should be changed. My concern with this PR is to get an evergreen master (and release branches), which (IMO) increases the development speed and scalability. Staging does not fall under that concern.

rfcs/0079-enforce-prs.md
matthiasbeyer: By what rules exactly?

rfcs/0079-enforce-prs.md
matthiasbeyer: > The group of people with push rights is too small.

No, it is too big. That's the whole point of this RFC - too many people can break master/release and push security issues.

7c6f434c: We do not have enough people with merge access to merge all of the slightly lower profile security fixes in a timely fashion. (Also, we lack enough people to triage security and general severity of bugfix releases, so we need to somehow speed up merging almost all of the minor updates).

Tricking people into merging non-obvious security issues will only become easier by increasing the amount of rote process.

rfcs/0079-enforce-prs.md
7c6f434c: Also, what does GitHub squash do?

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,78 @@
...
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This only affects nixpkgs committers.
+When pushing to a protected branch directly, they get the same message as everyone else and they have to push to an unprotected branch and create a Pull Request from there.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It might break the workflow of some committers which are only a small portion of the community.
```
7c6f434c, line 57: The question should be whether they do the majority of work for some specific areas, though.

Surely it is easy to me to go through PRs now that my activity has dropped and I rarely do more than one thing at once (which is often exotic enough to self-merge). People who try to quickly do a large amount of critical work will be more hindered than me.

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,78 @@
...
+
+Also, Pull Requests might take a bit of time before they are approved by somebody else, which shouldn't matter too much since the trust in committers is already very high and their Pull Requests are likely to be merged fast.
```
7c6f434c, line 59: Realistically, we have the norm that the committer PRs to leaf packages are unlikely to be reviewed ever unless someone has a very specific workflow to test.

rfcs/0079-enforce-prs.md
matthiasbeyer: Agreed. I don't see any conflicting points between your statement and mine. :smile:

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,78 @@
...
+In the last year (as of 2020-09-25), we had
+- 19471 commits on the master branch in nixpkgs
+- of which 10347 (53.14%) were merges
+- of which 9124 (46.85%) were non-merge commits
+
+<small>
+(`git log --oneline --since="1 year ago" --first-parent [--[no-]merges] | wc -l`)
+</small>
+
+This is an improvement over the previous statistic from 2019-04-13 (where 51.85% commits went directly to master) and proves the point that committers are able to work properly without pushing directly.
```
7c6f434c, line 36: Also, this growth looks more like an increase of long-tail contributions and improved efforts on merging these PRs, not like a change in workflows in the widely important areas.

rfcs/0079-enforce-prs.md
```diff
@@ -0,0 +1,78 @@
...
+
+By changing our branching workflow to a no-push-to-master workflow, we can achieve more security, stability and even more important: better scalability.
+
+# Detailed design
+[design]: #detailed-design
+
+In GitHub's [branch protection](https://github.com/NixOS/nixpkgs/settings/branch_protection_rules) rules, branch protection rules which require pull request reviews, include administrators, forbid force pushes and branch deletions must be created.
```
7c6f434c, line 43: Please write an actual detailed design of the proposed rules.

Right now different parts of the proposal leave incompatible impressions on what is actually proposed. Just no direct pushes? Review requriements? How much of them?

rfcs/0079-enforce-prs.md
7c6f434c: Note that the proposal as it stands will not achieve any usefully always-green master. We do not have _any_ pre-merge CI that would always catch either a Chromium build failure or _all_ of the possible rev-dep builds failures, and are not likely to have such in the forseeable future.

7c6f434c (commented):
Additionally, most of the previous discussions of that end up with the outcome «we need a working merge bot first».

Overall I find that this proposal is not yet in a shape where it brings anything new to the discussions that have previously happenned a lot of times, and unless significantly reworked is likely to end up in the same state — stalled because realistically likely improvements do not gain a reliable consensus of being large enough improvements to be worth further throughput problems. I am fine with this outcome, and hopefully the discussion will increase our chances of ever getting some merge automation (which, of course, will open some new and interesting security issues, but will still improve the overall safety on the balance due to faster processing of backlog of fixes).

rfcs/0079-enforce-prs.md
7c6f434c: If we do not have enough throughput, putting speedbumps to workflows of people providing the throughput is a bad idea. This proposal is doing that.

Ma27:
The problem I see is that `nixpkgs` is a fairly huge and diverse project with many different components, from e.g. high-impact code for the `stdenv` to small leaf packages that are only used by a few people which is why I'm skeptical that a "one-size fits all" solution is a good idea.

To join the discussion I'd like to share a few of my thoughts I had in the past about it[1]:

* Can you elaborate how you define an "ever-green" master? Is this reached if everything successfully evaluates (which is usually the case thanks to GrahamcOfBorg) or do you want to see all packages building? If the latter is meant, I'd like to emphasize that this is IMHO a desirable, but rather unrealistic plan. As I mentioned in the past[2], I don't think that it's the job of someone updating a core-package (like `gcc` or `glibc`) to fix every single, probably unused or outdated leaf-package, that's what we have stabilization periods ("Zero Hydra failures") for.

* Do you have some **actual** numbers of how often `master` broke badly (due to direct pushes) in comparison to all breakages?

* We already have a slowly growing PR queue, pushing *every single* change through it (including backports) will make the situation even worse IMHO. Also, with even more PRs, there's also a greater risk that fixes for security issues will get stuck. A while ago I started using tracking branches where I cherry-pick my own changes on top to be able to use them as PRed changes can take a while to be merged. For someone using NixOS on almost every machine every day, this is totally fine, but I wouldn't expect that from an average distro user.

* I think that there's a difference between e.g. backporting (or applying patches from folks who don't want to interact with GitHub) and pushing your own changes:
  * I consider a backport as part of the actual merge-process: a change has been reviewed and is needed on stable releases as well. Filling our crowded PR queue with such patches (unless the backport is non-trivial, of course!) is not a good idea.
  * During times where we have two supported stable releases (right now `release-20.09` and `release-20.03`) I tend to backport patches to 20.03 as well after having them tested. For instance, one of the last `matrix-synapse` releases fixed some security problems that were also on 20.03. If backporting will be more complicated, I'll really have to reconsider if I care enough for maintaining my stuff on releases I don't use anymore.
  * If someone sends you a patch (e.g. via email to avoid GitHub) you actually have a "four-eyes principle" as you review (and probably test) the code received as patch before pushing it, so strictly speaking this is also a pull request.

Instead I'd suggest something like this (note that these are just some ideas, not a detailed plan :)):

* Keep allowing trusted users (i.e. users with commit access) to push low-impact changes (such as updates to leaf-packages) and fairly important changes (e.g. fixes on staging) to avoid slowing down our process[3].

* Make sure that complex changes always go through a PR (e.g. library changes or core package updates just to name a few). Also we should make sure that packages that are "dangerous to update" go through a PR (for instance, I always file a PR for `matrix-synapse` to get some more testers before merging/backporting as they had some weird issues in the past that didn't affect everyone).

* Write up some guidelines to speed up our review process. For instance, I don't think it's helpful to ask humans to do easily automatable changes (e.g. doing the `pname`-change on an existing package where something else changed in a PR).

[1] Disclaimer: I'm sharing my **personal** experience as relatively active maintainer & committer, I don't intend to speak for all other developers. Also, it's possible that I forgot about something here :D 
[2] https://discourse.nixos.org/t/new-merge-policy-for-an-always-green-hydra/8889/19
[3] Of course we could also create a smaller group of "trusted users" with push-access.

blaggacao:
It is very clear to me that the policy proposed in this RFC, after an initial settling period, would shift the needs of core maintainers towards improving PR workflows.

This would put core maintainers into the same boat as "the rest of us" with regards to PR workflows and align interests, instead of preserving special priviliges/workflows which ultimately result in negative overall outcomes.

Oposing some of the oposing views expressed: hen or egg doesn't really (=lastingly) matter.

I strongly think, this RFC should move forward and provide _transitionary provisions_ just as any good legislation does.

7c6f434c:
> This would put core maintainers into the same boat as "the rest of us" with regards to PR workflows and align interests, instead of preserving special priviliges/workflows which ultimately result in negative overall outcomes.

Under the must-get-reviews scenario there are two cases: the amount of human reading increases (or, more realistically, more PRs are let to fall through cracks), with no way to change this by workflows; or people start approving without reading code. Do you claim that this does not happen? It is obvious to me that the less-frequent contributors will suffer first in the former case. Committers can trade reviews, after all…

(Note that true security benefits in case of compromised-committer scenario will never materialise because requiring two reviews on non-maintainer contributions will get cancelled in around 72 hours as unsustainable, and sockpuppets are cheap)

rfcs/0079-enforce-prs.md
kevincox: I checked a repo where people squash and GitHub's squash rebases the result of the squash

matthiasbeyer: This is IMHO the worst anti-pattern github introduced in the recent years, as this effectively destroy history!

kevincox:
Thanks for looping me in. I'm strongly in support of this idea but it is clear to me that there is a lot of infra work to do first. I've been (slowly) working on preparing a set of RFCs that will move us there. But it will take a long time, and non-trivial amounts of work.

My progress so far is [here](https://paper.dropbox.com/doc/nixpkgs-Merge-Queue-Overview--A95nIA~DMlRK59aAjRwYnG9HAg-MTSY8xKH6y1xDEwavyDNW), but I'll summarize the major points I think we need.

1. We need to define a policy for making breaking changes.
2. We need an easy way to mark many packages as "broken". (AKA, hydra and channels are green, but the package doesn't build).
3. We need an easy way to notify maintainers of broken packages.
4. We need some sort of bot to handle the merging (testing and bisection).

Furthermore if we want always green we need a policy for flaky packages and a way of handling arch-sensitive packages.

I'd love to continue to make progress here. I'm hoping to submit the breaking change policy as an RFC soon. I think it lays the groundwork for the rest. Although we will need to polish up the tooling before it can really take effect.

blaggacao:
> Under the must-get-reviews scenario there are two cases: the amount of human reading increases (or, more realistically, more PRs are let to fall through cracks), with no way to change this by workflows

I'd add: over time, the ecosystem would adapt towards better practices (principal motivation of this RFC) and find it's new optimum (most likely heavily backed by tooling and improved workflows)

> Committers can trade reviews, after all…

This is a problem, a code of conduct can counteract.

dasJ:
@blaggacao Of course, I'm in UTC+2, so whatever works for you (and is not at 4 in the morning for me). @matthiasbeyer is probably also interested

blaggacao:
@dasJ @matthiasbeyer @kevincox [Friday](https://www.timeanddate.com/worldclock/meetingtime.html?year=2020&month=10&day=23&p1=37&p2=250&p3=41)? [This Time](https://www.timeanddate.com/worldclock/meetingdetails.html?year=2020&month=10&day=23&hour=14&min=0&sec=0&p1=37&p2=250&p3=41)?

&rarr; https://meet.jit.si/NixOS-RFC0079

7c6f434c:
>I'd add: over time, the ecosystem would adapt towards better practices (principal motivation of this RFC) and find it's new optimum (most likely heavily backed by tooling and improved workflows)

Or, with a freshly organised PR disaster in place, handing out commit rights can finally reach the pace it should.

I mean, if you have a specific proposal to use an available GitHub Action for a mergebot with specific configuration, I will be glad to help you polish it, and I believe this should be possible to push through. What you propose, though, is breaking things to make a point. 

>> Committers can trade reviews, after all…
>
>This is a problem, a code of conduct can counteract.

Counteract what? Agreeing to help each other on an exchange-of-effort basis? The reviews will be honest, that is not a problem.

blaggacao:
> What you propose, though, is breaking things to make a point.

rather: ... to have a fair chance to escape the local optima.

It sounds as if I hadn't stressed the suggestion for a _transitory period_ for this policy decision. But holding back a policy decision, that in principle is right, because of the fact that it has practical (transitory) implications is not the kind of long term goal orientation I would expect from such policy decisions.

> Counteract what? Agreeing to help each other on an exchange-of-effort basis? The reviews will be honest, that is not a problem.

Counteract nixpkgs PR throughput being the space of private bargaining instead of following some sort of shared prioritization. Yes.

7c6f434c:
>> What you propose, though, is breaking things to make a point.
>rather: ... to have a fair chance to escape the local optima.
>
>It sounds as if I hadn't stressed the suggestion for a _transitory period_ for this policy decision. But holding back a policy decision, that in principle is right, because of the fact that it has practical (transitory) implications is not the kind of long term goal orientation I would expect from such policy decisions.

IT is not right in principle. There are contexts inside which it improves things, and contexts inside which it makes things worse. We are currently in the latter, with many acknowledged paths to improvement not looking like a leap of faith into a worse situation.

>Counteract nixpkgs PR throughput being the space of private bargaining instead of following some sort of shared prioritization. Yes.

So, forbidding volunteers to perform undisputably useful work for the project in the way they prefer. If any policy is wrong on principle, this is.

kevincox:
I can make that time. Happy to discuss and see the best way to move towards this goal.

8573:
I think I would support a proposal such as this only if it were to come with a Bors-style merge bot. Then concerns about, e.g., backports could be addressed by having the "trusted users" be designated as allowed to "r+" ("approve" to the bot) their own pull requests without four-eyes review (and similarly for allowing "updates to leaf-packages" by their maintainers), meaning that the bot would try merging the patches, test the merged code, and push it only if the tests pass on the merged code. If the patch doesn't cause breakage, this should be a "fire and forget" process for the trusted committer; if the patch does cause breakage, this contains the mess — as far as I see, it should make things easier for the committer overall.

@Ma27 wrote—
>Can you elaborate how you define an "ever-green" master? Is this reached if everything successfully evaluates (which is usually the case thanks to GrahamcOfBorg)

Doesn't Hydra still often spend several days failing to build the nixos-unstable channel? Just recently it was stuck for several days from 2020-10-14.

blaggacao:
Would you agree to switch roles and defend the opposite of your current position @7c6f434c ? I sense we are digressing from [the essence of this RFC](https://github.com/dasJ/rfcs/blob/enforce-prs/rfcs/0079-enforce-prs.md#summary) which argues with the 4 eyes principle. With your advantage in ecosystem experience, it should be relatively easy to help the authors to get details stretched out around this principle and transform this RFC into a solid proposal that carefully takes into account (and works around) the possible friction.

Ma27:
> Doesn't Hydra still often spend several days failing to build the nixos-unstable channel? Just recently it was stuck for several days from 2020-10-14.

Yes, but aren't many of those problems related to infrastructural problems like flaky tests, timeouts etc.?

7c6f434c:
> With your advantage in ecosystem experience, it should be relatively easy to help the authors to get details stretched out around this principle and transform this RFC into a solid proposal that carefully takes into account (and works around) the possible friction.

I can argue convincingly for bad ideas, but prefer not to.

The proposal looks like it expects to improve process security (without defining any threat model) by means that cannot, achieve ill-defined ever-green master either by means that actually cannot or according to a definition that is only very rarely violated today, puts abstract fairness above working workflows, rejects the only realistic way to improve fairness — transparent and swift process of granting commit access to people with proven commitment and experience, does not even consider the real bottlenecks, assumes that tooling can help with the bottleneck of having people read code, and tries to substantiate the points with data taken from Git history (which almost never works, and definitely doesn't this time).

If I started to fix all that, the process would be of little difference to me advising to scrap that, fork Nixpkgs, pick CI/automerge GitHub actions, test them out under different conditions, and propose and RFC to apply the same automation configuration to Nixpkgs. 

I guess feature include merge-on-CI-pass, ability to add some builds succeeding to the condition, ideally scoped permissions for non-committer maintainers.

(How would it go from there? See documentation format conversion, after a lot of debate and alternative proposals _some_ path in roughly the same direction gets finally chosen a year later)

Then I guess a cherry-pick bot could be proposed. Once the tooling is in place you have a chance to study what needs to be added to tooling so that people handling `staging` agree moving to some tooling-assisted PR-based workflow is not worse medium-term than their current approach. Note that making their workflow worse in the medium term is just bad, their work is critical for the project (in contrast to the talk about four eyes and whatever)

blaggacao:
@7c6f434c Please consider, joining the call, if you can find time. :+1:

rfcs/0079-enforce-prs.md
samueldr: From being pushed to.

blaggacao:
@NeQuissimus Did you put some (minimal) research into https://github.com/marketplace/actions/merge-pull-requests or was it just the next best search result?

domenkozar:
With actions it's simple to automate all these things, we can even decide to merge things without CI passing if that's desired.

I run a few things using https://github.com/stefanzweifel/git-auto-commit-action

rfcs/0079-enforce-prs.md
Profpatsch: I would not want the history of nixpkgs to have 50% merge commits to be honest. `magit` can barely display it as-is, no reason to clutter it with even more merge commits. If anything, we should encourage committers to use a rebase merge for 1-commit or 2-commit PRs that don’t introduce a critical change.

The PRs can still be found by searching Github for the commit id, and the default merge commit Github produces are rather simplistic (and don’t contain much information besides the PR number, not even a link to the PR).

rfcs/0079-enforce-prs.md
Profpatsch: Yes.

Also we need descriptions of every way the direct-push to master/branches is used by contributors at the moment, and what workflow to replace them with if applicable.

Profpatsch (changes requested):
I applaud the effort, and I agree with the security problem introduced by direct push access, but as it stands this proposal will require /a lot/ more fleshing out.

This includes talking to all release maintainers and all language set maintainers (e.g. @peti), to figure out their workflows and how we can change them to accommodate this RFC. I’m sure they see the problem, but we have to make sure to not throw any bricks between their feet.

Ma27:
> This includes talking to all release maintainers and all language set maintainers (e.g. @peti), to figure out their workflows and how we can change them to accommodate this RFC. I’m sure they see the problem, but we have to make sure to not throw any bricks between their feet.

Not really a "high-impact maintainer", but I'd be interested in such a discussion as well :)

7c6f434c:
@blaggacao as a pure point of proper workflows, «this time» in the link to timezone converter should be «2020-10-23 14:00 UTC» (or did I get the time wrong from the link?). Not insisting on the precise formatting, but it's always better to include one unambiguous description into the text and allow to get others via link…

(This message is neither accepting nor rejecting that invitation to participate, will see tomorrow afternoon)

Mic92:
Somebody who wants to shepherd this one?

alyssais:
As has been already pointed out, the four eyes principle is hugely impractical for Nixpkgs, when we can't even get four eyes on mostsmall PRs as it is.

So, let's assume that the only practical path forward is taken, and that motivation is dropped from the RFC.  And therefore we come up with some way of automatically applying patches once they have passed CI.  What do we still get out of it?

> Security vulnerabilities, breaking changes which are not tested, and regressions are easily introduced this way.

Security vulnerabilities and regressions are very unlikely to be caught by CI, so those aren't going to be helped by this change. Breaking changes which are not tested by the committer would still be caught by this (assuming the CI tests would catch them), so that would be our gain here.

> proves the point that committers are able to work properly without pushing directly.

This doesn't follow at all, as has already been pointed out.  It proves that committers are able to do _some_ work properly without pushing directly.

So, the benefit we'd _actually_ see from this change would be that changes had to pass some amount of tests before they were accepted. That's still a good benefit!

What are the drawbacks?

One is that if PRs have to be opened for every trivial change that would have previously been pushed to master, they'd all go right to the top of the PR list.  This would mean that the naive reviewer would end up directing reviews at the changes that need it least (trivial changes that some of the most trusted members of our community have decided are so insignificant that reviews need not be sought for them), and away from the changes that need it most (other small PRs that just need somebody to look them over and +1 them so they can be merged).

We could avoid this to some extent with a label — that could at least make it clear that these commits are low-priority for review, because they'll be auto-merged anyway.  But they'd still pollute the PR list and add just a little bit more friction to finding and reviewing good PRs from non-committers.

A better solution to this problem would be to have committers push to new branches named with a pattern like "automerge/*", and then have the apply bot look at those.  That way we'd avoid spamming the PR list. CI status would still be indicated on the commits, and ideally a committer could be notified by email if CI failed.

Overall I think mandatory CI for commits is a good idea, but we need to be careful about how it's implemented.  Mandatory review is unworkable until we can keep up with the demand for reviews we alreadyhave.

Finally, a general point:

The RFC is written in a way that feels very resentful towards and dismissive of the needs of committers.  Statements like "It might break the workflow of some committers which are only a small portion of the community." display a real lack of appreciation for the work committers do.  Nixpkgs committers aren’t some closed cabal of elites. Anybody can become a committer, and new committers are added all the time. Putting "Handle complaints from people who are used to direct pushing." in "Future work", is even worse — surely complaints should be handled now, before we make a change, and taken into account, rather than making a change and then dealing with the fallout from the people it affects.  I'm sure none of this was your intention, but it doesn't sound very friendly to the people whose volunteer workflows you are proposing to change here, and I don't think it'll help you achieve your aims.

NeQuissimus:
@blaggacao A search result. I use Mergify for my projects

matthiasbeyer:
> A better solution to this problem would be to have committers push to new branches named with a pattern like "automerge/*", and then have the apply bot look at those. That way we'd avoid spamming the PR list. CI status would still be indicated on the commits, and ideally a committer could be notified by email if CI failed.

This is basically a good path, but your implementation of that idea sounds like a workaround to the shortcommings of the github UI/UX. Maybe a variant of this would be to have a dedicated group which handles "easy-to-decide-on" package updates ("version + hash"-changes essentially). That group would push the individual patches to a dedicated branch, which then gets CI and merged ... daily, weekly,... whatever.
This would cause a minimal slow-down of simple package updates, true. The benefit would be that these simple package updates wouldn't spam the PRs anymore (only by new contributors, but that could be handled by a nice bot that closes the PR and gives them a tutorial on how to submit these trivial patches) and make the show happen in one place!

> The RFC is written in a way that feels very resentful towards and dismissive of the needs of committers. [...]
> t doesn't sound very friendly to the people whose volunteer workflows you are proposing to change here, and I don't think it'll help you achieve your aims

This was clearly not our intention with this PR and I apologize to everyone that feels offended by the wording in this RFC!

alyssais:
> Maybe a variant of this would be to have a dedicated group which handles "easy-to-decide-on" package updates ("version + hash"-changes essentially).

Ooh, I like this idea a lot!

> This was clearly not our intention with this PR and I apologize to everyone that feels offended by the wording in this RFC!

❤️

blaggacao:
> branches named with a pattern like "automerge/*"

Instead of a branch, I would suggest a label. Labels are a _generally better_ interface to any future improvement to ACL (integrate with available bots, fine grained control who can set what label, etc).

I'm trying to put something together based on a label query merge bot as an **_experiment_** with r-ryantm (with the help of @kevincox and @ryantm)

blaggacao:
Can we give the free tier of https://mergify.io/pricing a try? Hasn't to be a long term solution, but copycatting from an established company would probably positively inform our workflows.

7c6f434c:
>Instead of a branch, I would suggest a label. Labels are a _generally better_ interface to any future improvement to ACL (integrate with available bots, fine grained control who can set what label, etc).

Branches are an obviously generally better interface, though, because they can be set both using and avoiding the GitHub UI.

Can't the bots filter by the pushing person belonging to a team, or something?

blaggacao:
> Can't the bots filter by the pushing person belonging to a team, or something?

The emerging pattern that I see is label query based merge bots.

blaggacao:
Just wrote mergify contact:

> Hi, https://github.com/NixOS/nixpkgs is in bad PR distress. However, I'm not confident it will be possible to get somebody to sponsor. I'm just a regular community member, but would you be willing to sponsor a premium version for the NixOS organization if I could help broker a sponsorship acknowledgement here: https://nixos.org/donate.html ?

andir:
I would like to add a concern about lock-in that might not appeal to many/most but IMHO is still worth considering in the long run:

We might want to prefer what we can implement in a **portable** fashion. Something that we would be able to continue doing and/or using when Nixpkgs moved away from GitHub or some of the features (for example GH Actions) aren't available for "free" anymore. More importantly something that we can *change* instead of just hoping for the best intentions of some corp.

We are building our (F)OSS collection & Linux distribution with an ever growing stack of non-free supporting software. The list of dependencies and the amount of work required to migrate from these will not disappear magically by postponing these concerns forever.

blaggacao:
> We are building our (F)OSS collection & Linux distribution with an ever growing stack of non-free supporting software. The list of dependencies and the amount of work required to migrate from these will not disappear magically by postponing these concerns forever.

Extremely true. :heart: But I'll cite the already well known mnemonic: "our bottlenecks are elsewhere".

Sorry to offend, but I think it's the truth, that _has_ to be spelled out in this context:

NixOs has not an excellent track record of _getting things done_ in comparison with the top tier of open source projects of similar size (I'm looking at kubernetes specifically as a role model). I suspect, this is why it doesn't attract more vested investment in _people resources_ from the corporate world. So if we have to compromise for the sake of ecosystem sustainability and navigating our way through. Why not accept a considerate bargain?

matthiasbeyer:
@alyssais Maybe, if more people like this idea (about the dedicated team for minimal hash+version update patches), we can break this out into a dedicated RFC?

Sounds like a rather simple thing and self-contained problem to solve, doesn't it?

kevincox:
I second this sentiment. While I would like to stick to FOSS I think especially for non-critical experiments it is pragmatic to experiment with non-free options. We can then take what we learnt and either adopt or write free software once we know better what we need.

blaggacao:
@matthiasbeyer Maybe, if more people like this idea (about the dedicated team for minimal hash+version update patches), we can break this out into a dedicated RFC?

Please let's first experiment and gather data to avoid discussions that are revolving around local optima in the ecosystem advancement. :pray:

blaggacao:
> The RFC is written in a way that feels very resentful towards and dismissive of the needs of committers.

If (since) appreciation is the currency committers are paid with (I assume), can we make them a _extremely visible_ and _highly prized monument_ somewhere else (outside of RFC discussions &mdash; since those sentiments &mdash; safe to assume never being intended &mdash; might proliferate political correctness in discussions at the expense of being able to productively discuss the holy cows)?

7c6f434c:
@blaggacao 

>  I'll cite the already well known mnemonic: "our (current) bottlenecks are elsewhere".

Note that you need to think whether we are applying this to ends or to means. It applies more straightforwardly to the latter than to the former, and often reveals what speaker considers ends and what means.

(Yes, avoiding critical lock-in to non-FOSS which has some practices some people consider questionable is valuable in itself to some of us, including me)

> NixOs has not an excellent track record of getting things done in comparison with the top tier of open source projects of similar size (I'm looking at kubernetes specifically as a role model)

Which might be just a bit related to way broader-but-shallower mission, and way more diverse stakeholders. 

Nix* ecosystem is not just (and not supermajority) enterprise-ish Linux server deployment used by people at work, unlike Kubernetes.

> Please let's first experiment and collect hard data and experience in order to avoid discussions that are revolving around local optima in the ecosystem advancement.

Note that downside risk of some experiments is easier to estimate than of some others.

Collecting hard data «this breaks too many workflows» comes at a cost. 

> If (since) appreciation is the currency committers are paid with (I assume), can we make them a extremely visible and highly prized monument somewhere else

I am not sure I can explain why I deeply and sincerely making such statements is a bad idea without getting too negative, but I will try.

What you propose sounds like «take away a bit of self-determination by external pressure, in exchange for a shiny thing that explicitly does absolutely nothing». This is not a good trade…

blaggacao:
> Which might be just a bit related to way broader-but-shallower mission, and way more diverse stakeholders.
> Nix* ecosystem is not just (and not supermajority) enterprise-ish Linux server deployment used by people at work, unlike Kubernetes.

1. Let's start with the ends: make nix great again. That is: ensure ecosystem sustainability, diversity _and growth_ &rarr; get some fiat money injected &rarr; bring sane software packaging to a broader audience (and realize other spill over effects). Else, our "deeper" mission is just a hobby that does not live up to some sort of accompanying vision.
2. Why doesn't it play out as I would expect by pure measure of the _inherent technical value_ of the nix ecosystem? &rarr; sub-optimal coordination between different stakeholder groups &rarr; reduced social dynamic &rarr; reduced reach
3. How come coordination is sub-optimal? &rarr; relevant stakeholder groups for a successful software project are dauntingly underrepresented (either by lack of presence or &mdash; more importantly &mdash; by lack of advocates). I'm not lying by noting that committers (and their special interests &mdash; it is what it is) are a _very_ powerful lobby group in the ecosystem policy decision making. Does _any_ skewed representation of interests serve the broader goal of point 1? To my judgment, at least not good enough.
4. Who are the stakeholders, then? &rarr; It will be a good thing to conduct a truthful and _hopefully_ un-skewed stakeholder  analysis (present & desired) for the nix ecosystem and (at least formally) represent their stakes in the analysis section of future RFCs &mdash; and if they are at risk of getting under represented, you'd once again have to switch roles and defend their conflicting interests :wink: 

> What you propose sounds like «take away a bit of self-determination by external pressure, in exchange for a shiny thing that explicitly does absolutely nothing». This is not a good trade…

That would have been a complete misinterpretation that assumes careless thinking on my side on the basis that I'm refusing to accept the single primacy of the committer stakeholder group and interests over those of other vital stakeholders without even theologically (since they are not here in the RFC process) measuring their needs and interests.

For some of those stakeholders, not applying 4-eyes-principle (and taxing it as a bad idea "in principle") _might_ just subsume as inconsistency with general security values on other topics, such as attempts to bootstrap the `gcc` compiler (you name it). Such stakeholder would probably reason something around: "If there is a real mission behind that, it not only suffices to say A and make exceptions for the rest of the alphabet."

blaggacao:
So see you tomorrow in jitsi and let's conclude and agree on a path forward. If it does take less then a year to make some progress, that would be already an immensely valuable thing and a super rewarding experience for everyone.

&rarr; https://github.com/NixOS/rfcs/pull/79#issuecomment-713766359

7c6f434c:
> growth

Note that growth by abandoning that one thing X values in the ecosystem in the first place is often possible and also for network-effect reasons actively worse than ecosystem stagnation for X.

> How come coordination is sub-optimal? → relevant stakeholder groups for a successful software project are dauntingly underrepresented

Preferences of different group are in direct contradiction, therefore coordination is hard and cannot be done in a fair way on _some_ issues. The great value of Nix* community is ability to collaborate on something despite known unresolvable differences in other contexts.

> committers (and their special interests — it is what it is) are a very powerful lobby group

And your proposal makes things worse. We need more committers, and we need every group of contributors experienced enough to understand the scope of the project represented among them!

> It will be a good thing to conduct a truthful and hopefully un-skewed stakeholder analysis

Have you read the discussion of RFC SC rotation? You should.

> vital stakeholders

> not applying 4-eyes-principle (and taxing it as a bad idea "in principle") might just subsume as inconsistency

Not everyone in the world has a stake in Nix* ecosystem. That's fine. We do our best to help those who want, to learn. We sometimes tell people Nix* will currently provide no value for them, that's fine.

> bootstrap the gcc compiler

It's not security, it's predictability of build behaviour.

rfcs/0079-enforce-prs.md
matthiasbeyer: > If anything, we should encourage committers to use a rebase merge for 1-commit or 2-commit PRs that don’t introduce a critical change.

I'm not a fan of rebase merges, tbh. I'd rather propose a daily or weekly collecting-small-stuff-PRs. But that's another RFC and we shouldn't discuss this here, IMO.

dasJ:
Okay, so we (@7c6f434c @matthiasbeyer @lschuermann @kevincox @blaggacao) have now had our meeting call.

While this PR lacks direction, it is also limited by the current lack of tooling to reach this goal.

@kevincox is currently working on smaller RFCs for additions to the tooling, including the concept of merge trains.
His work also covers a lot more details and use cases, making it far easier to slowly approach some of the goal of this RFC.

Thank you all for your comments, they will help a lot in a future iteration of this RFC or might be picked into smaller RFCs.

kevincox:
I'm just going to try out what I see as the current state after the meeting today:

- With the current workflow simply requiring PRs for all changes will create unsustainable maintainer burden.
- The security benefit of having multiple approvers is desirable however would cause a substantial human cost that nixpkgs can't afford right now.
  - We should consider approaches to make reviews easier for the maintainers, in the hopes that it will allow us to pay this cost in the future.
  - We could consider an auto-merge bot that can merge with rules such as the following:
    - 2 approvers in the maintainers team allows merging.
    - Approval from maintainers of all affected packages allows merging. (Including self-merge!)
    - Approval from the maintainer of the root affect package and `min(count(maintainers(afected packages), log(affected packages))` additional approvals.
  - Better pre-merge testing can free some time that maintainers currently spend debugging hydra failures.

The TL;DR is that there are dependencies of this change that are useful on their own. When we have those in place we can reconsider this proposal.

7c6f434c:
Whatever our agreements and disagreements on the global roadmap, I am looking forward to see and support some of the proposals @kevincox is likely to submit in the first wave.

kevincox:
It is worth nothing one thing. We need to be careful because no matter how much maintainer time we save it will always be a cost to adding new security policy. We should keep our minds open that even though maintainers will always be overloaded at some point we will need to take a step back in productivity for core maintainers in order the help the community as a whole. However we need to do this at the right time and help make that step as small as possible.

7c6f434c:
> we will need to take a step back in productivity for core maintainers in order the help the community as a whole.

Well, it will never be for the benefit of community as a whole. There always will be contributors for whom faster merges of changes they care about, including some of the security fixes, matter more than review standards, of course provided that the committers are people who have some proven investment into the ecosystem.

You will _always_ need to define your threat model and argue carefully to have a convincing case that the benefit of the actual policy as implemented is worth anything,

>However we need to do this at the right time and help make that step as small as possible.

Well, if you want higher review standards I guess you could set up a fork, recruit people who are not committers and would like better standards, run a basic classifier of changes into PRs/unidentified change blocks, and publish a branch that tracks your progress in reviewing the master branch. I guess if you find anything convincingly bad that was not fixed Quite Soon, you will have a stronger case to rally people around reviews…

blaggacao:
> Well, if you want higher review standards I guess you could set up a fork, recruit people who are not committers and would like better standards, ...

No! It is preferable to gain support for greater flexibility in discussion to up the chances to move the _current_ ecosystem to a new steady state.

7c6f434c:
>> Well, if you want higher review standards I guess you could set up a fork, recruit people who are not committers and would like better standards, ...
>
>No! It is preferable to gain support for greater flexibility in discussion to up the odds to move the _current_ ecosystem to a new steady state.

A fork that tracks master with a lag based purely on extra review, has no extra commits, and raises noise in mainline about any problematic things that have been missed by the current processes, is still a part of the same ecosystem. Just like nixpkgs-unstable branch is.

If you end up catching problems regularly, you will gain leverage in discussions, both by demonstrating capability to run the missing part of the workflow and by bringing specific high-quality data.

Discussions should aim to properly appreciate all the evidence and all the preference mismatches, not just be flexible for flexibility sake.

blaggacao:
While I share your view on many leaf aspects, I wouldn't conclude to regress on the trunk(s) of the argument(s). A tree stands by its roots, roots grow off the soil. Nix* ecosystem is a tiny forest standing on fertile soil. After 20 years of building up the sap, it's probably the right time for a serious [mast year](https://sanantonioreport.org/mast-year-for-area-oak-trees-makes-for-massive-acorn-crop-well-fed-deer/) of fertile acorn spreading. Welcome flakes. Welcome change (in stakeholder structure).

7c6f434c:
> Welcome flakes. Welcome change (in stakeholder structure).

Flakes will be overused, people will make a mess, then a balance will be found. I expect Nixpkgs to become multiple flakes inside a monorepo once dust settles.

There _is_ change in the structure of decision making. But the aim is not to change everything quickly, it is to avoid destabilising the project, either by erratic change or by being stuck unable to make high-quality decisions.

Ericson2314:
I agree we shouldn't change too much too fast, but I also would *much* rather spend our "change budget" on things coming out of this like that could save serious core contributor time (auto-merge bot), rather than things like flakifying Nixpkgs which perhaps could increase adoption and look flashy but I don't believe address our core issues around the sustainability of the project.

7c6f434c:
I have high hopes that we can first get some of the tooling where change budget cost is low (implementation cost is another story, though)

blaggacao:
_Change Budget_ is a good mnemonic term for use in future policiy discussions. Note that it can not only be leveraged by tooling but also expanded by a strong shared vision.

blaggacao:
There seems to be a consensus about auto-merge bot.

I invite anyone interested in investing time and energy in making this happen to PM me on discourse with the goal to discuss (sheltered from public scrutiny) the constituing individual's feasibility to participate in the conformation of a special interest working group.

A special interest working group seems like a good enough answer to productively channel the dynamic this discussion has developped.

blaggacao:
Looks as if mergify is willing to help the community. @NeQuissimus would you be interested in helping to set up an example configuration to have a basis for discussion and see what's possible?

For conducting an experiment, I think it might help if it is narrowly scoped (and does not produce external effects on parties not yet accustomed with this workflow), so I would suggest to come up with a configuration for the subtree containing the things you maintain.

What do you think?

7c6f434c:
>What do you think?

I think that GitHub is an unfixable mess as they break their own API promises without updating documentation even in a reasonable time after the fact. Hence an external proprietary merge-bot with good reputation is probably not making things any worse than just the (bad for a ton of reasons, but expensive to change for another ton of reasons) fact of using GitHub, in my opinion.                                                                               
I like the idea of testing constrained to a relatively small subtree at first.

I am afraid that giving write access (which, I think, GitHub cannot limit to subtree) to the repository to an externally developed bot would require quite detailed write-up of what we know about all the moving parts. The knowledge might be imperfect but people will ask for what we do have…                                                                              
What questions I feel compelled to ask immediately:


Do I understand correctly that we are supposed to give the bot a moderately long-lived token granting push access?                                                                      
Does GitHub bundle this with any other permissions?

(Or does Mergify need a procedure that implies other permissions)

Does GitHub support limiting the token to subtrees?

Can the token be GitHub-side limited just to `master` ?

How token rotation works?

Do I understand correctly that a Mergify has long-supported option to never push anything touching anything outside the configured subtree?                                             
Can this be multiple subtrees?

How subtree reconfiguration works?

How many people could be managing the Mergify configuration?

blaggacao:
As a result of this RFC's discussion, part of the discussion evolves under the newly conformed [_SIG Workflow automation_](https://discourse.nixos.org/t/about-the-sig-workflow-automation-category/9673) and parts of the reverberations of this RFC's discussion are picked up in:

- [Geistesblitz: Empower our Trust Model through a Policy Decision Agent](https://discourse.nixos.org/t/geistesblitz-empower-our-trust-model-through-a-policy-decision-agent/9746)
- [Geistesblitz: Well defined JSON schema for ecosystem build tool outputs](https://discourse.nixos.org/t/geistesblitz-well-defined-json-schema-for-ecosystem-build-tool-outputs/9747)
- ANd a little more technical one: [Geistesblitz: Bors policy enforcement with revocation](https://discourse.nixos.org/t/geistesblitz-bors-policy-enforcement-with-revocation/9748)

_Please feel free delivering your own "Geistesblitz" or "Steam Engine" through the templates provided in the footer of the [SIG Workflow automation description page](https://discourse.nixos.org/t/about-the-sig-workflow-automation-category/9673). Keeping the discussion going is an important intrinsic motivator for all of us. Thank you all for your contributions! :heart:_

zimbatm:
NOTE: If you are a committer and want to avoid accidents, it's possible to disable the push on the client-side:
```sh
git remote set-url --push origin no_push
```
Or just for the master branch:
```sh
git config branch.master.pushRemote no_push
```

domenkozar:
I'd like to propose a different approach by taking smaller steps.

A good example of what motivated me is https://github.com/NixOS/nixpkgs/pull/112477 where the PR is stalled because there are no active maintainers of that stack.

The proposal is to use [an existing bot](https://github.com/OSS-Docs-Tools/code-owner-self-merge) to merge PRs approved by maintainers specified in [CODEOWNERS](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners).

This is simple to implement (a few lines of YAML + a token), most of the work is documenting the policy (RFC).

The benefits would be that we're slowly moving towards understanding better who is the maintainer of what part of nixpkgs and that we don't need to give out full commit access to those that want to merge things. On top of that it's all transparent.

I'd like to hear your thoughts about this as a reasonable way forward.

7c6f434c:
Does that necessiate a brief review of CODEOWNERS now that it will actually do something? In favour either way.

FRidh:
> This is simple to implement (a few lines of YAML + a token), most of the work is documenting the policy (RFC).

It would mean looking at https://github.com/NixOS/rfcs/pull/50 again and making the case for the suggested simpler method.

tazjin:
Strongly in favour of introducing an owners concept and automatic (or self) merges.

The benefit of the latter should also not be understated, it's a small psychological one that is known in large orgs.

Imagine the case of a PR that touches a range of different things and needs review by multiple people, but the "least significant reviewer" does their review last: In a model where completed reviews just "grant merge permission", the owner of the change (or a bot) can merge as soon as all reviews are done. In our current model, the owner of the change now needs to poke one of the other reviewers again (or find someone else willing to do the merge).

Not granting treewide commit access to every contributor is also great!

matthiasbeyer:
> Not granting treewide commit access to every contributor is also great!

And, because treewide commit access happened before, reducing the number of people who actually _have_ treewide access in the same step!

I love the idea of using CODEOWNERS - I just hope that the bot @domenkozar mentioned works well enough, so that if multiple files from different owners are touched, all owners must review (or at least that part is configurable with the bot and we can see what works best for us).

domenkozar:
> Does that necessiate a brief review of CODEOWNERS now that it will actually do something? In favour either way.

Skimming through it it's just a few people that don't already have commit access.

> It would mean looking at #50 again and making the case for the suggested simpler method.

Good point, all that feedback should be taken into the account.

>  I just hope that the bot @domenkozar mentioned works well enough, so that if multiple files from different owners are touched, all owners must review (or at least that part is configurable with the bot and we can see what works best for us).

It doesn't do that, although I wouldn't complicate it much, if we survived with the merge button we can start with a `Merge` comment as well (I don't think there's a good technical solution to communication issues).

matthiasbeyer:
> The big question then comes, who can become a maintainer?

As a first rule of thumb: If I add a new package, I should automatically be maintainer of that package (and by that I do mean `meta.maintainers` _AND_ codeowner).

It might also be a good idea to copy `meta.maintainers` of a package to codeowners as a starting point. For the non-package stuff in nixpkgs, I'd just use the top-10% (like the top 2 or 3) of the contributors for each file as a starting point and go from there... It won't be a one-step-change anyways, but rather a process of exploring what works best for the community.

Ekleog:
Right now, `meta.maintainers` is basically free-for-all. If any one who did one hash-bump update to a library with lots of dependents but few maintainers gets the ability to introduce a backdoor into this library, it's going to be a bad situation to be in.

> It doesn't do that, although I wouldn't complicate it much, if we survived with the merge button we can start with a Merge comment as well (I don't think there's a good technical solution to communication issues).

It'd definitely be possible, but it'd mean that any codeowner of any part of the code technically has the ability to land code literally anywhere in nixpkgs; so the security benefits of not giving everyone treewide access would be very limited… and it might actually become a security issue, because we'll start assuming that codeowners can only touch the stuff they are owners of and that would not be true. Because we'd start giving code ownership much more liberally than we currently do for the commit bit.

IMO the core of the idea is good, but we probably should 1. make sure that the bot requires review from at least one owner of each touch filed (otherwise it'd become a security liability) see paragraph above, and 2. start with code owners that are a subset of committers, and stay very deliberate in how we grant code ownership.

happysalada:
> Right now, `meta.maintainers` is basically free-for-all. If any one who did one hash-bump update to a library with lots of dependents but few maintainers gets the ability to introduce a backdoor into this library, it's going to be a bad situation to be in.

How about having a rule where if there is a single maintainer to the package, then one approval is enough to merge. If there are 2 or more, then you need at least 2 approvals. You can definitely still game the system by creating many accounts. I feel thought that it would take care of 90% of the cases where there would be a problem.

Another solution to the problem, would be to create an `activeMaintainer` attribute. People on it are using the package extensively and want to review/approve everysingle PR. (just thinking out loud here). That attribute could be called alternatively `codeOwners` or whatever you feel is best.

> IMO the core of the idea is good, but we probably should 1. make sure that the bot requires review from at least one owner of each touch filed (otherwise it'd become a security liability, see paragraph above) and 2. start with code owners that are (a subset of) committers, and stay very deliberate in how we grant code ownership.

Elaborating on the previous idea, you could make codeOwners required to review every PR. The more strict requirements would leave the more committed. The problem would arise when a codeOwner steps down and there is no one to replace him/her.

asymmetric:
Does anyone care to provide a summary of:

- the different points raised against this RFC
- why it was considered to "lack direction"

My impression is that an RFC that gets us one step closer to a Good Place<sup>TM</sup> got stuck in not so fruitful, very complex discussion, and that we let perfect be the enemy of good, as they say.

infinisil:
New RFC for this but slimmed down and updated: https://github.com/NixOS/rfcs/pull/156

PR #81, opened 2020-12-10
[RFC 0081] Show unmaintained packages

Lassulus:
[rendered](https://github.com/Lassulus/rfcs/blob/unmaintained-pkgs/rfcs/0081-show-unmaintained-pkgs.md)

### DIFF
--- a/rfcs/0081-show-unmaintained-pkgs.md
+++ b/rfcs/0081-show-unmaintained-pkgs.md
@@ -1,0 +1,56 @@
+---
+feature: show unmaintained packages
+start-date: 2020-12-10
+author: lassulus
+co-authors: Mic92
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+This RFC is about the policy (whether we should do this) not about implementation details (how we should do this)
+
+When evaluating a system (for example with nixos-rebuild) unmaintained packages should show up, so people are encouraged to maintain them.
+
+# Motivation
+[motivation]: #motivation
+
+Get all commonly used packages maintained. Get more maintainers, encourage users to contribute to nixpkgs.
+
+# Detailed design
+[design]: #detailed-design
+
+When evaluating a package a check would test `meta.maintainers` and raise a warning if it is empty. Add an option to disable it to speedup evaluation again.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+```
+package bzip2 is unmaintained: (https://github.com/nixos/nixpkgs/pkgs/tools/compression/bzip2/default.nix) it is required by nix, firefox, X11 ...
+```
+
+(not sure if we want to print the whole dependency list)
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Does evaluation take longer?
+People could be afraid to see how many unmaintained packages there are and switch to $distro.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Have a fancy website to show all currently unmaintained packages.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What parts of the design are still TBD or unknowns?
+
+# Future work
+[future]: #future-work
+
+Implement the feature :)
--- a/rfcs/0081-show-unmaintained-pkgs.md
+++ b/rfcs/0081-show-unmaintained-pkgs.md
@@ -52,5 +52,4 @@ What parts of the design are still TBD or unknowns?
 
 # Future work
 [future]: #future-work
-
-Implement the feature :)
+In the future, when packages usually have maintainers, we can think about enforcing packages to have a maintainer. So removing them if the only maintainer steps down and another doesn't show up.
--- a/rfcs/0081-show-unmaintained-pkgs.md
+++ b/rfcs/0081-show-unmaintained-pkgs.md
@@ -24,6 +24,7 @@ Get all commonly used packages maintained. Get more maintainers, encourage users
 [design]: #detailed-design
 
 When evaluating a package a check would test `meta.maintainers` and raise a warning if it is empty. Add an option to disable it to speedup evaluation again.
+By default we assume it's slow and should be opt-in (maybe a flag to nixos-rebuild). At a later point we can make it opt-out if impact on eval time isn't too big.
 
 # Examples and Interactions
 [examples-and-interactions]: #examples-and-interactions
--- a/rfcs/0081-show-unmaintained-pkgs.md
+++ b/rfcs/0081-show-unmaintained-pkgs.md
@@ -4,7 +4,7 @@ start-date: 2020-12-10
 author: lassulus
 co-authors: Mic92
 shepherd-team: Silvan Mosberger, Rok Garbas, Finn Behrens
-shepherd-leader: (name to be appointed by RFC steering committee)
+shepherd-leader: Rok Garbas
 related-issues: (will contain links to implementation PRs)
 ---
 
--- a/rfcs/0081-show-unmaintained-pkgs.md
+++ b/rfcs/0081-show-unmaintained-pkgs.md
@@ -3,7 +3,7 @@ feature: show unmaintained packages
 start-date: 2020-12-10
 author: lassulus
 co-authors: Mic92
-shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-team: Silvan Mosberger, Rok Garbas, Finn Behrens
 shepherd-leader: (name to be appointed by RFC steering committee)
 related-issues: (will contain links to implementation PRs)
 ---


ryantm:
The "fancy website" [basically already exists](https://repology.org/projects/?search=&maintainer=fallback-mnt-nix%40repology).

Lassulus:
> The "fancy website" [basically already exists](https://repology.org/projects/?search=&maintainer=fallback-mnt-nix%40repology).

this is indeed quite cool. But I would still prefer a integration in nix/nixpkgs/nixos.

On the repology page I can't see more than 200 packages at once (so creating a complete list seems not possible?). And I'm not entirely sure which packages I really use.

garbas:
> > The "fancy website" [basically already exists](https://repology.org/projects/?search=&maintainer=fallback-mnt-nix%40repology).
> 
> this is indeed quite cool. But I would still prefer a integration in nix/nixpkgs/nixos.
> 
> On the repology page I can't see more than 200 packages at once (so creating a complete list seems not possible?). And I'm not entirely sure which packages I really use.

This information is already in elasticsearch backend of search.nixos.org. We just need to find the right way to show it.

Lassulus:
> This information is already in elasticsearch backend of search.nixos.org. We just need to find the right way to show it.

This also sounds lika a good idea, but would be orthogonal to the "feature" proposed in this RFC.

garbas:
> When evaluating a system (for example with nixos-rebuild) unmaintained packages should show up, so people are encouraged to maintain them.

Wouldn't this be too verbose? I think that exposing unmaintained package is going to help people maintaining them, but displaying them would mostly annoy users.

Ma27:
> But that might also rub new users the wrong way: Oh, "half" of NixOS is unmaintained. I wonder what would be a good way to do this.

Just a random, probably not so well-thought, idea during lunch: what about something like `nix why-depends`, but for maintainers? This could e.g. be embedded into `nixos-rebuild`, so that interested people can see where help is needed when they run this dedicated command.

Lassulus:
> Wouldn't this be too verbose? I think that exposing unmaintained package is going to help people maintaining them, but displaying them would mostly annoy users.
> 
> Maybe a middle ground would be to show a summary. eg. Your system is using X unmaintained packaged. We're looking for volunteers to maintain them. Follow this link https://....

The important part would be to have a list of which packages I use are unmaintained. If I get a number and a list of not maintained packages I don't know how I'm afffected.


I would be happy to know if a package I use is unmaintained. I want to have an option to disable this, so if people get annoyed by it, could disable it with this option. We could of course also make this opt-in instead of opt-out. But I guess most people don't know how many packages in their system are maintained and will never find out if not shown to them.

Yes, showing people, that "half" of NixOS is unmaintained is exactly the purpose of this. Hiding it would not be a good solution IMHO :D

FRidh:
> Just a random, probably not so well-thought, idea during lunch: what about something like nix why-depends, but for maintainers? This could e.g. be embedded into nixos-rebuild, so that interested people can see where help is needed when they run this dedicated command.

With flakes it becomes more straightforward to do because you know of a derivation what inputs are used. Right now you could do, in case of your system, `nix show-derivation -r $(readlink -f /run/current-system)` to get the build-time closure of your system, however, maintainers are not part of the derivation and `show-derivation` shows a path, not an attribute, so you need to match paths with an evaluated Nixpkgs.

Anyway, users should not be bothered with these kinds of internal things; a user just wants to use the software. Whenever you use free and open source software you should be aware that things may not be maintained or can be unmaintained from one moment to the next, and that you're on your own. If you feel you only want to use software that is maintained, then use an evaluation test.

Lassulus:
> Anyway, users should not be bothered with these kinds of internal things; a user just wants to use the software. Whenever you use free and open source software you should be aware that things may not be maintained or can be unmaintained from one moment to the next, and that you're on your own. If you feel you only want to use software that is maintained, then use an evaluation test.

If users are to be expected, that there is unmaintained software, why not show it to them?
I think a warning like `unmaintained packages used: bzip2, curl, firefox, ....` would not be very annoying. But maybe we can disable the warning on stable systems by default. Fresh users probably use a stable channel anyway.

I also think a warning like this would encourage users to adopt unmaintained packages (at least I would). Maybe we can link to some introductory material of how to become a package maintainer or something along those lines.

Zimmi48:
It is my impression that the `meta.maintainers` field does not really reflect any "reality". What does it mean for a package to be maintained? In this RFC, it seems that you define it as "has a maintainer". But if the maintainer hasn't been active (hasn't updated or reviewed updates) recently, then is the package still actually maintained? And if there is no listed maintainer, does it really mean that the package is not maintained?

IMHO this RFC is incomplete as is and should consider the policy to implement on having this field reflect the "reality".

Lassulus:
> It is my impression that the `meta.maintainers` field does not really reflect any "reality". What does it mean for a package to be maintained? In this RFC, it seems that you define it as "has a maintainer". But if the maintainer hasn't been active (hasn't updated or reviewed updates) recently, then is the package still actually maintained? And if there is no listed maintainer, does it really mean that the package is not maintained?

Yes, I suggest here, that a package is maintained when is has an entry in the maintainers field. I don't want to talk about how to update this field here by other tools or manually (that would be another RFC IMHO). I believe the reality and the maintainers field will align when the effects of having people there or not are better visible, like with a warning during evaluation.

Your questions are good questions, but I don't want to answer them in this RFC, because I fear we will go down a rabbit hole from which we will not come back.

Zimmi48:
> I don't want to talk about how to update this field here by other tools or manually (that would be another RFC IMHO). I believe the reality and the maintainers field will align when the effects of having people there or not are better visible, like with a warning during evaluation.

I'm fine with this being a separate RFC, but I don't share your belief of a magical alignment once this field becomes more visible. In particular, your proposal *might* make more maintained packages get an official maintainer, but it definitely won't help clearing the inactive maintainers that are currently listed.

>Your questions are good questions, but I don't want to answer them in this RFC, because I fear we will go down a rabbit hole from which we will not come back.

I share your concern, but I also think that addressing those questions (possibly as part of another RFC) is a prerequisite for this one.

Lassulus:
> I share your concern, but I also think that addressing those questions (possibly as part of another RFC) is a prerequisite for this one.

I don't think they are so entangled. This RFC is about transparency and showing our current status to the users/maintainers/interested parties. The current state is the maintainers field. Letting maintainers decay is another topic which is also important but not required for showing this information.

also:

I nominiate @Kloenk as a shepherd
Does anyone else want to participate? please speak up :)

garbas:
I'd like to nominate myself for shepherding.

Infinisil:
Proof-of-concept NixOS module using https://github.com/NixOS/nixpkgs/pull/97023, which searches for unmaintained packages in `environment.systemPackages` (so no transitive dependencies, which is probably for the better)

```nix
{ config, ... }: {
_module.checks.unmaintainedPackages =
  let unmaintained = lib.filter (v: v ? pname && v ? meta.maintainers && v.meta.maintainers == []) config.environment.systemPackages;
  in {
    check = unmaintained == [];
    type = "warning";
    message = "These packages you're using don't have a maintainer: ${lib.concatMapStrings (v: "\n- ${v.pname}") unmaintained}";
  };
}
```

Since some maintainer-less packages are in the default environment.systemPackages, with an empty config this shows as
```
❯ nix-instantiate '<nixpkgs/nixos>' --arg configuration '{ fileSystems."/".device = "/"; boot.loader.grub.device = "/"; }' -A system
trace: warning: [unmaintainedPackages] These packages you're using don't have a maintainer: 
- bzip2
- gnused
- gnutar
```

Can be disabled by the user with
```nix
{
  _module.checks.unmaintainedPackages.enable = false;
}
```

spacekookie:
@Lassulus @Zimmi48 would either of you want to Shepherd this RFC?

Zimmi48:
Sorry, I don't have enough available time at the moment to be part of a shepherd team.

Mic92:
@Ma27 @mweinelt @Infinisil Does any of you want to be shepherd on this RFC?

Ma27:
Unfortunately not, but this is only due to a time limitation on my end.

Infinisil:
I can be shepherd

Mic92:
Ok. We have enough shepherds than: @Infinisil @garbas @Kloenk

Lassulus:
alright, since we now have a full team, lets find a date https://www.when2meet.com/?10788312-2saXp

Lassulus:
@garbas can you chose a date? :)

garbas:
@Lassulus done

Lassulus:
alright, so sunday, the 31. from 8PM to 9PM CET. We can meet on my jitsi instance: https://jitsi.lassul.us/rfc

Lassulus:
@garbas @Infinisil @Kloenk @Mic92 In around 2h the meeting is happening :)

Lassulus:
so, quick change, we will meet in mumble instead, so just run: `nix-shell -p mumble --run "mumble mumble://$USER@lassul.us"`

Lassulus:
# Meeting Notes 2021-01-31

- all agreed to have this information
    - if it does not slow down system build too much
    - if it does not increase ram usage too much
    - if it does not scare away potential maintainers/users
- potential other uses:
    - show licences used (warn on specific ones?)

show summary/percentage during nixos-rebuild. Probably at the bottom so people see it. With a command how to dig deeper. example:
`nix-metadata-analyze --maintainers /run/current-system`

Messaging should be very nice, like have a stallman for 100% GNU

## Implementation approaches

### Metadata for Nix derivations

Add special `derivationMeta` attribute to derivations, which is treated specially by Nix: The value of this attribute gets stored in the Nix database in association to the derivation it's part of, but without influencing the derivation hash.

This then allows querying this information after derivations are instantiated. A tool like

```
nix-transitive-meta /nix/store/<hash>-...
```

could be implemented.

### Eval-time output aggregation primop

Add a function like `builtins.aggregate` which can aggregate information during evaluation, outputting some folded result in the end. Something similar has been suggested in https://github.com/NixOS/nix/issues/1197

### External database for nixpkgs meta information

Have a central database populated by hydra containing all the meta information of nixpkgs packages. Identify packages based on either attribute path or uuids.

Uuids: Assign a uuid to every nixpkgs package, so that we can identify them. Like a derivation name, but without conflicts. This can then be used to match your system packages to ones in the central database.

Such a central database could also be used for other things:
- Storing the platforms the package builds on (see https://logs.nix.samueldr.com/nixos-dev/2021-01-12#4463655;)
- Tracking popularity of packages
- Licenses
- Maintainer mappings themselves

### Add a meta check for maintainership

Like how `meta.knownVulnerabilities` throws an error when used transitively, a similar check for `meta.maintainers` could show a warning when it's empty. The main problem with this is that we can't limit the number of warnings that are shown. Also, most derivations outside of nixpkgs never have any maintainers set, and would therefore also give a warning unless somehow detected.

### This Prototype
`nix eval --json '(with import <nixpkgs/nixos> {}; pkgs.lib.unique (builtins.map (p: p.name) (builtins.filter (p: (p.meta.maintainers or []) == [] && pkgs.lib.hasPrefix (toString (pkgs.path + "/pkgs")) p.meta.position) config.environment.systemPackages)))' | less`

# Task until the next meeting

- implement a prototype
- get ideas for shown messages

petabyteboy:
Implementation prototype in pure Nix, based on maintainers/scripts/find-tarballs.nix:

File find-unmaintained.nix
```
# This expression returns a list of all unmaintained packages used by ‘expr’.

with import <nixpkgs> {};
with lib;

{ expr, long ? false }:

let

  root = expr;

  message = if long then concatMapStrings (drv: ''
    Package ${drv.name} is looking for new maintainers${optionalString (drv ? meta) (let
      pos = builtins.unsafeGetAttrPos "maintainers" drv.meta;
    in " (at ${pos.file}:${toString pos.line})"
    )}
  '') uniqueUnmaintainedDependencies else ''
    ${toString (builtins.length uniqueUnmaintainedDependencies)} packages are looking for a maintainer
    Do [...] to list them all.
  '';

  uniqueUnmaintainedDependencies = map (x: x.file) (genericClosure {
    startSet = map (file: { key = file.name; inherit file; }) unmaintainedDependencies;
    operator = const [ ];
  });

  unmaintainedDependencies =
    filter (drv: ((drv.meta or {}).maintainers or null) == []) dependencies;

  dependencies = map (x: x.value) (genericClosure {
    startSet = map keyDrv (derivationsIn' root);
    operator = { key, value }: map keyDrv (immediateDependenciesOf value);
  });

  derivationsIn' = x:
    if !canEval x then []
    else if isDerivation x then optional (canEval x.drvPath) x
    else if isList x then concatLists (map derivationsIn' x)
    else if isAttrs x then concatLists (mapAttrsToList (n: v: addErrorContext "while finding unmaintained dependencies in '${n}':" (derivationsIn' v)) x)
    else [ ];

  keyDrv = drv: if canEval drv.drvPath then { key = drv.drvPath; value = drv; } else { };

  immediateDependenciesOf = drv:
    concatLists (mapAttrsToList (n: v: derivationsIn v) (removeAttrs drv ["passthru"]));

  derivationsIn = x:
    if !canEval x then []
    else if isDerivation x then optional (canEval x.drvPath) x
    else if isList x then concatLists (map derivationsIn x)
    else [ ];

  canEval = val: (builtins.tryEval val).success;

in message

```

```
[pbb@regolith:~/proj/nixpkgs]$ nix eval --raw "(with import ./. {}; import ./maintainers/scripts/find-tarballs.nix { expr = firefox; })" 2>/dev/null
34 packages are looking for a maintainer
Do [...] to list them all.

[pbb@regolith:~/proj/nixpkgs]$ nix eval --raw "(with import ./. {}; import ./maintainers/scripts/find-tarballs.nix { expr = firefox; long = true; })" 2>/dev/null
Package bzip2-1.0.6.0.1 is looking for new maintainers (at /home/pbb/proj/nixpkgs/pkgs/tools/compression/bzip2/default.nix:51)
Package gawk-5.1.0 is looking for new maintainers (at /home/pbb/proj/nixpkgs/pkgs/tools/text/gawk/default.nix:81)
Package gnused-4.8 is looking for new maintainers (at /home/pbb/proj/nixpkgs/pkgs/tools/text/gnused/default.nix:36)
Package gnutar-1.32 is looking for new maintainers (at /home/pbb/proj/nixpkgs/pkgs/tools/archivers/gnutar/default.nix:69)
Package patch-2.7.6 is looking for new maintainers (at /home/pbb/proj/nixpkgs/pkgs/tools/text/gnupatch/default.nix:55)
[...]

[pbb@regolith:~/proj/nixpkgs]$

```

Mic92:
Using the above script for NixOS with flakes:

```
$ nix eval --impure --raw --expr "(with import <nixpkgs> {}; import ./find-unmaintained.nix { expr = (builtins.getFlake (toString /etc/nixos)).outputs.nixosConfigurations.$(hostname).config.system.build.toplevel; long = true; })"
```

and without:

nix eval --expr "(with import <nixpkgs> {}; import ./find-unmaintained.nix { expr = (import <nixpkgs/nixos>).config.system.build.toplevel; long = true; })"

Mic92:
Can we have another meeting for this one?

Kloenk:
Best for me would be after the fifth of march. As then School is much less for me.

Lassulus:
lets plan another date: when2meet.com/?11230345-jwDak

Lassulus:
alright, so we will meet 2021-03-14 19:00 UTC (which is 20:00 in Europe/Berlin timezone). I guess we will do it on jitsi again.

Lassulus:
meeting link: https://jitsi.lassul.us/rfc81

Mic92:
```
# RFC 81 Meeting #2 Notes

## Opt-in or not

Under the assumption that this feature is expensive, we should make it opt-in by default.

In case this feature can be implemented more efficiently in the future, we can change this to opt-out.

Say max 1% eval time increase if opt-out.

An opt-in prototype could be implemented and included into nixpkgs without this RFC beeing done/accepted

## Limitations of current prototype

The current prototype is incapabale of finding packages which are "hidden" inside a string. example "${pkgs.firefox}/bin/firefox"

However it can find all buildInputs, and co. which should be good enough as a start.

## Interface proposals

### Opt-in

```
nixos-rebuild --show-unmaintained
```

Additionally we could inform the user about this flag at the end of a normal nixos-rebuild, like:

```
Did you know:
You can pass --show-unmaintained to nixos-rebuild 
to show unmaintained packages used in your system
```

This could be one of many "Did you know"s

### Opt-out

At the end of nixos-rebuild, display a count of unmaintaned packages like
```
Your system depends on 5 unmaintained packages
Run `nixos-rebuild rebuild --show-unmaintained` to see the full list
```

## Implement it by propagated this info in stdenv

stdenv can fill out a `$out/nix-support/unmaintained` which contains all the buildInputs and co. that don't have a maintainer, plus the `$out/nix-support/unmaintained` of them (making it recursive). This could be very efficient.

## Future work: CI checks for maintainership

Since the goal of this RFC is to have fewer unmaintained packages, in the future we could have a CI check that ensures all packages have a maintainer. Packages would then have to be removed when the last maintained removes themselves.
```

We can have the next meeting if we have a nixos-rebuild prototype as a PR.

Infinisil:
I discussed and worked a bit on this with @garbas today. Some points that were brought up:
- If we can have it be opt-out, we should only do so for people not using the stable channel. Reason being that stable could be outdated by many months, giving outdated information on which packages are maintained. In addition, people on unstable are more likely to step up to maintain packages.
- We should maybe show the date/commit of the maintainer state that is used in the output, so we can know how up-to-date it is
- With the search.nixos.org overhaul, backed by elasticsearch, it's now possible to do a single curl query to get up-to-date information on which packages don't have any maintainers, assuming you have an attribute path to search for

@garbas then created a simple example of how such a search.nixos.org query can be done. With a `query.json` file with contents
```json
{
  "from": 0,
  "size": 10000,
  "sort": [
    {
      "_score": "desc",
      "package_attr_name": "desc",
      "package_pversion": "desc"
    }
  ],
  "query": {
    "bool": {
      "filter": [
        {
          "term": {
            "type": {
              "value": "package",
              "_name": "filter_packages"
            }
          }
        }
      ],
      "must": [
        {
          "terms": {
            "package_attr_name": [
              "hello",
              "haskellPackages.hello",
              "haskellPackages.gtk2hs-hello"
            ]
          }
        },
        {
          "term": {
            "package_maintainers_set": {
              "value": "No maintainers"
            }
          }
        }
      ]
    }
  }
}
```

You can do a query for the given attribute paths with the following (the authorization key is public, don't worry)
```bash
curl 'https://nixos-search-5886075189.us-east-1.bonsaisearch.net/latest-19-20.09/_search' \
  -H 'Authorization: Basic ejNaRko2eTJtUjpkczhDRXZBTFBmOXB1aTdYRw==' \
  -H 'Content-Type: application/json' \
  -H 'Pragma: no-cache' \
  -H 'Cache-Control: no-cache' \
  --data @query.json
```

Meanwhile, I worked on a prototype for the idea of having an `$out/nix-support/unmaintained` be created by stdenv. However I implemented it a bit differently: To have a `$out/nix-support/attribute-deps` which contains a list of attribute paths that the derivation depends on (recursive build time deps, not runtime deps, for now). I implemented this with the commits [857efbc2](https://github.com/infinisil/nixpkgs/commit/857efbc2164e011d32f535c3bb81450c0835e8ac) and [d33a27b2](https://github.com/infinisil/nixpkgs/commit/d33a27b272a68752e67b54b2f09912d4786ae480). It's a bit messy with the whole splicing part, but it basically works (requires stdenv rebuild of course):
```bash
$ nix-build -A mbuffer
/nix/store/n3kx0839798la0k8sakvzwravb8grhsi-mbuffer-20200929
$ cat result/nix-support/attribute-deps | sort -u
binutils
pkgsHostTarget.openssl
pkgsHostTarget.perl
```

Once you have this information, it's very easy to query maintainer information for those packages. Doing it this way is very flexible, as you can use this list of attribute paths for not only maintainer information, but also e.g. for licenses or other things. It also allows integration with the above search.nixos.org query, therefore getting more up-to-date information than the user might have in their nixpkgs version.

However, the code is also easily adjustable to query maintainer information in the same evaluation.

Infinisil:
A tiny problem with this approach is that attribute changes can now influence derivations. So e.g. renaming `pkgs.foo` to `pkgs.bar` requires rebuilding all reverse-dependencies. Not sure if that's actually a problem, but something to keep in mind.

Here's another example:
```
```

7c6f434c:
>A tiny problem with this approach is that attribute changes can now influence derivations. So e.g. renaming `pkgs.foo` to `pkgs.bar` requires rebuilding all reverse-dependencies. Not sure if that's actually a problem, but something to keep in mind.

Does that mean that `linux_latest` and the corresponding versioned attribute would become different attributes? Same for LibreOffice picked by default and explicitly specified as Fresh/Still? That does sound undesirable.

Infinisil:
Oh yeah you're right, and that seems like a dealbreaker:
```
$ nix-instantiate --eval -A linux_latest.__path
[ "linux_latest" ]
$ nix-instantiate --eval -A linux_5_11.__path
[ "linux_5_11" ]
```

7c6f434c:
>Oh yeah you're right, and that seems like a dealbreaker:

Does that code need to be applied globally? Because if it can simply be enabled for a wrapper (so that user can apply it whenever) and maybe default-on for system instances, the rebuilds incurred are dirt cheap, and the user can easily have this information over the used environments (which is presumably the most interesting case, why go package-by-package when you have a summary for your system and your user profile and your project shell).

Mic92:
Closing for now until @Lassulus has time to work on a prototype.

PR #82, opened 2020-12-18
[RFC 0082] lib.experimental

fare:
I propose that we create a directory lib/experimental for experimental code,
also available as lib.experimental and pkgs.lib.experimental. Such code is
available for limited use within select parts of nixpkgs, with the explicit
opt-in approval of their respective maintainers, but not for widespread use
in other parts of nixpkgs.

### DIFF
--- a/rfcs/0082-lib-experimental.md
+++ b/rfcs/0082-lib-experimental.md
@@ -1,0 +1,129 @@
+---
+feature: lib.experimental
+start-date: 2020-12-17
+author: Francois-Rene Rideau
+co-authors: N/A
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+I propose that we create a directory lib/experimental for experimental code,
+also available as lib.experimental and pkgs.lib.experimental. Such code is
+available for limited use within select parts of nixpkgs, with the explicit
+opt-in approval of their respective maintainers, but not for widespread use
+in other parts of nixpkgs.
+
+# Motivation
+[motivation]: #motivation
+
+Some libraries we *must* use
+mandatorily unless it's impossible or at least unreasonable not to.
+Some libraries we *should* use
+unless there's a good reason not to.
+Some libraries we *may* use
+according to each contributors' preferences.
+Some libraries we *may not* use
+without an explicit endorsement by enough maintainers of the relevant packages.
+Some libraries we *shouldn't* use
+without a strong reason to do it.
+Some libraries we *mustn't* use
+and is only left temporarily for backward compatibility until complete migration off of it.
+
+I propose we create a directory `lib/experimental`
+to tag libraries as explicitly being part of the *may not* category.
+
+Creating this experimental section to nixpkgs makes it clearer to maintainers
+which code they may use where, and which code they may not.
+If you are the maintainer of a package, and understand the risks related to using
+experimental code, you may use this code or accept patches using it.
+But you may also freely reject it, and no one should impose you that code without your consent.
+In particular you understand that the code and its API may change, and
+you may have to update your code accordingly, or accept patches that do.
+Thus you agree to collaborate with the author of the experimental code
+while the code remains in the experimental section.
+
+Code that should *not* go into `lib.experimental` but directly into `lib` include:
+* code that is good quality, well tested, known to work, with a stable API,
+  good documentation, and no controversy among maintainers of nixpkgs.
+
+Code that should *not* go into `lib.experimental` and not into nixpkgs at all include:
+* code that bad quality, untested, not working, etc.
+* code that is not yet used by any package in nixpkgs nor by any extension to nixpkgs.
+* code that while cool has no purpose in nixpkgs as such and can well live outside it.
+
+Making these social constraints explicit avoid accidents and conflict that would happen
+if they were left implicit.
+
+# Detailed design
+[design]: #detailed-design
+
+Experimental library code should be placed under the `lib/experimental` directory,
+where it will be made available in the `pkgs.lib.experimental` attrset,
+that will *not* re-export "direct" versions of the code directly in a parent attrset.
+
+A README.md will explain the purpose of this directory, by quoting this RFC and linking to it.
+
+The process for taking things out of experimental is when there is a consensus to do so:
+when enough maintainers use this code rather than alternatives or workarounds,
+when its API is stable and well documented, with enough years of good use,
+when multiple maintainers are ready to service the code and it's not controversial anymore,
+and/or when a RFC blesses the code to be made a direct part of lib.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+The [Gerbil](https://cons.io) packages in the
+http://github.com/muknio/nixpkgs/archive/devel.tar.gz fork of nixpkgs use a
+prototype object system called `POP` as a more versatile and principled alternative
+to the traditional extension mechanisms in `lib.fixedPoints` or `lib.customisation`.
+The libraries or use to good effects in particular by
+[Glow](http://gitlab.com/mukn/glow) for local overrides.
+
+Under this RFC, the code for POP would go in `lib/experimental/POP.nix`
+and be available as `pkgs.lib.POP`.
+Package maintainers may opt into using it like Gerbil does,
+but contributors may not use it without the consent
+of the maintainers of the code they contribute to.
+
+If and when maintainers have adopted the code, and it becomes prevalent enough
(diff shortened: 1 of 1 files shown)


FRidh:
I suggest participating in https://github.com/NixOS/rfcs/pull/83 which is about creating a common interface for sub package sets, of which `gerbil-packages` is one.

rfcs/0082-lib-experimental.md
```diff
@@ -0,0 +1,129 @@
...
+
+Making these social constraints explicit avoid accidents and conflict that would happen
+if they were left implicit.
+
+# Detailed design
+[design]: #detailed-design
+
+Experimental library code should be placed under the `lib/experimental` directory,
+where it will be made available in the `pkgs.lib.experimental` attrset,
+that will *not* re-export "direct" versions of the code directly in a parent attrset.
```
roberth, line 66: This placement will not allow individual experimental functions to be added to existing lib files, even if they should reuse let bindings from those files. File-wide `let` definitions aren't used as much as they probably should though.

rfcs/0082-lib-experimental.md
```diff
@@ -0,0 +1,129 @@
...
+
+A README.md will explain the purpose of this directory, by quoting this RFC and linking to it.
+
+The process for taking things out of experimental is when there is a consensus to do so:
+when enough maintainers use this code rather than alternatives or workarounds,
+when its API is stable and well documented, with enough years of good use,
```
roberth, line 72: Ideally, all functions are documented in the Nixpkgs manual. Not all of them are, which is a separate issue, but well documented seems to imply that it's in the manual.
Otherwise, I don't see how they can get years of good use. Years does seem a bit inflexible. I'd hope we'll be able to promote functions sooner than 2 years.

rfcs/0082-lib-experimental.md
```diff
@@ -0,0 +1,129 @@
...
+[motivation]: #motivation
+
+Some libraries we *must* use
+mandatorily unless it's impossible or at least unreasonable not to.
+Some libraries we *should* use
+unless there's a good reason not to.
+Some libraries we *may* use
+according to each contributors' preferences.
+Some libraries we *may not* use
+without an explicit endorsement by enough maintainers of the relevant packages.
```
roberth, line 30: Should I read this as "forbidden"?

Nixpkgs doesn't only define packages. Most of the interesting use cases for `lib` functions are in `nixos/` (calling into `lib.modules`, `lib.types` and various functions to support the configuration logic). Some of the functions primarily support users' own projects, or at least that's how I think of `lib.cleanSourceWith`.

Usually when you choose to rely on someone's open source function, you accept that you may have to help improve it when the need arises. This is a mutually beneficial situation. You can make use of whatever beneficial qualities that code already has and when you need an improvement, the others will benefit as well.

When you decrease the appeal of certain functions, they will receive less use, and therefore less attention.
In my view, use of Nixpkgs is already quite pragmatic, in the sense that contributions arise from users' specific needs. So my questions is, what is the cost of keeping users away?

rfcs/0082-lib-experimental.md
```diff
@@ -0,0 +1,129 @@
...
+* code that is good quality, well tested, known to work, with a stable API,
+  good documentation, and no controversy among maintainers of nixpkgs.
+
+Code that should *not* go into `lib.experimental` and not into nixpkgs at all include:
+* code that bad quality, untested, not working, etc.
+* code that is not yet used by any package in nixpkgs nor by any extension to nixpkgs.
+* code that while cool has no purpose in nixpkgs as such and can well live outside it.
+
+Making these social constraints explicit avoid accidents and conflict that would happen
+if they were left implicit.
```
roberth, line 59: What is the cost of these accidents and conflict. Could you provide real-world examples?

Also note that small accidents and conflicts (human or otherwise) are part of how some software evolves.

roberth (commented):
I appreciate your intent to streamline Nixpkgs development in relation to `lib`, but I am concerned about the cost of the added process and the invisible cost of missed contributions and collaboration. Perhaps you could provide references for the conflicts or accidents that you're trying to resolve? I am not aware of any, but if they exist in sufficient numbers, that may justify it.

edolstra:

lheckemann:
@fare do you plan to continue with this RFC? If not, we would suggest closing it with the possibility of someone (else?) picking it up again later.

fare:
Yes, I'd like to see it through.

Alternative much simpler proposal: things in lib just get marked `### EXPERIMENTAL!` at the top of the file and their contents don't get merged at toplevel.

rfcs/0082-lib-experimental.md
fare: I mean that, assuming my pop library gets in, and someone wants to use it while sending a patch to some recipe that uses a different extension mechanism, I expect the recipe's owner to reject it and request the patch author to use the previously used extension mechanism instead. Alternatively, if daring, he'd ask the patch author to rewrite everything using pop.

rfcs/0082-lib-experimental.md
fare: Not sure about real-world examples in nixpkgs... maybe this is all overkill, but I was told not to make my pop.nix library part of the regular `lib/` for now, so I wanted a way to get it in without it being "regular".

FRidh (commented):
The `Motivation` lacks some rules as to what should *not* go in `lib.experimental`, and the `Detailed design` states some rules on when code from `lib.experimental` can be considered "stable", however, I am missing when we consider code to be a sufficiently good additional that it can go in `lib.experimental`. The `Motivation` states when its not badly written, unused or not fitting. But what if similar kind of code exists? Clearly a motivation needs to be given in the PR (as in every PR) and I think there still also needs to be ownership over the whole of `lib.experimental`, not just owners of the specific pieces.

Mic92:
To move this to the next phase we would need shepherds. Anyone open to?

Mic92:
If there are no shepherds for this RFC within two weeks, the rfc steering committee team will close this RFC.

lheckemann:
@fare or anyone else: if you wish to continue work on this RFC, please feel free to nominate shepherds and request a reopen :)

roberth:
I nominate myself as shepherd.

Any other takers? Perhaps @sternenseemann?

rfcs/0082-lib-experimental.md
```diff
@@ -0,0 +1,129 @@
...
+prototype object system called `POP` as a more versatile and principled alternative
+to the traditional extension mechanisms in `lib.fixedPoints` or `lib.customisation`.
+The libraries or use to good effects in particular by
+[Glow](http://gitlab.com/mukn/glow) for local overrides.
+
+Under this RFC, the code for POP would go in `lib/experimental/POP.nix`
+and be available as `pkgs.lib.POP`.
+Package maintainers may opt into using it like Gerbil does,
+but contributors may not use it without the consent
+of the maintainers of the code they contribute to.
```
sternenseemann, line 90: Something we should look into as well is how to enforce this. There's probably a nice way to do so using `scopedImport`.

sternenseemann:
> Any other takers? Perhaps @sternenseemann?

Sure.

sternenseemann:
Given the recent [row about flakes in `lib`](https://github.com/NixOS/nixpkgs/pull/173093), it'd be great if we could revive this RFC and find the third required sheperd: Then it'd be possible to have a flakes utility library in nixpkgs as many have wanted, but have the benefit that the ability to add experimental code to nixpkgs isn't exclusive to flake related functions.

edolstra:
The whole point of flakes is that experimental, controversial or hacky stuff doesn't need to go into Nixpkgs but can live in its own flake. A function like `callLocklessFlake` seems like a bit of a workaround for the current lack of builtin support for subflakes, but that's not something we want to be stuck with in the standard library forever.

fare:
Are flakes now considered production quality? If I have packages in main NixOS and they use an experimental library, should I remove all those packages from NixOS and create a flake for them? Create that library in an absurd place close to some of my packages rather than in lib/experimental?

roberth:
It's not feasible for `nixpkgs` to depend on flakes, certainly for a long time, so if we truly want to innovate in the beating heart of the ecosystem, we'd better not tell people to go off and do it elsewhere.

> The whole point of flakes is that experimental, controversial or hacky stuff doesn't need to go into Nixpkgs but can live in its own flake.

Is that really the whole point? If so, that only stresses the fact that Nixpkgs will remain at the center of the ecosystem and that we need to care for it. Caring includes having a means to deal with non-trivial changes!

Flakes will do more than just that, I know, but that doesn't diminish Nixpkgs' unique position as the ecosystem monorepo or the advantages that this architecture brings.
However green the grass looks on the polyrepo side, let's focus on our monorepo in this RFC discussion.

lheckemann:
>Are flakes now considered production quality?

no, they're still guarded behind experimental-features and have no interface stability guarantees.

PR #80, opened 2020-11-29
[RFC 0080] Change NixOS releases to YY.05,YY.11

jonringer:
This is meant to help minimize risk and work during the release stabilization period, by selecting a better time to branch-off from unstable.

Adapted from https://github.com/NixOS/rfcs/commit/8a3ba49393455fcf426488aef8b6d27d428c640c

Related discussions: 
 - https://discourse.nixos.org/t/20-09-release-retrospective/9851
 - https://discourse.nixos.org/t/what-should-stable-nixos-prioritize/9646

cc @worldofpeace @garbas @FRidh 

rendered: https://github.com/NixOS/rfcs/blob/ca06bb8dbb245f6a4c00311967e37753cd3d691a/rfcs/0080-nixos-release-schedule.md

PR #85, opened 2021-01-26
[RFC 0085] NixOS Release Stabilization: ZHF on master, new timeline

jonringer:
In similar spirit to #80, this RFC tries to change the nature in which we conduct the stabilization of a NixOS release. Without Desktop managers being a major risk (due to #80), the release should be free to focus on stabilizing package builds. This should enable us to perform ZHF mostly on master with minor adjustments to staging and a change to policies regarding certain "critical" or "high-risk" packages.

Overall, the goal of this RFC to reduce the pain associated with ZHF (release stabilization) at a slight expense to the staging branch workflow during the stabilization window.

I eluded to this RFC in https://github.com/NixOS/rfcs/pull/80#issuecomment-735366619

~~"Freeze" may be a little strong, as non-breaking changes are allowed. But I didn't want to use a term like "half-frozen" or "semifrozen"~~ "Restricted" has been chosen as the term to describe disallowing breaking change updates.

Related discussions:
    - https://discourse.nixos.org/t/20-09-release-retrospective/9851
    - https://discourse.nixos.org/t/what-should-stable-nixos-prioritize/9646

cc @worldofpeace @FRidh  @ryantm 

rendered: https://github.com/jonringer/rfcs/blob/85-ZHF-on-master/rfcs/0085-nixos-release-stablization.md


jonringer:
Added defintions section, and improved timeline clarity.

Updated new render

rfcs/0084-ZHF-on-master-and-freeze-dates.md
```diff
@@ -0,0 +1,104 @@
...
+
+Release team will retain the right to refine what constitutes "critical packages", as this may
+change over time, and will likely not warrant an RFC to caputre changes.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Normal staging development is disrupted for two weeks. (1 normal iteration)
+Changes to "critical packages" may have to wait for a month to be merged into staging.
+Master development will be uninterrupted, assuming all changes don't introduce new regressions.
```
Ma27, line 88: I think this assumption is too optimistic. However it's IMHO not such a big deal if regressions don't stem from critical packages (such as the ones you've listed above).

rfcs/0084-ZHF-on-master-and-freeze-dates.md
```diff
@@ -0,0 +1,104 @@
...
+| -1 Weeks | Merge second staging-next fix cycle |
+| -1 Weeks | Perform Branch-off, create release channels, create new beta / unstable tags |
+| -1 Weeks | ZHF transitions to "backporting" workflow |
+| -1 Weeks | Prepare for release, finish remaining issues |
+| 0 Weeks | Release! |
+| 0 Weeks | ZHF Ends |
+
+Actual release schedule may adjust dates slightly due to unforeseen events.
+
+Release team will retain the right to refine what constitutes "critical packages", as this may
```
Ma27, line 80: What does "Release team" mean? I know that a few people (including you IIRC) plan to "define" such a team rather than having two RMs with most of the workload, but I'm not sure if it's clear who's part of this at the moment.

rfcs/0084-ZHF-on-master-and-freeze-dates.md
```diff
@@ -0,0 +1,104 @@
...
+
+- "Freezing" a branch or package.
+  - Disallow breaking changes to branch or package. In SemVer terminology, disallow major version bumps.
+  - A release branch can be thought of being permanently "frozen" in this regard. Stable may be a better term, but it's already overloaded with a previous release branch or channel.
+- "Unfreezing" a branch or package.
+  - Allow for breaking changes to be made to branch or package.
+- "Critical packages"
+  - Important packages which have many dimensions of build or runtime behavior.
+  - Generally these will be packages which are a part of `stdenv`, or bootstrapping a system (e.g. systemd).
+  - Initially these will be: `stdenv.cc`,`binutils`, and `systemd` for this RFC.
```
Ma27, line 30: (nit) I guess at least `glibc` should be listed here as well.

rfcs/0084-ZHF-on-master-and-freeze-dates.md
```diff
@@ -0,0 +1,104 @@
...
+staging. In addition, I also propose to stabilize the release on master
+instead of the release branch, thus avoiding the need to backport every pull
+request, and take advantage of the stabilization that unstable normally
+receives.
+
+# Definitions
+[definitions]: #definition
+
+- "Freezing" a branch or package.
+  - Disallow breaking changes to branch or package. In SemVer terminology, disallow major version bumps.
```
Ma27, line 23: I'm wondering if we should define an exception here: breaking changes can be allowed if those are necessary to fix security issues or fix a critical bug (i.e. something that corrupts your data or crashes your system).

Note: while backporting single patches is the preferred way in that situation, those can be "breaking" as well.

rfcs/0084-ZHF-on-master-and-freeze-dates.md
```diff
@@ -0,0 +1,104 @@
...
+  - A release branch can be thought of being permanently "frozen" in this regard. Stable may be a better term, but it's already overloaded with a previous release branch or channel.
```
Ma27, line 24: I'm not sure if we should define such a detail here: from my observation this isn't the status quo and IMHO also a fairly debatable, but out-of-scope topic.

Ma27 (commented):
I like the idea :+1:

rfcs/0084-ZHF-on-master-and-freeze-dates.md
FRidh: >  - Generally these will be packages which are a part of `stdenv`, or bootstrapping a system (e.g. systemd).

Separate building/bootstrapping stdenv from that of a (NixOS) system? I'm of the opinion all stdenv deps, including stdenv itself should freeze at this point.

rfcs/0084-ZHF-on-master-and-freeze-dates.md
jonringer: Currently it means the release managers.

rfcs/0084-ZHF-on-master-and-freeze-dates.md
jonringer: It was implied that this would include everything `stdenv.cc`, but I could make it more explicit.

```
nix-repl> stdenv.cc.libc.name
"glibc-2.32-25"

nix-repl> stdenv.cc.bintools.name
"binutils-wrapper-2.35.1"
```

rfcs/0084-ZHF-on-master-and-freeze-dates.md
jonringer: Yea, and that's the exception I make for the release channel. But only if patching isn't a reasonable alternative.

rfcs/0084-ZHF-on-master-and-freeze-dates.md
jonringer: This was more meant to help convey that freezing only applied to breaking changes. No expects the stable channel to receive 0 updates, but people do expect the stable channel to rarely cause upgrade issues.

rfcs/0084-ZHF-on-master-and-freeze-dates.md
jonringer: This is mostly from my observation that the week prior to the 20.09 scheduled release date saw ZHF PRs slow down significantly. The "easy to fix" packages, usually have quick turn around, and the more involved package fixes were specific to the 20.09 branch; so non of the more active contributors really wanted to "pick up" the work.

rfcs/0084-ZHF-on-master-and-freeze-dates.md
jonringer: I would like to eventually create a "release team", as many people have expressed interest in being a part of a release; but it's much bigger of a ask to be "the release manager".

rfcs/0084-ZHF-on-master-and-freeze-dates.md
jonringer: Also, in practice, the branches diverge quite a bit if a particular package (or package set) had a series of updates from branch-off. It was sometimes not obvious what series of changes allowed for unstable to work, but the beta branch to still be broken.

Ericson2314:
I have historically been skeptical of slowing down `master`/`staging`/`staging-next`. But I now think the painfulness of the release process might justify it. Perhaps this we need to get good at making the deadlines and not killing ourselves doing things this way, hopefully improve CI so there's less "fixing things after they break" in general, and then we re-earn the privilege of not needing to slow down `master`/`staging`/`staging-next`.

jonringer:
> hopefully improve CI so there's less "fixing things after they break" in general,

I agree with this in general, but there's also issues like what we encounter with plasma which will be really hard to construct. This was a big factor for me to push DM packaging so far from the release date, we really do need the time on unstable for those scenarios to be stabilized.

ryantm:
I nominate myself to shepherd.

garbas:
I nominate myself to shepherd as well.

Mic92:
I nominate myself to shepherd as well.

Mic92:
@ryantm do you want to be shepherd leader?

alyssais:
I was worried when I saw this RFC, because as an unstable user I don't
like the idea of unstable having to slow down to accomodate stable.  But
having read it, it seems like the accomodations being asked for are very
small and reasonable, so I don't really see a problem with it, as long
as the set of frozen packages stays very small and focused.  The set
described here in the RFC is good, and some change at the RMs'
discretion is entirely reasonable.

I do have a couple of questions though:

First (and maybe this is very naive of me), why can't we just revert
problematic updates after branch-off, like the problematic systemd bump
from last time?  This seems to me like it would allow master to continue
at full steam ahead, while allowing the release managers to go ahead
without any recently introduced problematic changes.

> Updates to `staging` which will land before branch-off will be
> restricted to non-breaking updates

It's not clear to me what this means.  Is a git bump (for example) a
breaking update or a non-breaking update?  Maybe this could be added to
the terminology section.

7c6f434c:
@alyssais 

> why can't we just revert problematic updates after branch-off

As someone who also does not use the releases, I would assume that reverting on-branch means that some ZHF fixes have to be different on the release branch compared to the mainline, and so you cannot run a unified ZHF sprint without a mess of some things happenning only on one branch or something like that.

jonringer:
> First (and maybe this is very naive of me), why can't we just revert
problematic updates after branch-off, like the problematic systemd bump
from last time?

It can be very hard to capture all the changes that are related to a given bump.
For staging updates especially, where there may be many leaf packages that
are broken, it's sometimes not obvious how to capture all of the changes. Just
reverting a bump is not usually sufficient. Also the size of most systemd update
PRs is pretty large, so reverting the 30-100 commits can also be very error proned
and tedious.

> It's not clear to me what this means.  Is a git bump (for example) a
breaking update or a non-breaking update?  Maybe this could be added to
the terminology section.

"Disallow breaking changes to branch or package. In SemVer terminology, disallow major version bumps."
If the change is `git: 2.30.0 -> 2.30.1` or `git: 2.31.0`, then it's fine. Just worried about
`git: 2.30 -> 3.0.0`, in which there's likely to be some breaking changes which
would need a longer stabilization time to full fix everything.

ryantm:
> @ryantm do you want to be shepherd leader?

Sure, I can be leader.

alyssais:
> > It's not clear to me what this means.  Is a git bump (for example) a
> breaking update or a non-breaking update?  Maybe this could be added to
> the terminology section.
>
> "Disallow breaking changes to branch or package. In SemVer terminology, disallow major version bumps."
> If the change is `git: 2.30.0 -> 2.30.1` or `git: 2.31.0`, then it's fine. Just worried about
> `git: 2.30 -> 3.0.0`, in which there's likely to be some breaking changes which
> would need a longer stabilization time to full fix everything.

Maybe "breaking change" should be a top-level "definitions" entry rather
than defined under "frozen". :)


One more thing: the RFC should probably define where the canonical
home for the list of frozen packages will be, since it won't be changed
by RFC.  I'd vote for the manual.

andir:
I am not entirely sold on the idea yet. If it solves the stable problems I am all for it. If it works in practice remains to be tested.

The one change that should probably be made to this RFC is remove any listing of packages outside of the `Movation` section. Those  should be soley defined in the specific instance of ZHF. We could make a different RFC specifying a common set of packages with reasoning for each of them but this should really just specify the concept of ZHF on master without going too much into the direction of which packages are actually "critical" and which are not.

Another worry I have is: Communication
How do we inform all the maintainers/commiters in the project about the ZHF and the current policies towards what they are allowed/not allowed to merge?
Other projects have a ML for all the developers where each and everyone is subscribed for annoucements. The best we have is discourse which is pretty much mixed content and noisy for such information.

jonringer:
> Maybe "breaking change" should be a top-level "definitions" entry rather than defined under "frozen". :)

The best term would have been stable, but as I also mentioned, that term is already overloaded in the nixpkgs as it generally refers to the current release branch, not the upcoming release. For ZHF, I will probably use the phrase `frozen for breaking changes` to help emphasize that minor changes and new packages are still welcome.

If someone can think of a better term, I will gladly change it.

> If it works in practice remains to be tested.

Agreed, lets test it.

>  remove any listing of packages outside of the Movation section. Those should be soley defined in the specific instance of ZHF.

I disagree. Sure, not every "critical" package may be applicable to a release. But I would also like to avoid the situation of, "some people got in a room together and came up with this list without significant community involvement". I think a more appropriate time to revisit the list would be the release retrospective. This at least allows the community to participate, and major stakeholders can express their opinions. I tried to do [one for 20.09](https://discourse.nixos.org/t/20-09-release-retrospective/9851), which this RFC was one of the action items. Hopefully the next release will have more participation, and I still want to thank those who showed up.

I could also argue that creating an RFC to facilitate changes to the release process is only to involve the community more. However, just saying, "I think this would be better" without any input isn't a healthy way to do process changes. And this particular change should have community involvement since I would like to alter some merge criteria on unstable leading up to the release.

> We could make a different RFC specifying a common set of packages with reasoning for each of them but this should really just specify the concept of ZHF on master without going too much into the direction of which packages are actually "critical" and which are not.

I would like to avoid RFCs as much as possible. The process is slow (for good reason), and takes a fair amount of time and energy for the RFC committee to process. I think a retrospective would be a better format as the pain of the release would still likely be fresh, and certain packages would be easier to identify as being problematic. The list of critical packages can just be maintained in the [release wiki](https://github.com/NixOS/release-wiki).

> How do we inform all the maintainers/commiters in the project about the ZHF and the current policies towards what they are allowed/not allowed to merge?
Other projects have a ML for all the developers where each and everyone is subscribed for annoucements. The best we have is discourse which is pretty much mixed content and noisy for such information.

I mention in [future work](https://github.com/NixOS/rfcs/blob/66e401435eb6ca4805be40c3501a73389532709c/rfcs/0085-ZHF-on-master-and-freeze-dates.md#future-work) that I will be creating a discourse thread. And will likely ping someone to pin it for a while. I was probably also going to do a reddit post and discord announcement just referring to the discourse thread as well, but since reddit or discord isn't an official platform, I didn't mention them.

But NixOS communication problems aren't unique to this RFC, and slightly off-topic.

For enforce-ability,  I will be quite active (and hopefully the other RM, still TBD) and able to enforce it for most PRs. Most staging-targeted PRs take a while to review anyway, so I'm not too concerned that something will be merged while I was sleeping or away. I mentioned [here](https://github.com/NixOS/rfcs/pull/85#discussion_r564838630) that it's still somewhat of a desire to have a "release team", which would help distribute these responsibilities and own certain release issues.

rfcs/0085-ZHF-on-master-and-freeze-dates.md
```diff
@@ -0,0 +1,104 @@
...
+Master development will be uninterrupted, assuming all changes don't introduce new regressions.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Maintain the status quo regarding release dates
+
+Continue with cramming staging changes right before branch-off, keeping
+a significant amount of stabilization work and risk before a release, and likely
+continuing the trend of delayed releases.
```
Mic92, line 97:
```suggestion
continuing the trend of delayed releases.

## Better backport automation
A backport github action that can be would automatically create a backport of pull request. This would be also useful even if we decide to accept this pull request when a new NixOS version is released.
```

rfcs/0085-ZHF-on-master-and-freeze-dates.md
```diff
@@ -0,0 +1,104 @@
...
+
+# Summary
+[summary]: #summary
+
+To further minimize the amount of risk going into a NixOS release, I propose
+to set freeze dates regarding changes to critical packages being merged to
+staging. In addition, I also propose to stabilize the release on master
+instead of the release branch, thus avoiding the need to backport every pull
+request, and take advantage of the stabilization that unstable normally
+receives.
```
Mic92, line 17: This should also specify what happens with new packages/modules.

Mic92:
> > @ryantm do you want to be shepherd leader?
> 
> Sure, I can be leader.

@ryantm can you setup a [meeting](https://when2meet.com)

rfcs/0085-ZHF-on-master-and-freeze-dates.md
jonringer: Although I really like the idea of automated backports, I don't think it's too relevant to what the RFC is addressing. But if you want me to add it to the future work section I can, and that can be tackled at a later time.

rfcs/0085-ZHF-on-master-and-freeze-dates.md
jonringer: New packages are welcome. Changes to modules are welcome, assuming it doesn't cause regressions, and the module changes will probably need to have documentation so it's not additional work trying to curate the release notes.

I will modify the definitions section to say that "freezing" will still allow new package and module additions.

ryantm:
I've set up a meeting availability webpage:

https://www.when2meet.com/?11238994-7RwPh

@andir @jonringer @garbas @mic92 Can you please fill it out?

ryantm:
also cc @alyssais @Ericson2314 if you want to attend.

ryantm:
@andir @garbas @jonringer Are you able to meet? Can you fill out the meeting availability? https://www.when2meet.com/?11238994-7RwPh

Mic92:
@ryantm I think we need a new schedule. All dates are now in the past.

jonringer:
Just to put the timeline in context of this RFC:
Freeze for breaking changes in "critical" package will be in 3 weeks
Freeze for breaking changes in staging will be in 5 weeks
ZHF will begin in 7 weeks

jonringer:
@ryantm did some helpful rewording. If you found it difficult to understand, please try reading the RFC again :)

ryantm:
I believe the new version addresses all the comments of @Ma27, Mic92, @alyssais, and @andir. Please have a look!

https://github.com/jonringer/rfcs/blob/85-ZHF-on-master/rfcs/0085-nixos-release-stablization.md

In the interest of moving this along before it actually needs to come into affect:

I motion for the Final Comment Period (FCP) to begin with disposition to merge. @mic92 and @garbas please either approve or disapprove of the motion as soon as you can.

Mic92 (approved):

FRidh (approved):

rfcs/0085-nixos-release-stablization.md
```diff
@@ -0,0 +1,128 @@
...
+date](https://github.com/NixOS/nixpkgs/pull/101078).  This is not to
+discredit any individual, but demonstrate that certain packages have
+many dimensions of compatibility which need a longer time to stabilize
+on unstable before being included in a release.
+
+
+# Definitions
+[definitions]: #definition
+- Breaking change
+  - A changes that is likely to break downstream.
```
Ma27, line 52: `A change` I guess?

Ma27 (approved):

garbas (approved):

ryantm:
We are now in the Final Comment Period (FCP). Unless any major objections are raised, this will be accepted and merged in 10 days from now, on April 2nd!

rfcs/0085-nixos-release-stablization.md
jonringer: done. thanks :)

rfcs/0085-nixos-release-stablization.md
```diff
@@ -0,0 +1,128 @@
...
+## Release Critical Packages
+
+A list of Release Critical Packages will be maintained in the nixpkgs
+manual.  Release Critical Packages are important packages which have
+many dimensions of build or runtime behavior. Generally these will be
+packages which are a part of `stdenv`, or bootstrapping a system
+(e.g. systemd). The Release Managers are empowered to decide which
+packages are Release Critical Packages without additional RFCs.
+
+Initially these will be `stdenv.cc`,`binutils`, and `systemd`.
```
andir, line 75: Still not very happy about calling out any specific packages in here. The paragraph above is explicit enough IMHO.

rfcs/0085-nixos-release-stablization.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Restriction Timeline
+
+This timeline represents when certain branches and packages will be
```
andir, line 79: One thought: What are our escape hatches if the ZHF never completes? I am certain we will find a way if that happens just wanted to through the thought in here.

ryantm: Actually getting to zero Hydra failures is not a requirement for completing ZHF.

rfcs/0085-nixos-release-stablization.md
ryantm: Could you go into more detail about why you are opposed to this? I feel like it gives important context for future Release Managers when they decide what should be a Release Critical Package.

jonringer: Only three were mentioned, and each of them have significant build and runtime behavior which needs to be stabilized on master before branch-off.

stdenv also includes other packages such as `stdenv.shellPackage`, `stdenv.bootstrapTools`, and a few others. But I guess you could argue those are highly to cause stabilization pain if bumped as well. I'm open to changing the line to just read as
```suggestion
Initially these will be all packages included in `stdenv` and `systemd`.
```

rfcs/0085-nixos-release-stablization.md
jonringer: correct, generally it's a timeboxed effort, and one of the release tasks is to mark all remaining failing builds to be broken.

jonringer: There's even a script which helps aide in marking most packages as broken: https://github.com/NixOS/nixpkgs/pull/101812

ryantm (approved):

rfcs/0085-nixos-release-stablization.md
andir: A refactoring of nixpkgs or stdenv should not make this RFC obsolete. You can list those examples in an advisory section for release manager but including them in part of the "law" just ensures that it will be outdated at some point.

rfcs/0085-nixos-release-stablization.md
```diff
@@ -0,0 +1,128 @@
...
+discredit any individual, but demonstrate that certain packages have
+many dimensions of compatibility which need a longer time to stabilize
+on unstable before being included in a release.
+
+
+# Definitions
+[definitions]: #definition
+- Breaking change
+  - A change that is likely to break downstream.
+  - For projects following Semantic Versioning, disallow major version bumps.
```
andir, line 53: Do I read this right that it would still allow packages that have to be on a specific version to be useful (steam, spotify, slack, ...) because a remote server requires it are not affected by this rule even if they are major version upgrade? This scenario would probably qualify as bug fix and thus be fine?

rfcs/0085-nixos-release-stablization.md
Mic92: As an alternative should this just be linked to to the release maintainer wiki? https://github.com/NixOS/release-wiki

rfcs/0085-nixos-release-stablization.md
Mic92: I think we already have an unwritten policy for projects like discourse/skype/steam/spotify that allows backports to stable, even for major version to stay compatible with third parties.

andir: I am aware of that but now that we start introducing a written policy we might want to mention that packages that only work in combination with a remote service might be excluded.

jonringer: It's the defining the term. The "restricted" updates only apply to staging. So if the major version bump affects <100 (or 500 in stituations, e.g. python packages) rebuilds, then it's still free to be bumped. All PRs which target master will not be affected by any of these changes.

rfcs/0085-nixos-release-stablization.md
jonringer: I don't see systemd, gcc, binutils, or bootstrapping packages to be obsolete anytime soon.

I believe I also mentioned that the release retrospectives will be the appropriate time to revisit these lists and edit them. However, I think think it's safe to say that gcc, binutils, and systemd will be "high-risk" packages for the foreseeable future; and that bumping them immediately before ZHF cause more issues than it solves.

rfcs/0085-nixos-release-stablization.md
jonringer: Please refer to https://github.com/jonringer/rfcs/blob/85-ZHF-on-master/rfcs/0085-nixos-release-stablization.md#restriction-timeline, where "restrictions" are only placed on the staging and staging-next branches.

The goal is just to "limit potential breakages" and creating more work trying to stabilize master. Until we have a great process for stabilizing staging-next, this is the best compromise I could think of.

However, the current `staging-next` process has many short comings:
- Mostly core contributors are stabilizing it, and unable to dedicate time to fix all breakages
- The goal is for it to be "stable enough", and not necessarily be green-to-green

For ZHF and prevent trying to fix previously working packages, just disallowing updates which are likely to cause regressions seems like the most prudent course of action.

rfcs/0085-nixos-release-stablization.md
ryantm: @jonringer Do you prefer release-wiki to nixpkgs manual for storing a list of RCPs?

I thought about @andir's thing more, and I now do not object to removing the "Initially.." sentence.

jonringer: I would prefer the release-wiki. But either is okay.

ryantm: @jonringer I think you should change it to be the release wiki. It makes more sense.

davidak (approved):
Looks good

rfcs/0085-nixos-release-stablization.md
```diff
@@ -0,0 +1,128 @@
...
+| -3 Weeks | `master` | Focus on minimizing regressions in PRs |
+| -2 Weeks | `master` | Merge first `staging-next` fixes; begin second `staging-next` fix cycle |
+| -2 Weeks | `staging` | Unrestrict all breaking changes; new changes will not be present in `master` before branch-off |
+| -1 Weeks | `master` | Merge second `staging-next` fix cycle |
+| -1 Weeks | `staging-next` | Unrestrict all breaking changes; new changes will not be present in `master` before branch-off |
+| -1 Weeks | `master`, `release` | Perform Branch-off, create release channels, create new beta / unstable tags |
+| -1 Weeks | `master`, `release` | ZHF transitions to "backporting" workflow |
+| -1 Weeks | `release` | Prepare for release, finish remaining issues |
+| 0 Weeks | `release` | Release! |
+| 0 Weeks | | ZHF Ends |
```
das-g, line 101: Did I miss something, or does this not tell when "Restricted" mode begins and ends on `master`?

das-g (changes requested):

rfcs/0085-nixos-release-stablization.md
das-g: > so in best case they [the "Release Critical Package"] never break

I think the issue is less these packages themselves "breaking" (in the sense of not being buildable anymore or not functioning correctly when taken for themselves), but for changes to them to break many depending packages.

rfcs/0085-nixos-release-stablization.md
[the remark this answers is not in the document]
jonringer: The reference to Semantic Versioning was to further explain what a "breaking change" was, since semver already captures this concept stable releases in the form of a major version number bump, and the middle number bump for unstable releases.

But, all major version bumps are breaking changes in semver, which is more succinct and  aligned with the intent of the statement.

rfcs/0085-nixos-release-stablization.md
jonringer: You didn't miss anything. I mention [it above](https://github.com/NixOS/rfcs/pull/85#discussion_r602933101), but PRs which target master have a small enough scope that breaking changes can be more easily assessed, and thus don't bear the same potential to break builds as a staging PR.

>  if the major version bump affects <100 (or 500 in stituations, e.g. python packages) rebuilds, then it's still free to be bumped. All PRs which target master will not be affected by any of these changes.

The intent of this RFC is really to limit the amount of "potential breakages" right before a release. This is why breaking changes to `staging` shouldn't be merged. We constantly have to revisit downstream packages which are broken by such changes, and causes a lot of work which could be delayed til after the branchoff. "Non-breaking" staging changes are more aligned to potentially fixing issues and improving the quality of nixpkgs, so they should be allowed to be merged. Obviously not all packages follow semantic versioning, so changelogs should be reviewed and some downstream packages should be tested before merging. However, these tests would ideally happen on all PRs, however staging is particularly difficult as the magnitude of affected packages are much greater, and it's not feasible to thoroughly test every staging PR.

Since most of the risk should be avoided by not bundling 600+ rushed staging PRs immediately before ZHF, we should also be free to do ZHF on master as we should be progressing to a "more stable" state anyway. Also, the hundreds of backport PRs does pose a significant amount of "busy" work for contributors, and I go as far to say the backport PR process is a hindrance to ZHF participation.

jonringer:
> I have historically been skeptical of slowing down `master`/`staging`/`staging-next`. But I now think the painfulness of the release process might justify it. Perhaps this we need to get good at making the deadlines and not killing ourselves doing things this way, hopefully improve CI so there's less "fixing things after they break" in general, and then we re-earn the privilege of not needing to slow down `master`/`staging`/`staging-next`.

For clarification, this has no implication on master. The changes to master should be able to be captured for each PR, and we can effective move "green to green". This is more in regards to `staging`, where most packages we may build a downstream package, and run a test or two to verify it's not "absolutely broken"; but these often cause other packages to also break.

Also breakages introduced through staging PRs will likely need a staging PR to fix them, which makes the iteration for correcting such breakages much longer than  PRs which target `master`, in which "follow-up" PRs will also likely target master and the fixes can be merged "immediately".

rfcs/0085-nixos-release-stablization.md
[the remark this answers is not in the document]
das-g: > The reference to Semantic Versioning was to further explain what a "breaking change" was,

Yeah, that I understood. My point is that _disallowing_ such bumps (and/or breaking changes) should not be in the definition of "breaking change". Thus my suggestions
```suggestion
  - For projects following Semantic Versioning, major version bumps.
```
and alternatively
```suggestion
  - For projects following Semantic Versioning, major version bumps indicate breaking changes.
```

rfcs/0085-nixos-release-stablization.md
das-g: Ah, so the claim that
> Breaking changes to Release Critical Packages cannot be merged into master for 8 weeks

in the [current "Drawbacks" section](https://github.com/jonringer/rfcs/blob/e3d4d7d43fd0e3d32362905966f985b1c7d7a78d/rfcs/0085-nixos-release-stablization.md#drawbacks) follows _implicitly_ from the fact that such changes (both breaking _and_ concerning Release Critical Packages) would have to come to `master` _through_ `staging` or `staging-next` anyway rather than through regular PRs towards `master`?

jonringer: Correct, I guess the wording could be clearer, but yes, "critical packages" will most likely always need to go through `staging`. The later breaking change restriction to all `staging` PRs is meant to diminish  

"Critical Package" in this context would be a package which has significant downstream packages, and likely have multiple dimensions of compatibility (E.g. systemd's headers vs systemd's dbus messaging API). The ladder point about compatibility is the major concern, these behaviors are usually hard to test, or create tests for, and thus need to be used by unstable users to find issues and have additional `staging` iterations which can incorporate the fixes.

One example would be during the 20.09 release, where Plasma users could logout or change users because systemd had unintentionally made some login information hidden through its dbus API. Also, this behavior only occurred in nixpkgs due to plasma 5.19 + the most recent version of systemd at the time. Later versions of plasma didn't have this issue as they were developed against a more recent version of systemd. This should also be mitigated by RFC 80, in which our release timeline should be suitable for releasing newer versions of Plasma and Gnome, increasing the likelihood of kernel+systemd+desktopManager compatibility in general.

rfcs/0085-nixos-release-stablization.md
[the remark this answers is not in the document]
jonringer: Ah, I misunderstood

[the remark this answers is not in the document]
jonringer: applied

rfcs/0085-nixos-release-stablization.md
```diff
@@ -0,0 +1,128 @@
...
+The actual timeline will still be determined by the Release Managers
+and may be adjusted as needed. The current position in the timeline
+will be communicated on Discourse.
+
+| Weeks from Release | Branches Affected | Events |
+| --- | --- | --- |
+| -8 Weeks | | Gnome and Plasma(YY.11) packaging begins |
+| -6 Weeks | `staging-next`, `staging` | Restrict breaking changes to Release Critical Packages |
+| -4 Weeks | `staging-next`, `staging` | Restrict all breaking changes: allow only non-breaking updates and Desktop Manager changes |
+| -3 Weeks | `master` | (Day before ZHF) merge in `staging-next`, prep for ZHF |
```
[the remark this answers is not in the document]
jonringer: applied

rfcs/0085-nixos-release-stablization.md
```diff
@@ -65,14 +65,16 @@ on unstable before being included in a release.
 
 ## Release Critical Packages
 
-A list of Release Critical Packages will be maintained in the nixpkgs
-manual.  Release Critical Packages are important packages which have
+A list of Release Critical Packages will be maintained in the 
+[release wiki](https://github.com/NixOS/release-wiki).
+manual. Release Critical Packages are important packages which have
```
ryantm, line 70:
```suggestion
Release Critical Packages are important packages which have
```

rfcs/0085-nixos-release-stablization.md
jonringer: The applications in question (steam, spotify, discord, slack, etc. ) will generally target master, and can be backported. This RFC does not affect these use cases.

rfcs/0085-nixos-release-stablization.md
jonringer: List of release critical packages will now be curated on the release wiki

rfcs/0085-nixos-release-stablization.md
jonringer: Updated the wording to the "Drawbacks" section

rfcs/0085-nixos-release-stablization.md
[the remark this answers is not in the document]
das-g: Will commits of this PR be squashed before merging? If not you might want to fix the following:

61b40e8ea814f26d690ae348ab8a9a29b8d9c79d's commit message reads:
>  `Apply @davidak suggestion`

I guess autocomplete gave you the wrong username here. (I'm @das-g and the suggestion was by me.) For all I care, I don't need to be credited for that, though. Better describe what changed in the commit. (If suggestions for that are needed, I can come up with some.) If commits will be squashed and messages of intermediate commits discarded, don't bother, of course.

ryantm (changes requested):

rfcs/0085-nixos-release-stablization.md
```diff
@@ -0,0 +1,131 @@
...
+  - New packages and NixOS modules are welcome.
+- ZHF: Zero Hydra Failures
+  - Period in which packages are stabilized in preparation of a release.
+
+# Changes
+
+## Release Critical Packages
+
+A list of Release Critical Packages will be maintained in the 
+[release wiki](https://github.com/NixOS/release-wiki).
```
ryantm, line 69:
```suggestion
[release wiki](https://nixos.github.io/release-wiki/).
```

jonringer: Didn't realize this was a thing, awesome.

rfcs/0085-nixos-release-stablization.md
jonringer: applied

rfcs/0085-nixos-release-stablization.md
```diff
@@ -0,0 +1,128 @@
...
+This timeline represents when certain branches and packages will be
+restricted. It is not a complete timeline of all release activities.
+The actual timeline will still be determined by the Release Managers
+and may be adjusted as needed. The current position in the timeline
+will be communicated on Discourse.
+
+| Weeks from Release | Branches Affected | Events |
+| --- | --- | --- |
+| -8 Weeks | | Gnome and Plasma(YY.11) packaging begins |
+| -6 Weeks | `staging-next`, `staging` | Restrict breaking changes to Release Critical Packages |
```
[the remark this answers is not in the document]
ryantm: Since we defined Restricted above, I think we can just say:

```suggestion
| -6 Weeks | `staging-next`, `staging` | Release Critical Packages are Restricted |
```

ryantm:
The FCP comment period has ended. There has been some discussion and changes but I believe they are immaterial to the original intent/spirit of the RFC, so I do not think it is necessary for @garbas and @Mic92 to vote again on accepting the RFC.

We offer this RFC to @NixOS/rfc-steering-committee  for acceptance and merging!

jonringer:
Awesome, I'll do some announcements :)

jtrakk:
Rendered link is 404

jonringer:
> Rendered link is 404

Updated now that it has been merged

PR #86, opened 2021-02-09
[RFC 0086] NixOS generation APIs

miallo:
[rendered](https://github.com/miallo/rfcs/blob/86/rfcs/0086-nixos-generation-apis.md)

### DIFF
--- a/rfcs/0086-nixos-generation-apis.md
+++ b/rfcs/0086-nixos-generation-apis.md
@@ -179,9 +179,6 @@ scripts? (probably unnecessary, since usage in scripts probably limited)
 - Use `pager` to avoid spamming terminal?
 - Is the order fine?
 
-## nixos-version
-- Only accept known parameters (i.e. throw error on `nixos-version this --is $RANDOM`)?
-
 # Future work
 [future]: #future-work
 
--- a/rfcs/0086-nixos-generation-apis.md
+++ b/rfcs/0086-nixos-generation-apis.md
@@ -1,0 +1,195 @@
+---
+feature: nixos-generation-apis
+start-date: 2021-02-09
+author: Michael Lohmann <mial.lohmann@gmail.com>
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: https://github.com/NixOS/nixpkgs/pull/105910, https://github.com/NixOS/nixpkgs/issues/24374
+---
+
+# Summary
+[summary]: #summary
+
+On a typical system there are lots of generations. However rolling back to a
+specific one without rebooting is hard and it is even more difficult to get an
+overview of the available generations. This RFC proposes adding
+`nixos-rebuild --generation <number>` for rolling back to a specific generation
+and `nixos-rebuild list-generations` for showing details about the available
+generations.
+
+# Motivation
+[motivation]: #motivation
+
+The discoverability and usability of generations is limited on a running system.
+
+I had an issue, where I didn't notice a package breaking ~10 generations ago and
+while fixing it, I had to go back and forth multiple times.
+Since I didn't know about
+```sh
+sudo nix-env --switch-generation 12345 -p /nix/var/nix/profiles/system
+sudo /nix/var/nix/profiles/system/bin/switch-to-configuration switch
+```
+I repeated the rollback `n` times or rebooted and selected the respective
+generation in the bootloader.
+
+This RFC proposes to add
+```sh
+sudo nixos-rebuild switch --generation 12345
+```
+
+When doing this, it would be of great help for usage, if there was a way to show
+some some detail about the available generations, as is shown in the bootloader
+menu.
+
+# Detailed design
+[design]: #detailed-design
+
+## --generation
+Is basically a wrapper around the aforementioned `nix-env` command
+and the implementation is very much aligned with implementation of the `--rollback`
+flag.
+
+## list-generations
+Extract the relevant information about the generation from the
+`$profile-<generation>-link` directories and display it in a comprehensive way.
+This is similar to the creation of the
+[bootloader entries](https://github.com/NixOS/nixpkgs/blob/c14f14eeaf919c914e4dec2ce485a5bdc8dd4fec/nixos/modules/system/boot/loader/generations-dir/generations-dir-builder.sh#L50)
+but adds information like the configuration revision.
+
+Proposed information for each entry:
+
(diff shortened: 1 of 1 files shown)


miallo:
I don't know, if an RFC is the "right" thing, but I think, input from the comunity would be valuable and I didn't get answers to the remaining questions on [discourse](https://discourse.nixos.org/t/what-to-show-in-list-of-nixos-generations/11216).

For not force-pushing but still having a clean patch-set with 3 readable commits instead of the mess of the [PR](https://github.com/NixOS/nixpkgs/pull/105910), I have a [seperate branch](https://github.com/miallo/nixpkgs/tree/nixos-rebuild/clean-commits) for the implementation where the commits are clean. I will keep it updated and clean by force-pushing, so it stays readable and in sync

Mic92 (approved):
What is not clear to me, is what is causing these incompatibilities? We already have `init-interface-version` for systemd related incompatibilities.

Mic92:
What is not clear to me, is what is causing these incompatibilities? We already have init-interface-version for systemd related incompatibilities.

Mic92:
This looks like a straight forward feature to me. Is there anything in particular that you would want to discuss/solve with having this as a RFC?

rfcs/0086-nixos-generation-apis.md
```diff
@@ -0,0 +1,195 @@
...
+related-issues: https://github.com/NixOS/nixpkgs/pull/105910, https://github.com/NixOS/nixpkgs/issues/24374
+---
+
+# Summary
+[summary]: #summary
+
+On a typical system there are lots of generations. However rolling back to a
+specific one without rebooting is hard and it is even more difficult to get an
+overview of the available generations. This RFC proposes adding
+`nixos-rebuild --generation <number>` for rolling back to a specific generation
```
zimbatm, line 17:
```suggestion
`nixos-rebuild switch --generation <number>` for rolling back to a specific generation
```

rfcs/0086-nixos-generation-apis.md
```diff
@@ -0,0 +1,195 @@
...
+[54188](https://github.com/NixOS/nixpkgs/issues/54188).
+Though @edolstra [said it would be okay for now](https://github.com/NixOS/nixpkgs/issues/105910#issuecomment-754036275).
+
+
+## nixos-version
+
+Changing `nixos-version` so that a mistyped parameter throws an error improves
+error detection, but is a potentially breaking change, if there were scripts
+relying on mistyped flags. Though my opinion is, that it is better to throw
+meaningful errors instead of just doing something.
```
zimbatm, line 141: This looks like an independent change. But Is worth it nonetheless.

If scripts mistakenly passed an argument then it's up to them to fix their invocation.

rfcs/0086-nixos-generation-apis.md
```diff
@@ -0,0 +1,195 @@
...
+- Default value for configuration revision field if unset? (probably not)
+- Truncate revision (maybe if hash only)?
+- Put the `(current)` tag in it's own column?
+- Show only system-profile by default?
+- How to handle machine-readability?
+- Provide stable API (e.g. `--porcelain`) for futureprooving output changes for
+scripts? (probably unnecessary, since usage in scripts probably limited)
+- Use `column` format for human readability? => Evaluation is eager
+- Use `pager` to avoid spamming terminal?
+- Is the order fine?
```
zimbatm, line 187: All of these are good questions. The best way to answer them is to play with an implementation to get a feel for it.

Targetting the interface for human-consumption first will allow more flexibility and to change the output as needed. Then when it's stable, add a --json output for machine consumption.

zimbatm (approved):
This doesn't look controversial to me and could be implemented directly.

rfcs/0086-nixos-generation-apis.md
miallo: Yes, it sort of is independent, but the current implementation will just print the default value for any unrecognised flag and so in the implementation of list-generations I had to [implement a workaraound](https://github.com/NixOS/nixpkgs/blob/4ebca19b171dd5d0d0660ebfd9c413d24a70374e/pkgs/os-specific/linux/nixos-rebuild/nixos-rebuild.sh#L448-L455) that ignores the output if it is the default. I agree -  I probably should have put this in another RFC and kept the default behaviour here...

rfcs/0086-nixos-generation-apis.md
miallo: I thought about this and my thoughts were, to keep the interface as similar to the `--rollback` as possible. Also you could just as well `test`/ a generation. I thought about making `switch` the default (overridable) action but my main doubts came when @edolstra said
> There is an issue somewhere about replacing `nixos-rebuild` with a `nixos` command

and then I think `nixos --generation <n>` would be missing a verb ^^ (but maybe that is just personal preference...)

In general I got a bit unsatisfied with the name `switch` when I realised that the default behaviour is really doing a `build-and-switch` and here we really just want to switch to something already existing. This drove me a bit crazy when I learned about profiles while implementing this and I didn't know how to distinguish "switching to an existing profile" vs "creating a new generation of a profile". Actually I am thinking now, that the `--generation` flag is the only current possibility to just switch profiles - but I am probably overlooking something...

miallo:
> This looks like a straight forward feature to me. Is there anything in particular that you would want to discuss/solve with having this as a RFC?

It is mostly about the [unresolved questions](https://github.com/miallo/rfcs/blob/86/rfcs/0086-nixos-generation-apis.md#list-generations-4) for which I wanted to get some comments :)


> ignore this message. I reread the RFC.

I 100% agree - the title I chose isn't the best...

rfcs/0086-nixos-generation-apis.md
zimbatm: My bad. I forgot that the options are defined on the top-level.

In some sense, `--generation <n>` is a more general version of `--rollback` so I guess both flags should be mutually-exclusive.

miallo: I didn't enforce it with error messages, but the `--generation` is in an `else`-clause that doesn't get executed if both are given. Should I throw errors on all incompatible / useless combinations of commands? Right now they just don't do anything (i.e. `nixos-rebuild list-generations --rollback` will just list them and ignore the rest).

miallo:
I was also a bit unsure if this should be an RFC. Since I am relatively new to the Nix-environment I wanted more input from more experienced people. An example: I didn't know about profiles and only considered the system one in my first implementation...

Did I understand the [rfc steering comittee meeting summary](https://github.com/NixOS/rfc-steering-committee/issues/56) right that I should close this issue?

PR #88, opened 2021-03-13
[RFC 0088] Nixpkgs Breaking Change Policy

kevincox:
This specifies the expected workflow as a base for workflows and future automation.

[Rendered](https://github.com/kevincox/rfcs-1/blob/patch-1/rfcs/0088-nixpkgs-breaking-change-policy.md)

### DIFF
--- a/rfcs/0088-nixpkgs-breaking-change-policy.md
+++ b/rfcs/0088-nixpkgs-breaking-change-policy.md
@@ -3,20 +3,31 @@ feature: nixpkgs-breaking-change-policy
 start-date: 2020-09-24
 author: Kevin Cox
 co-authors: (TBD)
-shepherd-team: (TBD)
+shepherd-team: @blaggacao, @jonringer, @nrdxp
 related-issues: (TBD)
 ---
 
 # Summary
 
 This defines a policy for how to manage nixpkgs changes that break dependent
-derivations or tests. It specifies time frames and procedures so that
+derivations or tests. It specifies timeframes and procedures so that
 maintainers of dependencies and dependents can know what to expect from each
 other when making breaking changes.
 
+The direct intended result of this RFC is:
+
+- Maintainers of dependencies have a clear framework for handling changes that
+  break dependants.
+- Maintainers of dependants have a clear framework for how dependency breacking
+  changes will be handled.
+- NixOS channels will very rarely (ideally never) stall due to build or test
+  failures.
+
 This document is only focused on changes to derivation builds and tests it
 **does not** aim to make any opinion on changes that can break user machines or
-configurations.
+configurations. (While this is an interesting topic it is harder to put
+concrete rules on due to the difficulty of identifying these breakages. In
+order to be covered by this policy please but tests on your packages.)
 
 # Motivation
 
@@ -29,6 +40,15 @@ from:
 - Merge to `staging`, notify dependents then eventually merge to `master`.
 - Gather a large number of fixes in the PR branch then merge to `staging` or
   `master` once most or all dependencies are fixed.
+
+This adhoc nature has a number of downsides:
+
+- Change authors need to put effort into deciding how to handle each particular
+  change.
+- Maintainers of dependant packages don't know what to expect and have a hard
+  time keeping up with upstream breakages.
+- Merges to `master` or `staging` can result in broken channels which block all
+  updates including security updates.
 
 This aims to provide a uniform approach so that everyone involved knows what to
 expect and to allow further workflows including automation to be build on top
@@ -48,7 +68,7 @@ primary goals:
 - Avoid putting more burden than necessary on the dependency maintainer. If the
   maintainers of core derivations face toil proportionally to the number of
   transitive dependencies they will quickly become overloaded. These
-  maintainers are arguably the most critical to nixpkgs and  their load needs
+  maintainers are arguably the most critical to Nixpkgs and  their load needs
   to be kept manageable.
 - Avoid unnecessarily breaking packages for any period of time. There are a
   number of users on the `*-unstable` channels and it is annoying if packages
@@ -60,7 +80,9 @@ primary goals:
 
 ## Procedure
 
-The target branch for a merge is not affected by this policy. It will be picked, as it is today, according to https://nixos.org/manual/nixpkgs/stable/#submitting-changes-commit-policy.
+The target branch for a merge is not affected by this policy. It will be
+picked, as it is today, according to
+https://nixos.org/manual/nixpkgs/stable/#submitting-changes-commit-policy.
 
 1. The maintainer will prepare a PR with their intended changes.
 2. The maintainer should test a sample of dependent derivations to ensure that
@@ -88,12 +110,12 @@ sub-maintainers were notified.
 [broken](https://nixos.org/manual/nixpkgs/stable/#sec-standard-meta-attributes).
 8. The maintainer can now merge to the target branch.
 
-This procedure should not result in a failing package build in the target branch at any
+This procedure should not result in a failing channel build in the target branch at any
 point.
 
 # Drawbacks
 
-This delays the merge of “core” derivations as the author needs to wait for
+This delays the merge of "core" derivations as the author needs to wait for
 sub-packages to be tested and possibly fixed up to the 7 day threshold.
 
 # Alternatives
@@ -103,7 +125,7 @@ sub-packages to be tested and possibly fixed up to the 7 day threshold.
 A shorter grace period allows maintainers to move more quickly but requires
 maintainers of defendant derivations to jump to action quickly which is not
 always possible for volunteer driven work. For example if a maintainer is on
-vacation it can’t be expected that they respond in a couple of days.
+vacation they can't be expected to respond in a couple of days.
 
 ## Longer grace period
 
@@ -112,14 +134,15 @@ changes in dependencies and provide more time to search for replacement
 maintainers if the original maintainer has abandoned the derivation. The 7 day
 number was mostly arbitrary and can easily be changed in the future. A longer
 grace period would also result in more packages being fixed before the breaking
-change is shipped which would mean that end-users would experience less
-breakage on channels.
+change is shipped, which results in less breakage for end-users.
 
-# Unresolved questions ## Critical Packages and Tests
+# Unresolved questions
+
+## Critical Packages and Tests
 
 What if a breaking change breaks NixOS tests? There must be packages and tests
 so critical that we can not merge without them passing? In that case do we
-leave the PR open until fixed? or use the `staging` branch?
+leave the PR open until fixed?
 
 # Future work
 - Create a tool for automatically notifying maintainers of broken dependents
--- a/rfcs/0088-nixpkgs-breaking-change-policy.md
+++ b/rfcs/0088-nixpkgs-breaking-change-policy.md
@@ -14,6 +14,11 @@ derivations or tests. It specifies timeframes and procedures so that
 maintainers of dependencies and dependents can know what to expect from each
 other when making breaking changes.
 
+While this policy should be followed as much as possible once accepted it is
+primarily focused on getting agreement such that tools can be built that make
+implementing this policy very easy. See the [Future Work section](#future-work)
+for examples of the tools that will be built.
+
 The direct intended result of this RFC is:
 
 - Maintainers of dependencies have a clear framework for handling changes that
@@ -90,6 +95,8 @@ their change is not unnecessarily or unintentionally breaking. (Example: Ensure
 that a critical output file was not forgotten) Note that sometimes it **is**
 necessary to break all dependent packages, and the maintainer is not required
 to avoid this.
+    1. A sample of at least 100 packages or 3h of build time is recommend as a guideline.
+    2. The sample should include as many `passthru.tests` tests as is feasible.
 3. The maintainer will get the PR reviewed and approved.
     1. It is **recommended but not required** to have some maintainers of
     dependent packages involved in the review to review if the breakage is
@@ -110,8 +117,9 @@ sub-maintainers were notified.
 [broken](https://nixos.org/manual/nixpkgs/stable/#sec-standard-meta-attributes).
 8. The maintainer can now merge to the target branch.
 
-This procedure should not result in a failing channel build in the target branch at any
-point.
+This procedure should not result in a failing channel build in the target
+branch at any point. In particular NixOS channel constituents are considered
+blocking and are an exception to the 7 day breakage rule.
 
 # Drawbacks
 
@@ -138,11 +146,7 @@ change is shipped, which results in less breakage for end-users.
 
 # Unresolved questions
 
-## Critical Packages and Tests
-
-What if a breaking change breaks NixOS tests? There must be packages and tests
-so critical that we can not merge without them passing? In that case do we
-leave the PR open until fixed?
+None.
 
 # Future work
 - Create a tool for automatically notifying maintainers of broken dependents
--- a/rfcs/0088-nixpkgs-breaking-change-policy.md
+++ b/rfcs/0088-nixpkgs-breaking-change-policy.md
@@ -33,7 +33,7 @@ This document is only focused on changes to derivation builds and tests it
 **does not** aim to make any opinion on changes that can break user machines or
 configurations. (While this is an interesting topic it is harder to put
 concrete rules on due to the difficulty of identifying these breakages. In
-order to be covered by this policy please but tests on your packages.)
+order to be covered by this policy please put tests on your packages.)
 
 # Motivation
 
--- a/rfcs/0088-nixpkgs-breaking-change-policy.md
+++ b/rfcs/0088-nixpkgs-breaking-change-policy.md
@@ -94,7 +94,9 @@ point.
 This delays the merge of “core” derivations as the author needs to wait for
 sub-packages to be tested and possibly fixed up to the 7 day threshold.
 
-# Alternatives ## Shorter grace period
+# Alternatives
+
+## Shorter grace period
 
 A shorter grace period allows maintainers to move more quickly but requires
 maintainers of defendant derivations to jump to action quickly which is not
--- a/rfcs/0088-nixpkgs-breaking-change-policy.md
+++ b/rfcs/0088-nixpkgs-breaking-change-policy.md
@@ -4,6 +4,7 @@ start-date: 2020-09-24
 author: Kevin Cox
 co-authors: (TBD)
 shepherd-team: @blaggacao, @jonringer, @nrdxp
+shepherd-leader: @blaggacao
 related-issues: (TBD)
 ---
 
--- a/rfcs/0088-nixpkgs-breaking-change-policy.md
+++ b/rfcs/0088-nixpkgs-breaking-change-policy.md
@@ -1,9 +1,11 @@
-# nixpkgs Breaking Change Policy
-
-| feature          | start-date | author    | co-authors                                    | shepherd-team                                                   | shepherd-leader                                  | related-issues                             |
-| ---------------- | ---------- | --------- | --------------------------------------------- | --------------------------------------------------------------- | ------------------------------------------------ | ------------------------------------------ |
-| breaking_changes | 2020-09-24 | Kevin Cox | (find a buddy later to help out with the RFC) | (names, to be nominated and accepted by RFC steering committee) | (name to be appointed by RFC steering committee) | (will contain links to implementation PRs) |
-
+---
+feature: nixpkgs-breaking-change-policy
+start-date: 2020-09-24
+author: Kevin Cox
+co-authors: (TBD)
+shepherd-team: (TBD)
+related-issues: (TBD)
+---
 
 # Summary
 
--- a/rfcs/0088-nixpkgs-breaking-change-policy.md
+++ b/rfcs/0088-nixpkgs-breaking-change-policy.md
@@ -1,0 +1,124 @@
+# nixpkgs Breaking Change Policy
+
+| feature          | start-date | author    | co-authors                                    | shepherd-team                                                   | shepherd-leader                                  | related-issues                             |
+| ---------------- | ---------- | --------- | --------------------------------------------- | --------------------------------------------------------------- | ------------------------------------------------ | ------------------------------------------ |
+| breaking_changes | 2020-09-24 | Kevin Cox | (find a buddy later to help out with the RFC) | (names, to be nominated and accepted by RFC steering committee) | (name to be appointed by RFC steering committee) | (will contain links to implementation PRs) |
+
+
+# Summary
+
+This defines a policy for how to manage nixpkgs changes that break dependent
+derivations or tests. It specifies time frames and procedures so that
+maintainers of dependencies and dependents can know what to expect from each
+other when making breaking changes.
+
+This document is only focused on changes to derivation builds and tests it
+**does not** aim to make any opinion on changes that can break user machines or
+configurations.
+
+# Motivation
+
+Right now the procedure for making breaking changes varies widely by maintainer
+and is largely tribal knowledge. There are a wide variety of approaches ranging
+from:
+
+- After testing to confirm that a change is correct, merge and notify
+  dependents that they need to update their packages.
+- Merge to `staging`, notify dependents then eventually merge to `master`.
+- Gather a large number of fixes in the PR branch then merge to `staging` or
+  `master` once most or all dependencies are fixed.
+
+This aims to provide a uniform approach so that everyone involved knows what to
+expect and to allow further workflows including automation to be build on top
+of these procedures.
+
+## Past Discussions
+
+https://discourse.nixos.org/t/please-fix-all-packages-which-were-broken-by-the-qt-wrapping-changes/6444
+
+# Detailed design
+
+## Goals
+
+The following procedure will be used for making breaking changes. It has a few
+primary goals:
+
+- Avoid putting more burden than necessary on the dependency maintainer. If the
+  maintainers of core derivations face toil proportionally to the number of
+  transitive dependencies they will quickly become overloaded. These
+  maintainers are arguably the most critical to nixpkgs and  their load needs
+  to be kept manageable.
+- Avoid unnecessarily breaking packages for any period of time. There are a
+  number of users on the `*-unstable` channels and it is annoying if packages
+  in their configuration get broken.
+- Avoid breaking the `master` branch as much as possible. The `master` branch
+  is used by the `*-unstable` channels and breaking this branch means that no
+  updates can be pushed. This includes critical security updates as well as
+  regular feature updates.
+
+## Procedure
+
+The target branch for a merge is not affected by this policy. It will be picked, as it is today, according to https://nixos.org/manual/nixpkgs/stable/#submitting-changes-commit-policy.
+
+1. The maintainer will prepare a PR with their intended changes.
+2. The maintainer should test a sample of dependent derivations to ensure that
+their change is not unnecessarily or unintentionally breaking. (Example: Ensure
+that a critical output file was not forgotten) Note that sometimes it **is**
+necessary to break all dependent packages, and the maintainer is not required
+to avoid this.
+3. The maintainer will get the PR reviewed and approved.
+    1. It is **recommended but not required** to have some maintainers of
+    dependent packages involved in the review to review if the breakage is
+    justified.
+4. The maintainer will contact the maintainers of all dependent, broken
+packages (herein called sub-maintainers).
+    1. The sub-maintainers will be provided a link to the PR with the breaking
+    changes as well as any context that will help them resolve the breakages.
+    2. The sub-maintainers are expected to update the derivations that they
+    maintain within 7 days.
+        1. These changes should be merged to the target branch following the regular
+        nixpkgs PR process if they are backwards compatible. Otherwise they
+        will be merged into the branch of the breaking PR.
+5. The maintainer must provide a grace period of at least 7 days from when the
+sub-maintainers were notified.
+6. The maintainer must merge the target branch into their PR branch.
+7. The maintainer must mark all still-broken packages as
+[broken](https://nixos.org/manual/nixpkgs/stable/#sec-standard-meta-attributes).
+8. The maintainer can now merge to the target branch.
+
+This procedure should not result in a failing package build in the target branch at any
+point.
+
+# Drawbacks
+
+This delays the merge of “core” derivations as the author needs to wait for
+sub-packages to be tested and possibly fixed up to the 7 day threshold.
+
+# Alternatives ## Shorter grace period
+
+A shorter grace period allows maintainers to move more quickly but requires
+maintainers of defendant derivations to jump to action quickly which is not
+always possible for volunteer driven work. For example if a maintainer is on
+vacation it can’t be expected that they respond in a couple of days.
+
+## Longer grace period
+
+A longer grace period would give derivation maintainers more time to react to
+changes in dependencies and provide more time to search for replacement
+maintainers if the original maintainer has abandoned the derivation. The 7 day
+number was mostly arbitrary and can easily be changed in the future. A longer
+grace period would also result in more packages being fixed before the breaking
+change is shipped which would mean that end-users would experience less
+breakage on channels.
+
+# Unresolved questions ## Critical Packages and Tests
+
+What if a breaking change breaks NixOS tests? There must be packages and tests
+so critical that we can not merge without them passing? In that case do we
+leave the PR open until fixed? or use the `staging` branch?
+
+# Future work
+- Create a tool for automatically notifying maintainers of broken dependents
+- Create a tool for automatically marking broken packages as broken.
+- [This is a milestone on the way to implemented a queue-based merge process
+  for nixpkgs.](https://paper.dropbox.com/doc/MTSY8xKH6y1xDEwavyDNW)


rfcs/0000-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,75 @@
...
+- Avoid unnecessarily breaking packages for any period of time. There are a number of users on the `*-unstable` channels and it is annoying if packages in their configuration get broken.
+- Avoid breaking the `master` branch as much as possible. The `master` branch is used by the `*-unstable` channels and breaking this branch means that no updates can be pushed. This includes critical security updates as well as regular feature updates.
+
+## Procedure
+
+1. The maintainer will prepare a PR with their intended changes.
+2. The maintainer should test a sample of dependent derivations to ensure that their change is not unnecessarily or unintentionally breaking. (Example: Ensure that a critical output file was not forgotten) Note that sometimes it **is** necessary to break all dependent packages, and the maintainer is not required to avoid this.
+3. The maintainer will get the PR reviewed and approved.
+    1. It is **recommended but not required** to have some maintainers of dependent packages involved in the review to review if the breakage is justified.
+4. The maintainer will contact the maintainers of all dependent, broken packages (herein called sub-maintainers).
```
SuperSandro2000, line 43: How? Should I collect email addresses and write them an email? I think this is not practically possible.

rfcs/0000-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,75 @@
...
+This delays the merge of “core” derivations as the author needs to wait for sub-packages to be tested and possibly fixed up to the 7 day threshold.
+
+# Alternatives
+## Shorter grace period
+
+A shorter grace period allows maintainers to move more quickly but requires maintainers of defendant derivations to jump to action quickly which is not always possible for volunteer driven work. For example if a maintainer is on vacation it can’t be expected that they respond in a couple of days.
+
+## Longer grace period
+
+A longer grace period would give derivation maintainers more time to react to changes in dependencies and provide more time to search for replacement maintainers if the original maintainer has abandoned the derivation. The 7 day number was mostly arbitrary and can easily be changed in the future. A longer grace period would also result in more packages being fixed before the breaking change is shipped which would mean that end-users would experience less breakage on channels.
```
SuperSandro2000, line 65: In certain cases the package could also be dropped. I feel like this should be mentioned here.

rfcs/0000-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Unresolved questions
+## Critical Packages and Tests
+
+What if a breaking change breaks NixOS tests? There must be packages and tests so critical that we can not merge without them passing? In that case do we leave the PR open until fixed? or use the `staging` branch?
+
+# Future work
+- Create a tool for automatically notifying maintainers of broken dependents
+- Create a tool for automatically marking broken packages as broken.
+- [This is a milestone on the way to implemented a queue-based merge process for nixpkgs.](https://paper.dropbox.com/doc/MTSY8xKH6y1xDEwavyDNW)
```
SuperSandro2000, line 75: I have the fear that this link will bit rot over time.

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: The idea was to tag them on the PR. We have contact info in https://github.com/NixOS/nixpkgs/blob/master/maintainers/maintainer-list.nix so this can (and is expected to be) automated.

Ma27: I know this pops up quite often recently (or maybe I'm wrong and it's just selective perception), but this is something that can be solved by Hydra notifications or e.g. https://github.com/NixOS/hydra/pull/743.

Also I'd like to emphasize that there's a big difference between the stability of `master` and release branches, I mean there's a reason why `nixos-unstable` has `unstable` in its name and we have ZHF iterations after branch-off.

rfcs/0000-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+This defines a policy for how to manage nixpkgs changes that break dependent derivations or tests. It specifies time frames and procedures so that maintainers of dependencies and dependents can know what to expect from each other when making breaking changes.
+
+This document is only focused on changes to derivation builds and tests it **does not** aim to make any opinion on changes that can break user machines or configurations.
+
+# Motivation
+
+Right now the procedure for making breaking changes varies widely by maintainer and is largely tribal knowledge. There are a wide variety of approaches ranging from:
+
+- After testing to confirm that a change is correct, merge to `master` and notify dependents that they need to update their packages.
```
FRidh, line 17: The initial target branch that is chosen depends on several factors
- awareness of the staging branch and what its for
- knowing the amount of rebuilds needed and thereby whether it should target staging or master
- whether requires changes are on any of the possible target branches

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: We discussed this and there was much disagreement of when to drop vs mark broken. I think that dropping because the maintainer doesn't respond in 7 days is a bit aggressive (they could just be on vacation) so I think the best route is just make it a simple "mark broken" and we can leave when to delete derivations to another policy.

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: Do you have a preferred link? I could just leave it as a description but I figured the link didn't harm.

rfcs/0000-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,75 @@
...
+
+- Avoid putting more burden than necessary on the dependency maintainer. If the maintainers of core derivations face toil proportionally to the number of transitive dependencies they will quickly become overloaded. These maintainers are arguably the most critical to nixpkgs and  their load needs to be kept manageable.
+- Avoid unnecessarily breaking packages for any period of time. There are a number of users on the `*-unstable` channels and it is annoying if packages in their configuration get broken.
+- Avoid breaking the `master` branch as much as possible. The `master` branch is used by the `*-unstable` channels and breaking this branch means that no updates can be pushed. This includes critical security updates as well as regular feature updates.
+
+## Procedure
+
+1. The maintainer will prepare a PR with their intended changes.
+2. The maintainer should test a sample of dependent derivations to ensure that their change is not unnecessarily or unintentionally breaking. (Example: Ensure that a critical output file was not forgotten) Note that sometimes it **is** necessary to break all dependent packages, and the maintainer is not required to avoid this.
```
FRidh, line 40: Whether it is sometimes necessary to break all dependent packages is quite a claim and needs further motivation.

rfcs/0000-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,75 @@
...
+3. The maintainer will get the PR reviewed and approved.
+    1. It is **recommended but not required** to have some maintainers of dependent packages involved in the review to review if the breakage is justified.
```
FRidh, line 42: Why recommendation and not required?

rfcs/0000-nixpkgs-breaking-change-policy.md
Ma27: If I understand Sandro correctly here, this isn't supposed to be the "default measure", but a valid thing to do if the package is e.g. fairly old.

rfcs/0000-nixpkgs-breaking-change-policy.md
FRidh: GitHub also limits the amount of mentions in case many people are affected.

rfcs/0000-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,75 @@
...
+4. The maintainer will contact the maintainers of all dependent, broken packages (herein called sub-maintainers).
+    1. The sub-maintainers will be provided a link to the PR with the breaking changes as well as any context that will help them resolve the breakages.
+    2. The sub-maintainers are expected to update the derivations that they maintain within 7 days.
+        1. These changes should be merged to `master` following the regular nixpkgs PR process if they are backwards compatible. Otherwise they will be merged into the branch of the breaking PR.
+5. The maintainer must provide a grace period of at least 7 days from when the sub-maintainers were notified.
+6. The maintainer must merge `master` into their PR branch.
+7. The maintainer must mark all still-broken packages as [broken](https://nixos.org/manual/nixpkgs/stable/#sec-standard-meta-attributes).
+8. The maintainer can now merge to `master`.
```
FRidh, line 50: This assumes here that the change is sufficiently small it can go into master. With large changes that motivate this PR that is exactly not the case. The changes then go into staging, then staging-next and finally in master. That also gives an extra stabilization period on top of the proposed 7 days.

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: Part of the idea here is to allows us to move towards keeping the unstable branches green (at least from the building point of view). This allows security fixes and other changes to get pushed at any time. This policy won't get us all of the way there but it already in theory does as all broken packages should be marked as broken before getting merged to `master`.

Hydra notifications would be good, but they only apply to the merged branch. At that point it is too late, that broken package is already blocking changes which is bad for critical fixes and annoying to contributors (as their change gets delayed). Ideally this would be the same sort of tech, but applied somewhat pre-merge.

I don't know if there is an official stance on what `-unstable` means. Because the channel is always tested on hydra before being published. My understanding it is more that packages, including major versions will be updated frequently. I don't think it should mean "fails to build and release for days or weeks at a time".

rfcs/0000-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,75 @@
...
+A shorter grace period allows maintainers to move more quickly but requires maintainers of defendant derivations to jump to action quickly which is not always possible for volunteer driven work. For example if a maintainer is on vacation it can’t be expected that they respond in a couple of days.
+
+## Longer grace period
+
+A longer grace period would give derivation maintainers more time to react to changes in dependencies and provide more time to search for replacement maintainers if the original maintainer has abandoned the derivation. The 7 day number was mostly arbitrary and can easily be changed in the future. A longer grace period would also result in more packages being fixed before the breaking change is shipped which would mean that end-users would experience less breakage on channels.
+
+# Unresolved questions
+## Critical Packages and Tests
+
+What if a breaking change breaks NixOS tests? There must be packages and tests so critical that we can not merge without them passing? In that case do we leave the PR open until fixed? or use the `staging` branch?
```
FRidh, line 70: What is the scope of the change we are talking about? Staging is not for testing or dumping changes in the hope they get fixed; one is expected to follow their change and fix regressions especially when it has reached staging-next.

There are important NixOS tests and tests that are pretty unimportant.

rfcs/0000-nixpkgs-breaking-change-policy.md
Ma27: Well, the channel doesn't break on **any** package though. Most of the leaf-packages (and AFAIU that's what we're talking about here) can be totally broken (i.e. failing to compile or whatever) without preventing `nixos-unstable` from moving forward.

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: That makes sense to me, but I just wanted to keep it simple. In this policy the action for every package is simply to mark as broken. Then we can have some separate process for removing broken packages. I think trying to work a policy in here for when to delete is just unnecessary complication.

Note also that the intention is to create a tool to add the `broken = true` flags so we don't want to need to teach that tool to make this decision. It would be easiest to create a separate tool to raise PRs for old broken packages that will catch all packages, no matter why they were originally marked as broken.

FRidh (changes requested):
Good initiative to take this further. Please do limit the line length to make it easier to comment.

The RFC does not go sufficiently into size/scope of PR's this RFC covers nor how it relates to the different development branches (master/staging/staging-next) we have. Motivations are missing for the specific steps in the procedure. Maybe they should be added there or in a separate section.

When talking about what is an important and unimportant change it would be very helpful to first have defined what is considered core and what not. I think it is difficult to define the policy well in this RFC without defining package tiers.

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: That's a good point. For what I am aiming for I don't really care about a package if the channel config doesn't care about it. On the other hand is there a downside to marking these broken packages as broken? If there is it would be easy enough to update the RFC to define what dependents we care about and just ignore them completely (maybe still notify them, but no need to mark as broken).

kevincox:
> Good initiative to take this further. Please do limit the line length to make it easier to comment.

If that is desired I can hard-wrap it on the next push.

> The RFC does not go sufficiently into size/scope of PR's this RFC covers nor how it relates to the different development branches (master/staging/staging-next) we have. Motivations are missing for the specific steps in the procedure. Maybe they should be added there or in a separate section.

Right now I am focused on `master` and the `-unstable` channels which follow it. I am welcome to feedback on how this should apply to other branches but I am less familiar with them, (although IIUC they already require stability)
 
> When talking about what is an important and unimportant change it would be very helpful to first have defined what is considered core and what not. I think it is difficult to define the policy well in this RFC without defining package tiers.

This is being discussed here: https://github.com/NixOS/rfcs/pull/88#discussion_r593750923

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: I should probably remove the "all" but I'm sure there are cases where libraries make updates that change the build system or remove something. The point here is that the initial review should consider if the amount of breakage is reasonable. I will reword to make that more explicit.

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: Good point. I will update the target branch to be variable as it should still follow the correct procedures with that regard.

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: This is largely what this policy is trying to address. We are trying to write down what should be done to make these choices better known and more clear.

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: For reference I believe this is the current policy for picking a target branch: https://github.com/NixOS/rfcs/blob/master/rfcs/0026-staging-workflow.md#detailed-design

FRidh: That's the RFC. The implementation is in the Nixpkgs manual, https://nixos.org/manual/nixpkgs/stable/#submitting-changes-commit-policy

kevincox: Ah, thanks!

kevincox:
Updated:

* Hard wrap lines.
* Remove many instances of `master` and refer to "target branch" according to existing policy for picking the target branch.
* Update number in file name to the PR number.

rfcs/0000-nixpkgs-breaking-change-policy.md
SuperSandro2000: > We have contact info in

Those information are optional and not all people added a github name or an email address.

rfcs/0000-nixpkgs-breaking-change-policy.md
SuperSandro2000: I have mirrored it to web.archive.org https://web.archive.org/web/20210325015950if_/https://paper.dropbox.com/doc/nixpkgs-Merge-Queue-Overview-MTSY8xKH6y1xDEwavyDNW

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: Well if they aren't contactable than their package will be broken. I don't think we can really do anything about this if we don't have any handles to the maintainer.

SuperSandro2000: We should add something similar to ``if you want to be contacted you need to add contact information because otherwise it is not possible``.

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,124 @@
...
+their change is not unnecessarily or unintentionally breaking. (Example: Ensure
+that a critical output file was not forgotten) Note that sometimes it **is**
+necessary to break all dependent packages, and the maintainer is not required
+to avoid this.
+3. The maintainer will get the PR reviewed and approved.
+    1. It is **recommended but not required** to have some maintainers of
+    dependent packages involved in the review to review if the breakage is
+    justified.
+4. The maintainer will contact the maintainers of all dependent, broken
+packages (herein called sub-maintainers).
```
kevincox, line 74:
```suggestion
packages (herein called sub-maintainers).
    1. The sub-maintainers will be contacted preferably by the `github` field,
      followed by the `email` field of [maintainer-list.nix
      ](https://github.com/NixOS/nixpkgs/blob/master/maintainers/maintainer-list.nix).
      If neither of those fields are set than contacting the maintainer is not
      required and the package should be marked as broken after 7 days.
```

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: How does this sound? https://github.com/NixOS/rfcs/pull/88/files#r602721463

rfcs/0088-nixpkgs-breaking-change-policy.md
kevincox: Should we wait? Or remove this maintainer from the blocking list right away? What is the odds that they randomly notice that we are breaking their package?

sternenseemann (commented):
Overall, I'd say what this RFC lacks most of all is a clear definition of a “breaking change”. From reading it, I get the impression that it is used as “a change which causes a build failure in a reverse dependency”. Clarification of this would be important in my eyes.

Very important would be to clarify that this policy doesn't apply to the introduction of (partial) evaluation errors by e. g. changing an internal interface since we can't work around such a problem using meta attributes and the channel would be held back until all breakage would be fixed.

Additionally, I think we can not just ignore breakage for users, as outlined in a comment.

rfcs/0000-nixpkgs-breaking-change-policy.md
sternenseemann: I think removing old, unmaintained packages should be done independently of major updates of packages they depend on. Adding this as a valid option in the process adds extra need for discussion which is hindering if you are aiming to keep a defined schedule.

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,124 @@
...
+    1. The sub-maintainers will be provided a link to the PR with the breaking
+    changes as well as any context that will help them resolve the breakages.
+    2. The sub-maintainers are expected to update the derivations that they
+    maintain within 7 days.
```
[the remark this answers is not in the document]
kevincox: I think that for the purposes here we should consider the process of "how a patch is submitted" off-topic. If our current process is incapable of hitting the 7 day deadline then we should probably make improvements to that process (or can pick a longer deadline here). One other option is that there is a no-response deadline and an extension for in-progress fixes. For example 7 days to commit to a fix and 14 days to have the fix submitted. If all responses are merged in 7 days there is no need to wait longer.

But personally I don't feel the need to change this at this point. Also I explicitly said **at least** 7 days for the merge. The maintainer should use their own discretion.

> Additionally there should probably be a standardized way of keeping track of such PRs to help with this process.

How about suggesting that the PR description (and probably at least one of the commit messages) should link to the upstream PR? That way GitHub will link them together.

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,124 @@
...
+        1. These changes should be merged to the target branch following the regular
+        nixpkgs PR process if they are backwards compatible. Otherwise they
+        will be merged into the branch of the breaking PR.
+5. The maintainer must provide a grace period of at least 7 days from when the
+sub-maintainers were notified.
+6. The maintainer must merge the target branch into their PR branch.
+7. The maintainer must mark all still-broken packages as
+[broken](https://nixos.org/manual/nixpkgs/stable/#sec-standard-meta-attributes).
+8. The maintainer can now merge to the target branch.
```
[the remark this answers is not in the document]
kevincox: I am slightly hesitant to giving special privileges to changes that require dependent updates but maybe this is justified as packages with more dependents are likely more important.

I am also hesitant to put the full burden of merging these changes on an individual.

I guess this is blocking if this policy puts additional difficulties to getting the change merged. But as I see the additional burden is on the author/maintainer, not on the actual process of getting it merged.

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,124 @@
...
+breakage on channels.
+
+# Unresolved questions ## Critical Packages and Tests
+
+What if a breaking change breaks NixOS tests? There must be packages and tests
+so critical that we can not merge without them passing? In that case do we
+leave the PR open until fixed? or use the `staging` branch?
+
+# Future work
+- Create a tool for automatically notifying maintainers of broken dependents
```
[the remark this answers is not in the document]
kevincox: FWIW I have a very early "prototype" here. It needs proper testing to see how effective this solution is. I can work on finishing it up and propose both in parallel if you think it is critical.

I considered it future work because I think this policy is still effective even if it isn't automated. But maybe that is too much burden to impose temporarily.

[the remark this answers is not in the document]
kevincox: I think it would have to be building dependents and NixOS tests. Ideally it would be Hydra configured the same as the the channels that follow the eventual target branch. I think the main concern with Hydra is additional load.

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,124 @@
...
+- Create a tool for automatically marking broken packages as broken.
```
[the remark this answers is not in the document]
kevincox: That's great. It looks similar to what I had sketched out. I guess I can throw away my design and just make some tweaks to that script.

https://paper.dropbox.com/doc/nixpkgs-Merge-Queue-Easy-Mark-Broken--BHsIMZ2cLiBAvMfU66U0H9RIAg-A232tyO0zLDbUZ7zhAefX

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,124 @@
...
+
+# Summary
+
+This defines a policy for how to manage nixpkgs changes that break dependent
+derivations or tests. It specifies time frames and procedures so that
+maintainers of dependencies and dependents can know what to expect from each
+other when making breaking changes.
+
+This document is only focused on changes to derivation builds and tests it
+**does not** aim to make any opinion on changes that can break user machines or
```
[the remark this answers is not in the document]
kevincox: The fundamental reason I excluded this is that I don't think it is feasible to expect maintainers to know about all of the breakages that their change can cause in the dependent. At the end of the day I think the best we can do is say "If you liked it then you should have put a test on it".

spacekookie:
The RFC is open for shepherd nominations

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,124 @@
...
+
+This procedure should not result in a failing package build in the target branch at any
+point.
+
+# Drawbacks
+
+This delays the merge of “core” derivations as the author needs to wait for
+sub-packages to be tested and possibly fixed up to the 7 day threshold.
+
+# Alternatives ## Shorter grace period
```
asymmetric, line 97: Newline missing
```suggestion
# Alternatives

## Shorter grace period
```

lheckemann:
Nominations for the shepherd team are welcome! @kevincox maybe you have someone in mind who you think could help move this forward?

kevincox:
@blaggacao would you be interested and able?

kevincox:
@blaggacao would you be interested and able?

blaggacao:
I accept nomination.

blaggacao:
@nrdxp you have valuable experience (bors, hydra) to further this RFC, would you accept nomination for Shepherd?

Mic92:
We still need more nomination for this RFC to proceed.

rfcs/0000-nixpkgs-breaking-change-policy.md
kevincox: What are your suggestions? Should we have a list of tests that can't be marked broken? In that case I think we would need to block the initial PR,

rfcs/0088-nixpkgs-breaking-change-policy.md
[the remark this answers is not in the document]
jonringer: This could be a part of https://github.com/jonringer/basinix

jonringer:
I'll nominate myself as a shepard

kevincox:
Thanks Jonathan!

spacekookie:
Anyone else interested in being a shepherd for this RFC?

kevincox:
@nrdxp Can you send me your Matrix ID? I sent an invite to the email you have listed in maintainers-list.nix but I don't think it got through.

nrdxp:
yeah np, it's @timdeh:matrix.org

kevincox:
I believe that we have the requisite number of shepherds now. @blaggacao do you wish to be lead-shepherd?

blaggacao:
I can do that.

kevincox:
Hi everyone. I just pushed an update to the RFC that attempts to more explicitly answer the most common questions here. Please let me know if you still have any unaddressed questions.

I have also written a long-form article about where I hope this RFC allows us to go. While I hope that this RFC is self-describing and useful on its own it may help others to see the big picture spelt out as well. Please let me know if it would be helpful to bring any context from this post into this RFC: https://kevincox.ca/2021/08/21/nixpkgs-ci/

blaggacao:
Shepherds Meeting, Monday, 23th of August 2021

- How to practically derive new failures? @jonringer 
  - Open question for implementation of tooling. This RFC tries to agree on a policy that can then be implemented. @kevincox 
- Maybe start by suggesting a sample to build? @blaggacao 
  - 3 hours of build time / 100 packages. Only very general guidelines when interpreting the policy. Effectiveness of this policy depends heavily on tooling support. @kevincox
- Special role of passtru test? -> codify specific testing contracts in a complementary RFC. @jonringer 
  - e.g. "Define passtru tests to secure detection of a "breaking" change according to this RFC"
- Notification: Noise vs Signal ratio? @jonringer
  - Tooling will have to account for this problem in appropriate ways. The policy in this RFC ca be interpreted as a "principle right to be notified" under certain scenarios. @kevincox 
  - Maybe fundamentally conceptualize notification as a publisher endpoint that users can subscribe to and provide some optional additional notification middleware on top of that. @blaggacao 
- Automate detection of notifiable downstream packages shouldn't be too hard. @nrdxp
  - Though currently, none seem readily available.
- Until tooling is in place, it might be impractical to hard-enforce this policy. However, that does not hinder adoption of this RFC. @kevincox 
- Do we observe different message classes, such as "build failure", "test failure" that we should react/notify differently to? @blaggacao 
  - Possibly, but for the sake of simplicity they should all triggering the same 7 day grace period, if deemed eligible. @kevincox 


In addition to the above discussion the existing feedback was summarily reviewed. Most of it was reported as addressed. I'd encourage the parties that have given feedback in the past to revise the document under the light of the above discussion and the recent changes and potentially renew their questions / observations to the author.

Otherwise, in my opinion, we might be slowly approaching FCP.

jonringer:
I think most contributors who have tried to push for large changes, and have needed to wait for input from many individuals will agree that some convention about a grace period is needed. In general, this should only affect PRs against staging, as most PRs against master can generally be tested more thoroughly, and the effects of the changes usually carry much lower risk of breaking downstream packages (because there's significantly less of them).

I think we should move this into FCP.

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,155 @@
+---
+feature: nixpkgs-breaking-change-policy
+start-date: 2020-09-24
+author: Kevin Cox
+co-authors: (TBD)
+shepherd-team: @blaggacao, @jonringer, @nrdxp
```
Mic92, line 6:
```suggestion
shepherd-leader: @blaggacao
```

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,156 @@
...
+- Maintainers of dependants have a clear framework for how dependency breacking
+  changes will be handled.
+- NixOS channels will very rarely (ideally never) stall due to build or test
+  failures.
+
+This document is only focused on changes to derivation builds and tests it
+**does not** aim to make any opinion on changes that can break user machines or
+configurations. (While this is an interesting topic it is harder to put
+concrete rules on due to the difficulty of identifying these breakages. In
+order to be covered by this policy please but tests on your packages.)
```
Zimmi48, line 36:
```suggestion
order to be covered by this policy please put tests on your packages.)
```

lheckemann:
@jonringer has the shepherd team agreed to move this to FCP with disposition to accept? If so, let us know and we'll update the label.

blaggacao:
@lheckemann I confirm the shepherd team has agreed to move this to FCP with disposition to accept.

jonringer:
sorry, got buried under other notifications, yes. We would like to move forward, with disposition to accept.

Mic92:
We are now in the Final Comment Period (FCP). Unless any major objections are raised, this will be accepted and merged in 10 days from now, on November 13th!

rnhmjoj (approved):
I agree with pretty much everything proposed here. It's good to have some formal guidelines and documentation to point users to. Two considerations:

1. I think the grade period shouldn't be applied strictly. It should be a guideline or made flexible depending on the importance of the change and the sub-package: if a sub-package is not essential but important (some popular library, application, etc.) I'd rather delay a bit the update than to break it.  

2. I fear this could increase the burden on maintainers of mid to large packages (I mean those between being a leaf and core package). Those maintaining core packages are probably already well equipped and eat mass-rebuilds at breakfast, but not everyone else. It would be good if ofBorg, nixpkgs-review or some other new tool would automate doing some of the step in the proposed workflow before this RFC is enforced:
    - detecting broken packages
    - producing a list of contacts to ping the sub-maintainers
    - keep track of who replied
    - mark packages as broken

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,156 @@
...
+picked, as it is today, according to
+https://nixos.org/manual/nixpkgs/stable/#submitting-changes-commit-policy.
+
+1. The maintainer will prepare a PR with their intended changes.
+2. The maintainer should test a sample of dependent derivations to ensure that
+their change is not unnecessarily or unintentionally breaking. (Example: Ensure
+that a critical output file was not forgotten) Note that sometimes it **is**
+necessary to break all dependent packages, and the maintainer is not required
+to avoid this.
+    1. A sample of at least 100 packages or 3h of build time is recommend as a guideline.
```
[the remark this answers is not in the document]
jonringer: I'm okay with committers and active contributors to use my server: https://github.com/jonringer/server-configuration . If you're worried about compute. Since I'm not opening this up to the general public, it's not a solution, just a mitigation; but the offer is there.

Obviously, this is guideline is meant "within expectations" of a user's ability. Having some sort `nixpkgs-review` does make it a lot easier for reviewers to merge.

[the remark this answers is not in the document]
kevincox: This is why we added the time limit and called it a guideline. The idea is even if you only own a potato we hope that you can spare 3h of your potato or find someone else to do so for big changes. If you think 3h is still to muck to ask I'd be happy to consider other thresholds, it was picked mostly arbitrarily to provide a rough target.

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,156 @@
...
+    changes as well as any context that will help them resolve the breakages.
+    2. The sub-maintainers are expected to update the derivations that they
+    maintain within 7 days.
+        1. These changes should be merged to the target branch following the regular
+        nixpkgs PR process if they are backwards compatible. Otherwise they
+        will be merged into the branch of the breaking PR.
+5. The maintainer must provide a grace period of at least 7 days from when the
+sub-maintainers were notified.
+6. The maintainer must merge the target branch into their PR branch.
+7. The maintainer must mark all still-broken packages as
```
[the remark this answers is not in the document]
kevincox: Eventually it should be all packages. My not-so-secret goal is that one day every single package will be considered "channel blocking" unless it is marked broken, basically meaning that there are no unintentional breakages. 

It is expected that tooling will manage this. This policy is not going to be enforced until we have made it reasonably easy to comply. This RFC is a first step to agree agree upon a policy first, then we can build up the tooling to implement the policy.

7c6f434c:
Sounds like nothing FCP-breaking has happened.

rfcs/0088-nixpkgs-breaking-change-policy.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Goals
+
+The following procedure will be used for making breaking changes. It has a few
+primary goals:
+
+- Avoid putting more burden than necessary on the dependency maintainer. If the
+  maintainers of core derivations face toil proportionally to the number of
+  transitive dependencies they will quickly become overloaded. These
+  maintainers are arguably the most critical to Nixpkgs and  their load needs
```
Synthetica9, line 77:
```suggestion
  maintainers are arguably the most critical to Nixpkgs and their load needs
```

PR #89, opened 2021-03-14
[RFC 0089] Collect non-source package meta attribute

risicle:
~Don't know if there's a way of reserving a number - I just picked `0088` as it looked like the latest.~

~Edit: oh - is it the PR number?~ Fixed.

See https://github.com/risicle/rfcs/blob/ris-from-source/rfcs/0089-collect-non-source-package-meta.md

### DIFF
--- a/rfcs/0089-collect-non-source-package-meta.md
+++ b/rfcs/0089-collect-non-source-package-meta.md
@@ -110,7 +110,7 @@ The author is willing to spend a significant amount of time finding and marking
 non-source packages in nixpkgs.
 
 Addition of a simple accompanying boolean flag could allow the meaning of the
-`sourceProvenance` field to be switched to imply the declaration is
+`sourceProvenance` field to be changed to imply the declaration is
 "comprehensive" and that "source types" missing from the declaration are not
 present. This is something that could be added once a maintainer has thoroughly
 inspected a package, but should not place extra burden on someone wanting to
--- a/rfcs/0089-collect-non-source-package-meta.md
+++ b/rfcs/0089-collect-non-source-package-meta.md
@@ -40,8 +40,8 @@ For some users, these concerns are enough to deter them from using Nix entirely.
 Add a new `meta` attribute to non-source-built packages, `sourceProvenance`.
 The value of this attribute being a list of at least one entry from a
 collection of possibilities maintained in `lib.sourceTypes`. These possibilities
-can be of opaque type, but should have entries to represent at least "built from
-source", "binary native code", "binary bytecode" and "binary firmware".
+should have entries to represent at least "built from source", "binary native
+code", "binary bytecode" and "binary firmware".
 
 Packages built from source can be left as-is with the assumption of a missing
 attribute being the equivalent of `[ lib.sourceTypes.fromSource ]`.
--- a/rfcs/0089-collect-non-source-package-meta.md
+++ b/rfcs/0089-collect-non-source-package-meta.md
@@ -3,8 +3,8 @@ feature: collect-non-source-package-meta
 start-date: 2021-03-14
 author: Robert Scott
 co-authors: (find a buddy later to help out with the RFC)
-shepherd-team: (names, to be nominated and accepted by RFC steering committee)
-shepherd-leader: (name to be appointed by RFC steering committee)
+shepherd-team: @asymmetric, @aforemny, @alyssais
+shepherd-leader: @asymmetric
 related-issues: (will contain links to implementation PRs)
 ---
 
--- a/rfcs/0089-collect-non-source-package-meta.md
+++ b/rfcs/0089-collect-non-source-package-meta.md
@@ -37,29 +37,43 @@ For some users, these concerns are enough to deter them from using Nix entirely.
 # Detailed design
 [design]: #detailed-design
 
-Add a new `meta` attribute to non-source-built packages, `fromSource = false`.
-Leave other packages as-is with the assumption of a missing attribute meaning
-`true`.
+Add a new `meta` attribute to non-source-built packages, `sourceProvenance`.
+The value of this attribute being a list of at least one entriy from a
+collection of possibilities maintained in `lib.sourceTypes`. These possibilities
+can be of opaque type, but should have entries to represent at least "built from
+source", "binary native code", "binary bytecode" and "binary firmware".
+
+Packages built from source can be left as-is with the assumption of a missing
+attribute being the equivalent of `[ lib.sourceTypes.fromSource ]`.
+
+Multiple values present in a package's `sourceProvenance` would be used to
+mean that the package contains parts that fall under each of these categories.
+However, a "source type" not appearing in a package's `sourceProvenance` would
+_not_ necessarily mean that the package _doesn't_ contain parts which fall
+under that category - it could simply mean that a package hasn't been fully
+assessed yet. See "Future work" for discussion of adding the ability to make
+such a "comprehensive" declaration.
 
 Add a mechanism to allow `.nixpkgs/config.nix` to specify
 `allowNonSource = false` to prevent use of these packages in a similar manner
-to `allowUnfree`.
+to `allowUnfree`. An `allowNonSourcePredicate` parameter would allow the
+distinction to be customized, but the default predicate should take into account
+the possible hierarchical nature of `lib.sourceTypes` entries.
 
 # Alternatives
 [alternatives]: #alternatives
 
-I might have been tempted to collect the inverse, i.e. `isBinary = true` but
-this runs into problems with clunky terminology. In my mind, the kind of package
-that fails the transparency/malleability tests goes beyond what many people
-would argue is "a binary". For instance, many (most?) java packages in nixpkgs
-simply pull opaque `.jar`s - if not for their own app, they pull `.jar`
-dependencies from maven. These are not transparent or malleable, but it's quite
-an obtuse and disputable use of the term "binary" to describe them as such.
+The original proposal used a simple boolean attribute to declare whether the
+package contains any binary parts, mostly in an attempt to avoid having
+to go down the route of devising and debating an ontology of source types. This
+was deemed by the shepherding meeting to not embrace extensibility enough.
 
-I decided that those packages which _did_ pass these transparency/malleability
-tests had more in common than those that don't: that they are "from source", a
-form where users have as much ability to inspect and alter the result as the
-original author did.
+Another suggestion involved using a single value from a collection of "source
+types" to describe the source provenance in an effort to avoid complexity, but
+this appeared to have the disadvantages of requiring an ontology to be agreed
+upon and still yet not providing sufficient flexibility to cover many cases.
+The missing ability to express multiple provenances might even encourage
+maintainers to proliferate source types that represent combinations of others.
 
 There already exists a rather informally-applied convention of adding a `-bin`
 suffix to the package names of "binary packages". This is non-ideal because:
@@ -81,23 +95,23 @@ packages.
 # Drawbacks
 [drawbacks]: #drawbacks
 
-- Some maintainers may be upset by having their packages marked as
-  `fromSource = false`.
-- It could spur us to disappear into endless navel-gazing conversations about
-  what really counts as "from source" and what doesn't.
-- On the other hand, _not_ discussing where the line stands thoroughly enough
-  could cause the flag to be over-applied and thus become useless. Should we be
-  compiling all our fonts where e.g. fontforge files are available? If all of
-  these got marked as `fromSource = false`, all of a sudden users with
-  `allowNonSource = false` set may end up with no installable desktop.
+- It could spur us to disappear into endless navel-gazing conversations over
+  the `lib.sourceTypes` ontology.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-Exact attribute names are open for debate.
+Exact attribute names and contents of `lib.sourceTypes` are open for debate.
 
 # Future work
 [future]: #future-work
 
 The author is willing to spend a significant amount of time finding and marking
 non-source packages in nixpkgs.
+
+Addition of a simple accompanying boolean flag could allow the meaning of the
+`sourceProvenance` field to be switched to imply the declaration is
+"comprehensive" and that "source types" missing from the declaration are not
+present. This is something that could be added once a maintainer has thoroughly
+inspected a package, but should not place extra burden on someone wanting to
+simply flag up that they have spotted some binary element in the package.
--- a/rfcs/0089-collect-non-source-package-meta.md
+++ b/rfcs/0089-collect-non-source-package-meta.md
@@ -1,0 +1,103 @@
+---
+feature: collect-non-source-package-meta
+start-date: 2021-03-14
+author: Robert Scott
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Collect and maintain a new `meta` attribute in packages allowing users to easily
+identify and manage their preference for binary (more broadly "non-source")
+packages.
+
+# Motivation
+[motivation]: #motivation
+
+Different users have different expectations from a software distribution. We
+acknowledge that much with the collection of license information and the
+existence of the `allowUnfree` nixpkgs option, much as Debian maintains a
+separate `-nonfree` repository.
+
+Similarly, there are a number of different reasons users may have to disfavour
+those packages not built-from-source:
+
+- Transparency: an ever-growing concern with more focus than ever on
+  supply-chain attacks.
+- Malleability: being able to conveniently override packages with patches or an
+  altered build process is a key advantage of Nix, and for nixpkgs maintainers
+  it's not generally possible to backport security fixes to binary packages.
+
+For some users, these concerns are enough to deter them from using Nix entirely.
+
+# Detailed design
+[design]: #detailed-design
+
+Add a new `meta` attribute to non-source-built packages, `fromSource = false`.
+Leave other packages as-is with the assumption of a missing attribute meaning
+`true`.
+
+Add a mechanism to allow `.nixpkgs/config.nix` to specify
+`allowNonSource = false` to prevent use of these packages in a similar manner
+to `allowUnfree`.
+
+# Alternatives
+[alternatives]: #alternatives
+
+I might have been tempted to collect the inverse, i.e. `isBinary = true` but
+this runs into problems with clunky terminology. In my mind, the kind of package
+that fails the transparency/malleability tests goes beyond what many people
+would argue is "a binary". For instance, many (most?) java packages in nixpkgs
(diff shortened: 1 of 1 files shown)


ryantm:
Yes, it's the GitHub issue number.

rfcs/0089-collect-non-source-package-meta.md
```diff
@@ -0,0 +1,103 @@
...
+# Detailed design
+[design]: #detailed-design
+
+Add a new `meta` attribute to non-source-built packages, `fromSource = false`.
+Leave other packages as-is with the assumption of a missing attribute meaning
+`true`.
+
+Add a mechanism to allow `.nixpkgs/config.nix` to specify
+`allowNonSource = false` to prevent use of these packages in a similar manner
+to `allowUnfree`.
```
dotlambda, line 46: There should also be `allowNonSourcePredicate`.

rfcs/0089-collect-non-source-package-meta.md
```diff
@@ -0,0 +1,103 @@
...
+- Malleability: being able to conveniently override packages with patches or an
+  altered build process is a key advantage of Nix, and for nixpkgs maintainers
+  it's not generally possible to backport security fixes to binary packages.
+
+For some users, these concerns are enough to deter them from using Nix entirely.
+
+# Detailed design
+[design]: #detailed-design
+
+Add a new `meta` attribute to non-source-built packages, `fromSource = false`.
```
dotlambda, line 40: This doesn't seem flexible enough:
* Along the lines of the `unfreeRedistributableFirmware` license, there should be a way to label binary packages used in bootstrapping as such. Most users who don't want binary packages will be okay with their use in bootstrapping.
* Binary fonts (mentioned below) or graphics might also be acceptable.
Do you think `allowNonSourcePredicate` is enough for the latter?

risicle: Well, this hints slightly towards making it non-boolean, but if so, what should the options be? I think I considered finer classifications but saw it as something this could evolve into if/once we come to understand the problem better.

samueldr: A list of "types"? Because a package might contain different types!

```
{
  meta.sourceProvenance = with lib.sourceTypes; [
    sourceCode
    sourceFirmware
  ];
}
```

For an hypothetical driver with a code part, and a binary firmware part.


```
{
  meta.sourceProvenance = with lib.sourceTypes; [
    sourcePrebuiltFont
  ];
}
```

For a pre-built font.


(The naming here is clunky, I hope if gets the point across.)

dotlambda: I think a list of types is usually too complex. Hence I suggest usually not using them, even though some packages might require them.
We sometimes have lists of licenses, though without clear semantics. For this attribute, a list should be interpreted as being built from *all* of those types combined.

I would call the attribute `meta.builtFrom`.
A normal `xyz-bin` package would, even if parts of it are built from source, have
```
meta.builtFrom = lib.builtFrom.binary
```
A binary font could have
```
meta.builtFrom = lib.builtFrom.binaryAsset
```
And a bootstrapping package would have
```
meta.builtFrom = lib.builtFrom.binaryBootstrap
```
The default would be
```
meta.builtFrom = lib.builtFrom.source
```

samueldr: While I agree that they are more complex, I don't think it's an issue.

Otherwise we're pushing the complexity into pre-declaring all possible combinations as part of the library representing the different types.

We need granularity here. Otherwise we'll have issues deciding which descriptor to use, and end up using the wrong generic "it's a binary", which will make filtering against or for a specific descriptor needlessly harder.

kevincox: We can do something "weird" and have `lib.builtFrom.source = ["source"]` so that you can easily do one or `++` two together. Or maybe better is `lib.buildFrom.source = {source = true;}` and then you can `//` them together but it is a bit easier for a predicate to filter the ones that you are okay with (and removes the irrelevant ordering).

Also I think instead of `builtFrom` it probably makes sense to call it `nonSourceComponents` and list the types of things that were not built from source where "types" could be things like `code`, `assets`, `docs` and similar.

I see this is somewhat against the "Why not isBinary?" below but I think it kinda agrees. `nonSourceComponents = {}` is the "purest" form and well understood. Then you just have to document the deviations.

marius851000:
Will this also concerned vendored code ? If the option of using a list for the provenance of source information, might it be relevant to include information if the source have vendored code (I here mean code that isn't created by the author of the source file, and that is unmodified, so that may be compiled in another unit, like for example rust package, with the VendorSha256 entry). It might be relevant to indicate this, as it mean overriding the library in nixpkgs may not update other package that vendor the library.

risicle:
I think vendored code is beyond what I'd consider scope creep for this proposal.

spacekookie:
This RFC is open for shepherd nominations now btw!

alyssais:
Nominate myself.

rfcs/0089-collect-non-source-package-meta.md
alyssais: I really like samueldr's idea here.  Another benefit of it would be that I'd be able to say "I don't want proprietary software on my computer, but it's okay if data files are CC-BY-ND" or something.

risicle: It's nice, my worry is just what we'd _actually_ end up with given humans being humans, most people probably not _that_ invested in understanding a complex tagging scheme well enough to perfectly represent the situation for a particular package.

This concern probably comes from my background in openstreetmap and the long, heated discussions that take place on the "tagging" mailing list debating a scheme that can perfectly represent most every situation, yet which ends up bearing little relation to what people _actually_ map with in reality, because it's too complex and verbose.

I certainly think it's important to make the common cases have very concise representations, and also allow both coarse and fine granularity. If only fine granularities are allowed, it will deter most people from bothering to add an annotation at all. Coarse data is better than nothing and can always be refined.

Mic92:
We still need more nomination for this RFC to proceed.

asymmetric:
I’d like to be a shepherd on this one.

aforemny:
I nominate myself

Mic92:
We decided on the shepherd team: @asymmetric, @aforemny and @alyssais
with @asymmetric as shepherd leader.
@asymmetric could you propose a meeting (i.e. with https://www.when2meet.com/)

asymmetric:
@risicle could you edit the body of the PR and add a link to the [rendered text](https://github.com/risicle/rfcs/blob/ris-from-source/rfcs/0089-collect-non-source-package-meta.md)?

asymmetric:
@aforemny @alyssais here we go: https://www.when2meet.com/?12517947-RE67W

risicle:
Have added my availability, but I am not a critical participant - I'd probably just sit and make awkward jokes.

aforemny:
> @aforemny @alyssais @risicle here we go: https://www.when2meet.com/?12517947-RE67W

Thanks! I've added my availability just now

asymmetric:
Thanks everyone. As it stands, there is no single slot where we would all be available.

Can anyone shift their availability around, to make that possible?

If not, I'll schedule a meeting where all shepherds can be present (sorry @risicle)

aforemny:
> Thanks everyone. As it stands, there is no single slot where we would all (shepherds + author) be available.
> 
> Can anyone shift their availability around, to make that possible?
> 
> If not, I'll schedule a meeting where all shepherds can be present (sorry @risicle)

Unfortunately, I didn't see an opportunity where I could shift my availability so that all could be available. I have broadened my availability a bit more to help make that easier for someone else. :crossed_fingers:

risicle:
This is fine by me.

asymmetric:
OK, let's meet on Wednesday September 1 @ 3pm, Amsterdam time.

I'll post a jitsi link some time before the meeting.

rfcs/0089-collect-non-source-package-meta.md
asymmetric: I tend to agree with @risicle and @dotlambda here, in that I think we should at least start with a rougher schema, and wait until we've seen if we actually need a more refined one. I resonate with the risk @risicle sees, otherwise.

@dotlambda's proposal matches what I would do:

- If a package has both binary and source it will be marked as binary (i.e. we prefer a rougher schema, and err on the side of overapplying the "binary" label)
- it special cases bootstrap binaries and fonts
- Adds an `allowNonSourcePredicate`

I think this would be a great first step, and it would address the 2 points raised in the RFC:

- supply chain attacks
- patching

Other concerns, like wanting to only run free software + special casing some data files based on their license, I would leave out of this RFC, and potentially address in a follow-up one.

asymmetric:
If anyone else wants to participate, let me know, I'll send the link via email (to the email in your GH profile).

fzakaria:
Love this idea.

alyssais:
# Report from Shepherd Meeting, 2021-09-03

A couple of days ago we had a shepherd meeting.

## Category-based filtering

Our main concern was that as it stands, while this RFC is a step in the right direction, it isn't very useful by itself.  Anybody who wanted to restrict binary software on their computer would have to maintain a large allowlist in their predicate, because there's no way to exclude broad categories of things like fonts or bootstrapping packages.  We understand your concern about perfection being in the enemy of good when designing tagging schemes, but we feel there's a need for at least some form of more fine-grained filtering for the feature to be useful.

We evaluated all the potential solutions for this.

### List of types × binary-ness

This is the idea [suggested](https://github.com/NixOS/rfcs/pull/89#discussion_r593963892) by @samueldr.  This would be a sufficiently flexible solution to the problem.  One concern with it is that it looks like it does two things in the same attribute — knowing that a package contains a font is useful in situations where you don't care about whether it's built from source.

### Single value

This is the idea [suggested](https://github.com/NixOS/rfcs/pull/89#discussion_r595808264) by @dotlambda.  We identified a risk with this approach, which is that if we later needed to distinguish between two variations within a single type, we wouldn't be able to do that in a backwards-compatible way.

For example, suppose we start with one of the types being `binaryAsset`.  Later on, it becomes clear that we want to distinguish between fonts (which are really just programs, even though some people don't think of them that way), and other things like text files that really are just data.  So we invent two new categories, `binaryFont` and `binaryData`, and re-categorize all the categories marked as `binaryAsset` appropriately.  The problem is that now anybody who has an `allowNonSourcePredicate` like `pkg: pkg.meta.buildFrom == lib.builtFrom.binaryAsset` will suddenly find that all of the binary asset packages they were previously using will no longer pass the predicate check.  Even doing a deprecation warning for this would be tricky, and would have to happen somewhere inside `mkDerivation`.

In contrast, with the "List of types × binary-ness" approach, it would be easy to add more refined categories over time.  We would simply change `meta.sourceProvenance = with lib.sourceTypes; [ binaryAsset ]` to e.g. `meta.sourceProvenance = with lib.sourceTypes; [ binaryAsset binaryFont ]`,and document that `binaryFont` → `binaryAsset`.  To preserve compatibility with people still using `binaryAsset` in their predicate, we could even add a meta check ensuring that `binaryAsset` was also present in any list containing `binaryFont`.

### Extra meta attributes

Another idea I brought up during the meeting was to introduce separate boolean `meta` attributes describing what a package contains, e.g. `hasFont` or `hasExecutable`.  This would end up functioning similarly to "List of types × binary-ness".  The advantages to doing it this way would be that it doesn't combine type of artifact and built-from-source-ness into a single attribute, and that this approach would complement the `fromSource = false` approach this RFC currently proposes.  In fact, it could even be added later, after this RFC has been accepted.  It is also easily extensible in future, avoiding the issues with the single value approach.

The disadvantage to doing it this way is that it doesn't allow for distinguishing between cases where e.g. a package installs programs built from source code, but also fonts that were distributed as binaries in its source tarball.

## Next Steps

@risicle, we think that this RFC needs to at least consider how further filtering could help make this useful to users of Nixpkgs.  That doesn't mean that this RFC necessarily needs to define that mechanism for further filtering, though.  Just having `fromSource = false` marked appropriately in Nixpkgs would be a good step forward, even though it wouldn't be immediately useful to lots of people who're interested in this.

Specifically, we recommend you either

* Incorporate the list of types × binary-ness into the RFC
* Incorporate extra meta attributes into the RFC
* Proceed with the RFC as it stands, but mention in the RFC that an extra meta attributes approach would be the recommended way to add further filtering in future (because extra meta attributes builds on top of the meta attribute proposed in this RFC).

Additionally, the RFC should incorporate the `allowNonSourcePredicate`.

oxij:
I agree with the shepherds.

I assume the above implies that all bootstrap tools would be honestly marked as "bootstrap binary executable" and `allowNonSourcePredicate` would allow such things (and probably binary data and fonts) by default?

Also, let's not forget binary firmware.

alyssais:
During the meeting, I was arguing in favour of extra meta attributes, but since the meeting, I've been thinking about this a lot and I'm increasingly in favour of list of types × binary-ness.  I think that it would actually be useful to know binary-ness for different types of material installed by the package, which extra meta attributes can't provide.

But none of this changes my view that we shouldn't let perfection be the enemy of good, and adopting this RFC pretty much as-is right now, and adding extra meta attributes later, would be good enough.

Mic92:
So the state of this RFC is that the shepherd team expect changes from @risicle?

risicle:
Say wha?

aforemny:
> Say wha?

During the discussion here, it was identified that the approach outlined in this RFC is of limited usefulness with respect to building a system that disallows non-source packages. Most importantly, because there is no general way of excepting binary fonts and bootstrapping packages from the non-source predicate: one would have to declare each exception one-by-one.
As everyone so far agreed, _not_ marking those packages as "from source" when they clearly aren't only to circumvent this issue is not a solution, because it would allow binary packages to creep into the system that should be built _from source_.

To remedy this situation, we think it would be very helpful if this RFC would either

(a) adopt an implementation that would allow for such broader exceptions (ie. samueldr's suggestion, or some form of tagging), or
(b) is being updated to mention this problem and the alternate approaches without changing its proposed implementation, possibly outlining them as "future work."

We would also like very much if this RFC includes the `allowNonSourcePredicate` from the get-go.

I think [@alyssais's comment](https://github.com/NixOS/rfcs/pull/89#issuecomment-912994126) explains all this in greater detail. Please do reply if you have any questions about this, or if there is anything with our conclusion that you would like to discuss!

risicle:
Ok I will probably do (b) through expanding the "drawbacks" section as (a) rather puts the onus on me to make decisions about potentially complex schemas that I was rather trying to avoid in the first place. Anyone is still free to create their own RFC with their rival scheme fleshed out.

risicle:
While I agree that the "list of types" approach is better for the perfect capturing of a situation, what I don't understand about it is where the ability to express coarseness comes from.

I'm willing to bet that for 90% of packages that ever get this attribute, the only thing we will want to express about them is "it's a downloaded binary". Or perhaps "it's from source" in celebration of converting it from a former binary package. But if I mark a package with `with lib.sourceTypes; [ sourceCode ];`, am I also saying that it _doesn't_ include binary fonts (or... icons that weren't rendered from the source SVGs etc.)? I'm frankly not going to go through packages checking for that, but there's no way of expressing that this is a coarse classification I've made.

Using set-membership to denote the status of each type only allows us to express strict binary situations for each of those types.

risicle:
Just thought.. the above concerns could probably be addressed with an extra boolean flag denoting whether the source type declaration is "comprehensive" or not. So I'll probably adopt the `lib.sourceTypes` list approach and update this.

risicle:
Have updated with details of the multi-valued approach and `allowNonSourcePredicate`. Hopefully it still makes sense.

rfcs/0089-collect-non-source-package-meta.md
```diff
@@ -0,0 +1,117 @@
...
+Exact attribute names and contents of `lib.sourceTypes` are open for debate.
+
+# Future work
+[future]: #future-work
+
+The author is willing to spend a significant amount of time finding and marking
+non-source packages in nixpkgs.
+
+Addition of a simple accompanying boolean flag could allow the meaning of the
+`sourceProvenance` field to be switched to imply the declaration is
```
asymmetric, line 113:
```suggestion
`sourceProvenance` field to be changed to imply the declaration is
```

Switched read to me as meaning "changing to its opposite". YMMV :)

rfcs/0089-collect-non-source-package-meta.md
```diff
@@ -0,0 +1,117 @@
...
+  altered build process is a key advantage of Nix, and for nixpkgs maintainers
+  it's not generally possible to backport security fixes to binary packages.
+
+For some users, these concerns are enough to deter them from using Nix entirely.
+
+# Detailed design
+[design]: #detailed-design
+
+Add a new `meta` attribute to non-source-built packages, `sourceProvenance`.
+The value of this attribute being a list of at least one entriy from a
```
asymmetric, line 41:
```suggestion
The value of this attribute being a list of at least one entry from a
```

rfcs/0089-collect-non-source-package-meta.md
```diff
@@ -0,0 +1,117 @@
...
+collection of possibilities maintained in `lib.sourceTypes`. These possibilities
+can be of opaque type, but should have entries to represent at least "built from
```
asymmetric, line 43: What does opaque type mean in this context?

asymmetric (commented):
Thanks @risicle for the work you're putting into this. the RFC reads great!

Just to clarify, apart from getting this RFC through the finish line, are you still willing to go through (parts of) nixpkgs and mark non-exclusively-source packages as such?

I don't think this is a requirement, by the way - just wondering what practical outcomes we can expect to come out of this RFC. Even just arriving at consensus on the new attribute would be great.

rfcs/0089-collect-non-source-package-meta.md
risicle: I think I meant "you shouldn't expect any specific behaviour from it other than being able to check for equality". At time of writing I think I was leaving open the possibility of having values which were able to understand their own hierarchy to some degree, though the more I think of it, it would probably be better to embed all notions of hierarchy in the `allowNonSourcePredicate` function to allow customization there too.

risicle:
Yes, I'm willing to do a _non-comprehensive_ pass of nixpkgs looking for the low-hanging fruit and more obvious cases.

rfcs/0089-collect-non-source-package-meta.md
asymmetric: I would then rather change this to remove the reference to opaqueness, since it's rather.. opaque like this.

Do you agree?

asymmetric (changes requested):
I think that, apart from the nit I mentioned in this review, this is good to go.

@risicle can you make a quick edit addressing this point? I think after that, we can move ahead.

asymmetric:
Also, sorry for the delay everyone, life took over.

risicle:
Done.

asymmetric:
I propose this RFC move to FCP, with intention to merge.

aforemny:
> I propose this RFC move to FCP, with intention to merge.

I second that! :-)

alyssais:
I agree, let's go to FCP!

asymmetric:
@NixOS/rfc-steering-committee all shepherds have signaled their intention to move to FCP

risicle:
An initial implementation: https://github.com/NixOS/nixpkgs/pull/161098

PR #91, opened 2021-04-11
[RFC 0091] Pure Object Prototypes

fare:
In this [RFC 0091](https://github.com/fare-patches/rfcs/blob/pop/rfcs/0091-pure-object-prototypes.md), we propose to add Pure Object Prototypes, or POP, an object system, to the nixpkgs library. POP improves upon current Nix extension systems by supporting multiple inheritance and default values. These improvements solve modularity issues with existing extension systems. We actually use POP in production in a fork of nixpkgs to support local changes to some packages, by multiple people, with dependencies between changes.

POP can be also made to interoperate both ways with existing extension systems, with simple adapters.

### DIFF
--- a/rfcs/0091-pure-object-prototypes.md
+++ b/rfcs/0091-pure-object-prototypes.md
@@ -238,11 +238,13 @@ There is also a documentation file
 [pop.md](https://github.com/MuKnIO/nixpkgs/blob/devel/lib/pop.md)
 that discusses the background for the design of POP.
 
-Finally, I have been writing an essay that I intend to submit to some
-academic programming language conference, wherein
-I reconstruct the principles of Object-Oriented Programming
-based on my experience with POP and a similar library I wrote in Scheme:
-[Pure Object-Orientation, Functionally](https://github.com/metareflection/poof).
+Finally, I wrote an academic paper that was accepted at the
+Scheme and Functional Programming Workshop 2021,
+that explains the design behind POP.
+In this paper, I reconstruct the principles of Object-Oriented Programming
+based on my experience with POP and a similar library I wrote in Scheme,
+as well as an extensive bibliographical research on the topic:
+[Prototypes: Object-Orientation, Functionally](https://github.com/metareflection/poof).
 
 ## Notable though minor incompatibility
 
@@ -299,7 +301,7 @@ useCcl = {
 };
 ```
 
-Meanwhile, another extension might be about using a graphical debugging,
+Meanwhile, another extension might be about using a graphical debugger,
 including a graphical inspector:
 ```
 clGraphicalDebugging = {
@@ -429,10 +431,53 @@ extra debugging information.
 
 ## Interaction with modules
 
-NixOS has a notion of modules that has its own extensibility mechanism
-well distinct from the regular extension system.
-Someone ought to learn enough of both POP and modules to tell how it might be possible (or not)
-to have a unified mechanism that can handle modules as well as regular scopes, etc.
+NixOS has a notion of [modules](https://nixos.org/manual/nixos/stable/index.html#sec-writing-modules)
+that has its own extensibility mechanism well distinct from the usual extension systems.
+POP as it is proposed here does not attempt to compete with this notion of modules.
+Yet, it is conceivable to extend POP to build a better (though not 100% compatible)
+variant of NixOS modules.
+
+- The current module system has an `imports` system the effect of which is not very well defined.
+  That's one case where POP's multiple inheritance could give a good meaning
+  to defining dependencies between module definitions such that
+  each is combined only once in dependency order into the mix.
+
+- The NixOS module system actually defines three kinds of entities:
+  modules, configurations, types, each with its own semantics, merging algorithm, etc.
+  (Though, if you squint, modules could a special case of types, and
+  configurations be the type being define by the module system.)
+  The definition for a replacement of each of these entities would require
+  a different use of POP's meta-object protocol.
+
+- The merging algorithms used for modules, configs and types are much more elaborate
+  than the one used by POP by default (`mergeAttrset`), but
+  POP's meta-object protocol (`instantiateMeta`) supports specifying
+  an alternate `mergeInstance` mechanism (though currently the "toplevel" functions
+  `pop`, `basePop`, `kPop`, etc., all use the default mechanism and won't let you override it).
+  In this sense modules or something module-like could be done in POP,
+  and benefit from POP's features and approach.
+
+- Similarly, the `warning` mechanism could be achieved in a POP setting
+  by using the `topProto` feature of `instantiateMeta` to handle warnings
+  at the end of config merging. The same mechanism could more generally `finalize`
+  the values being defined from the result of the merges to something "user-visible".
+
+- POP has a system of dependencies and defaults that is a partial replacement for
+  the priority system of module options: the `defaults` in POP replace low-priority settings,
+  and the multiple inheritance of POP ensures that settings by children objects
+  (importing modules) will override those of parent objects (imported modules),
+  without having to assign priority numbers.
+  My experience with various configuration systems (e.g. SYSV init) is that
+  dependencies are a much more user-friendly way of managing priorities
+  than requiring users to put arbitrary priority numbers.
+  However, usability might require that we extend Nix to allow for cheap cycle detection
+  (see above regarding Missing Primitives for Performance),
+  so we can provide users with better error messages.
+
+Thus, while POP in its current state is not a replacement for the module system,
+it could serve as the basis for such a replacement,
+one that would provide better dependency management
+and a more systematic, more understandable, design.
 
 ## Backward compatibility wrappers
 
--- a/rfcs/0091-pure-object-prototypes.md
+++ b/rfcs/0091-pure-object-prototypes.md
@@ -22,7 +22,6 @@ local changes to some packages, by multiple people, with dependencies between ch
 POP can be also made to interoperate both ways with existing extension systems,
 with simple adapters.
 
-
 # Motivation
 [motivation]: #motivation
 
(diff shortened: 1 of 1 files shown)


rfcs/0091-pure-object-prototypes.md
```diff
@@ -0,0 +1,418 @@
...
+
+If extension `x` depends on extension `z`,
+then the author of `x` might be tempted to not define `x`,
+but instead define `zx = composeExtensions Z X`,
+so his users (starting with himself) don't have to
+manually do the composition every time they use `x`.
+However, if, independently, the author of extension `y` also depends on `z`,
+and also defines `zy = composeExtensions z y`,
+and if some later poor user wants to use both `zx` and `zy`,
+he can use `composeExtensions zx zy`, because `z` is then applied twice,
```
kevincox, line 80:
```suggestion
he can't use `composeExtensions zx zy`, because `z` is then applied twice,
```

rfcs/0091-pure-object-prototypes.md
```diff
@@ -0,0 +1,418 @@
...
+- Performance:
+  Computing the `precedenceList` for an object in general is
+  linear in the size of the inheritance graph (nodes plus arcs),
+  i.e. the number of transitive `supers` entries.
+  In the worst case, that's up to quadratic for each new object,
+  which is cubic in the total number of objects.
+  That can be slow.
+  That said, the same "computation" would have to be done by hand
+  by users who would want to achieve the same effect without automation,
+  so it's not really a drawback *given the desired effect*.
```
7c6f434c, line 366: … but automating handling this might encourage larger dependency graphs, making performance an issue.

(also, manual handling usually needs to be done once a few months with the result of the reevaluation «cached» in the form of an expression — here the cost is per-evaluation)

rfcs/0091-pure-object-prototypes.md
```diff
@@ -0,0 +1,418 @@
...
+`fix'`, `makeExtensibleWithCustomName`, `makeExtensible`,
+`makeOverridable`, `makeScope`, `makeScopeWithSplicing`,
+plus add-ons like `overrideDerivation`.
+Then, many languages and subsystems invent their own variant extension system.
+And while I haven't looked at modules in details,
+I'm told they also include a more elaborate variant in the same family.
+
+This maze of mostly similar yet subtly different yet incompatible constructs
+raises the barrier to entry to learning and using Nix.
+Therefore, to end this anarchy... let's create another extension system!
```
RosarioPulella, line 42: Maybe it would be more appropriate to say "to end this fragmentation"? [Anarchy](https://en.wikipedia.org/wiki/Anarchy) Seems like a imprecise term for an RFC and inaccurate term in this context.

RosarioPulella (commented):
Can you add a rendered link to the rfc?

rfcs/0091-pure-object-prototypes.md
```diff
@@ -0,0 +1,418 @@
...
+if not implicit conversion.
+
+# Future work
+[future]: #future-work
+
+- Use and enjoy POP in more subsystems of nixpkgs than Gerbil.
+- See how POP, with or without further improvements,
+  can or cannot fully replace other Nix extension systems.
+- In particular, see how POP, with or without further improvements,
+  may improve the situation for modules.
```
7c6f434c, line 416: I would classify comparison with the current module system as an unresolved question…

rfcs/0091-pure-object-prototypes.md
fare: I added these two remarks in the drawback discussion.

rfcs/0091-pure-object-prototypes.md
fare: Done!

rfcs/0091-pure-object-prototypes.md
fare: I added a paragraph there, too.

fare:
I added a link to the rendered page in the intro and in the issue discussion.

rfcs/0091-pure-object-prototypes.md
```diff
@@ -0,0 +1,442 @@
...
+## Invent an even better object system
+
+We can invent an even better object system and still include this one in nixpkgs in the meantime.
+
+## Implement objects at a deeper level
+
+We can make objects part of the language at a deeper level, as in Jsonnet,
+for better performance and/or better error messages.
+But if we can afford a user-level implementation, that is more flexible, and
+we don't deep magic to fix or extend the object system.
```
7c6f434c, line 408: «need» missing?

rfcs/0091-pure-object-prototypes.md
```diff
@@ -0,0 +1,442 @@
...
+  (multi-methods, method combinations, meta-object protocol, etc.),
+  at which point the design may have to evolve.
+  The library could be put in a staging area until it's considered stable.
+  See [RFC 0082 lib.experimental](https://github.com/NixOS/rfcs/pull/82).
+
+- Performance:
+  Computing the `precedenceList` for an object in general is
+  linear in the size of the inheritance graph (nodes plus arcs),
+  i.e. the number of transitive `supers` entries.
+  In the worst case, that's up to quadratic for each new object,
```
7c6f434c, line 365: Can we afford requiring `supers` entries to be objects with already computed `precedenceList`? Of course, built-in linear-time `uniqList` in Nix (current state in Nixpkgs: `uniqList` deprecated, but now there is `unique` which is roughly the same, and it is also quadratic…) could make things even better, but oh well

fare: We already require `supers` to be objects with already computed `precedenceList`, unless they are attrset that stand in for a trivial object with an empty list of supers.

Actually, that's one place that's problematic when there are circularities: we get an unhelpful error message, because we don't do a first circularity-detection pass, which would be quadratic or worse without some kind of eq set.

I added an unresolved issue on that topic.

Profpatsch:
I don’t think this is an actionable change that people feel there is a need for.

Can you structure the RFC in less of a rambling manner and go from a set of concrete problems in the existing code (with lots of examples) to how POP solves these issues in a cleaner manner?

For examples the [“Dependencies between extensions“](https://github.com/fare-patches/rfcs/blob/f6c951f7c84057cba0caaa56238cfabfd1b2dfc7/rfcs/0091-pure-object-prototypes.md#dependencies-between-extensions) section is horrid, there is no reason for the reader to try and understand it, since you have not even given a reason they should be interested in this proposal.

I think this needs a serious rework before anybody can be expected to read and understand it.

FRidh:
I recall your original PR, the other RFC and your feedback on https://github.com/NixOS/rfcs/pull/83 and now I went through this one. It reads a bit like a blog and I agree with @Profpatsch you need to make a stronger case earlier on. Start for example with a list of methods we have now (you did that) and a brief list with significant limitations they have that motivate switching to POP. I do now better understand what problems you resolve, but before adding a new system, I think it is important you also understand the existing ones, which includes the NixOS modules system.

- In the past we had `composableDerivation` which had a slight overlap with this. 
- There is a gist with some ideas with [language changes](https://gist.github.com/edolstra/29ce9d8ea399b703a7023073b0dbc00d)

If you want to propose POP for writing package sets, then I think it is important that it is able to do what we need it to do. That is what you need it for, dependencies between extensions, but also what is listed in https://github.com/NixOS/rfcs/pull/83 (which is far from ready...).

lheckemann:
This RFC is open for nominations; @fare do you have anybody in mind who could be interested in moving this forward?

7c6f434c:
I could nominate myself, I guess. I can also pre-disclose my position that this is likely to be valuable, but also won't move forward until there is a comparison with the module system so that trade-offs can be discussed.

fare:
Where do I learn about modules?

RosarioPulella:
The [NxOS manual](https://nixos.org/manual/nixos/stable/index.html#sec-writing-modules) might be a good place to start.

Also you can also look at some modules in https://github.com/NixOS/nixpkgs/tree/master/nixos/modules

Or search threw the [options](https://search.nixos.org/options) that come from modules in this nixpkgs repo.

7c6f434c:
Search through options is probably useless in this context, as most of the options are of the most primitive type not showing anything interesting about modules.

fare:
I read the module documentation at https://nixos.org/manual/nixos/stable/index.html#sec-writing-modules and while it is very interesting indeed, it's a very different mechanism from the extension system / object system that POP is designed to extend.
- The module system has an `imports` system the effect of which is not very well defined. That's one case where POP's multiple inheritance could give a good meaning to defining dependencies between module definitions such that each is combined only once in dependency order into the mix.
- The "NixOS module system" actually defines three kinds of entities: modules, configurations, types, each with its own semantics, merging algorithm, etc. (Though, if you squint, modules could a special case of types, and configurations be the type being define by the module system.) The definition for a replacement of each of these entities would require a different use of POP's meta-object protocol.
- The merging algorithms used for modules, configs and types are much more elaborate than the one used by POP by default (mergeAttrset), but POP's meta-object protocol (instantiateMeta) supports specifying an alternate mergeInstance mechanism (though the current "toplevel" functions pop, basePop, kPop, etc., all use the default mechanism and won't let you override it for now), so in this sense modules or something module-like could be done in POP and benefit from POP's features and approach.
- Similarly, the warning mechanism could be achieved in a POP setting by using the topProto feature of instantiateMeta to handle warnings at the end of config merging. The same mechanism could more generally "finalize" the values being defined from the result of the merges to something "user-visible".
- POP has a system of dependencies and defaults that is a partial replacement for the priority system of module options: the defaults in POP replace low-priority settings, and the multiple inheritance of POP ensures that settings by children objects (importing modules) will override those of parent objects (imported modules), without having to assign priority numbers.

So I'm not sure what the question or challenge exactly are with respect to "comparing POP and the NixOS module system". On the one hand, POP is a much simpler mechanism meant to replace the Nix extension system and not the NixOS module system. On the other hand, POP does have extension mechanisms with which the more elaborate aspects of the module system could be plugged in, yielding a variant of the module system that deals with imports as dependencies to be combined each a single time in a proper order.

fare:
Of course, in comparing POP and the existing module system:

- By hypothesis, if POP could be used to do exactly the same as the existing module system, that would be an extremely expensive way to achieve what you already have.
- If POP can do "better" somehow, it will necessarily be incompatible in various subtle or not subtle ways, that may or may not be "conservative" enough given the existing code base.

I'm actually slightly and happily surprised by how the mechanisms of the NixOS module system could by and large be plugged in as a layer on top of an unmodified POP (or rather, its instantiateMeta protocol) to yield a variant of this module system that might handle imports better.

Mic92:
We are going to close this RFC within 2 weeks time if there are no shepherds.

Mic92:
Sorry. This RFC did not seem reach the level of interests to find enough shepherds. Therefore we close it.

sternenseemann:
If this is pursued again, I can offer to be a shepherd. I have been lacking time in recent weeks which is why I've put off volunteering here.

Another thing I think would need to be pondered (maybe during discussion) is not only if POP could replace the module system potentially, but also if it could be faster (maybe with the help of a primop?). Module system evaluation speed is surely one of its main problems, to which e. g. [making configuration first class in nix](https://gist.github.com/edolstra/29ce9d8ea399b703a7023073b0dbc00d) has been proposed even.

PR #97, opened 2021-07-04
[RFC 0097] Unset read permission bit on /nix/store for other users

L-as:
This is effectively equal to `chmod o-r /nix/store`.

[Rendered](https://github.com/L-as/rfcs-1/blob/no-read-store-dir/rfcs/0097-no-read-store-dir.md)

### DIFF
--- a/rfcs/0097-no-read-store-dir.md
+++ b/rfcs/0097-no-read-store-dir.md
@@ -30,8 +30,8 @@ Some users might also want to do things like `chmod o-r /nix/store`, which
 gives you the interesting property that you can not access paths you do not
 already know of.
 Do note that given that all processes can by default read `/proc/cmdline`,
-they can still read your system's closure, which makes it an insufficient
-solution for security in many cases.
+`/run/current-system`, and many other places, they can still read your
+system's closure, which makes it an insufficient solution for security in many cases.
 This, however, is also entirely optional and is not the default in any way.
 
 # Detailed design
--- a/rfcs/0097-no-read-store-dir.md
+++ b/rfcs/0097-no-read-store-dir.md
@@ -29,6 +29,9 @@ This, however, should be optional, since the user should be able to do what they
 Some users might also want to do things like `chmod o-r /nix/store`, which
 gives you the interesting property that you can not access paths you do not
 already know of.
+Do note that given that all processes can by default read `/proc/cmdline`,
+they can still read your system's closure, which makes it an insufficient
+solution for security in many cases.
 This, however, is also entirely optional and is not the default in any way.
 
 # Detailed design
--- a/rfcs/0097-no-read-store-dir.md
+++ b/rfcs/0097-no-read-store-dir.md
@@ -11,7 +11,7 @@ related-issues:
 # Summary
 [summary]: #summary
 
-Set the permissions for /nix/store to 1771 instead of 1775, disabling reading the directory for other users.
+Set the permissions for /nix/store, /nix/store/.links and /nix/store/trash to 1771 instead of 1775, disabling reading the directory for other users.
 
 # Motivation
 [motivation]: #motivation
@@ -19,19 +19,25 @@ Set the permissions for /nix/store to 1771 instead of 1775, disabling reading th
 This means that you can not trivially see all the paths in the nix store, e.g. `ls` won't work
 on /nix/store without sudo unless you're in the nixbld group.
 
-Almost everything in NixOS needs access to /nix/store, which means that all your systemd services,
-your flatpak programs, your manually [bubblewrap](https://github.com/containers/bubblewrap)ed programs
-will almost certainly have access to /nix/store and see all of its contents.
-That means they will see your NixOS configuration, initrd secrets (unless your bootloader has support for it),
-all the programs you've built, and possibly also your important secrets if you
-(unfortunately) had to put them in the store!
+Almost everything in NixOS needs access to /nix/store, thus when sandboxing anything under NixOS,
+if you want to have a secure sandbox that can't access your NixOS configuration,
+initrd secrets (unless your bootloader has support for it), all the programs you've built,
+and possibly also your important secrets if you (unfortunately) had to put them in the store,
+you will need to restrict what store paths the sandbox can read.
 
-By simply removing the ability to read /nix/store (but not execute!), programs will only be able
-to access the store paths *which they already know the hash and name of*. This is a huge boon to security
-even if you don't have secrets in your store, as it will almost completely eliminate the above
-problems. That isn't to say it's a complete solution to security, but it will *allow* much more
-complete solutions to security, for example with bubblewrap, meaning that you can be sure that
-your sandboxed game won't suddenly scrape your store and send it off to NSA.
+By simply removing the ability to read /nix/store (but not execute), programs will only be able
+to access the store paths *which they already know the hash and name of*.
+In this case, even if a sandbox has access to the entire store, they will not be able to access any path they do
+not already know, and thus not be able to read any content they do not already know the hash of
+(for non-content-addressed paths, it would be a function of the derivation however).
+Essentially, knowing the hash will mean knowing the data.
+
+For non-sandboxed programs, much will not change if they can still read /run/current-system, /nix/var,
+etc., thus this change is only important for sandboxes where you can remove such information leaks.
+It isn't a big improvement by itself, but it is a small incremental hardening with few drawbacks that allows for better security in combination with sandboxes.
+
+NB: While this change is simple, it is not possible for end-users to do without changing Nix, since at the moment any
+`nix` command will reset the permissions back to 1775.
 
 # Detailed design
 [design]: #detailed-design
@@ -43,6 +49,11 @@ in [nix/src/libstore/globals.hh](https://github.com/NixOS/nix/blob/ba8b39c13003c
 in [nix/src/libstore/build/local-derivation-goal.cc](https://github.com/NixOS/nix/blob/6182ae689826554d915b4ed72e07f7978dc1d13c/src/libstore/build/local-derivation-goal.cc#L641), and
 in [nix/src/libstore/local-store.cc](https://github.com/NixOS/nix/blob/0a535dd5ac93576f7152d786464e330ae3d46b50/src/libstore/local-store.cc#L181)
 to 1771.
+
+/nix/store/trash and /nix/store/.links will also have to have their read bit removed, resulting in 0751.
+
+Currently, /proc/cmdline provides the path to the current system configuration, which is counter-productive in this case.
+This would have to be fixed, by e.g. either embedding the path into the initrd or using the generation number instead of the full path.
 
 Nothing else has to be done likely, since setting the store permissions to `1771` manually doesn't break
 anything other than what is mentioned in this document (though it is undone if you run a `nix` command).
@@ -56,7 +67,7 @@ The execute (x) bit allows us to `cd` to it and also access paths inside the sto
 E.g. `ls "$(readlink /nix/var/nix/profiles/system)"` will still work, since this is a directory
 inside the store, and not the store itself, but an unprivileged user can't `ls /nix/store` to find the system configuration.
 
-Note: A program could still have some idea of how the host machine is used by checking each store path that Hydra has ever built,
+Note: A sandboxed program could still have some idea of how the host machine is used by checking each store path that Hydra has ever built,
 then it could estimate whether the host machine has e.g. Tor on it.
 
 # Drawbacks
@@ -70,12 +81,30 @@ External tooling that does a traversal of the nix store (`find`, `du -s`, `ncdu`
 # Alternatives
 [alternatives]: #alternatives
 
-I had [this script](https://github.com/L-as/nix-misc/blob/e844a03ebf4cad4fc8eca0e52306788b70c2a60d/claybox.rb) that is a wrapper around bubblewrap before, but doing this
-is a lot cleaner, is system-wide, and is a lot faster, since bind mounting each individual store path with bubblewrap can be quite slow
-if you have many (it's O(n^2) IIRC).
+If Nix was made to not reset the permissions of /nix/store back to 1775, users who want this change could
+do it themselves by simply putting this into their configuration.nix:
+```nix
+{
+  system.activationScripts.chmod-store.text = ''
+    ${pkgs.util-linux}/bin/unshare -m ${pkgs.bash}/bin/sh -c '${pkgs.util-linux}/bin/mount -o remount,rw /nix/store ; ${pkgs.coreutils}/bin/chmod 1771 /nix/store'
+  '';
+}
+```
 
-Another alternative for users could be just doing `sudo unshare -m bash -c 'mount -o remount,rw /nix/store; chmod 1771 /nix/store'`,
-but this doesn't work right now because all of the `nix` executables set it to 1775 again.
+
+There is also currently the [systemd-confinement.nix](https://github.com/NixOS/nixpkgs/blob/93c9e5854d87b8e7eeafda2ead4b375d75500c80/nixos/modules/security/systemd-confinement.nix) module in NixOS, which makes use of systemd functionality
+that is functionally equivalent to what bubblewrap does, to make sure only the necessary store paths are mounted.
+
+This is obviously limited to systemd services, for non-systemd-services I had [this script](https://github.com/L-as/nix-misc/blob/e844a03ebf4cad4fc8eca0e52306788b70c2a60d/claybox.rb)
+that just wraps around bubblewrap, but 1) isn't convenient since you need to provide the derivation outputs you want to include and not the attributes for those derivations,
+and 2) it can get [quite slow](https://github.com/containers/bubblewrap/issues/384) once you have sufficiently many store paths you want to include. This could be fixed
+by fixing bubblewrap, but it does not seem to be [making progress](https://github.com/containers/bubblewrap/pull/385).
+
+Both of the above alternatives in addition have the problem that they can not be made aware of new store paths.
+In the case of the design specified in this document, you could pass a store path to a running sandboxed program and they
+would be able to access it normally.
+In addition, you would also be able to e.g. execute a setuid program from inside the sandbox that then has access to paths
+that the executer didn't, since the executed setuid program could read paths from a file with strict read permissions.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
@@ -85,5 +114,4 @@ There doesn't seem to be any.
 # Future work
 [future]: #future-work
 
-There could be more work in the future toward properly sandboxing systemd services and such by default,
-which could also make use of this vital change.
+There doesn't seem to be any.
--- a/rfcs/0097-no-read-store-dir.md
+++ b/rfcs/0097-no-read-store-dir.md
@@ -1,5 +1,5 @@
 ---
-feature: no-read-store-dir
+feature: nix-store-perms
 start-date: 2021-07-04
 author: Las Safin
 co-authors:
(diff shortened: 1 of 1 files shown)


rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,85 @@
...
+
+E.g. `ls "$(readlink /nix/var/nix/profiles/system)"` will still work, since this is a directory
+inside the store, and not the store itself, but you can't without sudo do `ls /nix/store` to find your system configuration.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+It might be a slight annoyance since shell completion won't work in the /nix/store anymore, e.g.
+if you have some hash 48914, you can't type `/nix/store/48914<tab>` to get the full path anymore.
+
```
nh2, line 59:
```suggestion

External tooling that does a traversal of the nix store (`find`, `du -s`, `ncdu`) will need `sudo` or explicitly given permissions.

```

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,85 @@
...
+[motivation]: #motivation
+
+This means that you can not trivially see all the paths in the nix store, e.g. `ls` won't work
+on /nix/store without sudo unless you're in the nixbld group.
+
+Almost everything in NixOS needs access to /nix/store, which means that all your systemd services,
+your flatpak programs, your manually [bubblewrap](https://github.com/containers/bubblewrap)ed programs
+will almost certainly have access to /nix/store and see all of its contents.
+That means they will see your NixOS configuration, initrd secrets (unless your bootloader has support for it),
+all the programs you've built, your usage activity, and possibly also your important secrets if you
```
nh2, line 26: What does `usage activity` mean here?

If it's about discovering which programs a user has installed, wouldn't that still be possible for a malicious containerised app that cannot list `/nix/store` but check the presence of known store paths, by constructing externally the last 10 years of output store paths built by Hydra, and `stat()`ing them on the local machine to check if they exist?

For example, such a program could check whether you have `tor-browser` installed by collecting the ~1000 `*-tor-browser` store paths that Hydra built, and check if you have any of them in the store.

L-as: You are right, I did not consider that.

L-as:
Fixed the parts about usage activity and added the drawback about tools needing more permissions.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,86 @@
...
+to 1771.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+Losing the read (r) bit means that you can't list the files inside the store.
+The execute (x) bit allows us to `cd` to it and also access paths inside the store.
+
+E.g. `ls "$(readlink /nix/var/nix/profiles/system)"` will still work, since this is a directory
+inside the store, and not the store itself, but you can't without sudo do `ls /nix/store` to find your system configuration.
```
kevincox, line 54:
```suggestion
inside the store, and not the store itself, but an unprivileged user can't `ls /nix/store` to find the system configuration.
```

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,86 @@
...
+(unfortunately) had to put them in the store!
+
+By simply removing the ability to read /nix/store (but not execute!), programs will only be able
+to access the store paths *which they already know the hash and name of*. This is a huge boon to security
+even if you don't have secrets in your store, as it will almost completely eliminate the above
+problems. That isn't to say it's a complete solution to security, but it will *allow* much more
+complete solutions to security, for example with bubblewrap, meaning that you can be sure that
+your sandboxed game won't suddenly scrape your store and send it off to NSA.
+
+# Detailed design
```
kevincox, line 36: Have you tested this design and does it appear to work with no further changes? That seems to be implied but it would be nice to note that explicitly.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,86 @@
...
+[alternatives]: #alternatives
+
+I had [this script](https://github.com/L-as/nix-misc/blob/e844a03ebf4cad4fc8eca0e52306788b70c2a60d/claybox.rb) that is a wrapper around bubblewrap before, but doing this
+is a lot cleaner, is system-wide, and is a lot faster, since bind mounting each individual store path with bubblewrap can be quite slow
+if you have many (it's O(n^2) IIRC).
+
+Another alternative for users could be just doing `sudo unshare -m bash -c 'mount -o remount,rw /nix/store; chmod 1771 /nix/store'`,
+but this doesn't work right now because all of the `nix` executables set it to 1775 again.
+
+# Unresolved questions
```
kevincox, line 77: What is the stance with "leaks" like `/nix/var/nix/profiles/system`? Should we expect to remove them over time or are they expected to live forever? While this will give some security (for example I build a derivation in a private directory others won't be able to find it) if we have a number of leaks that provide access to a huge swath of the derivations what is the point.

Other leaks that I am aware of:

- `/proc/cmdline`
- `/proc/{pid}/*`

rfcs/0097-no-read-store-dir.md
L-as: Yes it does work, until you run some `nix` command which will set it back to 1775.

rfcs/0097-no-read-store-dir.md
L-as: You'd disallow access to `/nix/var` instead of removing `/nix/var` (which would make no sense). This can be done through making a separate user namespace, then a separate mount namespace AFAIK. `/proc/{pid}/*` wouldn't be a separate problem if you make a separate pid namespace. This would basically just entail using bubblewrap or some other wrapper tool.

`/proc/cmdline` is indeed problematic, since it's available in all pid namespaces for some reason, you could however pass the system configuration to boot by making it a part of the initrd rather than choosing it by a kernel command-line parameter.

As I note in the document, the actual improvement only happens if you remove the leaks by using it in combination with some other sandboxing mechanism.
Essentially, with this system, knowing the hash of some data also means knowing the data itself. Thus, it isn't a huge improvement by itself, but it is a small incremental hardening with few drawbacks that allows for better security.

kevincox: Ah, so the immediate goal isn't to make the system secure against ordinary users, but make it easier to create effective sandboxes. Is thay understanding correct?

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,89 @@
...
+co-authors:
+shepherd-team:
+shepherd-leader:
+related-issues:
+---
+
+# Summary
+[summary]: #summary
+
+Set the permissions for /nix/store to 1771 instead of 1775, disabling reading the directory for other users.
```
zraexy, line 14: Don't forget about `/nix/store/.links` (The storage location for optimized files) and `/nix/store/trash` (Used to stage directories for GC).

rfcs/0097-no-read-store-dir.md
L-as: Yes, that is correct.

rfcs/0097-no-read-store-dir.md
L-as: Thanks.

rfcs/0097-no-read-store-dir.md
L-as: Do you have any recommendations for how to make this more clear in the document?

L-as:
I rewrote a lot of the text.

rfcs/0097-no-read-store-dir.md
L-as: How is the new explanation?

L-as: I actually found out that /proc/cmdline has the same permissions across all pid namespaces apparently, so just fixing the permissions should be enough!

kevincox: Thanks, I think it is clear now that the immediate benefits are primarily for sandboxed processes.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,119 @@
...
+Losing the read (r) bit means that you can't list the files inside the store.
+The execute (x) bit allows us to `cd` to it and also access paths inside the store.
+
+E.g. `ls "$(readlink /nix/var/nix/profiles/system)"` will still work, since this is a directory
+inside the store, and not the store itself, but an unprivileged user can't `ls /nix/store` to find the system configuration.
+
+Note: A sandboxed program could still have some idea of how the host machine is used by checking each store path that Hydra has ever built,
+then it could estimate whether the host machine has e.g. Tor on it.
+
+# Drawbacks
```
kevincox, line 75: 1. Some users will find it unacceptable that sandboxed programs will be able to probe for specific derivations. For example by probing for common source archives in the store the sandboxed application can likely identify a lot about the host system. This means that this level of access control is not sufficient for some use cases.
2. Leaks are easy to add and may largely remove the benefits of this approach. This makes it a somewhat weak solution as it will be easy to accidentally break the protection and there is no good way to audit against these leaks.

L-as: I'm not sure I agree with 2. but 1 is true yeah, for some use cases it isn't enough.
Let's say we're using bwrap to sandbox a program, for example some HTTP server, like so:
```
  env - bwrap
    --unshare-all
    --ro-bind /nix/store /nix/store
    --share-net
    --proc /proc
    --dev /dev
    --dir /tmp
    --setenv TMPDIR /tmp
    --ro-bind /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
    --ro-bind /etc/ssl/certs/ca-bundle.crt /etc/ssl/certs/ca-bundle.crt
    --ro-bind /etc/resolv.conf /etc/resolv.conf
    --hostname localhost
    --die-with-parent
    --new-session
    --bind . /data
    --chdir /data
    -- simple-http-server
```
Let's define a leak as the path to a built system configuration, or some similarly "large" derivation such that the closure is essentially everything of importance in the store.
Besides some files in /proc that can be fixed by fixing the permissions, how do you think you could unintentionally expose a leak?

kevincox: Files in `/etc` can link to fairly large closures, kernel APIs such as `/proc` can also have lots of info, binaries made available to the service may contain this information. The focus of my point is that we have no good way of detecting leaks, this makes me weary of relying on this for fairly untrusted applications.

For example `/proc/*/task/*/mountinfo` is somewhat dangerous and hard to just fix the permissions. Even things like `/proc/sys/kernel/modprobe` could be updated by a user to do something like logging and surprisingly reference a secret.

I'm not saying that this is a blocker. But I think they are important concerns to list so that their cost can be weighed against the benefit of the change.

L-as: We're considering this feature in the context of sandboxes, so let's continue from the example I made above (which albeit not completely secure is still good enough for this example).

We don't have access to the entirety of `/etc` in our sandbox, it doesn't make sense in any case. We do not want sandboxed programs to have access to anything in `/etc` that is not relevant to the sandbox. As such, we simply do not mount it.
As for `/proc/*/task/*/mountinfo`, that will inside the sandbox only show what is mounted inside the sandbox.

`/proc/sys` does indeed contain a couple of paths (notably to systemd and modprobe),  but for the sake of argument,
I will assume that there is some global environment that can not be hidden from sandboxes.
The problem is then that the user might not be aware that adding a path to this hypothetical environment will make it accessible to sandboxed programs, which are perhaps handled transparently by NixOS. Then the user could inadvertently add the path to something important, partially ruining the point of sandboxing. However, you could also argue that this behavior would actually be preferred, since what has access to the path should be able to read the path intuitively.

I can not actually think of any such global state that the user would want to add a path to that isn't `/proc/cmdline`, `/proc/config.gz`, etc., but of course that doesn't exclude it from existing.

I will add something about this tomorrow after I wake up (night here in Denmark right now).

A bit off-topic:
Regardless of this RFC, I think it's unfortunate that leaking information into sandboxes is so trivial in Linux.
`/proc/sys/kernel/modprobe` is of no relevance to most programs, yet it is present in `/proc`, which is universally available AFAIK (although namespaced). systemd does have a way of of restricting access to the non-PID related parts of `/proc` through `ProcSubset`, though it says this about it in the man page:
> Since these APIs are used frequently this option is useful only in a few, specific cases, and is not suitable for most non-trivial programs

edolstra:
I once tried removing read permission for the nixbld group (NixOS/nixpkgs@224d0d530241a666fc6d929019e07098230bd24e) because it gives some of the advantages of build sandboxing without needing a mount namespace. Unfortunately it broke our VM tests but that should be fixable.

I'm not opposed to this proposal, but it does have a "security through obscurity" vibe. Currently, you shouldn't store secrets in the store, and because Unix software has not traditionally treated paths as secrets, it would be dangerous to rely on the secrecy of a path name for access control.

Note that currently any user that has access to the Nix daemon socket can discover all store paths by doing `nix path-info --all`, or discover "secret" parts of a closure through `nix path-info -r` (e.g. if the `/run/current-system` closure contains a configuration file with a password). Of course, ideally, properly sandboxed services shouldn't have access to the daemon socket. (And we should probably make the daemon only accessible by the `users` group by default...)

L-as:
Thanks, the part about `/run/current-system` is expected though, since the expectation is that you'd be able to access every store path that is mentioned and its dependencies.

I get why you feel it has a "security through obscurity" vibe, since it *does* depend on hashes being secret.
The fundamental difference between a sandbox that uses something like the systemd-confinement.nix module and one that
just expects the store to be 1771, is that in the latter case, you can not trivially compute all store paths
that will be accessed, and that can be the source of unease, since it's harder to *prove* that some path isn't accessed
from inside the sandbox.

However, I think it's important to take into account that making this change won't *worsen* security, since
it doesn't force anyone to drop the use of alternative security measures. It is objectively a net improvement in
security, and essentially I think we need to consider whether it's worth the effort, which IMO is not a lot, since
it's just changing a few numbers, adding a few calls to chmod, and possibly some other small work.

rfcs/0097-no-read-store-dir.md
kevincox: >  As for /proc/*/task/*/mountinfo, that will inside the sandbox only show what is mounted inside the sandbox.

Yes, but if you map any file from a derivation it will reveal the full path, even if it behind a symlink IIRC. My point isn't that it is impossible to use correctly, but that it is easy to accidentally reveal something that you didn't intend to.

> I will add something about this tomorrow after I wake up (night here in Denmark right now).

Thanks. Again, I don't think this is a blocker as this change does technically improve security. I just want to make sure that we are also documenting the limitations and downside of this approach so that
1. People are extra careful if they do choose to rely upon this change.
2. People don't blindly assume that this is a sufficient solution to remove whatever previous mechanism they have, but instead consider the tradeoff carefully.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,119 @@
...
+[drawbacks]: #drawbacks
+
+It might be a slight annoyance since shell completion won't work in the /nix/store anymore, e.g.
```
ryantm, line 78: I am constantly relying on shell completion of hashes to investigate what is going on in NixOS, so I view this as a major step back in usability. Maybe there's some other way I could be working, like completing paths base on shell scrollback instead of path completion?

kevincox:
I think the key concern is: `because Unix software has not traditionally treated paths as secrets`.

If unix had always treated paths as a capability security mechanism kernel and software developers would design new interfaces and patches quite differently. With the current state of paths generally being considered mostly harmless to share with the entire system the risk of leaks is high, if not current interfaces then ones that may appear in the future. And without a good way to detect these leaks it is a very real threat to the security of sandboxes relying on "capability-based" `/nix/store` security.

Again, I don't think this is a blocker but I don't think this RFC is complete without thinking through the security risks that are present in this approach compared to options such as mounting only the needed paths that this RFC is suggesting could be replaced by this.

rfcs/0097-no-read-store-dir.md
L-as: I personally just copy the path in foot (my terminal) by pressing Ctrl+Shift+R.

7c6f434c: I think I even manage to have workflows that go both ways — sometimes it is nice to complete from a few letters, nicer than doing scrollback search then needing to do selection (or taking the mouse); sometimes it is less nice when zsh rechecks whether intermediate components in the path need expansion on Tab completion

L-as: You ought to use foot if possible, it's a great terminal, you basically select what to copy by searching for a prefix of it.

rfcs/0097-no-read-store-dir.md
L-as: I agree, I will add something about it.

rfcs/0097-no-read-store-dir.md
7c6f434c: That sounds basically what screen does anyway, maybe with more automagic for detecting the end, which will probably sometimes go wrong, even if not too often…

Nah, not even on the same scale as having a chance to affect my willingness of giving up my window manager configuration. (Then again, I don't even use urxvt-specific stuff much and sometimes end up working in fbterm…)

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,119 @@
...
+if you have some hash 48914, you can't type `/nix/store/48914<tab>` to get the full path anymore.
+
+External tooling that does a traversal of the nix store (`find`, `du -s`, `ncdu`) would need `sudo` or explicitly given permissions.
+
+# Alternatives
+[alternatives]: #alternatives
+
+If Nix was made to not reset the permissions of /nix/store back to 1775, users who want this change could
```
7c6f434c, line 86: Or maybe reset to the current mode or-ed with `0o1771`.  That, plus NixOS option to force the mode before running the daemon actually sounds like a first step that would be sufficient for people who actively want this RFC, no? Then people unsure about the implications can try switching completely safely (rolling back a Nix update might hit something unfortunate like a schema change…) And one can run systems differing only in this setting to cross-debug any interesting issues.

L-as: I personally don't think it should reset the permissions at all. It is IMO the user's responsibility to set the permissions for /nix/store.

7c6f434c: Hm, why the RFC proposes to keep the chmod in local-store.cc then?

L-as: Because this is simpler, but users should choose themselves, then I think they should have complete freedom wrt. that instead of arbitrarily or'ing with 1771.

7c6f434c: Intuitively I would expect that the installer ensures the defaults, and then Nix doesn't reset things, which seems to look the same in the normal case but skip one count of «what? why?» if the user actually wants to change the permissions. 

So in the first step there is not even a need to change the installer at all (but maybe release notes should include a mention)

L-as:
I made changes to the RFC about the drawbacks of depending on this for sandboxing and fixed some other things.

spacekookie:
Opening this RFC for nominations. Can nominated shepherds please respond whether they accept?

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,130 @@
...
+    ${pkgs.util-linux}/bin/unshare -m ${pkgs.bash}/bin/sh -c '${pkgs.util-linux}/bin/mount -o remount,rw /nix/store ; ${pkgs.coreutils}/bin/chmod 1771 /nix/store ; ...'
+  '';
+}
+```
+
+Traditional sandboxing could be seen as an alternative to this, however, they are not mutually exclusive,
+since this change is fundamentally just changing some permissions in NixOS.
+Improving the permissions for /nix/store will not decrease security for traditional sandboxes.
+
+# Unresolved questions
```
piegamesde, line 122: How does this affect other use cases of Nix, like on non-NixOS or single-user installations?

L-as: It won't be any different. Of course, on non-NixOS systems the permissions of /proc/cmdline, etc. are the responsibility of that distro, but everything else should be exactly the same.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,130 @@
...
+
+For a directory, being able to execute without being able to read it means
+that you can access paths inside the directory, if and only if you know the path beforehand,
+since you can not list the directory entries.
+
+In our case, with this change, all users that are not root and are not part of nixbld,
+will only be able to directly access paths inside `/nix/store` which they already know,
+i.e. the hash and the name.
+If you run some program that does not have access to the nix daemon under such
+a user, it will not be able to search the store for built system configurations,
```
piegamesde, line 30: What exactly do you mean with "search the store" and do applications actually do this? Because I read this as "applications that scrape the store for configuration" and I'm pretty sure I got that wrong.

L-as: I know of no applications that do this; it is just an example. So yes, applications that scrape the store for configuration, which is what doing `ls /nix/store` is. This is what this RFC is about, scraping the store.

piegamesde: Then I suggest removing that line and the following one. There are many legitimate use cases for listing the store (and they'll break and they should be mentioned + alternatives), but "looking for configuration" totally isn't one.

L-as: That is already in the RFC under Drawbacks.

edolstra:
@kevincox @7c6f434c Are you interested in shepherding this RFC? 

Also nominating myself.

kevincox:
I can be a shepherd.

Mic92:
We still need one more shepherd here.

7c6f434c:
Hm, I hoped this would get someone. OK then, I can be a shepherd.

L-as:
Do we have enough shepherds now? Perhaps we can organize a meeting on Matrix? I'm @Las:matrix.org if anyone is not sure.

kevincox:
I invited you to a room. Ping [@kevincox:matrix.org](https://matrix.to/#/@kevincox:matrix.org) for an invite.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,130 @@
+---
+feature: no-read-store-dir
+start-date: 2021-07-04
+author: Las Safin
+co-authors:
+shepherd-team:
```
Mic92, line 6:
```suggestion
shepherd-team: @kevincox @7c6f434c @edolstra 
shepherd-leader: @kevincox
```

spacekookie:
Shepherds for this RFC are @kevincox, @7c6f434c, and @edolstra.

Shepherd leader is @edolstra

Mic92:
@L-as  please also update the yaml frontmatter to incorporate the shepherds.

L-as:
> @L-as please also update the yaml frontmatter to incorporate the shepherds.

Can you elaborate? I thought I already did this.

7c6f434c:
@L-as technically speaking you have done it even before the shepherds have been confirmed, so at a glance there is no «list shepherds» commit after the RFC SC decision (as there are multiple RFCs changing their states, the reminders were sent as long as it is not immediately completely obvious they are not needed)

Mic92:
> > @L-as please also update the yaml frontmatter to incorporate the shepherds.
> 
> Can you elaborate? I thought I already did this.

Alright than.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,130 @@
+---
+feature: no-read-store-dir
+start-date: 2021-07-04
+author: Las Safin
+co-authors:
+shepherd-team: @kevincox @7c6f434c @edolstra
+shepherd-leader: @kevincox
```
sternenseemann, line 7:
```suggestion
shepherd-leader: @edolstra
```

https://github.com/NixOS/rfcs/pull/97#issuecomment-916076773

L-as: Didn't notice that the elected shepherd leader had changed... https://github.com/NixOS/rfcs/pull/97#discussion_r701091017

Mic92: @L-as I think we simply missed your yaml assignment when discussing this RFC.

Mic92:
@edolstra It seems like @L-as wants a another meeting on this or at least the new text should be reviewed.

L-as:
A review alone might be sufficient.

7c6f434c (approved):

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,78 @@
...
+- The default permissions if the store doesn't exist should be 1735 when the store is made by Nix or the NixOS installer.
+  This means that the nixbld group can't `ls` the directory.
+
+# Motivation
+[motivation]: #motivation
+
+Right now you can't set the permissions for `/nix/store`, since they'll be overwritten
+by Nix anytime you use `nix`.
+
+We want to `chmod g-r /nix/store`, because the `nixbld` group doesn't actually
```
kevincox, line 25:
```suggestion
`chmod g-r /nix/store` is beneficial because the `nixbld` group doesn't actually
```

Nit: Avoid first-person pronouns.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,78 @@
...
+need to read the directory. It only needs to be able to write and "execute" it.
+This, however, should be optional, since the user should be able to do what they want.
```
kevincox, line 27:
```suggestion
This, however, should be optional, since the user should be able to configure the permissions however they want.
```

Nit: define the subject more clearly.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,78 @@
...
+
+Some users might also want to do things like `chmod o-r /nix/store`, which
+gives you the interesting property that you can not access paths you do not
+already know of.
+Do note that given that all processes can by default read `/proc/cmdline`,
+`/run/current-system`, and many other places, they can still read your
+system's closure, which makes it an insufficient solution for security in many cases.
```
kevincox, line 34:
```suggestion
`/run/current-system` and many other places which reveals your system's closure, making this permission change an insufficient solution for security in many cases.
```

rfcs/0097-no-read-store-dir.md
kevincox: I consider this resolved.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,78 @@
...
+You should be able to do something like the following:
+```nix
+nix.store-perms = "xxxx";
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+If a user on a non-NixOS platform mistakenly sets the permissions for `/nix/store` to
+something else, it won't be reverted by Nix automatically.
```
kevincox, line 63:
```suggestion
something undesirable, it won't be reverted by Nix automatically.
```

Nit: Wording.

The fact that you can set the permissions to something "else" is the whole point.

kevincox (approved):

L-as:
I've committed the proposed changes.

7c6f434c:
As one of the shepherds of this RFC proposal, I formally propose Final Comment Period with disposition to accept.

As usual, the FCP shall start as soon as all of the shepherds confirm agreement and the corresponding summary announcement is posted here and on Discourse.

kevincox:
I support FCP with disposition to accept.

Profpatsch:
The text of the RFC was changed a lot in between.

I still don’t think it is ready to be accepted in this state.

e.g.

* it was mentioned that VM tests break earlier in this PR
* There are no real alternatives mentioned
* It doesn’t look like it has been thought a lot about what could break by doing this

In particular, what is the *actual* change that happens.
Will “there should be a NixOS option” mean it will be set to something that isn’t the current 
Will I be able to `ls -la /nix/store` in the default configuration in the future? 
What is the current setting? In what way does it provide a security risk?
What about sandboxing instead?

All of these I’d like to be clarified before accepting anything as fundamental as this.

7c6f434c:
> In particular, what is the actual change that happens.

Not much (which is good): stop enforcing permissions via Nix, and allow opting out from enforcing them via NixOS either, and do not make store group-readable by default when creating a store via Nix.

> There are no real alternatives mentioned

The alternative to «why we are even bothering to enforce this so much? let's stop» is not stopping, I don't think we need more.

> It doesn’t look like it has been thought a lot about what could break by doing this

Any changes of the defaults are for the store group users only, those are only supposed to be used during Nix builds. 

> Will I be able to ls -la /nix/store in the default configuration in the future?

In the near future yes, as there are options created but no change in the defaults; the longer term is future work for another RFC.

> In what way does it provide a security risk?

This change itself, as mentioned in RFC, is not large enough for a useful change in security. There is just a piece of functionality in Nix which is too rigid and without clear enough purpose and stands in the way of some more complex experiments not intended to become mainline defaults.

7c6f434c:
@L-as do you want to comment on questions by @Profpatsch or maybe update some writing to improve readability?

In particular, one thing maybe indeed we need to state clearly is that any changes in the name of implementing this RFC should create no new test failures (and if some VM tests somehow end up broken, the burden of fixing things or delaying the offending part is on the implementation PR)

@edolstra what is your current opinion on the current state of the RFC?

L-as:
I feel that your response is good enough. Not breaking existing tests is I think not worth mentioning. Any change should of course not break tests, no matter what that change is.

7c6f434c:
>Not breaking existing tests is I think not worth mentioning. Any change should of course not break tests, no matter what that change is.

Well, there are cases where existing tests are declared reliant on obsolete assumptions and removed, or partially fixed with least important ones left to languish (roughly along the lines of RFC 0088).

This is not what should happen here (I think the RFC as-is gives nothing to lighten the burden of justification of removing tests; it could explicitly add to it but I also hope this bar won't be cleared either way).

Maybe the PR name should be «Allow unsetting…» now.

I preserve my agreement to move to FCP with disposition to accept.

edolstra:
@7c6f434c Yeah I'm :+1: on FCP/accept.

tomberek:
How does this interact with https://github.com/edolstra/nix/commits/acls?

edolstra:
@tomberek It shouldn't have an effect, AFAIK ACL paths will work fine in a non-world-readable store.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,80 @@
...
+the permissions however they want.
+
+Some users might also want to do things like `chmod o-r /nix/store`, which
+gives you the interesting property that you can not access paths you do not
+already know of.
+Do note that given that all processes can by default read `/proc/cmdline`,
+`/run/current-system`, and many other places which reveal your
+system's closure, making this permission change an insufficient solution for
+security in many cases. This, however, is also entirely optional and is not
+the default in any way.
```
roberth, line 37: I assume that at least some filesystems are vulnerable to timing attacks, probing paths with `stat` to build prefixes of subdirectory paths until the full path is recovered; using the `x` bit to recover `r` for all paths, including ones that aren't running.

Nonetheless, it's good for practical hermeticity.

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,80 @@
...
+You should be able to do something like the following:
+```nix
+nix.store-perms = "xxxx";
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+If a user on a non-NixOS platform mistakenly sets the permissions for `/nix/store` to
+something undesirable, it won't be reverted by Nix automatically.
```
roberth, line 65: Why not restrict it to sensible or at least not terrible values? Seems quite feasible and the allow list could be extended if we discover a new use case.

7c6f434c: Because this «undesirable» includes the idea that most likely unfortunate mistakes are someon else's «just as planned» (I am not even sure 0000 is always useless; hmmm, now I am tempted to try once this is merged and trying is easy)

rfcs/0097-no-read-store-dir.md
7c6f434c: Yes, we hope that people either look at the list and start, like you, name extra leak vectors, or trust «insufficient for security» claim. It would be interesting to have a survey of path leak vectors, but this RFC is shorter than any useful survey like that…

kevincox: I agree. Any list we try to create in this RFC is likely incomplete so it is better to just state that as an assumption and accept it for the rationale here. If someone wants to prove that the list of leaks is empty (and expected to stay that way) they can propose a follow-up RFC to rely on that fact.

asymmetric:
Has this ever been implemented? When I look at [one of the lines](https://github.com/NixOS/nixpkgs/blob/8284fc30c84ea47e63209d1a892aca1dfcd6bdf3/nixos/modules/system/boot/stage-2-init.sh#L62) mentioned in the RFC, it hasn't been changed [since](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/stage-2-init.sh#L70).

rfcs/0097-no-read-store-dir.md
```diff
@@ -0,0 +1,80 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+There doesn't seem to be any.
+
+# Future work
+[future]: #future-work
+
+In the future we likely want to reduce the default permissions for `/nix/store` as much as possible.
```
roberth, line 80: Sometimes I use tab completion on a store path; saves some mousing when I fail to copy a path completely, or sometimes typing three or four letters of the hash is easy enough.

L-as:
I realised it wasn't actually that useful due to 7c6f434c's comments about being able to query all paths known from Hydra. Also, logs leak paths. Slightly better security, but only slightly.

roberth:
> Has this ever been implemented?

I don't think an RFC was strictly necessary for this (potential?) change. It'd be fine to introduce it as an opt-in hardening option and then perhaps create the RFC to call attention to it? Maybe I don't see the whole history though.

L-as:
Yeah looking back it didn't need an RFC, but if anyone implements this at least it's decided that it will be merged. It's a very low cost change.

PR #104, opened 2021-08-27
 [RFC 0104] Managing Hydra Configuration with Terraform

grahamc:
Manage hydra.nixos.org's project and jobset configuration as code with
Terraform, terraform-provider-hydra, a configuration repository, and a
continuous deployment process.

### DIFF
--- a/rfcs/0104-hydra-nixos-org-terraform.md
+++ b/rfcs/0104-hydra-nixos-org-terraform.md
@@ -1,0 +1,197 @@
+---
+feature: hydra-nixos-org-terraform
+start-date: 2021-08-27
+author: Graham Christensen
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Manage hydra.nixos.org's project and jobset configuration as code with
+Terraform, terraform-provider-hydra, a configuration repository, and a
+continuous deployment process.
+
+# Motivation
+[motivation]: #motivation
+
+Managing the configuration by hand can be troublesome when
+administrators, trying to solve a problem, tweak settings by hand without
+clearly communicating their changes and intent with other administrators.
+
+Users of various trust levels often request temporary jobsets to test
+high-impact pull requests. In many cases users are long-term, trustworthy
+contributors to the project. However expanding the group of
+administrators has the side effect of increasing the potential for
+confusion around Hydra configuration changes.
+
+Furthermore, Hydra has no memory of jobset configuration, history of
+changes, or audit log for administrative actions. Hydra is a
+security-critical tool and provides a root of trust for the project. In
+particular, the configuration of jobsets which originate channels in
+are in the critical path of determining what and when channel advances
+take place.
+
+Using a centralized configuration repository and continuous deployment
+process increases communication, transparency, and provides an auditable
+log of configuration changes.
+
+With these features, we are able to more liberally give contributors
+the privilege of managing projects and jobsets on Hydra.
+
+# Detailed design
+[design]: #detailed-design
+
+## Workflow
+[workflow]: #workflow
+
+
+* We will adopt Terraform and [terraform-provider-hydra](https://github.com/DeterminateSystems/terraform-provider-hydra)
+  for managing the configuration.
+* The relevant Terraform configuration files will be stored in a GitHub
+  repository named `NixOS/hydra-nixos-org-configuration`.
+* The repository will not allow direct pushes to the primary branch, and
+  will require pull requests. Appropriate CI steps should exist to validate
+  formatting and validity of the PR'd changes.
+* A BuildKite account will be created for the NixOS organization, and
+  a BuildKite runner will be configured on the NixOS server named `bastion`.
+* After merging changes, a BuildKite task will run `terraform plan`, prompt
+  for approval to continue, then run `terraform apply`.
+* The Terraform state will be stored in a private S3 bucket administered by
+  the NixOS Foundation.
+
+_A note about BuildKite:_ BuildKite is selected to maintain control of
+the worker and execution, while also providing transparency to users.
+The BuildKite agent is already packaged in NixOS. BuildKite is a paid
+service, but they readily allow OSS projects to use it for free.
+
+## Migration
+[migration]: #migration
+
+The Hydra provider for Terraform includes a tested import tool for
+converting a manually managed Hydra instance in to the corresponding
+Terraform configuration.
+
+With that tool, the process for migrating is as follows:
+
+1. The repository and BuildKite infrastructure is created and tested
+   against a test Hydra, verifying the workflows execute correctly.
+1. The `admin` and `create-projects` role is revoked from every account.
+1. A new user is created for Terraform to use. This user is granted `create-projects`.
+1. All projects have their `owner` changed to the new Terraform user.
+1. The import tool is run, producing a repository of `.tf` files.
+1. `terraform plan` is executed, verifying the generated `.tf` files are correct.
+1. The repository's BuildKite configuration is altered to operate against
+   hydra.nixos.org.
+1. The NixOS manual for release managers is updated to instruct Release Managers
+   to use this repository for their jobsets.
+
+## Access
+[acccess]: #access
+
+* All users which currently have the `create-projects` and `admin` roles
+  will be granted merge privileges on the repository.
+* Long term, trusted members of the project can request merge access. These
+  requests will be evaluated by the Infrastructure and Security team, and
+  the decision is ultimately up to the Infrastructure team.
+* Jobsets which are the origination point for channels will be maintained
+  in specially named files, and the GitHub Code Owners feature will be used
+  to require Infrastructure team members to approve changes to those jobsets.
+* The BuildKite pipeline will be totally public, allowing anyone to view
+  its execution logs.
+* All users who are allowed to _merge_ changes will be allowed to trigger
+  builds and unblock the `terraform apply` step.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+Contributors can send a pull request for desired Hydra configuration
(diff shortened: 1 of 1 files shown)


grahamc:
ping, the list of people with admin or create-projects on hydra:

@AmineChikhaoui
@rbvermaa 
@svanderburg
@zimbatm 
@edolstra

AmineChikhaoui:
This sounds great to me 👍🏼

AmineChikhaoui (approved):

edolstra:
This will require replacing the NixOps configuration for hydra.nixos.org with terraform. (I mean, we could use terraform only for the Hydra config, but it would be pretty ugly to have two different configuration management tools for one system.)

edolstra:
BTW I don't think issues pertaining to how we manage the nixos.org infrastructure require an RFC, since this only affects the NixOS infra team. An issue in the nixos-org-configurations repo is enough, IMHO.

edolstra:
On the technical merit, I'm not convinced that terraform-provider-hydra is the way to go. It seems to be a convergent rather than congruent approach, i.e. it tries to keep the configuration in the Hydra database in sync with the terraform definitions. This is rather error-prone. A better approach would be to move that configuration to `hydra.conf`. Then users, projects and jobsets can be managed in the same way as the rest of the system. (Ideally we could even get rid of the web interface for user/project/jobset configuration, but some non-hydra.nixos.org users might be attached to it.)

grahamc:
> This will require replacing the NixOps configuration for hydra.nixos.org with terraform. (I mean, we could use terraform only for the Hydra config, but it would be pretty ugly to have two different configuration management tools for one system.)

Not quite: none of hydra's hydra.conf would be impacted this is just projects and jobsets -- hydra.conf and nixops don't manage that at all.

> BTW I don't think issues pertaining to how we manage the nixos.org infrastructure require an RFC, since this only affects the NixOS infra team. An issue in the nixos-org-configurations repo is enough, IMHO.

Fair enough.

> It seems to be a convergent rather than congruent approach, i.e. it tries to keep the configuration in the Hydra database in sync with the terraform definitions.

While this is true, reading the project and jobset data from a file would also be convergent since the source of truth is the database. I don't think it would be a good idea to replace the database with a file, though:

1. Features like declarative jobsets depends on the database.
2. Changes to projects jobsets have a fairly different lifecycle to the entire system, and restarting the queue runner any time a jobset's configuration is tweaked would be weird.

There are other issues too, but I really don't think the idea has merit.

One of the benefits of having this configuration live outside of the nixos-org-configurations repository is the ability to grant merge privileges to a larger group of users without any possibility that their changes can impact more than the Hydra projects and jobsets.

As a final note, the motivations with introducing BuildKite here are grander scoped than just this text suggests. I could easily see moving channel origination to a buildkite task, exposing those logs publicly as well. This is something we've talked about before, but never came to a decision on.

AmineChikhaoui:
> On the technical merit, I'm not convinced that terraform-provider-hydra is the way to go. It seems to be a convergent rather than congruent approach, i.e. it tries to keep the configuration in the Hydra database in sync with the terraform definitions. This is rather error-prone. A better approach would be to move that configuration to `hydra.conf`. Then users, projects and jobsets can be managed in the same way as the rest of the system. (Ideally we could even get rid of the web interface for user/project/jobset configuration, but some non-hydra.nixos.org users might be attached to it.)

I also find the usage of hydra.conf odd. The experience with plugins like slack, bitbucket we had isn't great as whenever you change something you have to redeploy the machine and restart the queue-runner. It also a bit more tricky for automation as now instead of a self contained hydra JSON api call (through the terraform plugin), you have to redeploy the hydra instance through `nixops` and make sure that your deploy only impacts the creation/deletion of jobsets. Granted with nixops + pinned channel/flakes that's something possible but now you have to think about more possible failures.
So I think the proposed approach makes sense and is really orthogonal to the machine/hydra config we have in nixos-org-configurations repo which shouldn't be impacted or duplicated by the proposed RFC.

edolstra:
> While this is true, reading the project and jobset data from a file would also be convergent since the source of truth is the database.

No, the declarative config should supersede any (legacy) configuration in the database.

> Changes to projects jobsets have a fairly different lifecycle to the entire system, and restarting the queue runner any time a jobset's configuration is tweaked would be weird.

It could just reload hydra.conf without restarting.

I think allowing declarative user/project/jobset config in hydra.conf is a much simpler approach than a few thousand lines of not-obviously-correct Go code. And of course it's much more in line with the stateless approach to configuration management that we like to preach.

grahamc:
> I think allowing declarative user/project/jobset config in hydra.conf is a much simpler approach than a few thousand lines of not-obviously-correct Go code.

That may or may not be, I'm not convinced it is true. There would still be synchronization code, and dealing with cases like jobsets appearing and then disappearing. I don't believe this would be a simple task. This is also making me concerned about IDs for tables.

The Terraform provider has a few good things going for it. Its implementation is not particularly complicated, it is already written, tested (automatically and manually), and used by Hydra users. It is also a system we can deploy today and have fairly good confidence it will work. For another, the Terraform core manages state migration very well. This RFC is also empowering to our contributors by granting trust and tools to well trusted community members who can't deploy to the NixOS infrastructure.

Also, not that it matters to much but there is about 1k LOC of code written for that project. The rest is generated against the Hydra API's spec.

grahamc:
Another note, every six hours I import the hydra.nixos.org jobsets / projects as tf: https://github.com/DeterminateSystems/hydra-nixos-org-configuration/tree/main/config. This configuration applies cleanly, and if imported to a terraform statefile produces a clean plan. I regularly use this mirror to test Hydra locally.

ocharles:
I'm not sure it was mentioned yet, but if this is already managed by `nixops` - is having a Hydra resource/plugin for `nixops` also an option to consider?

grahamc:
On a technical  note, I wouldn’t take that on personally: nixops’ diff
engine is much less capable and implementing confirmation reconciliation
on nixops’ structure is not an easy task. Terraform’s architecture makes
this much easier.

However a key piece of this RFC is empowering the community to configure
and manage jobsets they need for their work. Having this tied to a full
deploy to the hydra machine does not seem like a good way to achieve
this. 

On August 29, 2021, GitHub ***@***.***> wrote:
> I'm not sure it was mentioned yet, but if this is already managed by
> nixops - is having a Hydra resource/plugin for nixops also an option
> to consider?
>

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/NixOS/rfcs/pull/104#issuecomment-907760093>, or
unsubscribe <https://github.com/notifications/unsubscribe-
auth/AAASXLA6IMCZUZXSFRO4FM3T7H44VANCNFSM5C6AOOGA>.
Triage notifications on the go with GitHub Mobile for iOS
<https://apps.apple.com/app/apple-store/id1477376905?ct=notification-
email&mt=8&pt=524675> or Android
<https://play.google.com/store/apps/details?id=com.github.android&referrer=utm_campaign%3Dnotification-
email%26utm_medium%3Demail%26utm_source%3Dgithub>.

blaggacao:
/cc @shlevy for `nomia`. I'd appreciate to hear his (general) opinion on this.

I'd also like to point loosely to https://github.com/input-output-hk/bitte-ci potentially for general inspiration. It manages jobsets via [cue files](https://github.com/input-output-hk/bitte-ci/blob/main/ci.cue)

Mic92:
@grahamc We decided to close this RFC as this is a better fit for an issue in https://github.com/NixOS/nixos-homepage/ as it does not affect the hydra source code itself.

grahamc:
I assume you mean nixos-org-configuration?

edolstra:
Yes, nixos-org-configuration.

andir:
> @grahamc We as the RFC steering committee decided to close this RFC as this is a better fit for an issue in https://github.com/NixOS/nixos-homepage/ as it does not affect the hydra source code itself.

Can you go into detail how hydra source code changes require RFCs?
Isn't this similar to how we retire old committers and how we invite users into the maintainers team? This is also "just" configuration.
I fail to see how there is a difference in governance there.

Regardless of that I would really like to see more smaller scope RFCs (just like this) that can be processed comparatively quickly and also be implemented in a timely manner. So far we have a lot of huge RFCs with a huge implementation and those aren't exactly ideal from my perspective.

L-as:
You could make an RFC for introducing mini-RFCs.

andir:
> You could make an RFC for introducing mini-RFCs.

IMHO "mini-RFCs" just follow the same process and are hopefully easier to agree on. I don't think the process needs changing for that. A few weeks of round trip are not an issue.

edolstra:
It's not so much about whether it affects the Hydra source code, but that how we manage the nixos.org infrastructure is primarily relevant to the infrastructure team and doesn't affect the wider community.

Mic92:
> > @grahamc We as the RFC steering committee decided to close this RFC as this is a better fit for an issue in https://github.com/NixOS/nixos-homepage/ as it does not affect the hydra source code itself.
> 
> Can you go into detail how hydra source code changes require RFCs?
> Isn't this similar to how we retire old committers and how we invite users into the maintainers team? This is also "just" configuration.
> I fail to see how there is a difference in governance there.

What I meant was fundamental design decisions affecting one of the projects in the nix community.

sternenseemann:
> It's not so much about whether it affects the Hydra source code, but that how we manage the nixos.org infrastructure is primarily relevant to the infrastructure team and doesn't affect the wider community.

The RFC specifically states an impact on the community (at least the narrower one of contributors) as a motivation:

> Users of various trust levels often request temporary jobsets to test
> high-impact pull requests. In many cases users are long-term, trustworthy
> contributors to the project. However expanding the group of
> administrators has the side effect of increasing the potential for
> confusion around Hydra configuration changes.

Mic92:
> The RFC specifically states an impact on the community (at least the narrower one of contributors) as a motivation:

Yes and I think everyone agrees that this is a good thing and it already happens with other part of the nixos org configuration such as fastly configuration that @zimbatm ported to terraform.
What I learned from talking with @edolstra, the actual discussions is more about whether this declarative form should be done in nix or with terraform - This is something the infrastructure team needs to decide on themselfs with as they need to maintain it in longterm. It would be odd if a shepherd team with no insight in the nixos org ops would make decisions like that.

PR #105, opened 2021-09-03
[RFC 0105] Nix flake labels

bryanhonof:
Currently a flake's only way to explain what it is that it contains is trough its description. Having labels could aid the end-user determine what a flake contains on an higher level.

[Rendered](https://github.com/bryanhonof/rfcs/blob/be7ed8c3308de9ac521bd21c07b6fb582fdac0dc/rfcs/0103-flake-tags-metadata.md)


rfcs/0103-flake-tags-metadata.md
```diff
@@ -0,0 +1,150 @@
...
+to one. Once that has happened, a NixOps plugin could identify itself to be a
+plugin by providing the tags `plugin` and `nixops`, inside their flake.
+
+The expected outcome is that flakes will be better identifyable through a simple
+interface.
+
+# Detailed design
+[design]: #detailed-design
+
+The `tags` attribute in a flake should NOT be mandatory, but optional. It is
```
jonringer, line 41: since flakes are heavily coupled to git, and `tag` is already a well-defined term when referring to version control, maybe the use of `label` would be more appropriate?

Otherwise I could see a scenario where someone says, "I cut a [git] tag for the product, make sure to update your [flake-referenced git] tags in your flakes. I also added another [flake] tag, to help with discoverability"

bryanhonof: That's a good comment, I hadn't thought about this myself. I'll make the edit when I get back behind my PC.

gytis-ivaskevicius (commented):
Few thoughts:

I find it funny that we got flakes definition change RFC before the introduction of flakes themselves RFC (Does this even need to be an RFC?)

`s/tag/label/g` (Not everything has been renamed yet) 

I very much do like the idea of tags as such:
- machine-learning
- system-config
- messaging
- containers
- cloud
- home-manager-config
- project-xyz-packages

But consider (just consider) replacing examples like `python3`/`c-plus-plus` due to them being oddly specific and summarizing derivation itself instead of the flake.

Sidenote:
It would be useful if we could evaluate flakes repository information (like repository name/owner) from within configuration. But this is probably out of scope for this RFC

edolstra:
The main problem with adding tags is that flakes are decentralized and uncurated. Without a central authority to organize tags, they're likely to become a mess where nobody agrees on what the preferred tag names are for a particular topic (e.g. do you use `c++` or `c-plus-plus`), or what the exact semantics are (e.g. does `python` mean any software written in Python, or a Python module?).

gytis-ivaskevicius:
@edolstra I don't think that there are any problems with it being decentralized and uncurated. The idea of labels/tags is definitely not new and we can take a look at other platforms for reference:
- twitter
- Google Ads
- SEO stuff
- Article websites (like medium.com)
- Photo galleries (like deviantart.com)

What they all have in common are non-standartized tags. 
One may ask how come it works well for sites as such? The answer is simple - due to authors themselves having an interest in tagging their content in an easily discoverable manner. 

**BUT, I since previous comment I gave it more thought and I think this should be approached differently:**
Pretty much all flakes are stored on Github or Gitlab, these platforms have labeling functionalities of their own. 
How about instead of implementing this as part of the flakes we modify search.nixos.org to query GitHub/GitLab APIs and grab tags/description/license/releases/whatever from there instead.

tomberek:
Part of the question here would be if it helps discoverability. Integration with nixos-search? Otherwise the name and description become de-facto search terms: https://github.com/NixOS/nixos-search/blob/af3b49421784fa557b90acffef08168d43d2a699/flake-info/src/elastic.rs#L17-L21

lucc:
@gytis-ivaskevicius & @edolstra python's pip also has labels or tags.  They just call it classifiers: https://pypi.org/classifiers/
They provide a long list of standard(?) classifiers.

I have not much experience apart from adding some to my python project for good measure. But maybe there is something to learn from the python guys here. They have a short section about it in a PEP:
https://www.python.org/dev/peps/pep-0301/#distutils-trove-classification

I did not look for any further info (blogs, email lists) to see how their experience was after some time.

lucc:
I thought about what I wanted to say before again, the idea is this:

- If we want we can implement a fixed set of predefined labels in nix itself (@edolstra nix would serve as a centralized place to define and manage the standard labels, this is the only thing the decentralized flakes can rely on)
- any string could be used as a label (not just the defined standard labels); they do not concern evaluation of the flake
- `nix flake check` could complain about non standard labels

bryanhonof:
@lucc 
> If we want we can implement a fixed set of predefined labels in nix itself 
> any string could be used as a label (not just the defined standard labels); they do not concern evaluation of the flake

I agree that it's not a good idea to enforce a predefined set of labels. However, I do think it'd be a good idea to have a sane set of default labels, or rather, recommended naming schemes. We could do this by providing a set of already well-known labels, just like the `meta.licenses` attribute works right now in the `mkDerivation` function. But don't keep the end-user from creating their own.

---

@gytis-ivaskevicius
> BUT, I since previous comment I gave it more thought and I think this should be approached differently:
Pretty much all flakes are stored on Github or Gitlab, these platforms have labeling functionalities of their own.
How about instead of implementing this as part of the flakes we modify search.nixos.org to query GitHub/GitLab APIs and grab tags/description/license/releases/whatever from there instead.

I thought about this as well, but I came to the conclusion that it'd be extremely annoying to maintain this way. Sure, git{hub,lab} support this at the moment, but there's examples of forges that don't. I think flakes should be self-describing, and the `description` attribute is already a good idea, but it might not be enough.

I'm also thinking, would it perhaps make sense to think about a `meta` attribute for flakes, just as `mkDerivation` has?

---

@edolstra 
> The main problem with adding tags is that flakes are decentralized and uncurated. Without a central authority to organize tags, they're likely to become a mess where nobody agrees on what the preferred tag names are for a particular topic (e.g. do you use c++ or c-plus-plus), or what the exact semantics are (e.g. does python mean any software written in Python, or a Python module?).

I agree with @gytis-ivaskevicius on this one. I think, over time, people will figure out what works best. And if we provide a sane set of defaults, that's already a bit of the burden of organising it to improve search-ability.

---

@tomberek 
> Part of the question here would be if it helps discoverability. Integration with nixos-search? Otherwise the name and description become de-facto search terms: https://github.com/NixOS/nixos-search/blob/af3b49421784fa557b90acffef08168d43d2a699/flake-info/src/elastic.rs#L17-L21

This is something I'm willing to research a bit about. But I can't give an answer on this right now.

andir:
Just a thought: How about making tags part of the registry entry of flakes? That way the operator of the registry can apply whatever tags that they seem correct. When you submit your flake for addition into a registry you can have a discussion with that group of people about what tags they use and which ones you'd want to add.

The actual flake wouldn't require any amount of changes and everyone can enforce the guidelines they like to follow for their registry.

AndersonTorres:
Flakes have such a thing as a meta attribute? Because those meta things as `description` or `labels` look a bit meta.

lheckemann:
This RFC is open for nominations, anyone who has ideas who would make a good shepherd (including themselves) is welcome to nominate them :)

gytis-ivaskevicius:
> Sure, git{hub,lab} support this at the moment, but there's examples of forges that don't. I think flakes should be self-describing, and the description attribute is already a good idea, but it might not be enough.

Self-hosted GitHub alternatives are not going to get included in nixos-search either way, which raises another question - do flake labels bring any value in any other use case aside from search engine integration? :thinking:

blaggacao:
I want to add to @edolstra 's commemt & @gytis-ivaskevicius reply:

Decentralized tags work.in github, ..., because while the creation of a tag is decentralized, the semantic of an existing task is not in isolation:

When you tag something, you discover other tagged projects.

spacekookie:
This RFC is open for nominations btw! Feel free to nominate anyone you think would be a good shepherd for this RFC (including yourself)

lheckemann:
Please nominate some potential shepherds; if we don't have anybody to drive the discussion we would move the RFC to draft status in 2 weeks (2021-12-15).

AndersonTorres:
What are the duties of a shepherd?

Because I want to suggest one more step: instead of a mere `labels` , create an umbrella `meta` attribute and put `description` and `labels` inside of it, instead of leaving them loose.

tomberek:
@AndersonTorres The shepherds guide discussion in order to ultimately recommend to accept/reject the RFC in an FCP (https://github.com/NixOS/rfcs/blob/master/rfcs/0036-rfc-process-team-amendment.md). This can involve obtaining more community input, exploring design decisions, implementation, collaborating on RFC modifications, and so on. I see it as distilling all the feedback and commentary into an actionable recommendation; this is much easier to achieve with a small focused shepherd team than in a large distributed forum.

AndersonTorres:
I candidate myself to shepherd this RFC.

8573:
> The main problem with adding tags is that flakes are decentralized and uncurated.

While I have no opinion on the present proposal, I note as prior art that Rust's Cargo supports two types of tags/labels: "keywords", which are uncurated, and "categories", which must be drawn from a predefined list (whether defined by the registry or by Cargo <i>per se</i> I don't know; with only one significant public registry, the distinction is blurred).

>  [...] they're likely to become a mess where nobody agrees on what the preferred tag names are for a particular topic (e.g. do you use `c++` or `c-plus-plus`), or what the exact semantics are (e.g. does `python` mean any software written in Python, or a Python module?).

8573:
> The main problem with adding tags is that flakes are decentralized and uncurated.

While I have no opinion on the present proposal, I note as prior art that packages for Rust's Cargo have two types of tags/labels: "keywords", which are uncurated, and "categories", which must be drawn from a predefined list (a [controlled vocabulary](https://en.wikipedia.org/wiki/Controlled_vocabulary) — whether defined by the registry or by Cargo <i>per se</i> I don't know; with only one significant public registry, the distinction is blurred).

>  [...] they're likely to become a mess where nobody agrees on what the preferred tag names are for a particular topic (e.g. do you use `c++` or `c-plus-plus`), or what the exact semantics are (e.g. does `python` mean any software written in Python, or a Python module?).

Cargo "keywords" do have these problems, and they can be slightly annoying, but I don't think they're seen as major problems in practice. Cargo "categories" have explicit definitions.

A separate point is that Cargo limits a package to having at most five of each type of label, forcing packagers to pick the most important and limiting the ability to use them for [spamdexing](https://en.wikipedia.org/wiki/Spamdexing).

AndersonTorres:
> While I have no opinion on the present proposal, I note as prior art that packages for Rust's Cargo have two types of tags/labels: "keywords", which are uncurated, and "categories", which must be drawn from a predefined list

There is any other tangible difference between keywords and categories besides "this is from a fixed list, the other is uncurated"? Because, seriously, I have no problem with an uncurated, community-driven list.

> Cargo "keywords" do have these problems, and they can be slightly annoying, but I don't think they're seen as major problems in practice. Cargo "categories" have explicit definitions.

We can recommend some labels more than others...

> A separate point is that Cargo limits a package to having at most five of each type of label

8 or 10 keywords should suffice.

AndersonTorres:
@bryanhonof what do you think about creating a `meta` attribute in the same spirit of `meta` attribute on Nixpkgs, and including `description` and `labels` on it?

spacekookie:
This RFC still needs more shepherd nominations. If you can think of anyone who you think would be good to guide this discussion (including yourself) please nominate them

tomberek:
Potentially related? https://github.com/NixOS/nix/issues/5879

(a name is basically a special label)

AndersonTorres:
> Potentially related? [NixOS/nix#5879](https://github.com/NixOS/nix/issues/5879)
> 
> (a name is basically a special label)

I read the name and I am seriously confused. `name` looks like a "metadata we are using outside metadata".

lheckemann:
We're marking this as a draft for now, given the lack of shepherds and the dependence on Flakes itself, which still lacks an RFC.

PR #106, opened 2021-09-23
[RFC 0106] Nix release schedule

edolstra:
This RFC proposes doing new Nix releases every six weeks, as [previously proposed on Discourse](https://discourse.nixos.org/t/nix-release-schedule-and-roadmap/14204).

[Rendered](https://github.com/edolstra/rfcs/blob/nix-release-schedule/rfcs/0106-nix-release-schedule.md)

### DIFF
--- a/rfcs/0106-nix-release-schedule.md
+++ b/rfcs/0106-nix-release-schedule.md
@@ -1,0 +1,91 @@
+---
+feature: Nix release schedule
+start-date: 2021-09-23
+author: Eelco Dolstra
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: N/A
+---
+
+# Summary
+[summary]: #summary
+
+Do a new Nix release every 6 weeks.
+
+# Motivation
+[motivation]: #motivation
+
+The last major Nix release, 2.3, came out in September 2019. Since
+then, there has been a huge amount of work on the Nix master branch,
+including major improvements to the new command line interface and new
+experimental features such as flakes, the content-addressable Nix
+store, and recursive Nix. This forces many users to use unstable Nix
+releases, which is undesirable.
+
+In the future, we should avoid having such long gaps between releases,
+since it’s bad for both contributors and users that there is an
+unbounded amount of time before a new feature shows up in a stable
+release. The thing that has historically caused long gaps between Nix
+releases is new experimental features landing in master that we
+weren’t quite sure about, and doing a new release meant having to
+support these features indefinitely. However, Nix 2.4 introduces an
+experimental-features setting that enables us to add such features in
+a way that they can be changed or removed, while still getting
+feedback from adventurous users. So long as experimental features
+don’t cause breakage in stable features, it’s fine to merge them into
+master and include them in a new release.
+
+# Detailed design
+[design]: #detailed-design
+
+* We do a new Nix release every 6 weeks. The release process is
+  already almost entirely automated so this is pretty easy.
+
+* The master branch should be kept in a releasable state at all times.
+
+* PRs should include release notes, if applicable. (Currently trawling
+  through the history to dig up interesting stuff for the release
+  notes is the most work in making a new release.)
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Infrequent releases give more stability to users. Users of Nix-stable
+have been blissfully isolated from all the code churn on master for
+the last two years.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Stick to the current release-when-it's-ready non-schedule.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* Should we still do maintenance releases (like 2.3.x)? Should there
+  be a long-term stability release (like 2.3 is now, de facto)?
+  Probably we should at least provide bug fixes for whatever Nix
+  release is used by the latest NixOS release.
+
+* Is 6 weeks the ideal interval between releases? It seems to work
+  well for Rust.
+
+* Should we keep using the current versioning scheme? For now we can
+  stick with the current scheme (i.e. only bumping the major version
+  if there are incompatible changes or major non-experimental new
+  features), but in the future we could switch to date-based versions
+  (e.g. Nix 21.07).
+
+# Previous work
+
+[RFC 0009](#9) proposed a rapid release policy where releases can be
+done at any time (e.g. on request) rather than on a fixed schedule. It
+wasn't feasible at the time because we didn't have a notion of
+experimental features, so we had to give such features some time to
+stabilize before doing a new release.
+
+# Future work
+[future]: #future-work
+
+N/A
--- a/rfcs/0106-nix-release-schedule.md
+++ b/rfcs/0106-nix-release-schedule.md
@@ -51,7 +51,7 @@ master and include them in a new release.
 # Drawbacks
 [drawbacks]: #drawbacks
 
-Infrequent releases give more stability to users. Users of Nix-stable
+Infrequent releases give somewhat more stability to users. Users of Nix-stable
 have been blissfully isolated from all the code churn on master for
 the last two years.
 
--- a/rfcs/0106-nix-release-schedule.md
+++ b/rfcs/0106-nix-release-schedule.md
@@ -39,14 +39,23 @@ master and include them in a new release.
 # Detailed design
 [design]: #detailed-design
 
-* We do a new Nix release every 6 weeks. The release process is
-  already almost entirely automated so this is pretty easy.
+* We do a new Nix release (named 2.5.0, 2.6.0, 3.0.0, ...) every 6
+  weeks. The release process is already almost entirely automated so
+  this is pretty easy.
 
-* The master branch should be kept in a releasable state at all times.
+* For each major release, we create a maintenance branch (named
+  e.g. `2.5-maintenance`) and corresponding Hydra jobset. Patch
+  releases (e.g. 2.5.1) are made as needed.
+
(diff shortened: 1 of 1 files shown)


rfcs/0106-nix-release-schedule.md
```diff
@@ -0,0 +1,91 @@
...
+
+* Should we keep using the current versioning scheme? For now we can
+  stick with the current scheme (i.e. only bumping the major version
+  if there are incompatible changes or major non-experimental new
+  features), but in the future we could switch to date-based versions
+  (e.g. Nix 21.07).
+
+# Previous work
+
+[RFC 0009](#9) proposed a rapid release policy where releases can be
```
ryantm, line 82: The link here does not work.

tomberek:
Nixpkgs/NixOS releases have mechanisms that encourage stability as well as ensuring it gets done. Concepts like release managers, ZHF, release channels (along with all the corresponding testing), backports policy, wiki about release process, and so forth help drive progress. I don’t have a solid idea formulated yet, but I think some analogous concept would be good to include in this RFC.

abathur:
I have no strong opinions on semver vs. calver. Thoughtful semver is useful, but not so great that it'd be worth regular debates over whether a release qualifies as an x/y/z bump instead of something a script generates.

I guess I'd prefer semver if it's a decision being made within 24h of release by a small, decisive group with good heuristics/razors. If it takes longer than that, it may be better for everyone to just automate it.

This is a little far afield, but I've daydreamed about "test-driven versioning." I can't really extoll its virtues--it's just something I vaguely intend to give a try someday--but it struck me while writing this that some of Nix's ~contracts (like store, CLI, and daemon compatibility) might make interesting candidates for test-enforced version promises.

rfcs/0106-nix-release-schedule.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* Should we still do maintenance releases (like 2.3.x)? Should there
+  be a long-term stability release (like 2.3 is now, de facto)?
+  Probably we should at least provide bug fixes for whatever Nix
+  release is used by the latest NixOS release.
+
+* Is 6 weeks the ideal interval between releases? It seems to work
```
L-as, line 71: Considering that Nix still sees regular development, 6 weeks seems fine to me.

rfcs/0106-nix-release-schedule.md
```diff
@@ -0,0 +1,92 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+Stick to the current release-when-it's-ready non-schedule.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* Should we still do maintenance releases (like 2.3.x)? Should there
+  be a long-term stability release (like 2.3 is now, de facto)?
```
L-as, line 67: I don't see the point of this. Ideally, the only backward-compatible changes that happen in Nix are in the interface, not the language and building mechanisms. Because of this, supporting a new Nix release in a project will likely never require large changes.

rfcs/0106-nix-release-schedule.md
```diff
@@ -0,0 +1,92 @@
...
+unbounded amount of time before a new feature shows up in a stable
+release. The thing that has historically caused long gaps between Nix
+releases is new experimental features landing in master that we
+weren’t quite sure about, and doing a new release meant having to
+support these features indefinitely. However, Nix 2.4 introduces an
+experimental-features setting that enables us to add such features in
+a way that they can be changed or removed, while still getting
+feedback from adventurous users. So long as experimental features
+don’t cause breakage in stable features, it’s fine to merge them into
+master and include them in a new release.
```
L-as, line 37: While this is true, a lot of smaller changes don't get an experimental feature, e.g. https://github.com/NixOS/nix/pull/4922. Perhaps there should be a generic experimental feature for enabling unstable minor changes such as this, so that we don't have to add an experimental feature for each?

Ma27:
So, I'd be up for it if everyone else involved is OK with that :)

tomberek:
I can assist with this. I've been focusing on Nix itself for a bit. I'll self-nominate.

I've got an updated flake fork of the nix-install-matrix lying around that can track and test various scenarios - throw it up on a "Nix installer status page"?

spacekookie:
Please update the RFC with the following shepherds: @Ericson2314, @regnat, @Ma27, @tomberek (leader). Thank you!

tomberek:
## Plan of Action

created a room at #nix-rfc-106:matrix.org for any discussion

My guess is that the schedule and intent is not controversial. Focus on establishing some policy and guidance the community can depend on.

1) Determine possible risks
2) Establish a few possible and acceptable controls.
3) Update RFC, with any policy-only controls
4) FCP
5) Merge
6) Implement technical controls in parallel with 3-5
7) Release another Nix version (apx: 2021 Dec 2013)

Ericson2314:
So based on struggles I am having updating a client from 2.3 to 2.4, one thing I would like to see more of is testing of mixed clients/daemons/sshes as part of release-readiness. We've made https://github.com/NixOS/nix/pull/5572 to test the **new client** with the **old daemon**, but I would also like to test the **old client** with the **new daemon**, and it is a but unclear what the best way to do that is.

tomberek:
@Ericson2314 There is a risk of forward/backward compatibility issues. This sort of thing can be added to an nix-install-matrix report. Having a 6 week release cycle will probably reduce these sorts of issues in the long run as we will be forced encounter breakages more often. Are you proposing some policy regarding new/old client/daemon pairings?

What other risks do we need to mitigate?
1) experimental -> non-experimental schedules
    - some policy for this would be helpful. Experimental -> RFC -> X backward-compat Nix versions -> on by-default 
2) what is a "releasable state"? What set of tests is enough?
    - https://github.com/NixOS/nix/blob/master/.github/workflows/test.yml
    -  mimic the Nixpkgs Tier support requirments?
    - `nix build -f default.nix checks.x86_64-linux`
    - `nix flake check`
    - some percentage of a possible nix-install-matrix
3) Who performs release tasks?
    - Eelco
    - backup release managers who are also responsible for some manual install + sanity checks?
4) Release notes requirement
    - https://github.com/NixOS/nix/blob/master/doc/manual/src/release-notes/rl-next.md).
    - I suspect that people don't quite know what to include or when it is required. Are we adding this as a requirement to merge PRs?

ck3d:
Regarding updating, I updated the daemon to 2.4pre and my services ran 2.3. After a few day I was confronted with https://github.com/NixOS/nix/issues/5299 . I spend hours to fix the broken entries in the nix store.

edolstra:
On 1: That's out of scope for this RFC IMHO. We should probably have another RFC to decide on an experimental features policy.

On 2: Releasable state means that https://hydra.nixos.org/jobset/nix/master must be green. This is a superset of `nix flake check`.

On 3: That's me currently since I have the GPG signing key.

On 4: We should update the PR template to ask for release notes / documentation updates, if appropriate.

rfcs/0106-nix-release-schedule.md
```diff
@@ -0,0 +1,92 @@
...
+* The master branch should be kept in a releasable state at all times.
+
+* PRs should include release notes, if applicable. (Currently trawling
+  through the history to dig up interesting stuff for the release
+  notes is the most work in making a new release.)
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Infrequent releases give more stability to users. Users of Nix-stable
```
regnat, line 54:
```suggestion
Infrequent releases give somewhat more stability to users. Users of Nix-stable
```

I’m not entirely convinced that having one really big release every two years gave more stability than a lot of small releases ;)

rfcs/0106-nix-release-schedule.md
regnat: Dunno for the long-term releases, but I think we’ll have to do maintenance releases anyways, if only because we might have to do some bug fixes on the latest release before the end of the 6w cycle

regnat (approved):

tomberek:
## Update Plan of Action

1) ~~Determine possible risks~~ complete
2) ~~Establish a few possible and acceptable controls.~~ adding clarification
3) Update RFC, with any policy-only controls
    a) update RFC to define "releasable state"
    b) updating issue template to nudge toward a more comprehensive test suite
4) FCP
5) Merge
6) ~~Implement technical controls in parallel with 3-5~~: out-of-scope and considered to be future work
7) Release another Nix version (apx: 2021 Dec 13)

@NixOS/rfc-steering-committee: recommend moving this to FCP after RFC edits mentioned above. Shepherd team has general consensus and is recommending adoption.

edolstra:
@tomberek I've updated the RFC to define "releasable state" and mention the issue template.

Mic92:
@tomberek do you think this RFC is now ready for FCP since the last commit?

tomberek:
> @tomberek do you think this RFC is now ready for FCP since the last commit?

Yes.

terlar:
When it comes to back-porting fixes, how many versions back would receive backports with this new more frequent releases? Will the recommended way be to move forward to next version instead?

Also should the nix versions automatically be updated in nixpkgs as well? I notice now we do have a release of Nix 2.5.0. But nixpkgs for example have yet to be updated with this new release.

edolstra:
Sorry, we forgot to announce the RFC on Discourse. I've done that now.

terlar:
What about my previous [questions](https://github.com/NixOS/rfcs/pull/106#issuecomment-995190943), are those things that should be part of the RFC?

E.g.
- What is the back-porting policy?
- Should the release process be connected in any way to nixpkgs?

edolstra:
We currently don't have a formal back-porting policy, so this RFC doesn't really change anything in that respect.

On the Nixpkgs release process, we probably want to maintain (i.e. back-port to) the default version used in the latest NixOS release.

edolstra:
The RFC has been merged. Thanks everybody!

PR #108, opened 2021-10-05
[RFC 0108] NixOS Container rewrite

Ma27:
[Rendered](https://github.com/Ma27/rfcs/blob/container-rfc/rfcs/0108-nixos-containers.md)
POC: https://github.com/NixOS/nixpkgs/pull/140669


rfcs/0000-nixos-containers.md
```diff
@@ -0,0 +1,463 @@
...
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* None that I'm aware of.
+
+# Future work
+[future]: #future-work
+
+* Write documentation for the new module.
+* Get the PR into a mergeable state.
```
L-as, line 462: What PR?

Ma27: https://github.com/NixOS/nixpkgs/pull/140669

Sorry, should've linked it here in the first place :)

rfcs/0000-nixos-containers.md
```diff
@@ -0,0 +1,463 @@
...
+  as address pool to distribute IPv4 addresses via DHCP to containers.
+
+Additionally, basic IPv6 support was implemented:
+
+* By specifying `::/64`, a [RFC4193 ULA prefix](https://tools.ietf.org/html/rfc4193) will be
+  allocated to the host-side interface.
+* With [`radvd`](https://github.com/radvd-project/radvd), containers can assign themselves addresses from this address prefix by utilizing
+  [RFC4862 SLAAC](https://tools.ietf.org/html/rfc4862)
+
+This is necessary since `systemd-networkd` doesn't support router advertisements with
```
arianvp, line 132: Hmm really? I thought networkd supported DHCPv6 just fine .

Ma27: I'm pretty sure that this was the case last time I checked, but I'll re-check to be safe (also, this was a few major releases of systemd ago, so perhaps that's just obsolete :))

rfcs/0000-nixos-containers.md
L-as: I just noticed the issue lol. In your YAML header, you refer to the issues like `#number`, but that's parsed as a comment, so it's stripped out in the rendered version... You probably want to use direct links anyway instead of just the issue number, since it's another repository.

roosemberth:
There's something itching my brain: Maybe we could explore link-local addressing (especially when using zones). Instead of requiring a central network manager (be dhcpcd, systemd-networkd, radvd, ...), containers could simply use a scope (e.g. the interface with the name of the zone) and llmnr.

I'm unable to formalize this as a section (also, simply creating a veth pair does not seem to automatically assign ll addresses).
Perhaphs it would be better to discuss the actual implementation of something like this in a separate RFC (or probably just an MR, since I don't expect much semantics to discuss), but I do believe this document should mention it (or at least discuss this here).

rfcs/0000-nixos-containers.md
Ma27: Done :)

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,467 @@
...
+
+### Default Mode
+
+If nothing else is specified, the [default settings of `systemd-nspawn`](https://github.com/systemd/systemd/blob/v247/network/80-container-ve.network) will
+be used for networking. To briefly summarize, this means:
+
+* A [`veth`](https://man7.org/linux/man-pages/man4/veth.4.html) interface-pair will be created,
+  one "host-side" interface and a container interface inside its own namespace.
+* A dynamically allocated prefix of `0.0.0.0/28` (or `0.0.0.0/24` for virtual zones) will be used
+  as address pool to distribute IPv4 addresses via DHCP to containers.
```
flokli, line 127: We don't use 0.0.0.0/28. This only instructs networkd to allocate a prefix in some private address range.

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,467 @@
...
+### "public" networking
+
+This is the most trivial networking mode. It is taken if the `PrivateNetwork`-option of the
+`.nspawn`-unit is set to `no`. In this case, the container has full access to the host's network,
+otherwise the container will run in its own namespace.
+
+### Default Mode
+
+If nothing else is specified, the [default settings of `systemd-nspawn`](https://github.com/systemd/systemd/blob/v247/network/80-container-ve.network) will
+be used for networking. To briefly summarize, this means:
```
flokli, line 122: I'm not sure if we should be too specific with the networkd config format internals here, and just describe how it'll look like from a birds-eye perspective.

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,467 @@
...
+
+Additionally, basic IPv6 support was implemented:
+
+* By specifying `::/64`, a [RFC4193 ULA prefix](https://tools.ietf.org/html/rfc4193) will be
+  allocated to the host-side interface.
+* With [`radvd`](https://github.com/radvd-project/radvd), containers can assign themselves addresses from this address prefix by utilizing
+  [RFC4862 SLAAC](https://tools.ietf.org/html/rfc4862)
+
+This is necessary since `systemd-networkd` doesn't support router advertisements with
+dynamically allocated prefixes.
```
flokli, line 137: Same same, I'm not sure if IPv6 prefix delegation on this is a thing/will be in the future. It shouldn't be part of an RFC.

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,467 @@
...
+
+Hosts will be available on the current system via the
+[`mymachines` `nss` module](https://www.freedesktop.org/software/systemd/man/nss-myhostname.html),
+which is already taken care of by `systemd-networkd`. This means that container names can be resolved to addresses like DNS names, i.e. `ping containername` works.
```
flokli, line 141: Hostnames of local container instances can be looked up on the host, but that's nothing networkd is taking care of. It's provided by nss-mymachines for all containers registered with systemd-machined.

Again, I don't think the RFC should cover systemd internals that much - it's probably sufficient to say they'll be available for local host resolution.

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,467 @@
...
+  * Deploy & reboot.
+  * See also https://github.com/Ma27/nixpkgs/blob/networkd-containers/nixos/tests/container-migration.nix as POC.
+
+## Imperative management
+
+`systemd` differentiates between "privileged" & "unprivileged" settings. Each privileged (also
+called "trusted") `nspawn` unit lives in `/etc/systemd/nspawn`. Since unprivileged container's
+don't allow bind mounts, these will be out of scope. Additionally, this means that
+`/etc/systemd/nspawn` has to be writable for administrative users and can't be a symlink to
+a store path anymore.
```
flokli, line 191: nspawn files can also be read from /run/systemd/nspawn and /var/lib/machines/. I don't see why /etc/systemd/nspawn needs to be writable.

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,467 @@
...
+* When activating a new config on the host, the following things happen:
+  * If the setting `Parameter=` in the container's `.nspawn`-unit is the only thing that has changed,
+    a `reload` will be done. This parameter contains the `init`-script for the container's NixOS
+    and only changes if the container's NixOS config changes.
+  * If anything else changes, `systemd-nspawn@container-name.service` will be scheduled for a restart
+    which effectively reboots the container.
+* This behavior can be turned off completely which means that the container where this is turned
+  off won't be touched at all on `switch-to-configuration`. Additionally, it's possible to always
+  force a `reload` or `restart`. See [Examples & Interactions](#examples-and-interactions) for
+  details.
```
flokli, line 231: This feels like a lot of (maybe unnecessary) complexity, especially the `nsenter` part.

Could we just activate the new configuration inside the container on a reload, and restart the container if any of the mounts in the .nspawn file has changed?

We could provide some option to always force a container restart.

rfcs/0108-nixos-containers.md
Ma27: > We don't use 0.0.0.0/28. This only instructs networkd to allocate a prefix in some private address range.

Yeah, it was my intention to describe that, but I see why this sentence can be misleading, will update that, thanks!

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,467 @@
...
+
+The main reason why `machinectl` cannot be used as a complete replacement are
+[imperative containers](https://nixos.org/manual/nixos/stable/index.html#sec-imperative-containers)
+and state getting lost after the `container@<container-name>.service` unit
+has stopped since `.nspawn` units aren't used.
+
+In the following section the design of a replacement is proposed with these goals:
+
+* Use [`networkd`](https://www.freedesktop.org/software/systemd/man/systemd.network.html) as the networking stack since `systemd-nspawn` is part of the same project and
+  thus both components are designed to work together and issues like no uplink until the container is fully booted will be fixed.
```
kevincox, line 62:
```suggestion
  thus both components are designed to work together and resolve issues like no uplink until the container is fully booted.
```

Nit: With "will be fixed" it isn't clear if this is *already* fixed in networkd+systemd-nspawn or if we expect that upstream will fix this issue because they work together. I think the new wording is more exact.

Ma27: Makes sense, applied, thanks!

Ma27:
No. I don't see how to integrate this here. The goal is to provide a thin wrapper with some "helpers" to use `systemd-{nspawn,networkd}`. With this approach, a bunch of systemd units & a store-path with a NixOS system is provided, that's it, images (such as OCI-compliant ones that are supported by systemd-nspawn for a few releases now) are out of scope.

In other words, this isn't about how to represent a container, but how configure systemd-nspawn & the networking around it.

I'm not sure if that fully answers your question, but I'm happy to discuss this further.

blaggacao:
> In other words, this isn't about how to represent a container, but how configure systemd-nspawn & the networking around it.

Perfect! I opened https://matrix.to/#/#nix-rfc-108:matrix.org for any not 100% on-topic or also future shepherds discussion.

Mic92:
I can volunteer for shepherding this one.

lheckemann:
I accept the nomination, though I won't be able to take the lead on this.

Ma27:
Would be perfectly fine for me :) 

Just wondering, are the other nominees @flokli & @arianvp interested in joining? :)

asymmetric:
@Ma27 would this allow us to do [this](https://www.wireguard.com/netns/), i.e.:

- create network NS
- control in which NS the container is created

There was [a PR](https://github.com/NixOS/nixpkgs/pull/71328) a while ago enabling this for the  existing setup.

Btw https://github.com/systemd/systemd/pull/14915 (the follow-up of https://github.com/systemd/systemd/pull/11710) hasn't been merged yet - is this relevant?

SuperSandro2000:
If there is still a shepherd needed I would like to nominate myself. I am using nixos-container on none NixOS which should hopefully be a valuable insight.

Mic92:
We still need one more shepherd because right now more than half of the shepherds are from the committee.

spacekookie:
Apologies for the label mishap, we actually need one more shepherd nomination because not more than half of the shepherds can be part of the RFCSC.

Current nominations: @Mic92, @lheckemann, @SuperSandro2000

ck3d:
Only for info, my 2 year old PR is somehow related to this RFC:
https://github.com/NixOS/nixpkgs/pull/67232

rfcs/0108-nixos-containers.md
zseri: why does the container need to be restarted when the port forwards change? How exactly do they affect the container side of things?

Ma27:
Pinging @andir, @flokli, @arianvp .

arianvp:
Thanks for the ping. I am very interested in this; but I was rather busy the past few months and NixOS was on the bottom of the list.   But I'm happy to try co-sheparding.

Mic92:
@Lassulus just told me, he would shepherd this RFC, and I think he would be a good fit.

Lassulus:
ok, lets try to find a date for the first call then: https://www.when2meet.com/?13905803-adV2a

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+* Start/Stop operations, logging into containers: this can be entirely done via [`machinectl(1)`](https://www.freedesktop.org/software/systemd/man/machinectl.html).
+* No configuration will be made via CLI flags. Instead, the option set from the
+  NixOS module will be used to declare not only the container's configuration, but also
+  networking. This approach is inspired by [erikarvstedt/extra-container](https://github.com/erikarvstedt/extra-container).
+
+But still, not all features from declarative containers are implemented here, for instance:
+
+* One has to explicitly specify whether to restart/reload a container when updating the config.
+  This is done on purpose to avoid duplicating the logic from `switch-to-configuration.pl` here.
+* IPv6 prefix delegation is turned off because `radvd`'s configuration is declaratively specified
```
genofire, line 201: Hmm, just in mind:
systemd has his own [prefix delegation](https://systemd.network/systemd.network.html#%5BDHCPPrefixDelegation%5D%20Section%20Options) (and [route advertisement](https://systemd.network/systemd.network.html#%5BIPv6SendRA%5D%20Section%20Options) ) why not use that?

https://systemd.network/systemd.network.html#%5BDHCPPrefixDelegation%5D%20Section%20Options

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
+---
+feature: NixOS Container rewrite
+start-date: 2021-02-14
+author: Maximilian Bosch <maximilian@mbosch.me>
+co-authors: n/a
+shepherd-team: n/a
+shepherd-leader: n/a
```
Mic92, line 7: We decided on a team:

```suggestion
shepherd-team: @Mic92, @lheckemann, @SuperSandro2000, @arianvp, @Lassulus 
shepherd-leader: @Lassulus 
```

Lassulus:
alright, seems like we have a date in the new year:

2022-01-02 13:00:00 CET

I suggest we meet on my jitsi instance (if its not broken on that day) https://jitsi.lassul.us/rfc-108

Lassulus:
> alright, seems like we have a date in the new year:
> 
> 2022-01-02 13:00:00 CET
> 
> I suggest we meet on my jitsi instance (if its not broken on that day) https://jitsi.lassul.us/rfc-108
> 
> EDIT: please thumbs up this post as ACK, otherwise I will message you as a reminder!

@Ma27 @Mic92 @lheckemann @SuperSandro2000 @arianvp 
this is today in ~2 hours

rfcs/0108-nixos-containers.md
arianvp: Had the same question.  But I think the answer is that we can't expect the host-side of the stack to use `systemd-networkd`  (where we'd enable prefix delegation on the host-side of the veth pair)

I'm confused about this section in general. How was this implemented before and what funtionality are we losing is not clear to me. It reads to me like   "it's up to the user to have some form of prefix delegation on the host side of the veth pair" which sounds fine. But wasn't that the case already in the old scenario?

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+  inside, however with no content.
+* To get a running NixOS inside, `/nix/store` is bind-mounted into it. As an `init` process,
+  the [stage-2 script](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/stage-2-init.sh)
+  is started which eventually `exec(2)`s into `systemd` and ensures that everything
+  is correctly set up.
+* The option `boot.isContainer = true;` will be automatically set for new containers as well.
+  This is necessary to
+  * avoid bogus `modprobe` calls since `nspawn` doesn't have its own kernel.
+  * as option to skip building a `stage-1` boot script when building a NixOS system for
+    the container.
```
lheckemann, line 100:
```suggestion
  * avoid building a `stage-1` boot script and initramfs as part of the container's NixOS system
```

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+
+If nothing else is specified, the [default settings of `systemd-nspawn`](https://github.com/systemd/systemd/blob/v247/network/80-container-ve.network) will
+be used for networking. To briefly summarize, this means:
+
+* A [`veth`](https://man7.org/linux/man-pages/man4/veth.4.html) interface-pair will be created,
+  one "host-side" interface and a container interface inside its own namespace.
+* A subnet from a [RFC1918](https://datatracker.ietf.org/doc/html/rfc1918) private IP range
+  is assigned to the host-side interface. IPv4 addresses will be distributed via DHCP to containers.
+* Analogous IPv4, a [RFC4193 IPv6 ULA prefix](https://tools.ietf.org/html/rfc4193) will be
+  assigned to the host-side interface, containers can assign themselves addresses from this
```
lheckemann, line 129:
```suggestion
* Analogous to IPv4, a [RFC4193 IPv6 ULA prefix](https://tools.ietf.org/html/rfc4193) will be
  assigned to the host-side interface. Containers can assign themselves addresses from this
```

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+  prefix by utilizing [RFC4862 SLAAC](https://tools.ietf.org/html/rfc4862).
+
+Hosts will be available on the current system via the
+[`mymachines` `nss` module](https://www.freedesktop.org/software/systemd/man/nss-myhostname.html).
```
lheckemann, line 133:
```suggestion
[`mymachines` `nss` module](https://www.freedesktop.org/software/systemd/man/nss-mymachines.html).
```

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+An example of how this can be done is shown in the [next chapter](#examples-and-interactions).
+
+### DNS
+
+The current implementation uses [`networking.useHostResolvConf`](https://search.nixos.org/options?channel=20.09&show=networking.useHostResolvConf&from=0&size=50&sort=relevance&query=networking.useHostResolvConf)
+to configure DNS via `/etc/resolv.conf` in the container. This option will be **deprecated** as
+`systemd` can take care of it:
+
+* If `networkd` is enabled via NixOS, [`systemd-resolved`](https://www.freedesktop.org/software/systemd/man/systemd-resolved.service.html) is enabled as well.
+  * By default, DNS for `resolved` will be configured via DHCP which is enabled in the [Default Mode](#default-mode) by default.
```
lheckemann, line 150:
```suggestion
  * By default, `resolved` will be configured via DHCP which is enabled in [Default Mode](#default-mode).
```

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+  * With only [Static networking](#static-networking) enabled, it is necessary to configure
+    configure DNS servers for resolved statically which can be done by setting e.g. DNS
+    servers via a `.network`-unit for the `host0` interface.
```
lheckemann, line 153:
```suggestion
  * With only [Static networking](#static-networking) enabled, it is necessary to configure
    DNS servers for resolved statically which can be done by setting DNS
    servers via a `.network` unit for the `host0` interface.
```

rfcs/0108-nixos-containers.md
lheckemann: https://github.com/systemd/systemd/blob/3b5816ba3362f046f317d4d20dc8bc3153a750fd/src/nspawn/nspawn.c#L4632-L4634
`/run/systemd/nspawn` is in fact trusted by default; however, this is not documented by upstream so I think we should open an issue to make sure this behaviour doesn't change unexpectedly. I'm also not sure to what extent this is useful, since AFAIK we want imperative containers to persist.

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+    named `host0` by default.
+  * As soon as the config is ready to deploy, the state directory in `/var/lib/containers` has to
+    be copied to `/var/lib/machines`.
+  * Deploy & reboot.
+  * See also https://github.com/Ma27/nixpkgs/blob/networkd-containers/nixos/tests/container-migration.nix as POC.
+
+## Imperative management
+
+`systemd` differentiates between "privileged" & "unprivileged" settings. Each privileged (also
+called "trusted") `nspawn` unit lives in `/etc/systemd/nspawn`. Since unprivileged container's
```
lheckemann, line 181:
```suggestion
called "trusted") `nspawn` unit lives in `/etc/systemd/nspawn`. Since unprivileged containers
```

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+  This is done on purpose to avoid duplicating the logic from `switch-to-configuration.pl` here.
+* IPv6 prefix delegation is turned off because `radvd`'s configuration is declaratively specified
+  when building the host's NixOS.
+
+Examples are in the next chapter.
+
+## Config activation
+
+By default, NixOS has to decide how to activate configuration changes for a container to avoid
+e.g. unnecessary reboots, but `reload`s aren't necessarily sufficient either because e.g.
```
lheckemann, line 209:
```suggestion
unnecessary reboots, but `reload`s aren't necessarily sufficient either because changes such as
```

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+new bind mounts require a reboot. The host's `switch-to-configuration.pl` implements it like
+this:
+
+* `systemctl reload systemd-nspawn@container-name.service` runs `switch-to-configuration test`
+  inside the container `container-name` using `nsenter`.
```
lheckemann, line 214: Why not `machinectl`?

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+* When activating a new config on the host, the following things happen:
+  * If the setting `Parameter=` in the container's `.nspawn`-unit is the only thing that has changed,
+    a `reload` will be done. This parameter contains the `init`-script for the container's NixOS
+    and only changes if the container's NixOS config changes.
```
lheckemann, line 218:
```suggestion
    and changes every time the container's NixOS config changes.
```

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+### Basics
+
+A container with a private IPv4 & IPv6 address can be configured like this:
+
+``` nix
+{
+  nixos.containers.instances.demo = {
+    network = {};
+    config = { pkgs, ... }: {
+      environment.etc."foo".text = "bar";
```
lheckemann, line 238:
```suggestion
      environment.systemPackages = [ pkgs.hello ];
```
If the purpose of this line is to illustrate that NixOS options go here, I'd suggest the probably most familiar option to all NixOS users :)

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+    };
+  };
+}
+```
+
+It's reachable locally like this thanks to the `mymachines`-feature of NSS:
```
lheckemann, line 244:
```suggestion
It's reachable locally like this thanks to systemd's `mymachines` NSS module:
```

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+    networkConfig.DHCP = "yes";
+    dhcpConfig.UseDNS = "no";
+    networkConfig.MACVLAN = "mv-eth1-host";
+    linkConfig.RequiredForOnline = "no";
+    address = lib.mkForce [];
+    addresses = lib.mkForce [];
+  };
+
+  # The host-side sub-interface of the MACVLAN. This means that the host is reachable
+  # within the (internal) network at `192.168.2.2`.
```
lheckemann, line 317:
```suggestion
  # The host-side sub-interface of the MACVLAN. This means that the host is reachable
  # at `192.168.2.2`, both on the physical interface and from the container.
```

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+{
+  config.nixpkgs = <nixpkgs>;
+  config.config = { pkgs, ... }: {
+    services.nginx.enable = true;
+    networking.firewall.allowedTCPPorts = [ 80 ];
+  };
+
+  # This implies that the "default" networking mode (i.e. DHCPv4) is used
+  # and not the host's network (which is the default for imperative containers).
+  config.network = {};
```
lheckemann, line 372: Maybe this should be called `container.network` to avoid confusion with the `networking` namespace available in all NixOS configurations.

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+
+#### Modify a container's config imperatively
+
+When `imperative-container.nix` is updated, it can be rebuilt like this:
+
+```
+$ nixos-nspawn update imperative --config ./imperative-container.nix
+```
+
+By default, it will be **restarted**. This can be overridden via `config.activation.strategy`,
```
lheckemann, line 398: Like `network`, this could be moved into `container.activation`?

lheckemann (approved):
Overall LGTM, just some phrasing, nits and minor improvements which I'd suggest.

arianvp:
> @Ma27 would this allow us to do [this](https://www.wireguard.com/netns/), i.e.:
> 
> * create network NS
> * control in which NS the container is created
> 
> There was [a PR](https://github.com/NixOS/nixpkgs/pull/71328) a while ago enabling this for the existing setup.
> 
> Btw [systemd/systemd#14915](https://github.com/systemd/systemd/pull/14915) (the follow-up of [systemd/systemd#11710](https://github.com/systemd/systemd/pull/11710)) hasn't been merged yet - is this relevant?

It's not relevant. That is for when networkd is _not_ running in the container. e.g. when the container is not a full OS container but a more classic application container.  Given we have a full NixOS instance running in the container we can configure the network namespace from inside the container using networkd.

Lassulus:
so all shepherds agreed this ready for FCP
@lheckemann 

thanks for everyone involved, keep up the good work!

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,460 @@
...
+
+* Start/Stop operations, logging into containers: this can be entirely done via [`machinectl(1)`](https://www.freedesktop.org/software/systemd/man/machinectl.html).
+* No configuration will be made via CLI flags. Instead, the option set from the
+  NixOS module will be used to declare not only the container's configuration, but also
+  networking. This approach is inspired by [erikarvstedt/extra-container](https://github.com/erikarvstedt/extra-container).
+
+But still, not all features from declarative containers are implemented here, for instance:
+
+* One has to explicitly specify whether to restart/reload a container when updating the config.
+  This is done on purpose to avoid duplicating the logic from `switch-to-configuration.pl` here.
```
m1cr0man, line 200: Hey there. Loving this proposal! I'm definitely going to try out the POC later.

I would like to use imperative containers to manage a fleet of NixOS installs out of band from the host, much as if they were full VMs. With old nixos-container, it was possible to create an imperative container and then update its configuration from within itself (login + nixos-rebuild) or by changing its profile in the filesystem. This has the benefit of not unnecessarily restarting the whole container. From what I understand of this proposal, nixos-nspawn would be the only way to update configuration, and I would need to make the decision myself on whether a reload or restart would be suffice.

My question is, how does this decision/flag (as described later on) impact the restart/reload of services within the container?

Ma27: > My question is, how does this decision/flag (as described later on) impact the restart/reload of services within the container?

By default the host's `/nix/store` is bind-mounted into a container (as it's the case with the current implementation). For your use-case I'd suggest to bind-mount `/nix/var/nix/profiles` as well, then you should be able to work with the profiles.

This is perfectly possible with the implementation, so I'd consider this an implementation detail. Do you agree?

Regarding activation: you can set `activation.strategy` to `none`, then `nixos-nspawn` won't do anything in terms of activating things. I just realized though that `nixos-container update` does a `systemctl reload` if the container is running, perhaps we should do this here as well?

rfcs/0108-nixos-containers.md
Ma27: We should probably cut the `nsenter`-part out because it's an (irrelevant) implementation detail.

`machinectl` isn't possible here because you can't pipe commands to `machinectl shell` and `systemd-run` will hang if you e.g. update the container's `systemd` (see e.g. https://github.com/NixOS/nixpkgs/issues/83432)

rfcs/0108-nixos-containers.md
Ma27: The primary issue is something different: I think that naming the container's config-attribute `config` was a poor choice actually because we now have to do some nasty hacks such as `config.config` (and thus also `config.network` which is indeed confusing).

I'll change it to let's say `system-config`, then the file would look like this:

``` nix
{
  nixpkgs = <nixpkgs>;
  network = { ... };
  system-config = { pkgs, ... }: {
    # ...
    networking.firewall.allowedTCPPorts = [ 80 ];
  };
}
```

rfcs/0108-nixos-containers.md
m1cr0man: Since my comment I took some time to read the POC PR too and deciphered what nixos-nspawn.py does. I agree that it's an implementation detail at this point, and adding the profiles bind mount is specific to my use case I should think.

I'm not sure if performing a `systemctl reload` in all cases makes sense either, as I'm unaware of the consequences of that for other attributes set within the npawn unit file (for example, would systemctl immediately make some change that doesn't require a container restart, if any such attributes exist?).

rfcs/0108-nixos-containers.md
Ma27: `systemd.nspawn(5)` explicitly documents that `/etc/systemd/nspawn` is trusted, so I suggest we stick to that:

```
Persistent settings files created and maintained by the administrator (and thus trusted) should be placed in /etc/systemd/nspawn/, while automatically downloaded (and thus potentially untrusted)
settings files are placed in /var/lib/machines/ instead (next to the container images), where their security impact is limited. In order to add privileged settings to .nspawn files acquired from the
image vendor, it is recommended to copy the settings files into /etc/systemd/nspawn/ and edit them there, so that the privileged options become available.
```

rfcs/0108-nixos-containers.md
Ma27: The `reload` only does a `switch-to-configuration` and thus only alters what happens *inside* the container. To activate the port-forward (via systemd), you can currently only reboot.

rfcs/0108-nixos-containers.md
```diff
@@ -0,0 +1,459 @@
...
+
+```
+$ nixos-nspawn update imperative --config ./imperative-container.nix
+```
+
+By default, it will be **restarted**. This can be overridden via `config.activation.strategy`,
+however only `reload`, `restart` and `none` are supported.
+
+Additionally, the way how the container's new config will be activated can be specified
+via `--reload` or `--restart` passed to `nixos-nspawn update`.
```
m1cr0man, line 401: From looking at the functionality within the POC, I think the "reload" and "none" strategies may be a little overloaded in the context of imperative containers managed like full systems of their own.

It's not clear if reload will be reloading configuration inside or outside of the container. I would expect it to be both, however the real behaviour is to `switch-to-configuration` within the container, and update the nspawn configuration on the host without actually applying any changes, since that would require a restart of the container. Also, since it's doing switch-to-configuration, all the actual services in the container will be reloaded _or restarted_ irrespective of this strategy.

I had a couple of suggestions initially, but in typing this out I think the easiest thing that could be done here is to rename `reload` to `switch`, in reference to `nixos-rebuild switch`. The consequences of this are now much more obvious and predictable given you know how nixos-rebuild works, and it maintains at least some consistency between the two commands (even though it's a flag in this case).

With that clarification, "none" does make a bit more sense as an activation strategy, but it is still a little misleading. _Something_ is being done to activate the config, it's just not immediately taking place. I think a name like "next-restart" or "boot" (again to be consistent with nixos-rebuild) would make more sense IMO.

I'm interested to hear your thoughts on this and how the strategy names came about in the first place :)

rfcs/0108-nixos-containers.md
Ma27: >  systemd has his own prefix delegation (and route advertisement ) why not use that?

It doesn't support dynamically allocated prefixes (i.e. `Address=::/64`) which is what's used by default.

> Had the same question. But I think the answer is that we can't expect the host-side of the stack to use systemd-networkd (where we'd enable prefix delegation on the host-side of the veth pair)

We talked about this in the shepherd's meeting, but for everyone: networkd is required on both sides.

> How was this implemented before and what funtionality are we losing is not clear to me. It reads to me like "it's up to the user to have some form of prefix delegation on the host side of the veth pair" which sounds fine. But wasn't that the case already in the old scenario?

It is a comparison to what declarative containers using networkd do (i.e. what's missing for imperative containers).

> FWIW; the upstream network file for the veth-pair does not enable prefix delegation either. it only enables DHCPServer (for ipv4) https://github.com/systemd/systemd/blob/main/network/80-container-vz.network

I know and since I was somewhat unhappy about it, I implemented it here :)

rfcs/0108-nixos-containers.md
Ma27: > I'm not sure if performing a systemctl reload in all cases makes sense either, as I'm unaware of the consequences of that for other attributes set within the npawn unit fil

It only executes what's in `ExecReload=` if that's set, so it only activates your configuration. Does that sound reasonable to you? :)

m1cr0man: Ok yeah, that makes sense then. It seems more consistent that way.

rfcs/0108-nixos-containers.md
m1cr0man: > reload was to remain consistent with the underlying operation (systemctl reload) for declarative containers and I decided to use the same things for imperative ones. switch is IMHO a bit ambiguous as it isn't clear how it differs from restart.

Ok yeah, I hadn't considered reading them as synonyms of `systemd reload/restart`, which does make far more sense now :) On that note, scratch my suggestion for that, however I would still be a fan of none -> boot, or possibly "onboot" to imply the event that will trigger activation.

Ma27:
As discussed in Sunday's meeting, 58ab950 makes it explicit that the old implementation will be deprecated (previously, it wasn't that explicit).

The changes before were mostly about nitpicks & naming. If any of that means that the FCP-phase should be longer, please let me know :)

hmenke (approved):

rfcs/0108-nixos-containers.md
lheckemann: >For your use-case I'd suggest to bind-mount /nix/var/nix/profiles as well

wouldn't that result in `nixos-rebuild` within the container changing the system profile of the host? That sounds disastrous.

Ma27: Not with `nixos-rebuild -p`. However fair, I guess I should actually test this -use-case first.

edolstra:
@Ma27 I don't think it warrants extending the RFC, so we've merged it. Thanks!

PR #111, opened 2021-10-27
[RFC 0111] Ensure Officially Hosted Communications are Public

nrdxp:
[Rendered](https://github.com/nrdxp/rfcs/blob/public-comms/rfcs/0111-public-only-comms.md)

### DIFF
--- a/rfcs/0111-public-only-comms.md
+++ b/rfcs/0111-public-only-comms.md
@@ -1,0 +1,61 @@
+---
+feature: public communication
+start-date: 2021/10/27
+author: nrdxp
+shepherd-team: TBD
+shepherd-leader: TBD
+---
+
+# Summary
+[summary]: #summary
+
+The NixOS community, up to this point, has _de facto_ operated in an open and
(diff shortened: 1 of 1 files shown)


nrdxp:
I'm closing for now. I was planning on taking the week and then redraft a bit, but #114 is basically what I was planning on. In any case, this has gotten completely out of hand, and my intention completely misinterpreted to the point now that I feel it necessary to write a blog post about how I came to this conclusion after nearly 2 1/2 years of thinking about what the biggest blockers in Nix adoption are. 

But the tl;dr is essentially that I realized that the major blockers aren't technical at all (I have 3 or 4 technically focused draft RFCs I've been toying with for a while) but social. This was to be the first in a series of possible policy driven RFCs as I see openness as a basic foundation we should all have been able to rally around. I think everyone would agree that a solid foundation is necessary to establish before working out the finer details.

That said, I find it concerning that some seem willing to bring the community down in flames for the sake of their political affiliation. I am not one of those actors, and rather than leaving this here to cause further derision, I will quietly exit the field for the time being. Those who are interested in my further comments can watch out for my blog post.

PR #113, opened 2021-10-29
Rfc 46: Link platform list

toonn:
My intention was to add a link to the official platform list. However, I didn't manage to find one in the manual, so I only have a couple typo fixes so far.

[This file](https://github.com/NixOS/nixpkgs/blob/8377a7bca967dba811899164d4218d1d4a24b483/lib/systems/supported.nix) was referenced from the RFC #112 discussion. Is this authoritative enough to link to as the official reference?

### DIFF
--- a/rfcs/0046-platform-support-tiers.md
+++ b/rfcs/0046-platform-support-tiers.md
@@ -49,7 +49,7 @@ the manual.
 achievement of the requirements defined in this RFC.
 
 * Discussions of platform support at the Tier 3 and above start using the
-vocubulary based on this RFC and ensure that the requirements of this RFC are
+vocabulary based on this RFC and ensure that the requirements of this RFC are
 met.
 
 ## Platform elements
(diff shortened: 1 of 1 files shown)


7c6f434c:
Some sections of the NixOS manual are auto-generated; so tier definitions + autogenerated list of systems by tier + some manual text about exceptional cases could be an option.

toonn:
Is there precedence for documentation being generated from nix files outside of `doc/`? That would avoid having to synchronize sections of the manual. Maybe it'd be slightly easier to consume if `supported.nix` was a thin nix shim on top of a `supported.json`?

Shall I go ahead and link to `supported.nix` from the paragraph at line 35? One issue I see with making this file the authoritative source on tiers is it doesn't include nearly all the tiers listed in this RFC.

7c6f434c:
https://nixos.org/manual/nixos/stable/options.html for one precedent.

I would expect that Nix code to convert the tier list to Markdown would not be too efficient, but would not be a problem as long as the platform list is relatively small. 

I think creating a parallel PR to add more tiers to `supported.nix` might be a good plan.

piegamesde:
This PR only fixed typos and did not add the link to the platform list. Is this intentional?

Mic92:
The list is here: https://github.com/NixOS/nixpkgs/pull/143732

toonn:
@piegamesde, the typo fixes kinda stood on their own. I didn't really want to claim multiple RFC PR numbers. The links to the list would be added once NixOS/nixpkgs#143732 is resolved. Discussion has somewhat halted on that so I don't have a clear way forward yet. I'll bring it up in chat a couple times.

PR #118, opened 2021-12-03
[RFC 0118] Extra semicolon as a warning instead of an syntax error

lucasew:
Local code flow can leave a semicolon at the end of the file and this should be a warning instead of an error.

[Rendered](https://github.com/lucasew/rfcs/blob/suggestion/semicolon/rfcs/0118-optional-final-semicolon.md)


KAction:
I oppose changing (complicating) the language for the sake of a bit easier programming. Also, it is slippery road, and can cause semicolon mess as in JavaScript (it has obscure rules when newline is interpreted as semicolon).

Any decent editor is capable of pre-save hooks, and something as simple as `sed '$ s/;$//' would solve the problem.

Ericson2314:
Yes semicolons are just used with attribute sets, `let..in` and `with;`. There is no general notion of a "statement".

spacekookie:
The RFC is open for shepherd nominations now

lheckemann:
Nominations are still open, if anyone would like to move this RFC forward this would be welcome.

lheckemann:
I nominate myself as a shepherd for this RFC (full disclosure: I'm not keen on this RFC passing, but aim to take the role of shepherd seriously and don't intend to "veto" or anything).

edolstra:
Since we only have one shepherd nomination at this time, we've decided to convert this RFC to draft status. If people are interested in shepherding, please let us know and we can un-draft it!

rfcs/0118-optional-final-semicolon.md
```diff
@@ -0,0 +1,80 @@
...
+make the parser happy, then you forgot the semicolon when you put the thing back to a let expression. This should be
+a recoverable syntax issue just raising a warning to remove that semicolon when all is over.
+
+This would make iterations a little faster and bring a bit less friction, and give a better clue about what is going on in this case.
+
+# Detailed design
+[design]: #detailed-design
+
+Allow an optional semicolon at the end of an expression raising a warning if it's provided.
+
```
Profpatsch, line 33: This is not a detailed design.

I would expect to get a good idea of how this will be implemented technically.

Note that I tried adding an optional semicolon to the parser a few years ago, but it was rejected because my implementation was naïve (just have a giant lookahead essentially). So I’d like to see a technical design here that is feasiblae.

rfcs/0118-optional-final-semicolon.md
```diff
@@ -0,0 +1,80 @@
...
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+=> 2+2
+4
+=> 2+2;
+warning: extra semicolon at the end of the expression
```
bobvanderlinden, line 40: For files, it should also show which file is having the extra semicolon.

rfcs/0118-optional-final-semicolon.md
```diff
@@ -0,0 +1,80 @@
...
+[drawbacks]: #drawbacks
+
+- Temporary unhandled edge case for secondary nix parsers like [rnix-parser](https://github.com/nix-community/rnix-parser) that by consequency affects [rnix-lsp](https://github.com/nix-community/rnix-lsp)
+
+# Alternatives
+[alternatives]: #alternatives
+
+There are no alternatives so far.
+
+The impact of not doing this is not critic.
```
bobvanderlinden, line 69: One alternative is changing the error from:

```
error: syntax error, unexpected ';', expecting end of file
```

to something like:

```
error: syntax error, unexpected ';', expecting end of file
Consider removing the ';' at the end of the file or expression.

       at «string»:1:4:

            1| 1+2;
             |    ^
```
---

That said, allowing people to write `;`, but outputting a warning might cause warnings in places when the parser touches a file lazily. Potentially warning about extraneous `;` in files that I didn't write myself.

lheckemann:

PR #92, opened 2021-04-26
[RFC 0092] Computed derivations

Ericson2314:
This is the spiritual successor of https://github.com/NixOS/rfcs/pull/40.

[Rendered merged](https://github.com/NixOS/rfcs/blob/master/rfcs/0092-plan-dynamism.md)

[Tracking issue](https://github.com/NixOS/nix/issues/6316)

PR #119, opened 2021-12-30
[RFC 0119] Formalize testing for nixpkgs packages

jonringer:
To better enable review software to run automatic testing, as well as help define when a PR is "tested enough"

rendered: https://github.com/jonringer/rfcs/blob/testing-rfc/rfcs/0119-testing-conventions.md

### DIFF
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -94,7 +94,7 @@ Continue to use current ad-hoc conventions.
 How far should testing go?
 - What consistitutes that "enough testing" was done to a package before a change was merged?
 
-Should `<packaga>.passthru.tests` be flat?
+Should `<package>.passthru.tests` be flat?
 For packages which have extremes in resource usage when testing (e.g. pytorch), it may
 be beneficial to have additional structure for the tests to denote expectations of resources
 and ownership of testing for upstream packages.
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -39,7 +39,7 @@ to help mitigate regressions from appearing in release channels.
 # Detailed design
 [design]: #detailed-design
 
-Standardize `tests.<name>` (previously `passthru.tests.<name>` as a mechanism of 
+Standardize `tests.<name>` (previously known as `passthru.tests.<name>`) as a mechanism of 
 more expensive but automatic testing for nixpkgs. As well as encourage the usage of
 `checkPhase` or `installCheckPhase` when packaging within nixpkgs.
 
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -3,8 +3,8 @@ feature: Defined conventions around testing of official Nixpkgs packages.
 start-date: 2021-12-29
 author: Jonathan Ringer
 co-authors:
-shepherd-team:
-shepherd-leader:
+shepherd-team: @Mic92 @Artturin @kevincox
+shepherd-leader: @Mic92
 related-issues:
   - [RFC 0088 - Nixpkgs Breaking Change Policy](https://github.com/NixOS/rfcs/pull/88)
 ---
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -29,7 +29,7 @@ package has additional tests which can help verify it's correctness.
 
 Breakages are a constant painpoint for nixpkgs. It is a very poor user experience to
 have a configuration broken because one or more packages fail to build. Often when
-these breakages occur, it is because the change had a large impact on the entirity
+these breakages occur, it is because the change had a large impact on the entirety
 of nixpkgs; and unless there's a dedicated hydra jobset for the pull request, it's
 infeasible to expect pull request authors to verify every package affected
 by a change they are proposing. However, it is feasible to specify packages that
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -39,7 +39,7 @@ to help mitigate regressions from appearing in release channels.
 # Detailed design
 [design]: #detailed-design
 
-Standardize `tests.<name>` (previously known as `passthru.tests.<name>`) as a mechanism of 
+Standardize `tests.<name>` (commonly set with the `passthru.tests.<name>` argument to `mkDerivation`) as a mechanism of 
 more expensive but automatic testing for nixpkgs. As well as encourage the usage of
 `checkPhase` or `installCheckPhase` when packaging within nixpkgs.
 
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -69,15 +69,6 @@ of the programs were linked correctly.
 - Assert behavior post installation (e.g. python's native extensions only get installed
 and are not present in a build directory)
 
-# Examples and Interactions
-[examples-and-interactions]: #examples-and-interactions
-
-This section illustrates the detailed design. This section should clarify all
-confusion the reader has from the previous sections. It is especially important
-to counterbalance the desired terseness of the detailed design; if you feel
-your detailed design is rudely short, consider making this section longer
-instead.
-
 # Drawbacks
 [drawbacks]: #drawbacks
 
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -1,6 +1,6 @@
 ---
 feature: Defined conventions around testing of official Nixpkgs packages.
-start-date: 2021-09-03
+start-date: 2021-12-29
 author: Jonathan Ringer
 co-authors:
 shepherd-team:
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -101,5 +101,5 @@ Onboarding of CI tools to support testing paradigms:
 
 Nixpkgs:
 - Add existing nixosTests to related packages
-- Updated testing clause on PR template
+- Update testing clause on PR template
 - Update contributing documentation
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -31,7 +31,7 @@ Breakages are a constant painpoint for nixpkgs. It is a very poor user experienc
 have a configuration broken because one or more packages fail to build. Often when
 these breakages occur, it is because the change had a large impact on the entirity
 of nixpkgs; and unless there's a dedicated hydra jobset for the pull request, it's
-infeasible to expect pull request author(s) to verify every package that's affected
+infeasible to expect pull request authors to verify every package affected
 by a change they are proposing. However, it is feasible to specify packages that
 are very likely to be affected by changes in another package, and use this information
 to help mitigate regressions from appearing in release channels.
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -1,0 +1,114 @@
+---
+feature: Defined conventions around testing of official Nixpkgs packages.
+start-date: 2021-09-03
+author: Jonathan Ringer
+co-authors:
+shepherd-team:
+shepherd-leader:
+related-issues:
+  - [RFC 0088 - Nixpkgs Breaking Change Policy](https://github.com/NixOS/rfcs/pull/88)
+---
+
+# Summary
+[summary]: #summary
+
+When updating or modifying packages, several conventions for testing regressions
+have been adopted. However, these practices are not standard, and generally it's not well
+defined how each testing method should be implemented. It would be beneficial to have
+an unambiguous way to say that a given package, and all downstream dependencies, have
+had as many automated test ran possible. This will give a high degree of certainty that
+a given change is less likely to manifest regressions once introduced on a release
+channel.
+
+Another desire of this rfc is also to have a way for various review tools
+(e.g. ofborg, hydra, nixpkgs-review) to have a standard way to determine if a
+package has additional tests which can help verify it's correctness.
+
+# Motivation
+[motivation]: #motivation
+
+Breakages are a constant painpoint for nixpkgs. It is a very poor user experience to
+have a configuration broken because one or more packages fail to build. Often when
+these breakages occur, it is because the change had a large impact on the entirity
+of nixpkgs; and unless there's a dedicated hydra jobset for the pull request, it's
+infeasible to expect pull request author(s) to verify every package that's affected
+by a change they are proposing. However, it is feasible to specify packages that
+are very likely to be affected by changes in another package, and use this information
+to help mitigate regressions from appearing in release channels.
+
+# Detailed design
+[design]: #detailed-design
+
+Standardize `passthru.tests.<name>` and `passthru.nixosTests.<name>` as a mechanism of 
+more expensive but automatic testing for nixpkgs. As well as encourage the usage of
+`checkPhase` or `installCheckPhase` when packaging within nixpkgs.
+
+Usage for `passthru.nixosTests.<name>`
+- Reserved for tests utilitizing the nixosTest utilties.
+  - Generally these are more resource intensive, and may require additional system features
+  such as kvm
+
+Usage for `passthru.tests.<name>`:
+- Running tests which include downstream dependencies.
+  - This avoids cyclic dependency issues for test suites.
+- Running lengthy or more resource expensive tests.
+  - There should be a priority on making package builds as short as possible.
+  - This reduces the amount of compute required for everyone reviewing, building, or iterating on packages.
+- Referencing downstream dependencies which are most likely to experience regressions.
+  - Most applicable to [RFC 0088 - Nixpkgs Breaking Change Policy](https://github.com/NixOS/rfcs/pull/88),
+as this will help define what breakages a pull request author should take ownership.
+
+Usage for mkDerivation's `checkPhase`:
+- Quick "cheap" tests, which run units tests and maybe some addtional scenarios.
+- Since this contributes to the "build time" of a package, there should be some
+emphasis on ensuring this phase isn't bloated.
+
+Usage for mkDerivations `installCheckPhase`:
+- A quick trivial example (e.g. `<command> --help`) to demonstrate that one (or more)
+of the programs were linked correctly.
+- Assert behavior post installation (e.g. python's native extensions only get installed
+and are not present in a build directory)
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This section illustrates the detailed design. This section should clarify all
+confusion the reader has from the previous sections. It is especially important
+to counterbalance the desired terseness of the detailed design; if you feel
+your detailed design is rudely short, consider making this section longer
+instead.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None? This is opt-in behavior for package maintainers.
+
+# Alternatives
+[alternatives]: #alternatives
+
+Continue to use current ad-hoc conventions.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+How far should testing go?
+- What consistitutes that "enough testing" was done to a package before a change was merged?
+
+Should `<packaga>.passthru.tests` be flat?
+For packages which have extremes in resource usage when testing (e.g. pytorch), it may
+be beneficial to have additional structure for the tests to denote expectations of resources
+and ownership of testing for upstream packages.
+
+# Future work
+[future]: #future-work
+
+Problem with onboarding more test to aspects of nixpkgs CI and processes is the increased
+need of compute, storage, and ram resources. Therefore, consideration of future work should
+take into consideration how much testing is feasible for a given change.
+
+Onboarding of CI tools to support testing paradigms:
+- ofborg
+  - Testing of <package>.passthru.tests is already done.
+  - Testing of downstream dependencies and their tests when minimal (e.g. <10 rebuilds?)
+- hydra
+  - Allow for derivations exposed to hydraJobs to also probe for `<drv>.passthru.tests`?
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -22,7 +22,7 @@ channel.
 
 Another desire of this rfc is also to have a way for various review tools
 (e.g. ofborg, hydra, nixpkgs-review) to have a standard way to determine if a
-package has additional tests which can help verify it's correctness.
+package has additional tests which can help verify its correctness.
 
 # Motivation
 [motivation]: #motivation
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -39,11 +39,11 @@ to help mitigate regressions from appearing in release channels.
 # Detailed design
 [design]: #detailed-design
 
-Standardize `passthru.tests.<name>` as a mechanism of 
+Standardize `tests.<name>` (previously `passthru.tests.<name>` as a mechanism of 
 more expensive but automatic testing for nixpkgs. As well as encourage the usage of
 `checkPhase` or `installCheckPhase` when packaging within nixpkgs.
 
-Criteria for `passthru.tests.<name>`:
+Criteria for `tests.<name>`:
 - Running tests which include downstream dependencies.
   - This avoids cyclic dependency issues for test suites.
 - Running lengthy or more resource expensive tests.
@@ -94,10 +94,10 @@ take into consideration how much testing is feasible for a given change.
 
 Onboarding of CI tools to support testing paradigms:
 - nixpkgs-review
-  - Run `passthru.tests` on affected packages
+  - Run `<package>.tests` on affected packages
   - Allow for filtering based upon requiredSystemFeatures
 - ofborg
-  - Testing of `<package>.passthru.tests` is already done.
+  - Testing of `<package>.tests` is already done.
 
 Nixpkgs:
 - Add existing nixosTests to related packages
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -39,16 +39,11 @@ to help mitigate regressions from appearing in release channels.
 # Detailed design
 [design]: #detailed-design
 
-Standardize `passthru.tests.<name>` and `passthru.nixosTests.<name>` as a mechanism of 
+Standardize `passthru.tests.<name>` as a mechanism of 
 more expensive but automatic testing for nixpkgs. As well as encourage the usage of
 `checkPhase` or `installCheckPhase` when packaging within nixpkgs.
 
-Usage for `passthru.nixosTests.<name>`
-- Reserved for tests utilitizing the nixosTest utilties.
-  - Generally these are more resource intensive, and may require additional system features
-  such as kvm
-
-Usage for `passthru.tests.<name>`:
+Criteria for `passthru.tests.<name>`:
 - Running tests which include downstream dependencies.
   - This avoids cyclic dependency issues for test suites.
 - Running lengthy or more resource expensive tests.
@@ -57,6 +52,9 @@ Usage for `passthru.tests.<name>`:
 - Referencing downstream dependencies which are most likely to experience regressions.
   - Most applicable to [RFC 0088 - Nixpkgs Breaking Change Policy](https://github.com/NixOS/rfcs/pull/88),
 as this will help define what breakages a pull request author should take ownership.
+- Running integration tests (E.g. nixosTests)
+  - Tests which have heavy usage or platform requirements should add the appropriate systemFeature
+    - E.g. `nixos-test` `kvm` `big-parallel`
 
 Usage for mkDerivation's `checkPhase`:
 - Quick "cheap" tests, which run units tests and maybe some addtional scenarios.
@@ -85,21 +83,23 @@ Continue to use current ad-hoc conventions.
 How far should testing go?
 - What consistitutes that "enough testing" was done to a package before a change was merged?
 
-Should `<package>.passthru.tests` be flat?
-For packages which have extremes in resource usage when testing (e.g. pytorch), it may
-be beneficial to have additional structure for the tests to denote expectations of resources
-and ownership of testing for upstream packages.
+Hydra: How would this look for hydra adoption and hydraChecks?
 
 # Future work
 [future]: #future-work
 
-Problem with onboarding more test to aspects of nixpkgs CI and processes is the increased
+One problem with onboarding more tests to the current nixpkgs CI and processes is the increased
 need of compute, storage, and ram resources. Therefore, consideration of future work should
 take into consideration how much testing is feasible for a given change.
 
 Onboarding of CI tools to support testing paradigms:
+- nixpkgs-review
+  - Run `passthru.tests` on affected packages
+  - Allow for filtering based upon requiredSystemFeatures
 - ofborg
   - Testing of `<package>.passthru.tests` is already done.
-  - Testing of downstream dependencies and their tests when minimal (e.g. <10 rebuilds?)
-- hydra
-  - Allow for derivations exposed to hydraJobs to also probe for `<drv>.passthru.tests`?
+
+Nixpkgs:
+- Add existing nixosTests to related packages
+- Updated testing clause on PR template
+- Update contributing documentation
--- a/rfcs/0119-testing-conventions.md
+++ b/rfcs/0119-testing-conventions.md
@@ -108,7 +108,7 @@ take into consideration how much testing is feasible for a given change.
 
 Onboarding of CI tools to support testing paradigms:
 - ofborg
-  - Testing of <package>.passthru.tests is already done.
+  - Testing of `<package>.passthru.tests` is already done.
   - Testing of downstream dependencies and their tests when minimal (e.g. <10 rebuilds?)
 - hydra
   - Allow for derivations exposed to hydraJobs to also probe for `<drv>.passthru.tests`?


rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
+---
+feature: Defined conventions around testing of official Nixpkgs packages.
+start-date: 2021-09-03
```
jonringer, line 3: oops, been dragging my feet on this for a while...
```suggestion
start-date: 2021-29-12
```

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+as this will help define what breakages a pull request author should take ownership.
+
+Usage for mkDerivation's `checkPhase`:
+- Quick "cheap" tests, which run units tests and maybe some addtional scenarios.
+- Since this contributes to the "build time" of a package, there should be some
+emphasis on ensuring this phase isn't bloated.
+
+Usage for mkDerivations `installCheckPhase`:
+- A quick trivial example (e.g. `<command> --help`) to demonstrate that one (or more)
+of the programs were linked correctly.
```
0x4A6F, line 68: There is also `testVersion` introduced in NixOS/nixpkgs#121896.

Synthetica9: Related: https://github.com/NixOS/nixpkgs/pull/144503

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+have a configuration broken because one or more packages fail to build. Often when
+these breakages occur, it is because the change had a large impact on the entirity
+of nixpkgs; and unless there's a dedicated hydra jobset for the pull request, it's
+infeasible to expect pull request author(s) to verify every package that's affected
+by a change they are proposing. However, it is feasible to specify packages that
+are very likely to be affected by changes in another package, and use this information
+to help mitigate regressions from appearing in release channels.
+
+# Detailed design
+[design]: #detailed-design
```
7c6f434c, line 40: Maybe expand on what are the intended changes to the status quo?

`passthru.tests` is a name documented inside the manual, however `nixosTests` are recommended to be also put there.

(also, if sorting by resource consumption, maybe this split is not needed?)

Are we encouraged to compromise on something in the name of more test coverage?

jonringer: Well, I wasn't sure what the status quo should be. My current thoughts are, "here is some addtional metadata you can add to ensure that people know how your package may break. Or add your package to the tests of other packages to ensure it's not broken."

Ekleog:
FWIW, I think part of the statu quo comes from https://github.com/NixOS/nixpkgs/pull/44439 (which switched to passthru.tests after that). That said, it'd probably be good to make it into a proper RFC, if only for documentation, now that we actually have a working RFC system! It'd also clarify which tests should go into which phase. (For instance, I'd maybe have put the `--help` into a full-blown test rather than an `installCheckPhase`, thinking that the environment would be closer to actual production environments this way)

That said, I'm not sure I feel really good about adding `passthru.nixosTests`: it'd require re-adjusting all automated tools like r-ryantm or ofborg to take this new attribute into consideration too. I'd probably feel better about this RFC adding a `passthru.tests.nixos.<name>` sublist. Though technically the previous RFC didn't allow such behavior, I think most tools would support such a change natively (at least my recollection of the changes needed to r-ryantm and ofborg are it should work naturally, as AFAIR nix recurses into the whole set and not a single layer when being asked to build a set)

Does that make sense? :)

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+Usage for `passthru.nixosTests.<name>`
+- Reserved for tests utilitizing the nixosTest utilties.
+  - Generally these are more resource intensive, and may require additional system features
+  such as kvm
+
+Usage for `passthru.tests.<name>`:
+- Running tests which include downstream dependencies.
+  - This avoids cyclic dependency issues for test suites.
+- Running lengthy or more resource expensive tests.
+  - There should be a priority on making package builds as short as possible.
```
mweinelt, line 55: I hear you! :smile:

jonringer: I think I originally wrote this about sage, and sageWithTests. Which took even longer :)

jonringer:
> That said, I'm not sure I feel really good about adding `passthru.nixosTests`: it'd require re-adjusting all automated tools like r-ryantm or ofborg to take this new attribute into consideration too.

That shouldn't be much work. And be adopted slowly, much like `passthru.tests`.

>  I'd probably feel better about this RFC adding a passthru.tests.nixos.<name> sublist. Though technically the previous RFC didn't allow such behavior, I think most tools would support such a change natively (at least my recollection of the changes needed to r-ryantm and ofborg are it should work naturally, as AFAIR nix recurses into the whole set and not a single layer when being asked to build a set)

Seems to be very similar to `nixosTests`. Also, `nix-build` will only build the top-level unless you do a `recurseForDerivations`. And that will only work if you do something like hydra-eval or `nix-env -qa`. If I do `nix-build -A python3`, It's not also going to build `python3.pkgs`, even though it's an attr.

Also, `nix build` will only build a derivation, `nix-build` will build a list, attr set, or derivation.

rfcs/0119-testing-conventions.md
7c6f434c: (looks sadly at NixOS/Nix#3600 being unmerged) although for a given situation with a package, speeding up separate tests on their own would also not hurt

7c6f434c:
> FWIW, I think part of the statu quo comes from NixOS/nixpkgs#44439 (which switched to passthru.tests after that). That said, it'd probably be good to make it into a proper RFC, if only for documentation, now that we actually have a working RFC system! It'd also clarify which tests should go into which phase.

I think the documentation proper, i.e. Nixpkgs manual, already describes quite a few of things about tests. So if we want to establish substantially new recommendation (even if on purely advisory level), we should say which parts are new, if we think the manual misses minor points of current practice, we could just fix it there, and if RFC is a better point of documentation than the manual, then maybe we should say it out loud so that this problem can be discussed on its own.

rfcs/0119-testing-conventions.md
Profpatsch: This should be YYYY-MM-DD instead

jonringer: Yep, sorry. American brain.

Ericson2314:
I add that `checkPhase` and `installCheckPhase` are always second-choice options, i.e. a separate derivation, no matter how tiny, is always preferable if it can be cleanly separated from the build.

Ericson2314:
Separately, I dislike the fact that `nixosTests` Currently violate the nixpkgs vs nixos layering. If we could make them them functions that are passed some nixos stuff I would prefer that. This is orthogonal to rest of the RFC, but if we are formalizing things maybe it's good time to think about this stuff.

jonringer:
> I add that `checkPhase` and `installCheckPhase` are always second-choice options, i.e. a separate derivation, no matter how tiny, is always preferable if it can be cleanly separated from the build.

On the contrary, a small test suite which can quickly identify whether or not a package should be considered "correct" I think is highly desired. Until we have tooling in place to do something like, "please test everything as well", then I'm going to still insist on unit tests as a good practice.

Maybe in another RFC, we can standardize around making a `passthru.tests.check` and `passthru.tests.installCheck` the normal.

> Separately, I dislike the fact that nixosTests Currently violate the nixpkgs vs nixos layering. If we could make them them functions that are passed some nixos stuff I would prefer that. This is orthogonal to rest of the RFC, but if we are formalizing things maybe it's good time to think about this stuff.

Sure... but nixpkgs is for packaging software, and nixos can be thought of as the practice of combining it. It's immensely useful to verify that it works as intended with other software.

It's a better user experience if we know that a nextcloud upgrade breaks the nextcloud+mysql integration. As that's one way that people are likely to leverage it.

The whole purpose of the RFC is to make changes with more certainty. From a committer's perspective, to change "I'm scared to press merge on something " to, "The build passed, the tests passed, the coupled packages passed, the nixos tests passed. I have a high degree of certainty that merging this is a good idea." Or even enable the ability to confidently do a "merge when CI passes".

jonringer:
> Separately, I dislike the fact that `nixosTests` Currently violate the nixpkgs vs nixos layering.

meanwhile https://github.com/NixOS/nixpkgs/pull/153551

AndersonTorres:
> I add that `checkPhase` and `installCheckPhase` are always second-choice options, i.e. a separate derivation, no matter how tiny, is always preferable if it can be cleanly separated from the build.

It looks like a good idea! 
I ponder about a `tests.nix` file and a `passthru.tests = import ./tests.nix;`.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+package has additional tests which can help verify it's correctness.
+
+# Motivation
+[motivation]: #motivation
+
+Breakages are a constant painpoint for nixpkgs. It is a very poor user experience to
+have a configuration broken because one or more packages fail to build. Often when
+these breakages occur, it is because the change had a large impact on the entirity
+of nixpkgs; and unless there's a dedicated hydra jobset for the pull request, it's
+infeasible to expect pull request author(s) to verify every package that's affected
```
kevincox, line 34:
```suggestion
infeasible to expect pull request authors to verify every package affected
```

Nit: Unnecessary.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+Standardize `passthru.tests.<name>` and `passthru.nixosTests.<name>` as a mechanism of 
+more expensive but automatic testing for nixpkgs. As well as encourage the usage of
+`checkPhase` or `installCheckPhase` when packaging within nixpkgs.
+
+Usage for `passthru.nixosTests.<name>`
+- Reserved for tests utilitizing the nixosTest utilties.
```
kevincox, line 47: This doesn't seem a meaningful distinction to me. Maybe we should flip this.

- Reserved for tests that are more resource intensive or that require additional system features such as kvm.
  - Generally these are tests using the `nixosTest` utilities.

Of course this makes the name less meaningful, but we can bikeshed that later.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+
+Usage for mkDerivation's `checkPhase`:
+- Quick "cheap" tests, which run units tests and maybe some addtional scenarios.
+- Since this contributes to the "build time" of a package, there should be some
+emphasis on ensuring this phase isn't bloated.
+
+Usage for mkDerivations `installCheckPhase`:
+- A quick trivial example (e.g. `<command> --help`) to demonstrate that one (or more)
+of the programs were linked correctly.
+- Assert behavior post installation (e.g. python's native extensions only get installed
```
kevincox, line 69: The "only" in this sentence is confusing me.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+your detailed design is rudely short, consider making this section longer
+instead.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None? This is opt-in behavior for package maintainers.
+
+# Alternatives
+[alternatives]: #alternatives
```
kevincox, line 87: One alternative is that we consider *all* dependent packages as tests. We can have dependent packages that are *just* tests for example a `testFoobar` package to test `foobar`.

Then a PR author would responsible that all dependents build (aka pass) or are marked broken.

The obvious issue here is that for packages with lots of dependents it becomes infeasible for the average author to run a tool that builds everything and marks failures as broken. I think it is worth mentioning this alternative because this RFC demonstrates a clean way to define an appropriate sample size. Then it is expected that nixpkgs-provided build resources can be used for the full build + mark as broken.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+
+Continue to use current ad-hoc conventions.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+How far should testing go?
+- What consistitutes that "enough testing" was done to a package before a change was merged?
+
+Should `<packaga>.passthru.tests` be flat?
```
kevincox, line 97: I vote yes. If you need some sort of differentiation you can use a naming convention such as `ownerSomething`. For now these conventions can be per-package and if we see enough of these arise we can consider unifying them or adding more structure once we have clear use cases.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+For packages which have extremes in resource usage when testing (e.g. pytorch), it may
+be beneficial to have additional structure for the tests to denote expectations of resources
+and ownership of testing for upstream packages.
```
kevincox, line 100: I think this differentiation can be solved by `nixosTests` vs `tests`.

kevincox:
I nominate myself and @Ericson2314 as shepherds.

Profpatsch:
> > Separately, I dislike the fact that nixosTests Currently violate the nixpkgs vs nixos layering. If we could make them them functions that are passed some nixos stuff I would prefer that. This is orthogonal to rest of the RFC, but if we are formalizing things maybe it's good time to think about this stuff.
> 
> Sure... but nixpkgs is for packaging software, and nixos can be thought of as the practice of combining it. It's immensely useful to verify that it works as intended with other software.
> 
> It's a better user experience if we know that a nextcloud upgrade breaks the nextcloud+mysql integration. As that's one way that people are likely to leverage it.

I think it’s fine in this context, because it is not *provided* to nixos, but nixos is used as an implementation detail to run the tests, which could in principle also be done by a different runtime.

In addition, due to the functional & lazy nature of nix it doesn’t actually incur a static dependency on nixpkgs in the derivation, so, there is no technical layering breakage either.

rfcs/0119-testing-conventions.md
jonringer: Native extensions only get installed. However, most test suites will consume the code in the build directory. So tests will fail because the compiled extensions will not be present.

I'm not sure how to word this better.

rfcs/0119-testing-conventions.md
jonringer: Yes, and this is the compromise with `passthru.tests` listing downstream dependencies. The idea is to list packages which are "likely to break with breaking changes".

For example, some packages may make use of many of systemd's features, however, other packages only really use libudev, which is much more stable. We could probably forego the libudev packages, and just list the packages which are using systemd's more advanced features.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+by a change they are proposing. However, it is feasible to specify packages that
+are very likely to be affected by changes in another package, and use this information
+to help mitigate regressions from appearing in release channels.
+
+# Detailed design
+[design]: #detailed-design
+
+Standardize `passthru.tests.<name>` and `passthru.nixosTests.<name>` as a mechanism of 
+more expensive but automatic testing for nixpkgs. As well as encourage the usage of
+`checkPhase` or `installCheckPhase` when packaging within nixpkgs.
```
legendofmiracles, line 44: I'm confused where to draw the line between what should be done in `installCheckPhase` and non-vm `passthru.tests`. When reviewing this has been a source of confusion, see: https://github.com/NixOS/nixpkgs/pull/153496 as an example. (Reading this I'm no longer convinced the decision to move it from installCheckPhase was the right one :>)


cc @thiagokokada so that you can add your thoughts

thiagokokada: Also, IMO, VM-less `passthru.tests` makes more sense for tests like the one described in this document: maybe we want to see how this particular package being build interacts with other packages. But if the idea is test only the current package, `installCheckPhase` generally makes more sense to me (of course, there could be exceptions).

7c6f434c: I think ideally we want expensive build steps spread among different derivations, but that's hard. Expensive tests are sometimes feasible to split, so splitting them is good when it doesn't create a mess.

(Cheap checks with no new deps do not follow this logic, of course)

Also, separated tests make it more practical to support packages with different level of «works» on different platforms, because we do not need to touch the main build causing  expensive rebuild on the «happy» platforms to add partial support (and its tests) on the more exotic ones. There is no single «package is broken», in the black-and-white view we have never had a non-broken LibreOffice packages, for example.

spacekookie:
A bit late to the party but this RFC is open for nominations :)

lheckemann:
Current nominations:
- @Ericson2314 not responded
- @kevincox accepted

More nominations to move this RFC forward are welcome!

7c6f434c:
I nominate: @cole-h  @Mic92 @Ekleog @Profpatsch @timokau @SuperSandro2000 (fully expecting most to say they do not have time right now)

rfcs/0119-testing-conventions.md
timokau: Separated tests also make it easier to debug failures, since you can easily access the package (which may build successfully but then fail the test). I think we should follow something like a "principle of least privilege", where it is best practice to always use the test that makes the fewest assumptions about its environment.

`checkPhase` assumes it is run within the build environment. That's a lot of assumptions. This increases build time for quick iteration, makes it harder to debug failures and also makes the test results less meaningful since the environment differs from the end-user's environments.

`installCheckPhase` has somewhat fewer assumptions than `checkPhase`, but still many of the same disadvantages. The results are more meaningful though, since the environment is more similar to the end-user's environment.

`passthru.tests` makes the fewest assumptions. If the tests pass in this "clean environment", they will likely pass in the user's environment as well (unless tests are actually broken by some environment variable or something similar). They can be run as-needed (but should be run (semi-)automatically in some reliable fashion). They make debugging easier.

`passthru.nixosTests` assumes access to a nixos VM, but does not assume a special build environment.

As a result, I'd suggest to pick `passthru.tests` > `passthru.nixosTests` > `installCheckPhase` > `checkPhase`. Practicality is king of course: If upstream assumptions make it trivial to run the test suite in `checkPhase` but hard to run them in `passthru.tests`, then its better to have some tests in `checkPhase` than to have no tests at all.

SuperSandro2000:
I don't think I have currently enough time to also take this task in addition to all my other tasks and life.

rfcs/0119-testing-conventions.md
abathur: I don't have much ~ecosystem perspective on this, but I do have a few scattered thoughts from authoring and maintaining a small number of packages (broadly, my thinking has been aligning with timokau over time):

1. I like `passthru.tests` at the front of the list. I can think of at least a few times I've been in the middle of some meaningful work and ended up derailed by an immediate or remote test or test-dependency failure:

    - I've started to see a check/test dependency on a linter or formatter as an antipattern in my own work. At least twice I've had a linter/formatter package turn up broken, or a bump in nixpkgs to trigger a rebuild that fails due to a lint/format check.
    
    - Recently I noticed that resholve's CLI arg parser dependency wasn't building on aarch64-darwin due to some failing tests. This was a pain since it was a platform I don't have readily available. I ended up overriding the dependency to disable checks for expediency, since resholve's own test suite (which I factored out into `passthru.tests` a while back) already exercises the CLI and should fail on borg/hydra if it did ultimately depend on whatever caused the test breaks.
    
2. I've thought for a while that it would be nice to have a low-friction mechanism for collecting example uses (say, a nudge to submit them + a CLI/form/template for doing so? Examples themselves might just be a valid invocation of some CLI, a sample file that exercises some parser, etc.) from package users. It occurs to me now that if there was a ~standard `passthru.tests.x` attr for these, it could at least drive the nudge (i.e. example, if you open a nix-shell with something that has fewer than N?)
3. I like prioritizing `installCheckPhase` over `checkPhase`. I have some minor insecurity about `checkPhase` providing false-confidence about something that `installPhase` or `fixupPhase` then breaks (especially since resholve presently runs in fixup).

rfcs/0119-testing-conventions.md
abathur: I assume this means: use a list and not a set? I guess two reasons not to flatten:
1. If the tests are slow, I'm sure it'll occasionally help someone to be able to target a specific one?
2. If there's enough speculative value in having some ~well-known attrs dedicated to a specific purpose (as in bullet 2 of https://github.com/NixOS/rfcs/pull/119/files#r805415330)?

kevincox:
Hello @cole-h @Mic92 @Ekleog @Profpatsch, please accept or reject your shepherd nominations.

cole-h:
I'm sorry, I must've let the previous ping slip through the cracks.

As much as I'd love to help out with this, I currently cannot spare any time.

Mic92:
Beeing pinged in the release notes definitely helped. I am in for this RFC.

abathur:
I've been considering moving a package's installCheck out into a derivation. At first I just wanted to re-use the package's source, but something in configure or build is necessary to prepare the tests as well. Since the package's test suite takes much longer to run than the build, I'm currently using a kludge like:

```nix
passthru.tests.upstream = package.overrideAttrs (old: {
  name = "${package.name}-tests";
  installCheckInputs = [ ... ];
  inherit doInstallCheck;
  installCheckPhase = ''
    : run tests

    rm -rf $out # no normal output
    touch $out # clean tests
  '';
});
```

This reminded me of something I wondered while working on a feature for resholve. 

For now, this feature categorizes executables based on whether they use an exec-like API or not. The (probably not achievable) aspiration is actually to categorize them based on whether they exec user-supplied args. The current implementation just maps packages through a function to analyze executables in `$out/bin/`, but I suspect there are (now or in the future) narrower tools that can analyze the source or IR to get better answers for specific languages (or sets of them, in the case of something like semgrep).

The value of any incremental improvement to how well resholve can answer this question will never justify re-building packages just to access the build directory. 

But I wondered whether there are, at Nixpkgs scale, enough interesting potential-uses of prepared build directories to justify preserving them as an output. The ability to separate out the test phases and arbitrarily run them without a rebuild probably qualifies.

Profpatsch:
I’m sorry I don’t think I can do a good job with shepherding, but I’d like to see this RFC merged.

AndersonTorres (changes requested):

rfcs/0119-testing-conventions.md
jonringer: I was trying to keep it consistent in plurality with the already used `passthru.tests`. Also `pkgs.nixosTests` is how it's exposed in nixpkgs.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This section illustrates the detailed design. This section should clarify all
+confusion the reader has from the previous sections. It is especially important
+to counterbalance the desired terseness of the detailed design; if you feel
+your detailed design is rudely short, consider making this section longer
+instead.
+
```
jonringer, line 80: Not needed.

```suggestion
```

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+# Future work
+[future]: #future-work
+
+Problem with onboarding more test to aspects of nixpkgs CI and processes is the increased
+need of compute, storage, and ram resources. Therefore, consideration of future work should
+take into consideration how much testing is feasible for a given change.
+
+Onboarding of CI tools to support testing paradigms:
+- ofborg
+  - Testing of <package>.passthru.tests is already done.
```
jonringer, line 111:
```suggestion
  - Testing of `<package>.passthru.tests` is already done.
```

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,114 @@
...
+
+Continue to use current ad-hoc conventions.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+How far should testing go?
+- What consistitutes that "enough testing" was done to a package before a change was merged?
+
+Should `<packaga>.passthru.tests` be flat?
```
jonringer, line 97:
```suggestion
Should `<package>.passthru.tests` be flat?
```

rfcs/0119-testing-conventions.md
jonringer: I meant for it to be as a single attr set. I wanted to avoid: `passthru.tests.scenarioA.variantB.test`

Artturin:
I nominate myself

kevincox:
I think we have enough shepherds then!

@Mic92, @Artturin and myself.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+these breakages occur, it is because the change had a large impact on the entirety
+of nixpkgs; and unless there's a dedicated hydra jobset for the pull request, it's
+infeasible to expect pull request authors to verify every package affected
+by a change they are proposing. However, it is feasible to specify packages that
+are very likely to be affected by changes in another package, and use this information
+to help mitigate regressions from appearing in release channels.
+
+# Detailed design
+[design]: #detailed-design
+
```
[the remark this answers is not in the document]
risicle: Think this technique was @jonringer 's idea in the first place :wink:

The main thing I'm slightly uncomfortable with is adding the reverse dependencies as arguments - I imagine it causing wierdness for people who do a lot of overriding. But I can't think of a sensible way it could be avoided.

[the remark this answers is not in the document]
jonringer: Ideally the benefit is for nixpkgs testing. For the overriding case, it should be that an override would not trigger a downstream dependency to influence the build as passthru gets pruned before instantiation.

I agree it's odd that a package is aware of it's downstream dependencies, but I'm not sure of another way of having a package be aware of them.

7c6f434c:
>I agree it's odd that a package is aware of it's downstream dependencies, but I'm not sure of another way of having a package be aware of them. 

Nominally, storing «selectors» as in lists of attribute names to find the crucial dependants is a way. Of course you get later and possibly weaker checking this way.

edolstra:
We have @Mic92, @Artturin and @kevincox as shepherds. Thanks! @Mic92 Do you want to be shepherd leader?

Artturin:
POC of building all the tests of a package

```bash
#!/usr/bin/env bash
arg="$1.tests"
tests=$(nix eval "$arg" --json | jq)
storepaths=$(jq -r ".[]" <<< "$tests")
drvs=()

for path in $storepaths; do
    drvs+=("$(nix show-derivation "$path" | jq -r 'to_entries[] | .key')")
done

for drv in "${drvs[@]}"; do
    echo "building $drv"
    nix build "$drv"
    echo "built $drv"
done
```

jonringer:
@Artturin 

Another option:
```
#!/usr/bin/env bash
attr=$1

nix-build -E "with import ./. { }; lib.attrValues ($1.passthru.tests or {})"
```

Artturin:
> @Artturin
> 
> Another option:
> 
> ```
> #!/usr/bin/env bash
> attr=$1
> 
> nix-build -E "with import ./. { }; lib.attrValues ($1.passthru.tests or {})"
> ```
> 
> This will also allow for the builds to occur in parallel, as the entire list is passed to nix

didn't think of it not being parallel

here's a improved ver

```bash
#!/usr/bin/env bash
arg="$1.tests"
tests=$(nix eval "$arg" --json | jq)
storepaths=$(jq -r ".[]" <<< "$tests")
drvs=()

for path in $storepaths; do
    drvs+=("$(nix show-derivation "$path" | jq -r 'to_entries[] | .key')")
done

echo "building "${drvs[@]}""
nix build "${drvs[@]}"
echo "built ${drvs[@]}"
```

terlar:
This could work as well:
```
arg="$1.tests"
drvs=("$(nix eval --raw "$arg" --apply 'x: builtins.concatStringsSep " " (map (builtins.getAttr "drvPath") (builtins.attrValues x))')")

echo "building "${drvs[@]}""
nix build "${drvs[@]}"
echo "built ${drvs[@]}"
```

If Nix had a builtins of something like `nixpkgs.lib.pipe`, but that took the thing to pipe as last argument):
```
(builtins.? [builtins.attrValues (map (builtins.getAttr "drvPath")) (builtins.concatStringsSep " ")])
```

abathur:
@edolstra I asked on Matrix why `nix-build -A <pkg>.tests` wouldn't be workable as an alternative to what the comments above are picking at and it seems to come down to at least 3 answers:
- It doesn't work in the new `nix build` CLI.
- It only builds the top level, and there's some discussion of nesting some semantics within this attrset.
- There's likely a need for asserting better control over what gets built for cases like nixpkgs-review where they may only want to build a fraction of what's included or skip known-expensive ones.

This just made me wonder if these kind of toolchain-focused use-cases drive enough ~ecosystem value to merit direct support, even if it's tucked away somewhere that normal end-users won't be tempted?

Mic92:
I created #community-rfc119:nixos.dev to coordinate a first meeting.

Artturin:
> This could work as well:
> 
> ```
> arg="$1.tests"
> drvs=("$(nix eval --raw "$arg" --apply 'x: builtins.concatStringsSep " " (map (builtins.getAttr "drvPath") (builtins.attrValues x))')")
> 
> echo "building "${drvs[@]}""
> nix build "${drvs[@]}"
> echo "built ${drvs[@]}"
> ```
> 
> If Nix had a builtins of something like `nixpkgs.lib.pipe`, but that took the thing to pipe as last argument):
> 
> ```
> (builtins.? [builtins.attrValues (map (builtins.getAttr "drvPath")) (builtins.concatStringsSep " ")])
> ```

https://github.com/NixOS/nix/pull/6333 would make it possible to use apply with nix build

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+- Quick "cheap" tests, which run units tests and maybe some addtional scenarios.
+- Since this contributes to the "build time" of a package, there should be some
+emphasis on ensuring this phase isn't bloated.
+
+Usage for mkDerivations `installCheckPhase`:
+- A quick trivial example (e.g. `<command> --help`) to demonstrate that one (or more)
+of the programs were linked correctly.
+- Assert behavior post installation (e.g. python's native extensions only get installed
+and are not present in a build directory)
+
```
Artturin, line 71: some programs are hard to test automatically so how about creating a new meta attribute like `testingInstructions`

kevincox: I strongly oppose this idea.

IIUC the goal of this RFC is to make it easier for changes to upstream packages to be tested. The end goal is that we have automatic tooling that can test packages, notify maintainers of breakages and eventually mark as broken if the maintainers are unable to fix the package in time. Adding required manual testing puts unacceptable levels of burden on core package maintainers (that are depended on by hundreds or thousands of other packages).

I think a `testingInstructions` attribute may be an interesting and useful idea but I think it would serve a different purpose as the formalized testing specified by this RFC.

TL;DR I don't want to require people to manually test loads of packages, if you want your package not to break due to changes in dependencies you need automated tests.

7c6f434c: I guess `testingInstructions` would fit here if it came with «replacing this with a test that looks robust will be accepted even if the test implementation is ugly», but I do not believe such a commitment would be accepted into the attribute semantics…

Mic92:
I would like to schedule a first meeting:
https://www.when2meet.com/?15384463-X7OwB

Please fill out when you are available @jonringer @Artturin @kevincox

Mic92:
We will meet up Saturday the 30th April to discuss this topic.

Artturin:
https://github.com/NixOS/nixpkgs/pull/119942
will be useful because then we won't have to add the package to its own args
https://github.com/NixOS/nixpkgs/pull/119942/files#diff-83080343d86eefa6f2dc5fe92f5131a86267717ed97e1d74392d0a002d49ed2c

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+  - This avoids cyclic dependency issues for test suites.
+- Running lengthy or more resource expensive tests.
+  - There should be a priority on making package builds as short as possible.
+  - This reduces the amount of compute required for everyone reviewing, building, or iterating on packages.
+- Referencing downstream dependencies which are most likely to experience regressions.
+  - Most applicable to [RFC 0088 - Nixpkgs Breaking Change Policy](https://github.com/NixOS/rfcs/pull/88),
+as this will help define what breakages a pull request author should take ownership.
+
+Usage for mkDerivation's `checkPhase`:
+- Quick "cheap" tests, which run units tests and maybe some addtional scenarios.
```
roberth, line 62: Not all test suites are quick or cheap, but running them should be a priority over quickness. If we can make running them in a separate derivation easy, that's worth considering, but it seems that the human overhead would not be worth it _in the general case_.
A lot could factor into this, so I think we should make this less prescriptive.

```suggestion
- Preferably quick "cheap" tests, which run units tests and maybe some addtional scenarios.
```

rfcs/0119-testing-conventions.md
roberth: Maybe start with the phrase about the build directory?

rfcs/0119-testing-conventions.md
roberth: Perhaps we could record merge conditions instead?
Some packages don't need manual testing, so all that's needed for a merge is a review of the changelog, but this information has not been recorded yet.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+Problem with onboarding more test to aspects of nixpkgs CI and processes is the increased
+need of compute, storage, and ram resources. Therefore, consideration of future work should
+take into consideration how much testing is feasible for a given change.
+
+Onboarding of CI tools to support testing paradigms:
+- ofborg
+  - Testing of `<package>.passthru.tests` is already done.
+  - Testing of downstream dependencies and their tests when minimal (e.g. <10 rebuilds?)
+- hydra
+  - Allow for derivations exposed to hydraJobs to also probe for `<drv>.passthru.tests`?
```
roberth, line 105: This could be implemented in `release.nix` files.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+# Future work
+[future]: #future-work
+
+Problem with onboarding more test to aspects of nixpkgs CI and processes is the increased
+need of compute, storage, and ram resources. Therefore, consideration of future work should
+take into consideration how much testing is feasible for a given change.
+
+Onboarding of CI tools to support testing paradigms:
+- ofborg
+  - Testing of `<package>.passthru.tests` is already done.
```
roberth, line 102: The technical goal of `passthru` is to set attributes in the package that are not in the derivation, a solution to the problem of `(pkg // { x = y; }).overrideAttrs f` losing `x`.
Since #92, the notion of package needs to be more decoupled from derivation. In other words, `mkDerivation` won't be in charge of the package (attrset) anymore. In this new world, it does not make sense to expose all `mkDerivation` arguments as attributes on the package, so the need for "passthru" becomes conceptually redundant.
As such, we should not call the `tests` attribute on a package `passthru.tests`.
Before #92, conflating the two was convenient, but it will be breaking and confusing instead.

Mic92:
# Shepherd meeting notes from 30th April for RFC 119

* Tests of the updated package should be become mandatory for acceptance of a PR
    * Contributor documentation needs to be extended
* Check nixos tests if have `requiredSystemFeatures = [ "nixos-test" ]` set 
  rather than having a seperate nixosTests attribute 
* New tests should be added affected packages via `passthru.tests`
    * Future work: Check that existing tests are added as passthru.tests
* Extend nixpkgs-review to support passthru.tests
    * Nixos tests need to evaluate seperate (memory)
    * Other tests could be build right away

jonringer:
Updated RFC with feedback from meeting, let me know if I missed anything.

rfcs/0119-testing-conventions.md
jonringer: I removed this line, and added it to unresolved section, as it's more of a hydra concern than anything.

Also, the ultimate goal of hydra is to provide channel updates for nixpkgs, in which these tests are more "nice to have" than anything else. 

Also, I'm not sure if the official hydra should be building all of the tests, if the tests are only meant to influence whether  a change should be merged. The official hydra could be populating the cache with a lot of never-to-be-used-again builds.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+[design]: #detailed-design
+
+Standardize `passthru.tests.<name>` and `passthru.nixosTests.<name>` as a mechanism of 
+more expensive but automatic testing for nixpkgs. As well as encourage the usage of
+`checkPhase` or `installCheckPhase` when packaging within nixpkgs.
+
+Usage for `passthru.nixosTests.<name>`
+- Reserved for tests utilitizing the nixosTest utilties.
+  - Generally these are more resource intensive, and may require additional system features
+  such as kvm
```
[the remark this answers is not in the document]
jonringer: We decided in the meeting to use `requiredSystemFeatures` to influence the intensity of a review.

It re-uses existing conventions, and largely achieves the same goal. The burden of filtering the tests will be on the tooling side, not the maintainer side.

rfcs/0119-testing-conventions.md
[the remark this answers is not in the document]
jonringer: usage of `pkgs` is frowned upon in nixpkgs from a hygiene perspective, I would rather not involve it

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+Onboarding of CI tools to support testing paradigms:
+- nixpkgs-review
+  - Run `passthru.tests` on affected packages
+  - Allow for filtering based upon requiredSystemFeatures
+- ofborg
+  - Testing of `<package>.passthru.tests` is already done.
+
+Nixpkgs:
+- Add existing nixosTests to related packages
+- Updated testing clause on PR template
```
jonringer, line 104: Future work should use present tense
```suggestion
- Update testing clause on PR template
```

rfcs/0119-testing-conventions.md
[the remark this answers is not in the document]
roberth: Some in-nixpkgs provisions may still be good, as it can shrink the interface between the tooling and nixpkgs.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+
+Criteria for `passthru.tests.<name>`:
+- Running tests which include downstream dependencies.
+  - This avoids cyclic dependency issues for test suites.
+- Running lengthy or more resource expensive tests.
+  - There should be a priority on making package builds as short as possible.
+  - This reduces the amount of compute required for everyone reviewing, building, or iterating on packages.
+- Referencing downstream dependencies which are most likely to experience regressions.
+  - Most applicable to [RFC 0088 - Nixpkgs Breaking Change Policy](https://github.com/NixOS/rfcs/pull/88),
+as this will help define what breakages a pull request author should take ownership.
```
7c6f434c, line 54: Is this the only place corresponding to «tests have to pass»?

(BTW what about a test removal procedure? Explicit maintainer approval?)

jonringer: In the future work, there's mentions to add it to the PR template, and it's already part of ofborg.

7c6f434c: I think all of this is vague enough that it remains unclear whether the RFC establishes a norm that tests should be run completely, and have to pass (hmmm, what about platform discrepancies…)

jonringer: This is more meant to set the expectation. There's already somewhat of a convention, and I would like for that to be more explicit and expected.

`nixpkgs-review` goes a long way to finding regressions, but the problem is totality. For some changes, there may only be a few packages which may realistically be affected by the change, so I don't want to build 500+ packages to review the build implications of a change, I just the relevant package and the few direct downstream dependencies. Even better if there's a testcase for the package itself.

7c6f434c: So basically you do not want to implement something with the full extent of the point from the meeting notes?

> Tests of the updated package should be become mandatory for acceptance of a PR 

(Here my question can be treated as procedural: for the text «updated according to the meeting notes» I am fine both with the text that clearly claims what the point in the meeting notes claims, or with an explicit comment that the meeting note point as written is too far-reaching; I just want clarity what interpretation of RFC gets covered by the eventual decision)

jonringer: > So basically you do not want to implement something with the full extent of the point from the meeting notes?

ofborg already runs `passthru.tests` if the commit message is formatted correctly. I think I'm mis-understanding what you're trying to say. My opinion is that the existing process already enforces the opt-in testing behavior. The main issue right now is that the usage of `passthru.tests` is the exception, not the norm.

> BTW what about a test removal procedure?

This is a non-goal for the RFC. Personally, I believe this should be left up to the maintainer(s) to decide. The additional tests should be providing value to the review process, if they don't then they should probably be removed, but this can be decided by the maintainers.

> Explicit maintainer approval?

Also a non-goal. Just concerned with the ability to test a PR, which may aide in the decision making process to merge a PR.

7c6f434c: Well, do we ask that tests are _run_ or that tests pass fully at least on one platform (maybe not the first try etc.…)?

jonringer: > Well, do we ask that tests are _run_ or that tests pass fully at least on one platform (maybe not the first try etc.…)?

I think a failing test should be ran, and if it fails, it should be an indicator that the package is unhealthy in context of the PR

7c6f434c: I agree with the position, and I think it could be made less vague in the RFC. Maybe put that statement that «tests should be run and a failing test is an indicator that the package is unhealthy in the context of the PR.» as a top-level statement in the detailed design?

rfcs/0119-testing-conventions.md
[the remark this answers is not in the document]
davidak: **nixpkgs-review** builds all reverse dependencies. could we just use this mechanism in a `testers.testBuildReverseDeps`? but there can be thousands :grimacing: 

cc @Mic92

rfcs/0119-testing-conventions.md
davidak: I would like to have a `CONTRIBUTING.md` in a packages directory with documentation for package maintainers and contributors, like how to test, how to update...
In addition to that, a `README.md` with documentation for users.

kevincox (approved):

Mic92:
We are now in the Final Comment Period (FCP). Unless any major objections are raised, this will be accepted and merged in 10 days from now, on June 6th!

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+of nixpkgs; and unless there's a dedicated hydra jobset for the pull request, it's
+infeasible to expect pull request authors to verify every package affected
+by a change they are proposing. However, it is feasible to specify packages that
+are very likely to be affected by changes in another package, and use this information
+to help mitigate regressions from appearing in release channels.
+
+# Detailed design
+[design]: #detailed-design
+
+Standardize `passthru.tests.<name>` as a mechanism of
```
roberth, line 42: The `<name>` syntax suggests that the namespace is flat, but it isn't. Both ofborg and `nix-build` respect the `recurseForDerivations` attribute (aka `recurseIntoAttrs`), which is a valuable behavior
 - for authors, adding an attribute is easier than merging attribute sets
 - for "testers", the computation of the set of attribute names is lazier, improving evaluation performance when they only run a specific test or set of tests

`passthru` is an implementation detail of `mkDerivation` that must not be part of this specification. You could mention it _once_ as a note to guide package authors, and you can use it in examples with `mkDerivation`, but it must not be part of the package specification. (See also for an attempt to define "package" https://github.com/NixOS/nix/issues/6507)

```suggestion
Standardize `<pkg>.tests` as a mechanism of 
```

roberth (changes requested):
It's important to omit `passthru` as described in the comment and my earlier comment https://github.com/NixOS/rfcs/pull/119/files#r857168303

Short process question, wouldn't it make sense for the rfc author to address all comments _before_ entering the final comment period? Feel free to ignore this if the author just forgot to push.

edolstra:
@Mic92 Since FCP has passed, do you want to go ahead and merge this, or does it need further discussion?

kevincox:
I see one unresolved comment. Maybe we should wait for a response or to integrate that into the RFC?

roberth:
My objections against `pkg.passthru.tests` in favor of `pkg.tests` have not been discussed or implemented yet.

rfcs/0119-testing-conventions.md
jonringer: won't `$tests` be populated in each build environment if this is true? the inclusion of `passthru` was because those get pruned, and is the current convention. (not to say that current conventions aren't flawed, and should be changed)

jonringer:
> My objections against `pkg.passthru.tests` in favor of `pkg.tests` have not been discussed or implemented yet.

I'm not very knowledgeable about the implications of #92, so I'll defer to your expertise. I made a mention that it was previously `passthru.tests` as the convention.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+of nixpkgs; and unless there's a dedicated hydra jobset for the pull request, it's
+infeasible to expect pull request authors to verify every package affected
+by a change they are proposing. However, it is feasible to specify packages that
+are very likely to be affected by changes in another package, and use this information
+to help mitigate regressions from appearing in release channels.
+
+# Detailed design
+[design]: #detailed-design
+
+Standardize `tests.<name>` (previously known as `passthru.tests.<name>`) as a mechanism of
```
roberth, line 42:
```suggestion
Standardize `tests.<name>` (commonly set with the `passthru.tests.<name>` argument to `mkDerivation`) as a mechanism of 
```

`passthru.tests` will stick around in `mkDerivation`, so "previously known" wasn't quite appropriate.

Other than that, all good.

rfcs/0119-testing-conventions.md
```diff
@@ -0,0 +1,105 @@
...
+to help mitigate regressions from appearing in release channels.
+
+# Detailed design
+[design]: #detailed-design
+
+Standardize `tests.<name>` (previously known as `passthru.tests.<name>`) as a mechanism of 
+more expensive but automatic testing for nixpkgs. As well as encourage the usage of
+`checkPhase` or `installCheckPhase` when packaging within nixpkgs.
+
+Criteria for `tests.<name>`:
```
roberth, line 46: `tests.<name>` suggests the type `attrsOf package`, whereas in practice we make use of `nix-build`'s attrset traversal capability to build NixOS tests. They're approximately attrset trees with derivations at the leaves; elaborated in the GitHub suggestion below.

Nested attribute sets of packages have already been implied earlier in the discussion, but haven't made it explicitly into the document yet.
Noteworthy advantages of nested attribute sets
 - easier to write than merges (`//`)
 - not susceptible to name collisions
 - allows the selection of groups of derivations by path
 - lazier. `//` forces the computation of all the attribute names, which may be slow when an expression library is involved

If you don't want to go into the technical details, I guess you could roughly summarize this as "tools should align with `nix-build`'s behavior", but it seems like some elaboration of that would be expected for "Formalize testing for nixpkgs packages".
The following is not a complete description of `nix-build` and allows some divergence (in the mathematical sense as well), as I believe defaulting `recurseForDerivations` to `false` is counter-intuitive. Furthermore, `nix-build`'s behavior can be exploited to limit the default set of tests that a user can be expected run, whereas an asynchronous tool can run all tests.

```suggestion
Criteria for `tests.<name>`:
- Each `test.<name>` should be a derivation, or an attribute tree with derivations at the leaves.
   - `nix-build -A pkg.tests` only recurses into attrsets below `tests.<name>` that also have the attribute `recurseForDerivations = true`, as set by `recurseIntoAttrs`.
   - Other tools may recurse into attrsets that do not define `recurseForDerivations`, but must adhere to `recurseForDerivations = false;`.
   - Tools may allow the user to specify an attribute, in which case the traversal starts at that attribute, regardless of any `recurseForDerivations` in the specified attribute or its ancestry.
   - Function-valued attributes should be ignored.
```

jonringer: > Nested attribute sets of packages have already been implied earlier in the discussion, but haven't made it explicitly into the document yet.
Noteworthy advantages of nested attribute sets

    easier to write than merges (//)
    not susceptible to name collisions
    allows the selection of groups of derivations by path
    lazier. // forces the computation of all the attribute names, which may be slow when an expression library is involved

I don't think it's going to be super common for the need to override tests in one package from another, which would require (//). The goal of this is to have a compromise between building everything, and building what's most likely to fail.

For example, if there's an update to openssl, then I would expect to run the downstream packages which have the most rebuilds (e.g. 1000+ rebuilds) and maybe some NixOS tests. But this will all be defined within openssl's expression, I can't think of use case where I would want to (//) more attrs into the tests attrset, I would just add them directly to openssl's `passthru.tests`.

From a PR reviewer's standpoint, if I see that someone has a PR updating openssl, I just want to be able to do `nix build -f . openssl.tests` and have enough faith that it will cover the most likely regressions. If there was a regression found later, then we can add that package to tests, and improve it over time. 

My other issue with nested tests, is that these tests are meant to act more like a CI/CD. If anyone one of them fails, then that failure should be addressed. It's not going to be a "well the simple tests pass, but this one one failed, I guess it's fine". A test shouldn't exist, if it's failure state isn't going to indicate that the package is unhealthy.

roberth: > I don't think it's going to be super common for the need to override tests in one package from another, which would require (//).

This is kind of what we're doing with `nixosTests` though. We're taking tests from one place and using it them in a package.

> [> allows the selection of groups of derivations by path]
>
> "if we get to the point where we feel it's necessary to start organizing the tests with nested attrsets"

This was a distraction. Neither of us care about organizing the tests into clever hierarchies.

I just want to throw whatever expression in there and expect all of it to run in `nix-build`, ofborg, etc.

By throwing out nested attrsets we need to start caring about the hierarchy, because we will have to flatten it.
If I change a test in `nixosTests` from a single test to a test matrix, that should have no ill effect on the packages that reference it. If we disallow nesting, it will.

> My other issue with nested tests,

Well, this is about test selection again, viewing nested attrsets as a means only to that end, which it is not.

It's really about making NixOS test linking easy and robust.

---------

I guess I could change https://github.com/NixOS/nixpkgs/pull/176557 to _always_ return `attrsOf derivation`; also when no `matrix` is defined: `{ test = drv; }`, and flattening the attrsets when it has multiple axes. That would make the PR a breaking change rather than a refactor with added functionality, so I'll scope that out for the PR itself. I also don't want to be associated with such a needless breaking change.

Honestly though, I think it's a bad choice, as we're painting ourselves into a corner with this unnecessary restriction. Our tooling has no problem dealing with nested attrsets...

>> lazier. // forces the computation of all the attribute names, which may be slow when an expression library is involved

I've seen a developer question his life choices when iterating on expressions that had a false dependency on an expensive attrset's keys. The expression they iterated on had no dependency on those keys, but they had to be computed because they `//`-ed it to their `packages` attrset. This added 30 seconds to their cycle time, making unnecessary strictness take up the majority of their cycle time. They'd been passionate about Nix and they had plenty of experience, but weren't _intimately familiar_ with the evaluator to figure out that `//` was ruining their workflow.

Please reconsider flattening.

jonringer: If you were just concerned about maintaining a package, and you were just concerned about the long-term maintenance of that package? what would you want the "api" to be for the test suite?

I'm not advocating that nested attrsets should be disallowed, I'm saying they are a "code smell". Since nested attrsets already exist in nixosTests, then  I think that's fine for that to carry over. And having a large test suite in nixosTests should probably not be frowned upon if the tests provide value. 

I am saying that (//) and large test suite should be avoided, because the desire of this rfc to make PR acceptance more automated, and large testing requirements will get in the way of that.

kevincox:
FCP is complete. The RFC has been accepted!

vcunat:
Now the "rendered" link in the first post won't work anymore.  I suggest to edit it, for UX of later visitors.

jonringer:
Yes, thank you :)

PR #110, opened 2021-10-18
[RFC 0110] Add "inherit-as-list" syntax construct to the Nix language

r-burns:
[Rendered](https://github.com/r-burns/nixos-rfcs/blob/rfc-inherit-as-list/rfcs/0110-inherit-as-list.md)

PR #120, opened 2022-01-12
[RFC 0120] Discourage nested `with` expressions

fogti:
[Rendered](https://github.com/zseri/rfcs/blob/no-nested-with/rfcs/0120-no-nested-with.md)

This has come up while trying to implement alternative interpreters/evaluators for the Nix language, and similar tools which need to perform static analysis on Nix expressions.

This idea is a compromise. The original idea was to forbid `with` expressions completely, which proved to be undesirable and impratical. A related idea was the introduction of a `with-only` expression, but it (1) wouldn't completely solve the problem and (2) introduce another keyword with no apparent immediate benefit.

This issue was discussed quite a few times on various channels, and this is an attempt to move closer to a solution.

cc @sternenseemann

### DIFF
--- a/rfcs/0120-no-nested-with.md
+++ b/rfcs/0120-no-nested-with.md
@@ -1,0 +1,98 @@
+---
+feature: no-nested-with
+start-date: 2021-01-11
+author: Ellen Emilia Anna Zscheile (@fogti)
+co-authors: @sternenseemann
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Disallow or discourage usage of multiple `with` expressions covering
+the same expression / forbid nesting `with` expressions, even indirectly.
+If infeasible in general (e.g. forbidding it in all nix expressions),
+this can be limited to nixpkgs.
+
+# Motivation
+[motivation]: #motivation
+
+It makes static analysis of nixpkgs easier, because as soon as `with`
+expressions are nested, it becomes basically impossible to [statically
+deduce where any free variable comes from] without implementing a
+full-blown nix evaluator including lazy evaluation, which is difficult as
+soon as `with` expressions and mutually recursive imports are involved
+(e.g. as currently present in `nixpkgs/lib/systems/{inspect,parse}.nix`).
+
+# Detailed design
+[design]: #detailed-design
+
+At least warn about any nested usage of `with` expressions,
+at least when they get evaluated, possibly even when they get parsed.
+After a grace period ("real time", probably a few months or years, not as in
+"a timeout when interpreting" because that wouldn't make sense),
+abort instead.
+
+When the full approach (aborting instead of warning) is applied,
+any reference to any free variable can be easily resolved
+to the enclosing `with` expression "scope-include", and because
+this `with` expression couldn't then be enclosed by another
+one, even indirectly, no lookup ambiguity exists.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+```nix
+pkgs: {
+  # allowed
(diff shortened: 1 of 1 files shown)


L-as:
Related: #110

gytis-ivaskevicius:
For the record let me explain what @L-as means by noting `inherit-as-list` RFC. Assuming this RFC gets implemented - there is no need for `with-only` expressions due to syntaxes as such:
```nix
# Basic usecase:
meta = {
  inherit homepage; # inherit in attribute set
  license = with licenses; [ inherit (lib.licenses) mit  ]; # inherit in list
}

# Not so obvious syntax
let inherit (lib) mapAttrs filterAttrs; in {
  a = mapAttrs (a: b: xyz) abc;
  b = filterAttrs (it: xyz) abc;
}

```

rfcs/0120-no-nested-with.md
```diff
@@ -0,0 +1,90 @@
...
+e.g.
+* Mutually recursive imports combined with `with` expressions also make static
+  analysis harder, because they require lazyness at the level of scope lookups,
+  which is difficult to implement corrently.
+* Check if any use case is too much negatively impacted by this.
+
+# Future work
+[future]: #future-work
+
+Unknown.
```
gytis-ivaskevicius, line 90: Nickel. :smirk:

zseri: Is Nickel really related to this proposal, in a stronger relation than basically any static analyzer for Nix?

gytis-ivaskevicius: I meant replacing nix with [Nickel](https://github.com/tweag/nickel). Meant it mainly as a joke due to me seeing quite a few Nix language related changes/proposals in the past little while

rfcs/0120-no-nested-with.md
```diff
@@ -0,0 +1,94 @@
...
+With this approach, any reference to any free variable can be easily
+resolved to the enclosing `with` expression "scope-include", and
+because this `with` expression couldn't then be enclosed by another
+one, even indirectly, no lookup ambiguity exists.
+
+# Detailed design
+[design]: #detailed-design
+
+At least warn about any nested usage of `with` expressions, at least when they get evaluated,
+possibly even when they get parsed. After a grace period, abort instead.
```
zseri, line 38: This currently partially contradicts the summary and is shorter than the summary. Needs to be fixed.

lheckemann:
This RFC is now again open for nominations :)

kevincox:
This RFC has not gotten any nominations which appears to show lack of interest from the community. In order to drive this RFC forward I recommend trying to help find shepherds by reaching out to people you know and posting in discourse or Matrix.

If no shepherds can be found we will convert this RFC to draft to be on hold until we can find enough interested participants.

tomberek:
To re-iterate the previous comment, this RFC will require some shepherds in the next two weeks otherwise will be turned into a draft awaiting further interest.

zseri:
A simpler alternative to this would be to "just" forbid mutual imports combined with `with`-expressions, because these prevent the import structure from forming a DAG, *and* have no unambiguous translation to explicit fixed points (`lib` in `nixpkgs` is an unambiguous and explicit fix point); which is the case which triggered the problem in the first place, because "no unambiguous translation" means that interpreting nix files separately or on a one-by-one basis is hard, because the scope is temporarily ambiguous...

AndersonTorres:
Related:
https://github.com/NixOS/nixpkgs/pull/294383
https://github.com/NixOS/nixpkgs/issues/292468#issuecomment-2016454242
https://github.com/NixOS/nixpkgs/pull/298316/files

fogti:
As I am no longer part of the Nix community and this PR partially stemmed from a misunderstanding of the behavior of `with` expressions in case that stuff "imported" that way would "shadow" other stuff: I'm going to abandon this (but others can feel free to continue on this on their own, if they want, and reuse this (in part or such, etc.))

PR #122, opened 2022-03-07
[RFC 0122] IPFS CID optionally on narinfo in binary caches

lucasew:
The idea is to provide the CID of the nar file from the binary cache optionally to allow reducing bandwidth costs and in some cases increase efficiency by allowing users to download the binary cache nar files over IPFS

[Rendered](https://github.com/lucasew/rfcs/blob/binary-cache-ipfs/rfcs/0122-binary-cache-ipfs.md)

### DIFF
--- a/rfcs/0122-binary-cache-ipfs.md
+++ b/rfcs/0122-binary-cache-ipfs.md
@@ -3,8 +3,8 @@ feature: binary-cache-ipfs
 start-date: 2022-03-07
 author: lucasew
 co-authors: (find a buddy later to help out with the RFC)
-shepherd-team: (names, to be nominated and accepted by RFC steering committee)
-shepherd-leader: (name to be appointed by RFC steering committee)
+shepherd-team: John Ericson, Tom Bereknyei, Kevin Amado
+shepherd-leader: Tom Bereknyei
 related-issues: (will contain links to implementation PRs)
 ---
 
--- a/rfcs/0122-binary-cache-ipfs.md
+++ b/rfcs/0122-binary-cache-ipfs.md
@@ -1,0 +1,84 @@
+---
+feature: binary-cache-ipfs
+start-date: 2022-03-07
+author: lucasew
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+In binary caches add an extra property on narinfo to reference the IPFS CID of the nar file
+
+# Motivation
+[motivation]: #motivation
+
+IPFS is still not a present reality on the mainstream Nix ecosystem, altough it's not reliable to store long term data, it can reduce bandwith costs for both the servers and the clients but the question is where the NAR file could be obtained in IPFS.
+
+Its not espected that, for example, cache.nixos.org would run a IPFS daemon for seeding but it could just calculate the hash using `ipfs add -nq $file` and provide it on the narinfo so other nodes can figure out alternative places to download the NAR files, even closer than a CDN could be.
+
+Parallel binary caches could arise for regions that internet connectivity is a problem and a local distribution is preferred. If the payload is properly signed it shouldnt be a problem to prove that given path comes originally from given binary cache.
+
+# Detailed design
+[design]: #detailed-design
+
+A narinfo file is a file provided by the binary cache server that provides metadata for an existent path in the binary cache. It has information about the nix store path, which compression algorithm is used, hashes, sizes, references, a signature and a relative direct path to download the compressed NAR file.
+
+It has the sha256 hash of the file but from that it's still not possible to find out where to download it on the IPFS network so, to make it possible, the CID is required.
+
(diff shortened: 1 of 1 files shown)


Ericson2314:
We should be able to use the existing CA field for this. That has many other benefits, too. That is what we did in our IPFS Nix work.

rfcs/0122-binary-cache-ipfs.md
```diff
@@ -0,0 +1,84 @@
...
+[summary]: #summary
+
+In binary caches add an extra property on narinfo to reference the IPFS CID of the nar file
+
+# Motivation
+[motivation]: #motivation
+
+IPFS is still not a present reality on the mainstream Nix ecosystem, altough it's not reliable to store long term data, it can reduce bandwith costs for both the servers and the clients but the question is where the NAR file could be obtained in IPFS.
+
+Its not espected that, for example, cache.nixos.org would run a IPFS daemon for seeding but it could just calculate the hash using `ipfs add -nq $file` and provide it on the narinfo so other nodes can figure out alternative places to download the NAR files, even closer than a CDN could be.
```
kevincox, line 21: One little concern is that a given file doesn't have exactly one CID. Depending on how you chunk the file you can get effectively unlimited different CIDs. This isn't a problem when the CID distributor starts the seed and the CID stays live on the network because whatever CID is advertised will be fetched. However for the case like this is matters a lot, because different settings will result in a would-be seeder generating the wrong CID.

IIUC the current default for `ipfs add` is fixed-size blocks of 262144B each (aka `size-262144`). However for a nixpkgs cache where subsequent versions of a derivation may be largely similar it may make more sense to do a smarter chunker based on a rolling hash.

Anyways, the exact chunking mechanism is bikeshedding, but what do we want to do about this? I see a few main options.

1. Put the chunker into the narinfo so it can be reproduced. (I don't know if there is a well defined standard format but current go-ipfs uses strings like `size-262144` and `rabin-2048-65536-131072` which are pretty easy to understand and unlikely to be ambiguous.)
2. Declare a chunker upfront and expect people to use it. (We can revert to 1 in the future by adding the chunker information later).
3. Convince cache.nixos.org to also run an IPFS node that advertises the CIDs that are advertised in the narinfo files.

nixinator: rsync has a pretty interesting algorithm for syncing files https://stackoverflow.com/questions/1535017/rolling-checksums-in-the-rsync-algorithm  , there maybe something in that, However probably not directly portable to IPFS and chunking.

I'd vote for 3! and get that working today (or perhaps tomorrow) and think about options 1/2 for the day after tomorrow (or some point in the future).

Thanks for your detailed analysis of this, my understanding of Nars on IPFS has increased!

kevincox: This is basically equivalent to the Rabin chunking. But the biggest problem isn't what algorithm to use but know to know what algorithm to use.

lucasew: For this we could do like how we already do with hashes, like sha256:something

AFAIK ipfs has symbol friendly names for the chunking methods

tomberek: Other possibilities of chunking with casync: https://discourse.nixos.org/t/nix-casync-a-more-efficient-way-to-store-and-substitute-nix-store-paths/16539

kevincox: I really don't care about the chunking algorithm. Please stop discussing this here.

What I care about is that we record the chunking algorithm in a way that someone who wishes to advertise this path can do so.

edolstra:

Ericson2314:
I supposed I could shepherd this, but really I want and should soon be able to write a counter-proposal RFC for the work we did in 2020. So perhaps there ought to be one shepherd team for two "competing" RFCs (though it's really more prioritizing features than actually disagreement).

Ericson2314:
You might take a look at https://github.com/NixOS/nix/pull/3727, which I locally fixed conflicts with. (Tests however, are broken, still debugging.)

That goes a few steps in trying to put the narinfos in IPFS as IPLD rather than files too, but this should be complementary:

 - We should make a new JSON narinfo format for "regular" binary caches too, as the current line-oriented file makes backwards compatible evolution too hard. (We can simply upload both types of narinfo for compatibility with old versions of Nix if we like.)

If we do that we can also share lots of code between both approaches:

 - All the "how do I talk to IPFS" legwork can of course be shared.
 - The JSON serialization can be shared between a "native" IPLD narinfo and legacy file version
 - Code to deal with getting the file data to/from IPFS can be shared.

kamadorueda:
I nominate myself as a shepherd

Ericson2314:
Looks like we have the required number? :)

Ericson2314:
#nix-rfc-122:matrix.org

edolstra:
Any updates on the status of this RFC?

lucasew:
We (or I) need to build a proof of concept. Maybe we will pivot this RFC to an LRU-based cache proxy approach at the beginning and iterate to a p2p approach if necessary, but I am without time to test it now, I am very busy because of the end of the semester.

The plan is to apply that prototype to an organization to reduce internet usage with things people often need, so, that prototype should be working until the end of the year, or I definitely will not get my degree by the end of the year xD.

lheckemann:
Sounds good! On behalf of the Steering Committee, I'd like to suggest moving the RFC to draft status until then --- any objections?

PR #123, opened 2022-03-12
[RFC 0123] Flake names

schuelermine:
# [RFC 0123] Flake names

[Rendered](https://github.com/schuelermine/nix-rfcs/blob/master/rfcs/0123-flake-names.md)


rfcs/0123-flake-names.md
```diff
@@ -0,0 +1,54 @@
...
+[design]: #detailed-design
+
+A new supported property for flakes is introduced, `name`.  
+Running `nix flake metadata` on a flake that declares this field displays it at the top.  
+The derivation that contains the flake’s content is called `flake-source-${name}` or, if a short revision identifier is available, `flake-source-${name}-${shortRev}`.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+A flake that is in the registry should have its name match the registry identifier.
```
bew, line 34: Wouldn't that prevent the use of alternate names, or having multiple versions of the same flake in the registry?
(I use both)

schuelermine: That’s a good point

schuelermine: Maybe it’d be a good rule for the global registry, though, but it’d probably be followed anyways, and would only hinder edge-cases

edolstra:
Note that flakes had a name field but it was removed (80c36d4562af71a90c67b3adb886a1003834890e) because there was no clear use case / semantics. It also raises questions like whether names should be globally unique, and what should happen if they're not.

The RFC mentions discoverability but doesn't really explain how a `name` attribute helps here. Concretely, what does the `name` attribute provide to https://search.nixos.org/flakes that the repo URL doesn't already?

See also https://github.com/NixOS/nix/issues/5879 - we thought about bringing back `name` to be used for default flake output attributes (i.e. `packages.<system>.<name>` instead of `packages.<system>.default`).

>  The derivation that contains the flake’s content is called `flake-source-${name}` or, if a source control tag (e.g. git tag such as V3) is available, `flake-source-${name}-${tag}`.

This has the problem described [here](https://github.com/NixOS/nixpkgs/pull/153386#issuecomment-1004715841), namely that it reduces content-addressability by duplicating identical source trees when fetched from different flakes/tags.

schuelermine:
1) In this case, the clear use case is usage in the derivation name for better UX. In my opinion the name should be entirely decorative or only used for default behaviors.
2) I find non-unique names are a non-issue, they should be disregarded, and are unlikely to be common. Currently, all derivation names are non-unique. Granted, it could confuse people into believing two derivations are of the same flake, but this is still an improvement over not knowing the source at all.
2) I thought the name could be used in the future for a more user-friendly UI. Discoverability of flakes isn’t necessarily helped as the global registry provides it, but it could make flake command usage easier and therefore more discoverable. However, I agree in hindsight that this is a mistaken inclusion, I will strike it from the RFC.
3) I see, the tag is possibly a bad idea. But why is changing the name bad for content addressability? It’s all about coercing derivations with different source expressions but the same result, right?

JJJollyjim:
Are flakes not still an experimental feature that we're waiting for an RFC to formalise? It seems odd for this to come first.

schuelermine:
I don’t know how else it would be possible to suggest this feature.

edolstra:

lheckemann:
@schuelermine maybe you have some people in mind who would be suitable as shepherds? Feel free to nominate them, if so :)

schuelermine:
@lheckemann I think I’m not familiar enough with the Nix development team & community to confidently nominate anyone in particular. The only people I know are ones I’ve more “seen” than “met”, and I don’t know of their specialties. How would I go about nominating someone?

rehno-lindeque:
Minor use case: https://github.com/rehno-lindeque/flake-help

I [use `flake.description`](https://github.com/rehno-lindeque/flake-help/blob/dd85539a7c1a0530f9084869958c64cf914116c2/flake.nix#L28-L32) to print a help message. With this change I could also print `flake.name`.

jeff-hykin:
Following the nix-discourse (making flakes indexable/discoverable), I'd like to propose simple change to make this forward-compatible and modular. 

`inputs` and `outputs` are wonderfully elegant/generic names. Tacking on `name`, `description`, etc one after another seems like it could easily devolve into a name/pname/version issue, as a bunch of hardcoded edgecases that can never be removed without breaking things.

So, I'd like to propose a `static` field (name can be anything, I'm just going to call it "static" here)
This would fit in nicely with inputs/outputs and be easy to explain "flakes have only three top-level keys"

```nix
{
    static = {
        name = "anything";
        description = "blah blah blah";
        # arbitrary additional data
        license = "GPLv3";
        keywords = [ "image" "photo" "app" "image-editing" "image-viewing"  "python" ];
        code_source = "https://github.com/some_user/some_repo";
        cve_reports = "";
        bug_reports = "";
        homepage = "";
        icon = "https://";
        latest_sha256 = "";
        bin_names = [ "pip" "python" "python3" ];
    };
    inputs = {
        nixpkgs = { url = "github:nixos/nixpkgs/nixos-unstable"; };
    };
    outputs = inputs : { };
}
```

The important aspects being;
1. `static` contains non-functional data (read "indexable/searchable data" )
2. The structure can be expanded as-needed in a forward-compatible way
3. Unlike inputs or outputs, `static` can be automated (ex: generated from json) for version bumps and hash changes, which helps solve the [unmaintainability problem](https://discourse.nixos.org/t/nixpkgss-current-development-workflow-is-not-sustainable/18741)
4. Rather than must-have requirements, the structure can be flexibly checked against standards. For example:
  - nix-search could have its own "baseV1" standard that requires name, description, and license. <br> e.g. "Nix Search supports all flakes that comply with the baseV1 standard"
  - [Lazamar's version search](https://lazamar.co.uk/nix-versions/) could require having a version string
  - NixOS could have an "ApplicationV1" standard that requires additional static fields to be considered an "app" flake
  - New standards can build-upon or fork from this without breaking core nix functionality, since there's no guarantee a flake will even have a `static` field.
  - Npm packages could have a requirement for `is-npm-package`, could require `npm-version` that follows semver pattern, and could have more strict requirements on `name`
  - Even singular packages could have their own standard such as marking whether something was a LTS release or not
5. This allows systems outside of nix to leverage/understand packages without a full nix evaluation engine

thufschmitt:
> So, I'd like to propose a static field (name can be anything, I'm just going to call it "static" here)

Just bike-shedding on the name (because it’s too tempting), this kinda matches the `meta` attributes for packages, and I think it could be a good name here too.

(And regardless of the color of the bike, I agree that keeping this things scoped is a good idea, if only to keep the schema tidy)

schuelermine:
I think `static` would conflict with the existing `description` field. Of course, it could be integrated, but that’d seriously expand the scope of this RFC. Maybe make your own RFC about a `meta` or `static` flake attribute?

milahu:
> * `static` contains non-functional data (read "indexable/searchable data" )

thats sounds overly limiting. scrapers must parse the `.nix` file anyway, and evaluation should be cheap enough

what if my `flake.nix` file has a million outputs, and each output has a different name/description/...?

schuelermine:
>
> what if my `flake.nix` file has a million outputs, and each output has a different name/description/...?
> 

You wouldn’t need to evaluate `outputs`, though, right? Or is the no-thunks rule also active inside `outputs`?

jeff-hykin:
> > * `static` contains non-functional data (read "indexable/searchable data" )
> 
> thats sounds overly limiting. 
> 

Why? If you need functional behavior, use outputs. Otherwise use static. I don't see how this limits anything.

On the flip side static is absolutely needed to index things efficiently. Starting in 2020 I spent months trying to index nixpkgs and make them searchable. But the absolute bane of that endeavor was dynamically generated package names, versions, and other metadata. Some packages changed versions depending on what host-OS was performing the query, some were using fetch just for their version number. That makes it impossible to index what's available because we can't even list of all of what's available in an automated way. I mean there's a reason other package managers require things like a package.json in a serialized format. At minimum it's a security risk to be running random package code just to find a package name.

Things like name cannot both be scrape-able and generated on-demand using stdenv, fetch, file read, etc. 


>  scrapers must parse the `.nix` file anyway, and evaluation should be cheap enough

In theory sure, but there's a reason nix-env -i is so slow that I can ripgrep the result faster. 

I can parse a nix file performantly basically anywhere, even client side in a browser thanks to the tree sitter WASM module. I don't need a file system, or a nix binary to parse a nix file. That's a pretty low threshold. Converting the entire nix binary to WASM, and emulating a file system and network connection, sadly isn't "cheap enough" in comparison. It's just so much easier to build a parser than it is an evaluator.


> what if my `flake.nix` file has a million outputs, and each output has a different name/description/...?
> 

Perfect! The old system (just a hard coded top-level description field) wouldn't be able to handle that case. And while this static system technically could handle that case by having a  sub-packages field and putting them all in there, I would question if flakes were ever intended to represent a grouping of a million packages rather than one package per flake.

jeff-hykin:
>  Maybe make your own RFC about a `meta` or `static` flake attribute?

That's the plan 👍 I asked for related discussions on the discourse and this was the only one that popped up so far.

> I think `static` would conflict with the existing `description` field. Of course, it could be integrated, but that’d seriously expand the scope of this RFC

I was hoping a set-attribute of static would be comparable in effort to string-attribute of name, but maybe you're right since there's more to consider.

milahu:
> But the absolute bane of that endeavor was dynamically generated package names, versions, and other metadata.

we could cache these strings in flake.lock, which already has the "no eval" format (json) that you want
otherwise, just exclude packages from your search-index that are too expensive to eval? (and notify maintainers)

when we introduce a new standard, then we get 3 types of packages:
1\. the "good" packages that follow the new standard
2\. the "neutral" packages that use the old standard, but metadata is fast to eval
3\. the "bad" packages that use the old standard, but metadata is slow to eval

jeff-hykin:
@milahu I think you're right that would be a lot better. That would allow arbitrarily complex functions to generate the metadata, and then the output would be easy to index 👍

kevincox:
This RFC needs more shepherds! Please nominate those who you think could help pull this forward.

schuelermine:
I nominate @alyssais and @edolstra for shepherd for this NFC

rfcs/0123-flake-names.md
```diff
@@ -0,0 +1,89 @@
...
+shepherd-leader: None
+related-issues: None
+---
+
+# Summary
+[summary]: #summary
+
+Flakes can declare the field `name`.  
+It represents the name of the flake.  
+The derivations for a flake are no longer called `source`, but use the flake name.
```
thufschmitt, line 16:
```suggestion
The store paths corresponding to a flake source are no longer called `source`, but use the flake name.
```

(I’m nitpicking a bit here, but technically these aren’t derivations, they are just paths that are copied directly to the store)

rfcs/0123-flake-names.md
```diff
@@ -0,0 +1,89 @@
...
+$ nix flake metadata /example
+Name:          example
+Resolved URL:  git+file:///example
+Locked URL:    …
+$ nix flake show /example
+example (git+file:///home/anselmschueler/Code/example?rev=b0&rev=c714c8624f5d49a9d88e6e24550dd88515923c18)
+└───lib: unknown
+```
+
+# Drawbacks
```
thufschmitt, line 69: The big drawback here (where “big” would probably need to be refined, I’ve no idea how much of an issue it is in practice) is that since the name is part of the store path, two flakes with the same content but a different name will end-up in a different place in the store − which itself means that everything that depends on them would have to be rebuild.

rfcs/0123-flake-names.md
```diff
@@ -0,0 +1,89 @@
...
+
+Flake-centric workflows often end up with a lot of derivations named “source”, and it’s difficult to navigate this.
+This can also make flake-centric commands friendlier and easier to approach.
+
+# Detailed design
+[design]: #detailed-design
+
+A new supported property for flakes is introduced, `name`.  
+The derivation that contains the flake’s content is called `flake-source-${name}`  
+If a source control tag (e.g. git tag such as V3) is available, it is `flake-source-${name}-${tag}`.
```
thufschmitt, line 29: This isn’t really possible without breaking purity afaik: The same git commit can have or not a tag attached to it (since tags are mutable), and evaluating a flake from that commit with or without the tag would yield a different result

rfcs/0123-flake-names.md
```diff
@@ -0,0 +1,89 @@
...
+[summary]: #summary
+
+Flakes can declare the field `name`.  
+It represents the name of the flake.  
+The derivations for a flake are no longer called `source`, but use the flake name.
+
+# Motivation
+[motivation]: #motivation
+
+Flake-centric workflows often end up with a lot of derivations named “source”, and it’s difficult to navigate this.
```
thufschmitt, line 21: That’s probably worth a bit more explanations: for most intents and purposes, the store is supposed to be an opaque thing, and not meant to be directly “navigated”.

Now it is a very leaky abstraction, so my remark probably doesn’t really hold in practice, but it would be good to make this explicit (maybe show some concrete examples of where having all the sources be called “source” is an issue)

thufschmitt (commented):
My 2cts on that:

- I’m generally in favor of having a `name` attribute in the the flakes schema (or `meta.name`), regardless of whether Nix uses it directly or not.
- I’m not really fond of having it leak into the store, unless there’s a good case made to show that it won’t cause too many rebuilds (which I’ve no idea how to assess tbh)

Note that part of the store-naming problem should be solved by https://github.com/NixOS/nix/pull/6530 which will cause the flake evaluation to happen in-place (for the toplevel-flake) rather than in the Nix store

rfcs/0123-flake-names.md
schuelermine: But flakes with a different name will necessarily  have different content? Or are the flake.* files not copied to the store?

rfcs/0123-flake-names.md
schuelermine: It's an issue if you want to know what a nix command is doing at the moment. It often just displays “building XXXX-source”

rfcs/0123-flake-names.md
schuelermine: Good point!

rfcs/0123-flake-names.md
schuelermine: I thought I had removed this section? Someone already raised this issue.
Yes, that's a good point.

rfcs/0123-flake-names.md
thufschmitt: Fair point (I had a convoluted scenario in mind, but that’s probably way out of the scope of this RFC now that I think of it). 

That’s actually highlighting an interesting design challenge: There’s currently a neat separation between the fetching code and the code that knows about what the flake is. `inputs.foo = something` gets translated to a call to `builtins.fetchTree something`, and its result is then fed to the `outputs` function. And the `fetchTree` part doesn’t know (nor has to know) that it’s fetching a flake.
But if the name gets defined in the flake (and used in the output path), then the `fetchTree` logic must be a bit more convoluted:
- It has to know whether it is fetching a flake
- When fetching a flake, it must get its `flake.nix`, evaluate it to find its `name`, and only then put it in the store with the correct `name`.

rfcs/0123-flake-names.md
thufschmitt: I might be wrong, but that exact message shouldn’t happen (not because of flakes at least). This message happens when Nix builds a derivation, but flake sources aren’t derivations, they are directly instantiated during the evaluation (with `builtins.fetchTree`) using a different code path with amongst other things a more informative display.

What you might get though is something like “fetching source from https://cache.nixos.org” (because although they aren’t derivations, they can be substituted from the binary cache). Probably we could improve the message in that case.

schuelermine: Oh, that’s good to know, sorry for not knowing that.

schuelermine:
Are there any other contexts at the moment where the name of a store path depends on the *contents* of the store path instead of the expression that generates it?

rfcs/0123-flake-names.md
schuelermine: About the store path being meant to be opaque: In my opinion that’s just practically not true at all, and the fact that derivation names exist at all proves it. If anything, this is useful for people that want to spelunk in the depths of what Nix does.

I also regularly use store paths to figure out what package or other source provides some binary in my profile. I have a shell function `goto-bin` for going into the directory of the real path of an executable, and I sometimes skim `realpath` output to see a derivation name.

rfcs/0123-flake-names.md
schuelermine: Wouldn’t it be better to have this conversation outside the review widget? I feel like this aspect of the discussion is somewhat hidden & non-chronological.

schuelermine:
@thufschmitt I feel like this wouldn’t cause many expensive rebuilds, since the things defined *in* flakes don’t depend on the flake itself at all.

thufschmitt:
> Are there any other contexts at the moment where the name of a store path depends on the _contents_ of the store path instead of the expression that generates it? I’m not sure if flakes work that way right now

On top of my head, this can happen in four situations:
1. Paths explicitly added to the store with `nix-store --add`/`nix store add-path` or equivalent
2. Paths added in the store as part of the evaluation (`"${./foo}"`, the result of the `builtins.fetch*` family of functions, flake inputs)
3. Fixed-output derivations (all the `nixpkgs.fetch*` functions in particular)
4. The experimental content-addressed derivations (which aren’t really used in the wild yet)

> Can anybody confirm if the store path of a flake is determined before or after fetching it?

Technically it is determined after. The logic of the fetcher is roughly:
1. Fetch the input in a temporary location
2. Serialize that as a NAR archive (more-or-less a tar, but normalizing all the filesystem metadata)
3. Hash that archive, and use that hash to compute the final store path
4. Extract it in the final store path

> I feel like this wouldn’t cause many expensive rebuilds, since the things defined in flakes don’t depend on the flake itself at all.

You’re right, that probably wouldn’t change a lot

schuelermine:
I have rewritten the RFC slightly to incorporate some of the drawbacks mentioned and for better formatting

schuelermine:
After reading this

> The RFC mentions discoverability but doesn't really explain how a name attribute helps here. Concretely, what does the name attribute provide to https://search.nixos.org/flakes that the repo URL doesn't already?

again, I’ve come up with some things: It makes the purpose of the flake clearer, as a flake could be mirrored from a different URL that doesn’t suggest the flake’s name already. It could also make it faster to make out the relevant part of the URL in flakes that have it in there. In general, the URL can be seen as an entirely functional device used for fetching the flake, which only happens to usually contain the flake name or suggest its purpose.

schuelermine:
@lheckemann Are RFC shepherds supposed to be people who are interested in the RFC getting accepted, or just people familiar with the matter? My previous nomination was a bit uncreative and not fruitful, and sadly, I’m still not clear on how to choose good shepherd candidates…

alyssais:
I'm afraid I'll have to decline the nomination.  I don't use flakes, so I don't think I would be able to do a good job making sure this RFC is the best it can be.

Additionally, I don't think we should be doing RFCs for changes to flakes, until flakes themselves have gone through the RFC process and are no longer considered an experimental feature.  Until then I think changes to flakes should be discussed on the Nix issue tracker.  (I appreciate that the Nix issue tracker is a difficult place to get a response. :/)

Ma27:
IMHO the proper solution would be to actually do flakes via RFCs rather than having a few blessed people doing what they want :shrug:

schuelermine:
Just wanted to make a note here that I realize now that nominating edolstra must’ve seemed super arrogant—I was at a loss for how exactly this shepherds thing worked so I just copied shepherds from a (closed) Flake RFC, since it’s said that shepherds should be people familiar with the area touched by the RFC. When I wrote that comment I didn’t actually know what role they played in the Nix community. I’m sorry if that came across as arrogant.

tomberek:
A note from RFC Steering Committee: a flake RFC is a work-in-progress, but is not required for this RFC to have discussion or to even proceed if shepherds would like to. Considering the potential dependency on a flake RFC, it can be reasonable to not have a resolution until that is complete.

edolstra:
@schuelermine No problem, anybody is free to nominate anybody for an RFC since they can always say no. I'll be happy to help shepherd this RFC.

edolstra:
Anybody else interested in shepherding this? Otherwise I suggest turning this into a draft RFC in 2 weeks until there is more interest.

rfcs/0123-flake-names.md
roberth: > * When fetching a flake, it must get its `flake.nix`, evaluate it to find its `name`, and only then put it in the store with the correct `name`.

This can be solved by adding `name` to the lock file.
We can use the same trick for other fetch-related parameters, like whether to fetch submodules or not, whether to git-crypt decrypt, etc, all of which would also be nice to configure in `flake.nix`.
`fetchTree` remains unaware of flake-ness.

rfcs/0123-flake-names.md
```diff
@@ -0,0 +1,91 @@
...
+example (git+file:///home/anselmschueler/Code/example?rev=b0&rev=c714c8624f5d49a9d88e6e24550dd88515923c18)
+└───lib: unknown
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This may cause clutter and additional maintenance.
+- Since this changes the output of nix flake metadata and nix flake show, it might cause scripts that read this output to break.
+- This requires a significant change to the way flakes are handled.
```
roberth, line 73:
```suggestion
```

All changes come with an implementation cost. I don't think this RFC is special, because it seems like an addition that doesn't require action by the ecosystem, especially if the name is allowed to be unset.

rfcs/0123-flake-names.md
```diff
@@ -0,0 +1,91 @@
...
+- This treats the Nix store as a user-facing part of Nix, which is generally not intended.
```
roberth, line 74:
```suggestion
```

The store has always been user-facing. Users interact with it, store paths show up in logs, etc.

Sometimes you don't have to think much about the store and that's good.

How will users will have to pay more attention to the store after this change?

rfcs/0123-flake-names.md
```diff
@@ -0,0 +1,91 @@
...
+Flakes can declare the field `name`.  
+It represents the name of the flake.  
+The store paths corresponding to a flake source are no longer called `source`, but use the flake name.
+
+# Motivation
+[motivation]: #motivation
+
+- Flakes generate store paths named “source”, and it’s difficult to navigate this when manually inspecting the store.
+- This metadata can be used to make flakes more approachable and usable, in particular, it can be output instead or with the URI, making the nix commands friendlier.
+
```
roberth, line 23:
```suggestion
- Unique names are useful for the purpose of finding duplicate things caused by diamond dependencies. (E.g. `a -> b -> d1`, `a -> c -> d2`. Without a unique name (`d`), it is not possible to correlate `d1` and `d2` reliably.)

```

 - Most relevant when https://github.com/NixOS/nix/issues/5570 remains unimplemented. "Userspace" experimentation may yield some insights into solutions for 5570.

blaggacao:
I had and still have an interest in this and may volunteer to join the shepherds.

May the author reactivate the momentum?

I'd be also interested in exposing `self.name` since that seems to be missing in small projects where the flake name percolates all through to the binary name.

schuelermine:
I am still interested in this RFC. This RFC was put into the draft state because I could not produce RFC shepherds. If you’d like to be a shepherd, I can nominate you. I’m not sure how to “reactivate momentum” on this RFC.

kevincox:
All that it takes to reopen this RFC is to find shepherds. Since we have a nomination it may make sense to bring out of draft and try to find a few more.

schuelermine:
@blaggacao would you be interested in being a shepherd for this proposal?

blaggacao:
With https://discourse.nixos.org/t/why-is-there-a-default-overlay-in-flakes/24422/2, I would. :+1:

schuelermine:
I nominate @blaggacao to be a shepherd of this RFC!

PR #125, opened 2022-05-09
[RFC 0125]: Use an intermediary representation (bootspec) for bootloader backends

grahamc:
Bootspec is a set of memoized facts about a system's closure.
These facts are used as the primary input for bootloader backends like
systemd-boot and grub, for creating files in `/boot/loader/entries/`
and `grub.cfg`.

In this proposal we create a stable, comprehensive, and machine-parsable
definition of a NixOS Generation as an intermediate representation (IR)
between the NixOS system definition and the bootloader management tools.

[rendered](https://github.com/grahamc/rfcs/blob/bootspec/rfcs/0125-bootspec.md)
(thanks @ryantm)

### DIFF
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -3,8 +3,8 @@ feature: bootspec
 start-date: 2022-05-09
 author: Graham Christensen
 co-authors: Cole Helbling
-shepherd-team: (names, to be nominated and accepted by RFC steering committee)
-shepherd-leader: (name to be appointed by RFC steering committee)
+shepherd-team: @lheckemann, @06kellyjac, @GovanifY, @RaitoBezarius
+shepherd-leader: @lheckemann
 related-issues: https://github.com/NixOS/nixpkgs/pull/172237
 ---
 
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -1,0 +1,212 @@
+---
+feature: bootspec
+start-date: 2022-05-09
+author: Graham Christensen
+co-authors: Cole Helbling
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Bootspec is a set of memoized facts about a system's closure.
+These facts are used as the primary input for bootloader backends like systemd-boot and grub, for creating files in `/boot/loader/entries/` and `grub.cfg`.
+
+In this proposal, we create a stable, comprehensive, and machine-parsable definition of a NixOS Generation as an intermediate representation (IR) between the NixOS system definition and the bootloader management tools.
+
+# Motivation
+[motivation]: #motivation
+
+NixOS's bootloader backends don't support a uniform collection of features and design decisions, partially due to the complexity of implementing the features.
+Using a statically parsable bootspec definition reduces the work involved in implementing bootloader support.
+
+If we survey the current set of bootloader and feature matrix, we see a bit of a smattering:
+
+| Bootloader                  | Generation limits | Initrd Secrets | Multiple Profiles | Specialisation | Custom Labels |
+| --------------------------- | ----------------- | -------------- | ----------------- | -------------- | ------------- |
+| systemd-boot                | YES               | YES            | YES               | YES            | YES           |
+| grub                        | YES               | YES            | YES               | YES            | YES           |
+| generic-extlinux-compatible | YES               |                |                   |                | YES           |
+| raspberrypi                 | YES               |                |                   |                |               |
+| init-script                 |                   |                |                   | YES            |               |
+| generations-dir             |                   |                |                   |                |               |
+
+One reason the matrix is not filled out is the technical difficulty of implementing these various features.
+The current API for detecting most of this information is by globbing directories looking for specific files.
+
+By forcing implementations to spelunk the filesystem, we create a complicated problem in multiple ways:
+
+- Introducing or improving bootloader features is complicated and requires significant effort to implement it for the major bootloaders.
+- The relatively complicated filesystem-based behavior involves reimplementing similar logic across Perl, Python, and Bash.
+- Our current tools can only support limited workflows, and are difficult to extend.
+
+## Supporting Externalized Bootloader Backends
+[externalized-bootloaders]: #externalized-bootloaders
+
+Our NixOS-provided tooling is sufficient for most use cases; however, there are more complicated cases that are not adequately covered by our tools and would not be appropriate to include in NixOS directly.
+Let's use SecureBoot as an example.
+
+A naive, single-user implementation of SecureBoot may have signing keys on the local filesystem, and every `nixos-rebuild boot` call signs the relevant files.
+This is a valid method of implementation, but is insufficient for larger deployments.
+
+An enterprise deployment of SecureBoot may have a centralized signing service with careful auditing about what is signed.
+It is more likely that the signed bootloader files come prebuilt and presigned from upstream and an unsigned file means a violation of policy.
+This is also a valid implementation, but is too heavy for a single user.
+
+There are infinitely many policy choices and implementations possible, and a good solution here is to allow deployers of NixOS to implement their own bootloader management tools.
+By creating a well defined specification of generations and the boot-relevant data we enable this external development.
+
+## What about systemd's Bootloader Specification?
+[systemd-bootloader-specification]: #systemd-bootloader-specification
+
+Systemd's bootloader specification is a good format for a different problem.
+A single NixOS generation can contain multiple bootable systems and options, with additional features unique to NixOS built on top.
+Most Linux distributions don't deal with many unique and ever-changing bootables.
+This proposal is specifically to deal with the collection of bootables and improve our ability to interface with.
+
+# Goals
+[goals]: #goals
+
+- Enable a more uniform bootloader feature support across our packaged bootloaders.
+  Concretely, making it relatively straightforward to convert most of the empty spaces in the feature matrix to YES's.
+- Enable users of NixOS to implement custom bootloader tools and policy without needing to dive through the system profiles, and without patching Nixpkgs / NixOS.
+- Define a stable specification of a generation's boot data which internal and external users can rely on.
+  Changes to the specification should go through an RFC.
+
+### Non-Goals
+[non-goals]: #non-goals
+
+- Rewriting the existing bootloader backends to actually fill out the feature matrix.
+  The goal of this RFC is to make the feature development *easier*, not actually do it.
+- Supporting SecureBoot.
+  The authors of this RFC have done work in this regard, but this RFC is not about SecureBoot.
+
+# Proposed Solution
+[proposed-solution]: #proposed-solution
+
+- Each NixOS generation will have a bootspec (a JSON document) at `$out/boot.v1.json` containing all of the boot properties for that generation.
+  NixOS's bootloader backends will read these files as inputs to the bootloader installation phase.
+    - The bootloader installation phase is relatively unchanged from the way it is now.
+      The bootloader backend will have an executable that is run against a collection of generations, and the backend is any of the currently supported backends plus an "external" backend which the user can define.
+- The bootloader backends will avoid reading data from the other files and directories when possible, preferring the information in the bootspec.
+- A bootspec synthesizing tool will be used to synthesize a bootspec for generations which don't already have one.
+  This tool will be shared across all of the bootloader backends, helping produce more uniform behavior.
+- Existing bootloader backends will be updated to read properties from the bootspec, removing most if not all of their filesystem-spelunking code.
+
+### Bootspec Format v1
+[format-v1]: #format-v1
+
+
+Using the following JSON:
+
+```json5
+{
+  // Version of the specification used in the document
+  "schemaVersion": 1,
+
+  // Path to the stage-2 init, executed by the initrd
+  "init": "/nix/store/xxx-nixos-system-xxx/init",
+
+  // Path to the initrd
+  "initrd": "/nix/store/xxx-initrd-linux/initrd",
+
+  // Optional path to a tool to dynamically add secrets to an initrd
+  "initrdSecrets": "/nix/store/xxx-append-secrets/bin/append-initrd-secrets",
+
+  // Path to the kernel image
+  "kernel": "/nix/store/xxx-linux/bzImage",
+
+  // Kernel commandline options
+  "kernelParams": [
+    "amd_iommu=on",
+    "amd_iommu=pt",
+    "iommu=pt",
+    "kvm.ignore_msrs=1",
+    "kvm.report_ignored_msrs=0",
+    "udev.log_priority=3",
+    "systemd.unified_cgroup_hierarchy=1",
+    "loglevel=4"
+  ],
+
+  // The label of the system. It should contain the operating system, kernel version,
+  // and other user-relevant information to identify the system. This corresponds
+  // loosely to `config.system.nixos.label`.
+  "label": "NixOS 21.11.20210810.dirty (Linux 5.15.30)",
+
+  // Top level path of the closure, in case some spelunking is required
+  "toplevel": "/nix/store/xxx-nixos-system-xxx",
+
+  "specialisation": {
+    // <name> corresponds to <name> in specialisation.<name>.configuration.
+    // Note: a specialisation's bootspec document should not contain any specialisations.
+    "<name>": {
+      // bootspec is the path to a bootspec document
+      "bootspec": "/path/to/a/bootspec.v1.json",
+    }
+  }
+}
+```
+
+### Risks
+[risks]: #risks
+
+- Some of the bootloader backends are quite complicated, and in many cases have inadequate tests.
+  We could accidentally break corner cases.
+- The bootloader backends are inherently a weak point for NixOS, as it is our last option for rolling back.
+  We cannot roll back a broken bootloader.
+  This and the previous point are risks, but also help demonstrate the value of reducing the amount of code and complexity in the generator.
+
+### Milestones
+[milestones]: #milestones
+
+- Create and package the backwards compatibility synthesizer as a standalone tool.
+  A version of it already exists, but it is not standalone.
+- Generate the bootspec files as part of building the system closure.
+- Update the bootloader backends to use bootspec as its primary source of installation data.
+- Implement a NixOS module which supports external bootloader tooling.
+
+# FAQ
+[faq]: #faq
+
+**Does this involve patching any of the bootloaders?**
+
+No: this work is all about creating a cleaner interface for the tools we maintain which generate the files and configuration for our supported bootloaders.
+
+**Why can't we solve this with the module system?**
+
+Boot-menu generation is by definition cross-generation.
+Bootloader backends, by definition reconfigure the boot partition and records, and are supposed to apply retroactively to old configurations.
+This regeneration happens way beyond the lifecycle of the module system.
+
+**How is this easier than `if (-f "$out/append-initrd-secrets")`?**
+
+The main problem with that approach is there are a lot of files that you need to look for and examine in specific ways.
+We end up duplicating a lot of work across bootloader backends to "learn" about the system.
+By creating a source of truth for this information the bootloader backends can focus more on the work that is unique to their problem.
+
+Doing this work now may also make it easier to refactor a NixOS system's `$out` , reducing the number of symlinks and "noise" if we wanted to do that.
+
+**Does this interrupt the work to make stage-1 based on systemd?**
+
+No.
+Stage-1 won't use this document at all.
+
+**What about memtest and other boot options?**
+
+That should be left to configuration passed to the bootloader backend.
+There could be any number of "extra" bootables like this, and importantly they have nothing to do with a specific NixOS system closure.
+Concretely, a user who enables memtest probably wants the most recent memtest, and not one memtest from every generation.
+
+# Open Questions
+[open-questions]: #open-questions
+
+- Should there be a general-purpose "meta" or "extensions" field which allows arbitrary extension of the data? What should this look like?
+
+# Future Work
+[future]: #future-work
+
+- Completing the migration from filesystem-spelunking into using the bootspec data.
+- Implementing a NixOS module for supporting externalized bootloader backends.
+- Implementing a base level of SecureBoot support.
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -90,7 +90,7 @@ This proposal is specifically to deal with the collection of bootables and impro
 
 - Each NixOS generation will have a bootspec (a JSON document) at `$out/bootspec/boot.v1.json` containing all of the boot properties for that generation.
   NixOS's bootloader backends will read these files as inputs to the bootloader installation phase.
-  A bootspec document named `boot.v1.json` must have the `schemaVersion` set to `1`. Any other value makes the document invalid.
+  A bootspec document named `boot.v1.json` must have the `schemaVersion` set to `1`. Any other value makes the document invalid. Generally, the file must be named `boot.v${schemaVersion}.json`.
     - The bootloader installation phase is relatively unchanged from the way it is now.
       The bootloader backend will have an executable that is run against a collection of generations, and the backend is any of the currently supported backends plus an "external" backend which the user can define.
 - The bootloader backends will avoid reading data from the other files and directories when possible, preferring the information in the bootspec.
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -85,20 +85,22 @@ This proposal is specifically to deal with the collection of bootables and impro
 - Supporting SecureBoot.
   The authors of this RFC have done work in this regard, but this RFC is not about SecureBoot.
 - Integrating Bootspec into an existing bootloader backend is not expected to perfectly produce the exact same result. For example, it is likely that the text describing the entries in the menu may differ.
+- Specifying how to discover generations. This is desirable, but should not be tied to bootspec directly since bootspec may be useful with diverse discovery mechanisms.
 
 # Proposed Solution
 [proposed-solution]: #proposed-solution
 
 - Each NixOS generation will have a bootspec (a JSON document) at `$out/boot.json` containing all of the boot properties for that generation.
   NixOS's bootloader backends will read these files as inputs to the bootloader installation phase.
-  A bootspec document named `boot.json` must have a `"v1"` key that contains the body of the specification.
-  Generally, the key must follow the format of `v${schemaVersion}`.
+  A v1 bootspec document must have a key `"org.nixos.bootspec.v1"` that describes a bootable NixOS configuration.
     - The bootloader installation phase is relatively unchanged from the way it is now.
       The bootloader backend will have an executable that is run against a collection of generations, and the backend is any of the currently supported backends plus an "external" backend which the user can define.
 - The bootloader backends will avoid reading data from the other files and directories when possible, preferring the information in the bootspec.
 - A bootspec synthesizing tool will be used to synthesize a bootspec for generations which don't already have one.
   This tool will be shared across all of the bootloader backends, helping produce more uniform behavior.
 - Existing bootloader backends will be updated to read properties from the bootspec, removing most if not all of their filesystem-spelunking code.
+- The file paths referred to in the bootspec document should be references to Nix store paths, so that the presence of the bootspec document in the store (assuming no breakage of the Nix store invariants through modifications outside Nix or bugs in Nix) guarantees the presence of the files required for boot.
+  Bootspec documents should not be copied out of the store to avoid breaking this invariant.
 
 ### Bootspec Format v1
 [format-v1]: #format-v1
@@ -109,17 +111,21 @@ Using the following JSON:
 ```json5
 {
   // Toplevel key describing the version of the specification used in the document
-  "v1": {
+  "org.nixos.bootspec.v1": {
     // (Required) System type the bootspec is intended for (e.g. `x86_64-linux`, `aarch64-linux`)
     "system": "x86_64-linux",
 
-    // (Required) Path to the stage-2 init, executed by the initrd
+    // (Required) Path to the stage-2 init, executed by the initrd (if present)
     "init": "/nix/store/xxx-nixos-system-xxx/init",
 
     // (Optional) Path to the initrd
     "initrd": "/nix/store/xxx-initrd-linux/initrd",
 
-    // (Optional) Path to a tool to dynamically add secrets to an initrd
+    // (Optional) Path to a tool to dynamically add secrets to an initrd.
+    // Consumers of a bootspec document should copy the file referenced by the "initrd" key to a writable location, ensure that the file is writable, invoke this tool with the path to the initrd as its only argument, and use the initrd as modified by the tool for booting.
+    // This may be used to add files from outside the Nix store to the initrd.
+    // This tool is expected to run on the system whose boot specification is being set up, and may thus fail if used on a system where the expected stateful files are not in place or whose CPU does not support the instruction set of the system to be booted.
+    // If this field is present and the tool fails, no boot configuration should be generated for the system.
     "initrdSecrets": "/nix/store/xxx-append-secrets/bin/append-initrd-secrets",
 
     // (Required) Path to the kernel image
@@ -144,27 +150,26 @@ Using the following JSON:
 
     // (Required) Top level path of the closure, in case some spelunking is required
     "toplevel": "/nix/store/xxx-nixos-system-xxx",
+  },
+  // The top-level object may contain arbitrary further keys ("extensions"), whose semantics may be defined by third parties.
+  // The use of reverse-domain-name namespacing is recommended in order to avoid name collisions.
 
-    // (Optional) Mapping of specialisation names to their bootspec document
-    "specialisation": {
-      // <name> corresponds to <name> in specialisation.<name>.configuration.
-      // Note that it is not valid for a specialisation to be a different version than the main document.
-      "<name>": {
-        // A full Bootspec v1 document, _without_ the `"v1"` key.
-        // Note that it is not valid for a specialisation to have further specialisations.
-      }
-    },
-
-    // (Optional) User-specified metadata
-    "extensions": {
-      // <namespace> corresponds to a user-provided namespace
-      // to reduce the chances of collision when using
-      // multiple extensions at once
-      "<namespace>": {
-        // Any kind of representable JSON is valid here
+  // (Optional) Specialisations are an extension to the specification which allows bundling multiple variants of a NixOS configuration with a single parent.
+  // These are shaped like the top level; to be precise:
+  //  - Each entry in the toplevel "org.nixos.specialisation.v1" object represents a specialisation.
+  //  - In order for the top-level document to be a valid v1 bootspec, each specialisation must have a valid "org.nixos.bootspec.v1" key whose value conforms to the same schema as the toplevel "org.nixos.bootspec.v1" object.
+  //  - The behaviour of nested specialisations (i.e. entries in "org.nixos.specialisation.v1" which themselves contain the "org.nixos.specialisation.v1" key) is not defined.
+  //  - In particular, there is no expectation that such nested specialisations will be handled by consumers of bootspec documents.
+  //  - Each specialisation document may contain arbitrary further keys (extensions), like the top-level document.
+  //  - The semantics of these should be the same as when these keys are used at the top level, but only apply for the given specialisation.
+  "org.nixos.specialisation.v1": {
+    // Each key in this object corresponds to a specialisation as defined by the `specialisation.<name>` NixOS option.
+    "<name>": {
+      "org.nixos.bootspec.v1": {
+        // See above
       }
     }
-  }
+  },
 }
 ```
 
@@ -229,3 +234,4 @@ Concretely, a user who enables memtest probably wants the most recent memtest, a
 - Completing the migration from filesystem-spelunking into using the bootspec data.
 - Implementing a NixOS module for supporting externalized bootloader backends.
 - Implementing a base level of SecureBoot support.
+- Addressing the known weaknesses of this specification (no support for multiple initrds, janky initrd secrets mechanism, lack of device tree support) in a later version of the spec.
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -86,7 +86,7 @@ This proposal is specifically to deal with the collection of bootables and impro
 # Proposed Solution
 [proposed-solution]: #proposed-solution
 
-- Each NixOS generation will have a bootspec (a JSON document) at `$out/boot.v1.json` containing all of the boot properties for that generation.
+- Each NixOS generation will have a bootspec (a JSON document) at `$out/bootspec/boot.v1.json` containing all of the boot properties for that generation.
   NixOS's bootloader backends will read these files as inputs to the bootloader installation phase.
     - The bootloader installation phase is relatively unchanged from the way it is now.
       The bootloader backend will have an executable that is run against a collection of generations, and the backend is any of the currently supported backends plus an "external" backend which the user can define.
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -84,7 +84,7 @@ This proposal is specifically to deal with the collection of bootables and impro
   The goal of this RFC is to make the feature development *easier*, not actually do it.
 - Supporting SecureBoot.
   The authors of this RFC have done work in this regard, but this RFC is not about SecureBoot.
-- Integrating Bootspec into an existing bootloader backend is not expected to perfectly produce the exact same result. For example, it is likely that the text describing the entrties in the menu may differ.
+- Integrating Bootspec into an existing bootloader backend is not expected to perfectly produce the exact same result. For example, it is likely that the text describing the entries in the menu may differ.
 
 # Proposed Solution
 [proposed-solution]: #proposed-solution
@@ -170,7 +170,7 @@ Using the following JSON:
 - Create and package the backwards compatibility synthesizer as a standalone tool.
   A version of it already exists, but it is not standalone.
 - Generate the bootspec files as part of building the system closure.
-- Update the bootloader backends to use bootspec as its primary source of installation data.
+- Update the bootloader backends to use bootspec as their primary source of installation data.
 - Implement a NixOS module which supports external bootloader tooling.
 
 # FAQ
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -5,7 +5,7 @@ author: Graham Christensen
 co-authors: Cole Helbling
 shepherd-team: (names, to be nominated and accepted by RFC steering committee)
 shepherd-leader: (name to be appointed by RFC steering committee)
-related-issues: (will contain links to implementation PRs)
+related-issues: https://github.com/NixOS/nixpkgs/pull/172237
 ---
 
 # Summary
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -16,6 +16,8 @@ These facts are used as the primary input for bootloader backends like systemd-b
 
 In this proposal, we create a stable, comprehensive, and machine-parsable definition of a NixOS Generation as an intermediate representation (IR) between the NixOS system definition and the bootloader management tools.
 
+This document describes Bootspec v1.
+
 # Motivation
 [motivation]: #motivation
 
@@ -88,6 +90,7 @@ This proposal is specifically to deal with the collection of bootables and impro
 
 - Each NixOS generation will have a bootspec (a JSON document) at `$out/bootspec/boot.v1.json` containing all of the boot properties for that generation.
   NixOS's bootloader backends will read these files as inputs to the bootloader installation phase.
+  A bootspec document named `boot.v1.json` must have the `schemaVersion` set to `1`. Any other value makes the document invalid.
     - The bootloader installation phase is relatively unchanged from the way it is now.
       The bootloader backend will have an executable that is run against a collection of generations, and the backend is any of the currently supported backends plus an "external" backend which the user can define.
 - The bootloader backends will avoid reading data from the other files and directories when possible, preferring the information in the bootspec.
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -84,6 +84,7 @@ This proposal is specifically to deal with the collection of bootables and impro
   The goal of this RFC is to make the feature development *easier*, not actually do it.
 - Supporting SecureBoot.
   The authors of this RFC have done work in this regard, but this RFC is not about SecureBoot.
+- Integrating Bootspec into an existing bootloader backend is not expected to perfectly produce the exact same result. For example, it is likely that the text describing the entrties in the menu may differ.
 
 # Proposed Solution
 [proposed-solution]: #proposed-solution
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -145,8 +145,8 @@ Using the following JSON:
     // <name> corresponds to <name> in specialisation.<name>.configuration.
     // Note: a specialisation's bootspec document should not contain any specialisations.
     "<name>": {
-      // bootspec is the path to a bootspec document
-      "bootspec": "/path/to/a/bootspec.v1.json",
+      // A full Bootspec v1 document.
+      // Note that it is not valid for a specialisation to have further specialisations.
     }
   }
 }
--- a/rfcs/0125-bootspec.md
+++ b/rfcs/0125-bootspec.md
@@ -66,7 +66,7 @@ By creating a well defined specification of generations and the boot-relevant da
 Systemd's bootloader specification is a good format for a different problem.
 A single NixOS generation can contain multiple bootable systems and options, with additional features unique to NixOS built on top.
 Most Linux distributions don't deal with many unique and ever-changing bootables.
-This proposal is specifically to deal with the collection of bootables and improve our ability to interface with.
+This proposal is specifically to deal with the collection of bootables and improve our ability to treat the collection consistently.
 
 # Goals
 [goals]: #goals


rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,212 @@
+---
+feature: bootspec
+start-date: 2022-05-09
+author: Graham Christensen
+co-authors: Cole Helbling
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
```
grahamc, line 8:
```suggestion
related-issues: https://github.com/NixOS/nixpkgs/pull/172237
```

ryantm:
[rendered](https://github.com/grahamc/rfcs/blob/bootspec/rfcs/0125-bootspec.md)

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,212 @@
...
+There are infinitely many policy choices and implementations possible, and a good solution here is to allow deployers of NixOS to implement their own bootloader management tools.
+By creating a well defined specification of generations and the boot-relevant data we enable this external development.
+
+## What about systemd's Bootloader Specification?
+[systemd-bootloader-specification]: #systemd-bootloader-specification
+
+Systemd's bootloader specification is a good format for a different problem.
+A single NixOS generation can contain multiple bootable systems and options, with additional features unique to NixOS built on top.
+Most Linux distributions don't deal with many unique and ever-changing bootables.
+This proposal is specifically to deal with the collection of bootables and improve our ability to interface with.
```
ryantm, line 67: interface with what?

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+}
+```
+
+### Risks
+[risks]: #risks
+
+- Some of the bootloader backends are quite complicated, and in many cases have inadequate tests.
+  We could accidentally break corner cases.
```
7c6f434c, line 156: Is it supposed that initial rewrite should produce exactly same menu? Should it be an option to compare two ways of generating things at activation time, so that users have easier time checking if their setup is properly supported after rewrite?

grahamc: It should produce a nearly identical menu, though minor details like the formatting of the information in the text may be different.

rfcs/0125-bootspec.md
grahamc: I'm not totally sure what I was thinking there. Perhaps this:

```suggestion
This proposal is specifically to deal with the collection of bootables and improve our ability to treat the collection consistently.
```

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,212 @@
...
+
+- Rewriting the existing bootloader backends to actually fill out the feature matrix.
+  The goal of this RFC is to make the feature development *easier*, not actually do it.
+- Supporting SecureBoot.
+  The authors of this RFC have done work in this regard, but this RFC is not about SecureBoot.
+
+# Proposed Solution
+[proposed-solution]: #proposed-solution
+
+- Each NixOS generation will have a bootspec (a JSON document) at `$out/boot.v1.json` containing all of the boot properties for that generation.
```
Hoverbear, line 89: Given that you already require the document to be `json` and have a `"schemaVersion": 1,` field, perhaps this is redundant?

grahamc: I could imagine having multiple `boot.x.json` documents in the output to create very easy migrations between breaking changes, and the `scehmaVersion` in the document as a run-time verification of what exactly you're looking at. What do you think?

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,212 @@
...
+
+  // Top level path of the closure, in case some spelunking is required
+  "toplevel": "/nix/store/xxx-nixos-system-xxx",
+
+  "specialisation": {
+    // <name> corresponds to <name> in specialisation.<name>.configuration.
+    // Note: a specialisation's bootspec document should not contain any specialisations.
+    "<name>": {
+      // bootspec is the path to a bootspec document
+      "bootspec": "/path/to/a/bootspec.v1.json",
```
roosemberth, line 146: Where is this path expected to be? In the nix store? In the file directory?

I can imagine tools parsing this file outside a system with a viable nix-store.

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,212 @@
...
+```
+
+### Risks
+[risks]: #risks
+
+- Some of the bootloader backends are quite complicated, and in many cases have inadequate tests.
+  We could accidentally break corner cases.
+- The bootloader backends are inherently a weak point for NixOS, as it is our last option for rolling back.
+  We cannot roll back a broken bootloader.
+  This and the previous point are risks, but also help demonstrate the value of reducing the amount of code and complexity in the generator.
```
06kellyjac, line 159: Could 2 bootloader partitions be an option to mitigate this risk? An A/B style thing where someone uses an external bootloader and also one of our more tested options as a fallback like systemd-boot/grub

It wouldn't have to be default but could be opt in

grahamc: I don't think solving the problem like that is in scope for this RFC. There could be other work around making the bootloader A/B'd, but in the risks section for this RFC I think we should be focusing on ways to mitigate the risks of this specific RFC, not the systemic risk of a bad bootloader which NixOS already has.

K900 (commented):
The `raspberrypi` bootloader should probably go away in some form, it's a horrible hack. Another thing for my bucket list...

rfcs/0125-bootspec.md
7c6f434c: Specific new risks (rewriting all the things) could probably be mitigated by at least encouraging support for before/after comparisons at the generate-the-menu-time. Hopefully the users with rare corner cases know they have rare corner cases and will use whatever consistency checks are offered…

rfcs/0125-bootspec.md
Hoverbear: If the file name and `schemaVersion` differ, is it an assertion error?

rfcs/0125-bootspec.md
grahamc: I'm inclined to move to embedding the full bootspec for each specialisation into the parent document, to address problems like this. I've written a patch to do that on my implementation PR, and will update the tools and RFC to match: https://github.com/NixOS/nixpkgs/pull/172237/commits/33fa3bd4409d47872ef88454eadad5950d5f691c

samueldr: Embedding the full document is the way to go imo.

grahamc: I've also updated the bootspec datatypes and synthesis tool to match: https://github.com/DeterminateSystems/bootspec/pull/2 and corresponding secureboot implementation: https://github.com/DeterminateSystems/bootspec-secureboot/pull/141 I can update the RFC later today / tomorrow morning.

spacekookie:
Hey everyone, this RFC is open for shepherd nominations now

rfcs/0125-bootspec.md
grahamc: I don't know about an assertion error, but it should be considered an error of some sort. Probably better behavior would be to discard it and attempt to use a different boot.vN.json.

grahamc, line 89: Imagining a scenario where we have a v2 schema, we should probably still generate a v1 document. In this scenario we'd have a growing number of boot.vX.json documents. It'd probably be good to move these into a subdirectory.

```suggestion
- Each NixOS generation will have a bootspec (a JSON document) at `$out/bootspec/boot.v1.json` containing all of the boot properties for that generation.
```

grahamc: https://github.com/DeterminateSystems/bootspec-secureboot/pull/143 & https://github.com/NixOS/nixpkgs/pull/172237/commits/9b4df1d50f6df092cb195d630972ab36835d4f49

rfcs/0125-bootspec.md
```diff
@@ -84,6 +84,7 @@ This proposal is specifically to deal with the collection of bootables and impro
   The goal of this RFC is to make the feature development *easier*, not actually do it.
 - Supporting SecureBoot.
   The authors of this RFC have done work in this regard, but this RFC is not about SecureBoot.
+- Integrating Bootspec into an existing bootloader backend is not expected to perfectly produce the exact same result. For example, it is likely that the text describing the entrties in the menu may differ.
```
cole-h, line 87:
```suggestion
- Integrating Bootspec into an existing bootloader backend is not expected to perfectly produce the exact same result. For example, it is likely that the text describing the entries in the menu may differ.
```

samueldr:
I self-non-nominate. I suspect my nomination is otherwise likely (given my work/interest in the topic). I know from experience that I won't be able to attend to the duties for shepherding this RFC.

I'm still open to ad-hoc questions and opinions though.

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,216 @@
...
+  We cannot roll back a broken bootloader.
+  This and the previous point are risks, but also help demonstrate the value of reducing the amount of code and complexity in the generator.
+
+### Milestones
+[milestones]: #milestones
+
+- Create and package the backwards compatibility synthesizer as a standalone tool.
+  A version of it already exists, but it is not standalone.
+- Generate the bootspec files as part of building the system closure.
+- Update the bootloader backends to use bootspec as its primary source of installation data.
```
lheckemann, line 171: nit
```suggestion
- Update the bootloader backends to use bootspec as their primary source of installation data.
```

lheckemann (approved):
LGTM overall.

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,216 @@
...
+  "initrd": "/nix/store/xxx-initrd-linux/initrd",
+
+  // Optional path to a tool to dynamically add secrets to an initrd
+  "initrdSecrets": "/nix/store/xxx-append-secrets/bin/append-initrd-secrets",
+
+  // Path to the kernel image
+  "kernel": "/nix/store/xxx-linux/bzImage",
+
+  // Kernel commandline options
+  "kernelParams": [
```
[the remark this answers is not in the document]
7c6f434c: But kernel command line is also used to pass configuration to early init, so it is a bit less limited.

A list gives an option to do at least some filtering.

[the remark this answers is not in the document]
samueldr: There's no drawbacks to keeping this a list, while if it was squished in a string it would lose a bit of the meaning of the input. At least that's what I see here.

> A list gives an option to do at least some filtering.

It's probably too late to filter at this point; this is an input for tools that serialize the generation's boot configuration. But true, it wouldn't be as easy to filter otherwise.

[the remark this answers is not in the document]
7c6f434c: I assume that a) there will be unpredictable mix of data for kernel / PID1 / activation, b) the type choice here will propagate into some random place.

So a list costs little, allows asserting lack of spaces in entries if we want (pretty sure _something_ will be broken by them), and increases chances to filter out a single specific option before bootspec being written. Overall, a win, in my opinion.

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,218 @@
...
+      "kvm.report_ignored_msrs=0",
+      "udev.log_priority=3",
+      "systemd.unified_cgroup_hierarchy=1",
+      "loglevel=4"
+    ],
+
+    // The label of the system. It should contain the operating system, kernel version,
+    // and other user-relevant information to identify the system. This corresponds
+    // loosely to `config.system.nixos.label`.
+    "label": "NixOS 21.11.20210810.dirty (Linux 5.15.30)",
```
Atemu, line 141: Some bootloaders (GRUB for instance) also record the date when the generation was made (probably mtime of the generation symlink?). Would that be included in the label or a separate field?

LunNova:
Does this intend to support multiple bootloaders at the same time/is that in scope for this? With EFI there's no reason you can't have multiple in the ESP, and currently we use rEFind and systemd boot but set up rEFind imperatively. Baking in support for a list of bootloaders to set up would be cool.

LunNova:
I was also working on [a NixOS module](https://github.com/LunNova/nixos-configs/blob/dev/modules/exported/efi-tools.nix) recently that puts some EFI tools from packages into the ESP, which can then be picked up by any bootloader that looks in EFI/tools. I hope to upstream this eventually, but it currently feels very hacky as it updates the contents of the ESP during an activation script.

Would this fit into this model?

samueldr:
> Does this intend to support multiple bootloaders at the same time/is that in scope for this? With EFI there's no reason you can't have multiple in the ESP, and currently we use rEFind and systemd boot but set up rEFind imperatively. Baking in support for a list of bootloaders to set up would be cool.

When talking about the scope of the RFC in an earlier phase, it was discussed that this was strictly about *describing the pertinent data about a generation*. How the data ends-up being used is out of scope. Note that this does not prevent or make this any easier than it currently is [except for the fact that the source of truth is better defined].

samueldr:
> I was also working on [a NixOS module](https://github.com/LunNova/nixos-configs/blob/dev/modules/exported/efi-tools.nix) recently that puts some EFI tools from packages into the ESP, which can then be picked up by any bootloader that looks in EFI/tools. I hope to upstream this eventually, but it currently feels very hacky as it updates the contents of the ESP during an activation script.
> 
> Would this fit into this model?

See [the section about memtest](https://github.com/NixOS/rfcs/pull/125/files#diff-709266622ec29f9cabb5a5b4e6f93792851db9cfebd72bdf78f3032b437d8764R202-R206).

It is assumed that *extra tools* live outside of the scope of the *generation lifecycle*. In other words, there is no need to "lock" the extra tool with a specific generation.

Though, if you have tools or utilities that are intrinsically linked to a NixOS generation, please expand on the use case. It might be worthwhile to know more about it.

LunNova:
Okay! I guess I'll wait and see what happens when this gets implemented, and try to get support for a list added at that point if it isn't done already.

Not really sure where to fit in the extra tools in the ESP thing, maybe the tool installer gets treated as another bootloader to install, maybe the current approach with an activation script is fine.

lheckemann:
I nominate myself as a shepherd for this RFC.

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,218 @@
...
+[format-v1]: #format-v1
+
+
+Using the following JSON:
+
+```json5
+{
+  // Toplevel key describing the version of the specification used in the document
+  "v1": {
+
```
arianvp, line 113: Don't we need to know the target system architecture in the bootspec?

grahamc: I don't know that you would, but perhaps we would. What for?

roosemberth: I don't think the arch should be included in the present document.

I could see a 'universal media' in the future where the bootloader can filter the derivations compatible with the system architecture, but that seems too far-fetched to be included in v1.

lheckemann:
I'm nominating for the shepherd team:
 - @LunNova, @roosemberth, @Hoverbear who have made comments resulting in changes
- @govanify who has done some signing-related work on the GRUB NixOS module
- @lukegb who implemented specialisations for systemd-boot
- @nh2, @xaverdh who have done some other work on the GRUB module
- @pinpox who wants Secure Boot, since this helps a great deal with implementing it, and shepherding this RFC is a great way to help move it forwards

Please comment here to let us know whether you can and want to shepherd.

GovanifY:
Thanks for the nomination, I would love to be a shepherd for this RFC!

RaitoBezarius:
> I'm nominating for the shepherd team:
> 
> * @LunNova, @roosemberth, @Hoverbear who have made comments resulting in changes
> * @GovanifY who has done some signing-related work on the GRUB NixOS module
> * @lukegb who implemented specialisations for systemd-boot
> * @nh2, @xaverdh who have done some other work on the GRUB module
> * @pinpox who wants Secure Boot, since this helps a great deal with implementing it, and shepherding this RFC is a great way to help move it forwards
> 
> Please comment here to let us know whether you can and want to shepherd.

I would also be interested into following this work as a shepherd as I am planning to do some substantial work on Secure Boot in the future.

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,218 @@
...
+That should be left to configuration passed to the bootloader backend.
+There could be any number of "extra" bootables like this, and importantly they have nothing to do with a specific NixOS system closure.
+Concretely, a user who enables memtest probably wants the most recent memtest, and not one memtest from every generation.
+
+# Open Questions
+[open-questions]: #open-questions
+
+- Should there be a general-purpose "meta" or "extensions" field which allows arbitrary extension of the data? What should this look like?
+
```
roberth, line 212:
```suggestion

- Should there be a non-derivation metadata facility that carries useful metadata for which it isn't desirable to cause rebuilds, redeployments or new boot entries - specifically data such a git commit hash or a date of last modification.

```

roosemberth:
Thanks for the nomination, I've never been a shepherd before but I'm excited to give it a try :rocket:!

LunNova:
@lheckemann Thanks for the nomination, I would like to accept but due to time constraints it would not be reasonable for me to do so.

lheckemann:
Wow, that gives us 5 accepted nominations to form a team of 4!
Myself (volunteering as leader), @06kellyjac, @GovanifY and @RaitoBezarius will be the shepherd team, @roosemberth you're welcome to join the meetings as well of course :)

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,218 @@
+---
+feature: bootspec
+start-date: 2022-05-09
+author: Graham Christensen
+co-authors: Cole Helbling
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
```
kevincox, line 7:
```suggestion
shepherd-team: @lheckemann, @06kellyjac, @GovanifY, @RaitoBezarius
shepherd-leader: @lheckemann
```

kevincox (commented):
Thanks @lheckemann for volunteering to lead the shepherd team. Please consider setting up a chat room and meeting to move this RFC forward.

grahamc:
Matrix room exists: https://matrix.to/#/#rfc-0125:nixos.org

grahamc:
Note that the Matrix room existed before it was in the NixOS matrix space, and is definitely welcome for anyone to join who is interested. If the shepherds want to make a separate room, we could do that.

rfcs/0125-bootspec.md
grahamc: The tooling / automation is going to have to identify a collection of bootspec documents to use. It could be a directory of symlinks, or a database, or a tool run in CI -- but wherever it comes from should probably provide this sort of information.

K900: FWIW, I've implemented the current logic (well, an amalgamation of the current logic from different scripts) as a reusable component here: https://gitlab.com/K900/bootis/-/tree/main/nixos-list-generations

rfcs/0125-bootspec.md
grahamc: I think we should add the architecture.

lheckemann:
## Shepherd meeting 2022-08-18

Present: @grahamc, @raitobezarius, @06kellyjac, @govanify, @lheckemann, anonymous observer

The shepherd team unanimously agrees that this is a desirable RFC and would like to move it to final comment period (FCP) once the following points are addressed:

- The initrd field should become optional.
- An optional `meta` field should be added, specified as a JSON object with arbitrary contents, to allow experimentation within the confines of the spec.
- "optional" should be clarified to mean that the field may be _absent_ (not null)
- We agreed that things like A/B bootloader schemes are out of scope
- Determining which boot specs are used by a bootloader config generator (and unifying this across boot loader implementations) is desirable, but out of scope
- The system should be included as a field; an example use of this is generating EFI binaries to place in the (architecture-dependent) removable device fallback path.
- Before we enter the FCP, the draft implementation should:
   - support the current version of bootspec
   - work

rfcs/0125-bootspec.md
RaitoBezarius: @roberth Could you clarify what you envision by a "non derivation metadata facility"? Would that be a generalization of bootspec so that bootspec would be only a special case of it? (my interpretation)

We discussed it during our meeting, but I think we had some different interpretations of your comments.

rfcs/0125-bootspec.md
RaitoBezarius: What was mentioned by @lheckemann during the meeting was (paraphrased): putting the arch helps to do the decision of "do I need BootX64 or BootX32, etc. artifacts?" IIRC. (correct me if I'm wrong)

samueldr:
Thinking out loud:

> We agreed that things like A/B bootloader schemes are out of scope

It makes sense that it would be out of scope, since the A/B handling (or even boot counting handling) lives outside the lifecycle of a generation. The generation itself does not really need to care about it *for data to present to the bootloader*.

That is, the generation likely needs to be aware that it needs to signal that the boot was successful, but stating that the boot is successful is not a property that can be represented in the bootspec, since it is runtime stateful data.

Even signalling that the generation can/can't signal that a boot is successful is stretching it, and if needed for the first draft of such schemes should be handled in the `meta` field. I think it is fair to assume that in such a scheme, the bootloader can safely assume that the generation will intrinsically signal its success, that if a generation does not signal its success it is a failure condition at runtime.

rfcs/0125-bootspec.md
roberth: By adding a field such as the git revision or last modified date, you're creating new generations that are otherwise identical to preceding ones when you make changes to unrelated files. My open question aims to prevent that from happening.
(The need for this came up in https://github.com/NixOS/nixpkgs/pull/179772)

> Would that be a generalization of bootspec so that bootspec would be only a special case of it?

Possibly.

> > non-derivation

I suppose the metadata _can_ be in a derivation, but this derivation should not be used for determining the equivalence of generations. Assuming we keep using the output path of `system.build.toplevel` for the purpose of generation equivalence, the metadata could be put in a separate derivation that isn't part of `system.build.toplevel`.
There's a number of ways you could implement this sort of thing.

Making `toplevel` not top level isn't great. I could rephrase this as

```suggestion

- How should we identify generations as being equivalent? This will help keep boot menus concise and allows activation to be cut short, especially on systems with frequently changing metadata, such as the revision of a repository that contains components not deployed to the system.
```

This allows other solutions such as a "fingerprint" field containing a hash of the essential values so that duplicate generations can be detected by the bootloader menu generator (or other tool that benefits from a "shortcutting" behavior).

Implementation idea (not necessarily part of bootspec; more of a NixOS implementation detail. `fingerprint` existence and type (string) _is_ bootspec though.)
```
let spec0 = config.system.build.bootspec; # assuming this is where NixOS puts the almost-final bootspec stuff
     specHash = builtins.hashString (builtins.toJSON (config.system.build.bootspec // { meta = null; }));
     spec1 = spec0 // { fingerprint = specHash; };
in {
  # assuming somethign like https://github.com/NixOS/nixpkgs/pull/148456/commits/4ec415cff9bcbaef08e0d900406a5c7181e71881
  config.system.systemBuilderCommands = ''
    ln -s ${json.generate "bootspec.json" spec1} $out/bootspec.json
  '';
}
```

lheckemann:
We had an RFC meeting at nixcon's hackday. We came to:

1. We'll have the "extension" field for extra metadata. Tools should strive to use their own namespaces within the extensions field, and the RFC should state this.
2. This RFC is strictly defining an interface, not the behavior of specific use of this interface.
3. We should merge the implementation of bootspec into NixOS before this RFC is merged under an experimental option to make it easy to iterate on and test / use. Bootspec will then be enabled unconditionally after the RFC is merged.

RaitoBezarius:
While finalizing the PR for experimental bootspec in nixpkgs and using a schema-defined bootspec, an interesting question arised.

Can specialisation have a different bootspec version or should they use the same version necessarily? What about migrations? Do we need them? Do we need backward-compat? I misread the spec potentially but I feel like this is going to be a question after N generations for N very large such that *2* versions (or at least, an extended bootspec, wink wink secureboot or tpm2 or whatever) start existing.

RaitoBezarius:
> We had an RFC meeting at nixcon's hackday. We came to:
> 
> 1. We'll have the "extension" field for extra metadata. Tools should strive to use their own namespaces within the extensions field, and the RFC should state this.
> 2. This RFC is strictly defining an interface, not the behavior of specific use of this interface.
> 3. We should merge the implementation of bootspec into NixOS before this RFC is merged under an experimental option to make it easy to iterate on and test / use. Bootspec will then be enabled unconditionally after the RFC is merged.

I cannot remember the name of the extension field we decided on: `extension` or `extensions` ?
Note that I adopted `extensions`, but I have no strong feeling on this.

RaitoBezarius:
Just a status update in terms of implementation over https://github.com/DeterminateSystems/bootspec (set of libraries and tooling around bootspecs) and https://github.com/nix-community/lanzaboote (consumer of bootspec for SecureBoot over systemd-boot):

> * The initrd field should become optional.

`initrd` is not optional AFAIK.

> * An optional `meta` field should be added, specified as a JSON object with arbitrary contents, to allow experimentation within the confines of the spec.

`meta` is named `extensions` at the moment, it is materialized by a `Option<Extension>` where `Extension` is a generic type, defaulting to `HashMap<String, serde_json::Value>`, which I feel is covering all the cases for now.

The biggest issue is that, a missing `extensions` provide a `None` value, rather than an empty HashMap in the case where `Extension = HashMap<X, Y>` which is a bit of a shame, but we are sorting out things in https://github.com/DeterminateSystems/bootspec/pull/68#discussion_r1053687508 — please join the discussion if you are interested.

> * "optional" should be clarified to mean that the field may be _absent_ (not that it may be null)

We need to go and double check this is the case, for example: `extensions: null` might provide `extensions: None`.

> * We agreed that things like A/B bootloader schemes are out of scope

FWIW, https://github.com/baloo may provide valuable insights on how to achieve this. (as a reference for any reader in the future.)

> * The system should be included as a field; an example use of this is generating EFI binaries to place in the (architecture-dependent) removable device fallback path.

This is not done for now, plus, we are noticing that for example, we use `$esp/nixos` to store kernels & initrd, this path is not really documented and is an implementation-detail.

7c6f434c:
> `initrd` is not optional AFAIK.

In which sense? I see multiple options…

* The chosen backends demand it
* NixOS configured without custom modules would always provide it
* No published Nix-managed system boots without initrd

I am not qualified to judge (1), I find (2) plausible, and (3) false.

What is the best way to reflect this in terms of what fields of the spec are optional?

lheckemann:
@7c6f434c we [clarified this in the matrix channel](https://matrix.to/#/!CDsPGxqljDUrcevkKh:grahamc.com/$35mQ-LVnNhUmTSpm2TeRG2hpkaX3qI3ft6fT0aIi5U0?via=grahamc.com&via=matrix.org&via=nixos.org): the TODO is for changing the spec to allow generations without initrds and Ryan thought this was for changing the implementation but not the spec.

RaitoBezarius:
Bringing this discussion to the community attention: https://github.com/DeterminateSystems/bootspec/pull/68#discussion_r1055005709 (some discussion was done too with @lheckemann over Matrix channel.)

RaitoBezarius:
Proposed changes for this RFC: https://github.com/grahamc/rfcs/pull/7

- implements *optional* initrd
- implements `system` field
- implements `extensions` field
- clarify the meaning of optional
- clarify what is *required*

rfcs/0125-bootspec.md
```diff
@@ -0,0 +1,231 @@
...
+**Does this interrupt the work to make stage-1 based on systemd?**
+
+No.
+Stage-1 won't use this document at all.
+
+**What about memtest and other boot options?**
+
+That should be left to configuration passed to the bootloader backend.
+There could be any number of "extra" bootables like this, and importantly they have nothing to do with a specific NixOS system closure.
+Concretely, a user who enables memtest probably wants the most recent memtest, and not one memtest from every generation.
```
hesiod, line 221: Another option (just an idea that came to my mind) would be to allow special "generation-less bootspec files" that have alternative fields. Instead of `kernel`, `kernelParams`, `initrd`, `init`, `system`, these files would simply have an `efi` field pointing to a valid EFI executable file.

That would allow a unified view of "bootable things" as an input to the bootloader backend: Either a generation (a directory containing a generation bootspec file) or a standalone bootable (a directory containing a standalone bootspec file).

hesiod:
I've tested some things in the existing (non-bootspec) systemd-boot bootloader backend. Reading through the RFC, I came across a few possible issues regarding compatibility of the bootspec specification with some possible bootloader features.
Specifically, the following comments primarily have the needs of the systemd-boot backend in mind, including the corresponding requirements in order correctly generate "Type #1 Boot Loader Entries" according to the [UAPI Boot Loader Specification](https://uapi-group.org/specifications/specs/boot_loader_specification/#type-1-boot-loader-specification-entries).
I hope I didn't miss any prior discussion on the points I'm raising, I looked through the previous comments but perhaps I missed something.

#### `initrd` field as a list of images
There already was some discussion regarding whether the `initrd` field should be required. However, I'd propose that it should ideally be a list instead of a single field, supporting any number of initrd entries. Why?

From the UAPI spec: "`initrd` is the initrd cpio image to use when executing the kernel. This key **may appear more than once** in which case all specified images are used, in the order they are listed."

In other distributions (e.g. Arch), this feature is typically used for specifying microcode images. NixOS currently appends the microcode image to the beginning of the initrd (as the first file in the archive) if I understand correctly.
While this is off topic for this RFC, it might make sense to move to the multiple initrd approach to microcode updates in the future (it would save ~5MB per generation as the microcode image could be shared between generations).
By making the  bootspec `initrd` field a list, both microcode-as-initrd systems and initrd-less systems could be supported in the future.

Another use case for additional initrd images would be DSDT modification ([Arch wiki](https://wiki.archlinux.org/title/DSDT#Using_a_CPIO_archive)), however this is not implemented in NixOS at the moment.

Note: I'm not 100% sure how bootloaders other than systemd-boot handle multiple initrds. However, even if they didn't support multiple initrds directly, it should always be possible to generate a combined image by concatenating the images.

#### Device Trees and Device Tree Overlays
Are device trees and device tree overlays supported at all by bootspec? Or is bootspec limited to typical x86 systems, not supporting device trees for now?

#### `machine-id`
Should bootspec provide a field containing the `machine-id`?
Currently, systemd-boot reads this at runtime from `/etc/machine-id`. When generating type #1 entries, `machine-id` is supposed to uniquely identify the OS installation.
I guess the underlying question is whether the generations are tied to a specific installation instance (`machine-id`). Currently this is not the case, but hypothetically NixOS could somehow generate the machine ID as part of the generation, configured by the user (comparable to `networking.hostId` used by zfs).
For example, if I were to "re-install" NixOS by wiping everything but `/nix` (thus keeping the system profiles in `/nix/var/nix/profiles`), generating a new `machine-id` in the process, should the generated bootloader entries consider the entries for the old generations (with the old machine ID) as distinct to the entries for the new generations (with the new machine ID)?

#### Discovery of Bootspec Entries
While the RFC formalizes the discovery of specializations, I am assuming the discovery of installable generations itself is left unspecified as out of scope?
I am referring to the following process in the systemd-boot backend (Python pseudocode), which I'm assuming is similar for the other backends:
```python
default_profile = Profile("/nix/var/nix/profiles/system")
# Profiles generated by nixos-rebuild --profile-name <name>
named_profiles = [ Profile(named_path) for named_path in directories("/nix/var/nix/profiles/system-profiles") ]
profiles = [ default_profile ] + named_profiles
generations = []
for profile in profiles:
    generations.append(find_directories(profile.path + "[0-9]+-link"))
```

#### Semantics of `initrdSecrets`
Would it make sense to specify the basic semantics and interface of the `initrdSecrets` field?
While the other fields specify well-known paths and parameters, the `append-initrd-secrets` tool is really only used inside NixOS, so the semantics are currently only defined within the NixOS repository.
For example:
"`initrdSecrets` specifies the path to an executable that will be called for each installed generation. The path to the installed initrd [or perhaps a temporary path?] is passed as an argument to the executable, and the executable may modify the initrd as required. If the `initrdSecrets` field is defined, execution of the tool is mandatory for the backend. If the process finishes with a nonzero exit status, the backend has to consider the generation uninstallable and abort."
By the way (probably off topic to this RFC), it might also be possible to replace the `initrdSecrets` entirely mechanism with simply generating another, additional initrd image for the secrets (if `initrd` is a list, see above). That is more or less what `append-initrd-secrets` is doing, anyway - the script simply appends another CPIO image to the end of the initrd, which should be the same as specifying multiple initrds.

hesiod:
It turns out that the only other bootloader other than systemd-boot that supports initrdSecrets, GRUB, already uses a multiple initrd approach ([relevant part of `install-grub.py`](https://github.com/NixOS/nixpkgs/blob/723a8c9eac09c976b55dbad19d7dff0cc2f1035b/nixos/modules/system/boot/loader/grub/install-grub.pl#L451-L479)) - instead of appending to the existing initrd, it passes a temporary file to `append-initrd-secrets`.
However, I failed to consider that a bootspec file can't reference a non-store file (that would only be generated during activation) in the hypothetical `initrd` list.
With this in mind, I'm wondering if it might make sense to generalize the `initrdSecrets` field a bit to something like `initrdExtensionGenerator` to reflect the meaning of the field more accurately: An executable that may add arbitrary additional initrd images.

Yet another option would be to somehow move the initrd secrets logic entirely to the bootspec domain: instead of generating an `append-initrd-secrets` script per generation that tries to copy the secrets as specified in the generation's configuration, the required secret paths could be specified entirely within bootspec. The bootloader backend could then handle generation of the secrets image. After all, the [existing script definition](https://github.com/NixOS/nixpkgs/blob/4a6cd14f37b22c9d2f90e3774b8ea9601025ad6b/nixos/modules/system/boot/stage-1.nix#L403-L450) is quite simple and does not really contain additional logic beyond copying the files (except the compressor arguments which aren't terribly important for the secrets image).

In summary, as you can probably tell from my comments on `initrdSecrets`, I'm not quite fond of the current solution in general. Of course, when strictly considering bootspec as specifying the status quo, the specification is totally fine. Perhaps an improved version of `initrdSecrets` can be found for a version 2 of bootspec. OTOH, it would be nice if bootspec could help along the way of solving some issues with initrdSecrets, e.g. see https://github.com/NixOS/nixpkgs/issues/85563 and https://github.com/NixOS/nixpkgs/issues/85000

Majiir:
> #### Device Trees and Device Tree Overlays

I have an ARM machine running NixOS. It has an older version of UBoot that doesn't support `extlinux.conf`, so I'm using `generations-dir`. To assemble the kernel image, I have a script using `ubootTools` like this:

```
cat ${config.system.build.kernel}/zImage ${config.hardware.deviceTree.package}/* > zImage
mkimage \
  # ...some options omitted
  -d zImage \
  $out
```

Meanwhile, [`generic-extlinux-compatible` looks at the system closure](https://github.com/NixOS/nixpkgs/blob/16a907b/nixos/modules/system/boot/loader/generic-extlinux-compatible/extlinux-conf-builder.sh#L80-L83) and adds `FDT`/`FDTDIR` directives.

So yeah, device trees in bootspec would be great! 

> #### `machine-id`
> 
> [...] Currently this is not the case, but hypothetically NixOS could somehow generate the machine ID as part of the generation, configured by the user (comparable to `networking.hostId` used by zfs). For example, if I were to "re-install" NixOS by wiping everything but `/nix` (thus keeping the system profiles in `/nix/var/nix/profiles`), generating a new `machine-id` in the process, should the generated bootloader entries consider the entries for the old generations (with the old machine ID) as distinct to the entries for the new generations (with the new machine ID)?

The same system closure can be deployed to multiple machines in principle, so including `machine-id` in bootspec might not be appropriate. (Then again, the same goes for `networking.hostId`.) Any change here should also account for scenarios where the system closures for multiple machines are built from a common host.

lheckemann:
Thanks for these additional points, @hesiod!

@RaitoBezarius and I were able to meet up and have a chat about these points, as well as continuing our previous discussion from the Matrix room.

Here's a summary of what we talked about (many digressions omitted):

#### Device trees

These are definitely relevant, but we would like to get an initial version of bootspec finalised; handling device trees is neither trivial nor necessary for bootspec to be useful. In the meantime, the extension field allows experimenting with ways to handle them, so we can work out how to do this in a future revision of bootspec.

#### Multiple initrds

Similar to device trees, this is definitely relevant and we will need them sooner or later, but we think they're not a hard requirement for getting bootspec v1 done and useful.

#### `machine-id`

(as @majiir also wrote -- their comment wasn't there at the time!)

We consider that a NixOS toplevel isn't necessarily tied to a specific installation: the same closure may be booted on different machines, and thus shouldn't specify the machine ID. Since bootspec is about representing the bootable artifacts of a NixOS toplevel, the machine ID would be out of scope.

I can see a use case for bundling a machine-id in a NixOS configuration, but I think this is best addressed by custom config (outside the scope of this RFC) --- maybe even setting `environment.etc.machine-id.text` would be enough.

#### Semantics of initrd secrets

We definitely need to address this before we can accept this RFC.

Idea: maybe we should have a specification of the secrets, instead of a program to run? e.g. 
```json
{
  "/etc/ssh/ssh_host_ed25519_key": {
    "path": "/etc/ssh/initramfs_host_ed25519_key",
    "mode": "0600"
  }
}
```

If it's a program as it currently is, cross is harder (how would we `nixos-install` on x86 to set up an ARM device?).
Programs are more versatile (could fetch secrets from a server, etc?), but if a secret is going to be written to an initramfs on the boot partition anyway, it's probably okay for it to be on the filesystem.
Long-term, the right solution here is probably to use systemd credential images (systemd sysexts), for which multiple initrd support is extremely relevant, but we think the best way forwards is to cover functionality included in NixOS as it is now, to build a foundation for future work.
The disadvantage of this change is that it would require further engineering, possibly delaying us further; maybe that should be pushed back to v2 as well.
If we do decide not to do anything of the sort, we should specify the current behaviour: it's the path to a program which should be invoked with the path to a writable initrd as its only argument, and which adds files from outside the Nix store to said initrd.

#### Where do extensions go?

We had previously talked about where the extension field should actually be (on the Matrix channel). The conclusion we drew yesterday was that:
- Extensions should be top-level keys
- Specialisations should be an extension which we define along with bootspec itself; each specialisation can contain the same keys as the top level, so extensions can work
- Maybe we should rename `v1` to `org.nixos.bootspec.v1` to make the expectation of namespacing clear.


#### What a bootspec might look like with the suggested changes

(inventing a "menu icon" extension for demonstration purposes)
```json
{
  "org.nixos.bootspec.v1": {
    "init": "/nix/store/azzngrynz4wb71p9s799iymzjy4y24gh-nixos-system-nixos-23.05pre-git/init",
    "initrd": "/nix/store/ssbrk7ssa9bkdvjlh275wq80y6ihwv3b-initrd-linux-5.15.96/initrd",
    "initrdSecrets": {
      "/etc/ssh/ssh_host_ed25519_key": {
        "mode": "0600",
        "path": "/etc/ssh/initramfs_host_ed25519_key"
      }
    },
    "kernel": "/nix/store/crmgbd4a6ysp8avnh7mk8mcf0vjfndgj-linux-5.15.96/bzImage",
    "kernelParams": [
      "loglevel=4"
    ],
    "label": "NixOS Stoat 23.05pre-git (Linux 5.15.96)",
    "system": "x86_64-linux",
    "toplevel": "/nix/store/azzngrynz4wb71p9s799iymzjy4y24gh-nixos-system-nixos-23.05pre-git"
  },
  "org.nixos.specialisation": {
    "work": {
      "org.nixos.bootspec.v1": {
        "init": "/nix/store/vpa1xzbqal7z56v6b5yc53w5v1skia71-nixos-system-nixos-23.05pre-git/init",
        "initrd": "/nix/store/ssbrk7ssa9bkdvjlh275wq80y6ihwv3b-initrd-linux-5.15.96/initrd",
        "initrdSecrets": {
          "/etc/ssh/ssh_host_ed25519_key": {
            "mode": "0600",
            "path": "/etc/ssh/initramfs_host_ed25519_key"
          }
        },
        "kernel": "/nix/store/crmgbd4a6ysp8avnh7mk8mcf0vjfndgj-linux-5.15.96/bzImage",
        "kernelParams": [
          "loglevel=4"
        ],
        "label": "NixOS Stoat 23.05pre-git (Linux 5.15.96)",
        "system": "x86_64-linux",
        "toplevel": "/nix/store/vpa1xzbqal7z56v6b5yc53w5v1skia71-nixos-system-nixos-23.05pre-git"
      },
      "xyz.lahfa.ryan.lanzaboote.v1": {
        "menu-icon": "/nix/store/j13pl3ychwf6m5j1pym9p9bbp8s5ibkz-logo.png"
      }
    }
  },
  "xyz.lahfa.ryan.lanzaboote": {
    "menu-icon": "/nix/store/1gdsdm2wa0sajbc5z856dbab49j953bp-logo.png"
  }
}
```

Of course, none of this is final, and it only represents what @RaitoBezarius and I discussed and considered to be the best approach. Feedback welcome!

RaitoBezarius:
> #### Semantics of initrd secrets
> ...
> Long-term, the right solution here is probably to use systemd credential images (systemd sysexts), for which multiple initrd support is extremely relevant, but we think the best way forwards is to cover functionality included in NixOS as it is now, to build a foundation for future work. The disadvantage of this change is that it would require further engineering, possibly delaying us further; maybe that should be pushed back to v2 as well. If we do decide not to do anything of the sort, we should specify the current behaviour: it's the path to a program which should be invoked with the path to a writable initrd as its only argument, and which adds files from outside the Nix store to said initrd.

I am down for the engineering work on this if there is a "fast enough" consensus on it (i.e. giving us time to wrap up this RFC for 23.05).

cc @nikstur because I know you hate the current semantics of initrd secrets. :)

RaitoBezarius:
Can I ask remaining shepherds to express their (dis)agreement with regards to latest changes? @06kellyjac @GovanifY

Also, @nikstur expressed he wanted to get rid of the `label` field because it was not really meaningful, to elaborate on this matter, consider Lanzaboote context where we generate boot menu entries using those labels, it seems like having a `nixosVersion` and `kernelVersion` field to build the `label` dynamically through `/etc/os-release` for example, is a better option.

(If you want to elaborate more on this, please feel free :P.)

7c6f434c:
> it seems like having a nixosVersion and kernelVersion field to build the label dynamically through /etc/os-release for example, is a better option.

What is nice about `label`, it is forking-agnostic. A generation constructs the label however is desired, so the non-NixOS Nix-based systems can easily adopt the format.

RaitoBezarius:
> > it seems like having a nixosVersion and kernelVersion field to build the label dynamically through /etc/os-release for example, is a better option.
> 
> What is nice about `label`, it is forking-agnostic. A generation constructs the label however is desired, so the non-NixOS Nix-based systems can easily adopt the format.

I agree but as we do not have yet a non-NixOS Nix-based system using bootspec, it is hard to understand and cater properly for their needs and the extensions field is there for fully experimenting for any need that can be ratified in next revisions of bootspec.

7c6f434c:
> I agree but as we do not have yet a non-NixOS Nix-based system using bootspec, it is hard to understand and cater properly for their needs and the extensions field is there for fully experimenting for any need that can be ratified in next revisions of bootspec.

The point is not to try guessing external needs, it is to avoid overfitting on NixOS where it is _obviously_ NixOS specific and also clearly doable while the bootspec JSON is being generated ayway.

It's not like we have a NixOS release using bootspec either.

RaitoBezarius:
> > I agree but as we do not have yet a non-NixOS Nix-based system using bootspec, it is hard to understand and cater properly for their needs and the extensions field is there for fully experimenting for any need that can be ratified in next revisions of bootspec.
> 
> The point is not to try guessing external needs, it is to avoid overfitting on NixOS where it is _obviously_ NixOS specific and also clearly doable while the bootspec JSON is being generated ayway.

Reading a kernel version or some "OS" version is probably something that is not too much NixOS-specific, but the current way forces you to assume that the toplevel contains the data you need `$toplevel/nixos-version` or reading the link of `$toplevel/kernel` and parsing the path.

Whereas `label` is completely opaque, I would rather have it as a extra optional field called `pretty_name` but I'm not sure it provides any real value for bootspec "developers".

> It's not like we have a NixOS release using bootspec either.

Yes, but bootspec has been running on unstable since last release and experience has been accumulated on its actual usage.

If this RFC gets merged, it is highly likely (I won't jinx it promise) that bootspec will be part of 23.05.

7c6f434c:
Looking at the example, I see nothing that cannot be simply generated at the build time and put into `$toplevel/…/bootspec.v1.json`. Sure, not now, but later this could become the preferred option with the current approach used as a fallback.

What's the point for the boot menu generator to have visibility into the semantics of the label? The label is an opaque way in which the generation prefers to describe itself, yes.

RaitoBezarius:
> Looking at the example, I see nothing that cannot be simply generated at the build time and put into `$toplevel/…/bootspec.v1.json`. Sure, not now, but later this could become the preferred option with the current approach used as a fallback.
> 
> What's the point for the boot menu generator to have visibility into the semantics of the label? The label is an opaque way in which the generation prefers to describe itself, yes.

Well, for example, for sorting purposes. :)
See: https://github.com/systemd/systemd/blob/main/src/fundamental/bootspec-fundamental.c for a good example of the constraints we have to operate with.

7c6f434c:
Well, PRETTY_NAME = label is the first choice…

_Sorting_ is based more on generation number / generation symlink timestamp, no? (Timestamps will be useful in the future when you support mixing multiple independent profiles with their own generation sequences…)

RaitoBezarius:
> Well, PRETTY_NAME = label is the first choice…

For selecting a potential good name, yes. But then, after,

https://github.com/systemd/systemd/blob/main/src/boot/efi/boot.c#L1792-L1849

And we actually see problems in lanzaboote with these methods (duplicates kernel information in the title, etc, etc.)

> _Sorting_ is based more on generation number / generation symlink timestamp, no? (Timestamps will be useful in the future when you support mixing multiple independent profiles with their own generation sequences…)

Semantically, generation number is not really an IMAGE_ID I think?

If I pull out lanzaboote's EFI:
```
00018600: 4944 3d6c 616e 7a61 0a50 5245 5454 595f  ID=lanza.PRETTY_
00018610: 4e41 4d45 3d4e 6978 4f53 2053 746f 6174  NAME=NixOS Stoat
00018620: 2032 332e 3035 7072 652d 6769 7420 284c   23.05pre-git (L
00018630: 696e 7578 2036 2e31 2e31 3629 0a56 4552  inux 6.1.16).VER
00018640: 5349 4f4e 5f49 443d 4765 6e65 7261 7469  SION_ID=Generati
00018650: 6f6e 2039 392c 2042 7569 6c74 206f 6e20  on 99, Built on 
00018660: 3139 3730 2d30 312d 3031 0a00 0000 0000  1970-01-01......
```

---

```
ID=lanza
PRETTY_NAME=NixOS Stoat 23.05pre-git (Linux 6.1.16) (the bootspec label)
VERSION_ID=Generation 99, Built on 1970-01-01
```

I don't feel like this is optimal for running a proper menu yet, but this might be also related to systemd-boot.
Anyway, my point is simply to say that having this data as structured is probably interesting because of some difficulties we currently are running into and @nikstur has some documentation (I only reported them and I know others lanzaboote users ran into them).

Currently, we are only supporting systemd-boot, I am interested into having more bootloaders on-board, we have a (WIP) GRUB PR in-flight, we would like to have extlinux at some point too, and I am afraid they will come also with their own challenges, under SecureBoot contexts.

7c6f434c:
Well, I guess if you change the current spec towards a split that is fine enough to map easily to the notions of the bootloaders supported in NixOS (regardless of current bootspec support), it also makes sense. I guess it's nicer if the names are tied to the semantics on the botloader side, not to NixOS.

RaitoBezarius:
I would like to invite to FCP letting aside label-related changes and considering only the final suggested changes: moving layout structure of the file & semantics of initrd secrets as per https://github.com/NixOS/rfcs/pull/125#issuecomment-1455052279.

lheckemann:
Sorry, we forgot to respond to one of your points @hesiod:
>Discovery of Bootspec Entries

>While the RFC formalizes the discovery of specializations, I am assuming the discovery of installable generations itself is left unspecified as out of scope?

Yes -- this is what we decided in the 2022-08-18 meeting. Let's put that in the RFC text for clarity.

RaitoBezarius:
Discovered a potential problem in Cuelang (https://github.com/cue-lang/cue/issues/2299) which prevent us from validating specialisations using the new format, I suppose.
I don't think it should be a blocker for the FCP IMHO.

06kellyjac (approved):

amjoseph-nixpkgs (approved):
This is really convenient for `kexec`-based bootloaders (petitboot, ownerboot, etc).

> So yeah, device trees in bootspec would be great!
:+1: 
> we can work out how to do this in a future revision of bootspec.
:+1:

GovanifY (approved):

lheckemann (approved):
Moving to FCP, with disposition to accept!

edolstra:
Do I understand correctly that this makes NixOS depend on Cue? Since Cue is essentially a competing configuration language, making NixOS depend on that might send an unintended message (namely that Nix or the NixOS module system are inadequate). What do we need Cue for that we can't do with the NixOS module system?

K900:
I'd argue that not using Nix for validation is actually the right call here, because it makes it easier for other systems to adopt bootspec should they want to (and I believe some people were vaguely interested in doing that before, at the very least from the Guix side).

RaitoBezarius:
> Do I understand correctly that this makes NixOS depend on Cue? Since Cue is essentially a competing configuration language, making NixOS depend on that might send an unintended message (namely that Nix or the NixOS module system are inadequate). What do we need Cue for that we can't do with the NixOS module system?

We only use it for schema validation, CUE is completely optional for this specification, you can bring your own schema validator if you need to.

See manual on it: https://nixos.org/manual/nixpkgs/unstable/#id-1.5.8.11.5

RaitoBezarius:
Also, we cannot validate using the NixOS module because we are writing parts of the file in the derivation builder script. Therefore, we would need to run the NixOS module verification at that moment, which is painful for no reason.

RaitoBezarius:
> > We only use it for schema validation, CUE is completely optional for this specification, you can bring your own schema validator if you need to.
> 
> Why does the validator need to be executed on activation? This would imply a (afaik) new dependency on Go for every NixOS system.

We did not want to ship broken configurations, but you are totally right, this is a problem.
We are more in implementation land, but, would an option to enforce validation be acceptable to you?

pasqui23:
Can't we use [JSON schemas](http://json-schema.org/) for validation or is this more complex?

RaitoBezarius:
It is more complicated than JSON schemas enable to express.

edolstra:
>  We are more in implementation land, but, would an option to enforce validation be acceptable to you, disabled by default?

Yeah that sounds good. I didn't even consider the build-time dependency on Go, but that's something to be avoided in the base system if at all possible.

PR #126, opened 2022-05-26
[RFC 0126] Standardize special properties

schuelermine:
[Rendered](https://github.com/schuelermine/nix-rfcs/blob/rfc-0126/rfcs/0126-standardize-special-properties.md)

### DIFF
--- a/rfcs/0126-standardize-special-properties.md
+++ b/rfcs/0126-standardize-special-properties.md
@@ -1,0 +1,81 @@
+---
+feature: standardize-special-properties
+start-date: 2022-05-26
+author: Anselm Schüler
+co-authors: None
+shepherd-team: None
+shepherd-leader: None
+related-issues: None
+---
+
+# Summary
+[summary]: #summary
+
+Rename attrset attribute names treated specially by Nix to start with a double underscore.
+Add a builtin function `guardSpecial` that returns `null` for double underscore prefixed strings.
+
+# Motivation
+[motivation]: #motivation
+
+Nix treats many attributes specially with unique behaviours. Some of these are prefixed with double underscores to disambiguate them from everything else. But some aren’t: `recurseForDerivations` isn’t, `type`, when set to `"derivation"`, changes visual output.
+
+It would also be a good idea to avoid accidentally using reserved special attribute names. Therefore I suggest `guardSpecial` that can be used in dynamic attribute names to avoid this.
+
+# Detailed design
+[design]: #detailed-design
+
+Properties treated specially by the Nix tools and Nix language are renamed to start with a double underscore. Instances where this is not necessary are `__functor` and `__toString`. `recurseForDerivations` is renamed `__recurseForDerivations` and `type` is renamed `__type`.
+
+A new builtin function is introduced, `builtins.guardSpecial`.
+If called on a string, it returns the string if it does not start with a double underscore, and `null` otherwise. This can be used by semantically charged non-generic functions that update attrs dynamically to avoid unwanted interactions.
+
+Nixpkgs & Nix need to be updated to reflect this. Therefore, it may be advisable to have a grace period during which unprefixed special properties are still supported, but warn.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+```nix
+rec {
+  set-val-to-null = k: attrs: attrs // {
+    ${guardSpecial k} = null;
+  };
+  x = set-val-to-null "__functor" { __functor = _: _: { }; };
+  y = set-val-to-null "y" { __functor = _: _: { }; };
+}
+```
+evaluates to
+```nix
+{
+  set-val-to-null = «lambda»;
+  x = { __functor = «lambda»; };
+  y = { y = null; __functor = «lambda»; };
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This change is backwards-incompatible
+- This requires judging if certain properties count as special attributes
+
+# Alternatives
+[alternatives]: #alternatives
+
+- `guardSpecial` could be introduced on its own, thereby complicating its implementation
+- These could be left alone
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+I honestly don’t have a full list of these special names, I only listed the ones I know about.
+
+It’s also unclear what exactly counts. I think attrsets passed to functions as de-facto named arguments shouldn’t count (e.g. for `derivation`). You could even argue the outputs of `listToAttrs` should be prefixed!
+
+The name of `guardSpecial` might need some work.
+
+# Future work
+[future]: #future-work
+
+This eases introduction of any future special names.
+
+It can also be assumed the implementation of `guardSpecial` won’t need to be updated if this standard naming convention is adhered to.


rfcs/0126-standardize-magic-properties.md
```diff
@@ -0,0 +1,81 @@
...
+[motivation]: #motivation
+
+Nix treats many attributes specially with unique behaviours. Some of these are prefixed with double underscores to disambiguate them from everything else. But some aren’t: `recurseForDerivations` isn’t, `type`, when set to `"derivation"`, changes visual output.
+
+It would also be a good idea to avoid accidentally using reserved special attribute names. Therefore I suggest `guardMagic` that can be used in dynamic attribute names to avoid this.
+
+# Detailed design
+[design]: #detailed-design
+
+Properties treated specially by the Nix tools and Nix language when dealing with arbitrary are renamed to start with a double underscore. Instances where this is not necessary are `__functor` and `__toString`. `recurseForDerivations` is renamed `__recurseForDerivations` and `type` is renamed `__type`.
```
fricklerhandwerk, line 27:
```suggestion
Properties treated specially by the Nix tools and Nix language when dealing with are arbitrarily renamed to start with a double underscore. Instances where this is not necessary are `__functor` and `__toString`. `recurseForDerivations` is renamed `__recurseForDerivations` and `type` is renamed `__type`.
```

rfcs/0126-standardize-magic-properties.md
```diff
@@ -0,0 +1,81 @@
...
+# Summary
+[summary]: #summary
+
+Rename attrset attribute names treated specially by Nix to start with a double underscore.
+Add a builtin function `guardMagic` that returns `null` for double underscore prefixed strings.
+
+# Motivation
+[motivation]: #motivation
+
+Nix treats many attributes specially with unique behaviours. Some of these are prefixed with double underscores to disambiguate them from everything else. But some aren’t: `recurseForDerivations` isn’t, `type`, when set to `"derivation"`, changes visual output.
```
fricklerhandwerk, line 20: I have a hard time understanding the exact problem and its severity.

Could you please make an explicit list of the magic attributes you are aware of, explain what they are for, why they are magic, and link to relevant documentation and source code?

rfcs/0126-standardize-magic-properties.md
```diff
@@ -0,0 +1,81 @@
...
+  set-val-to-null = «lambda»;
+  x = { __functor = «lambda»; };
+  y = { y = null; __functor = «lambda»; };
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This change is backwards-incompatible
```
Ma27, line 58: I don't think it's such a good idea to make it actually breaking: IIRC one of the benefits Nix advertises is that you can instantiate old Nix expressions and still get the same result (but that also means that the behavior of the expression shouldn't change).

schuelermine: I see, but then this reduces to just `guardMagic`.

rfcs/0126-standardize-magic-properties.md
schuelermine: The problem is that there’s a bunch of specially treated attribute names that mean attrsets aren’t neutral data structures.
Maybe magic is a bad word.
As I say further down, I don’t know a full list, since these seem to be things that aren’t considered together in the documentation.

`type` is magic because if `type = "derivation"`, the data is displayed completely differently in the Nix repl.
By renaming them to start with `__`, I hope to make them more obvious, more segregated from normal attributes, and easier to deal with.

A problem this could fix could arise when a user who does not know how a process works inputs data into it, which is unknowingly processed to create an attribute that unexpectedly alters the result’s behaviour.
None exist ATM AFAIK, but they could.

rfcs/0126-standardize-magic-properties.md
schuelermine: Also, this doesn’t stop old expressions from building, since `__toString` and `__functor` are the only ones here that aren’t user-facing, and they’re not renamed.

Ma27: When grepping for `recurseForDerivations` in `nixpkgs`, there's a bunch of places where it's explicitly used, so what do you mean with "user-facing"?

schuelermine: Their special effect is outside the expression language. AFAIK `recurseForDerivations` tells nix commands to search for derivations deeper down, but it’s only being set in Nixpkgs, not used by it. Renaming it without renaming it in nixpkgs or using an old nix expression will not affect directly building any particular derivation, but only auxiliary things like searching. It’s unlikely that many nix expression directly check `recurseForDerivations`.
The other ones, on the other hand, if changed, would actually stop some Nix expressions from evaluating. `__functor` concerns the in-expression behaviour of these values.

7c6f434c: Technically speaking, _adding_ synonyms that start with `__` for all the magic attributes and deciding that Nixpkgs will use the underscore-based things from now on is more than `guardMagic`.

Evaluation-affecting magic seems to be indeed underscored; `nix` command behaviour is nominally still experimental so some changes are possible with proper warning.

Ma27: > Evaluation-affecting magic seems to be indeed underscored; nix command behaviour is nominally still experimental so some changes are possible with proper warning.

But doesn't that affect nix-* commands usually as well?

bew:
> I have updated this PR to say "special" instead of "magic", I think that’s better.

Please also change the 'Rendered' link in first post (:

rfcs/0126-standardize-magic-properties.md
vcunat: Attrsets by themselves are (almost?) neutral, I believe, but some functions consuming them treat some attributes specially.

schuelermine: With the exception of `__functor`. And precisely these, in my opinion, should be dunder(__)-prefixed

rfcs/0126-standardize-magic-properties.md
schuelermine: It does. In my opinion that’s not as much a breaking change as a change in how some *derivations* build.

rfcs/0126-standardize-magic-properties.md
pennae: > type is magic because if type = "derivation", the data is displayed completely differently in the Nix repl.
By renaming them to start with __, I hope to make them more obvious, more segregated from normal attributes, and easier to deal with.

[it's even more magic than that](https://github.com/NixOS/nix/blob/a72e75e6b6db5506f0932d8f1d507cdf3a466130/src/libexpr/eval.cc#L2207). changing `type` is a flag day event, and for that reason alone probably not worth it. also note that `recurseForDerivations` isn't (afawk( honored by the new nix commands at all either.

schuelermine: That’s a good point. Looking at the code though it seems not so hard to change. Don’t the new commands try to build non-derivation store path results anyways? But yeah, this would be a big change, probably not worth it. Also, AFAIK nix-command honors recurseForDerivations in legacyPackages *only*.

edolstra:

rfcs/0126-standardize-special-properties.md
```diff
@@ -0,0 +1,81 @@
...
+[motivation]: #motivation
+
+Nix treats many attributes specially with unique behaviours. Some of these are prefixed with double underscores to disambiguate them from everything else. But some aren’t: `recurseForDerivations` isn’t, `type`, when set to `"derivation"`, changes visual output.
+
+It would also be a good idea to avoid accidentally using reserved special attribute names. Therefore I suggest `guardSpecial` that can be used in dynamic attribute names to avoid this.
+
+# Detailed design
+[design]: #detailed-design
+
+Properties treated specially by the Nix tools and Nix language are renamed to start with a double underscore. Instances where this is not necessary are `__functor` and `__toString`. `recurseForDerivations` is renamed `__recurseForDerivations` and `type` is renamed `__type`.
```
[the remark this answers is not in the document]
schuelermine: Wouldn’t that have to be a completely different RFC?

Note that I’m open to closing this for good, it’s a much bigger change than I had anticipated when I wrote it up.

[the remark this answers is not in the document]
pennae: to expand on the "any evidence of this causing a problem" bit: *changing* attribute names would be incredibly hard in the very dynamic language nix is. we can find all instances of e.g. `drvPath` in code, but which of these are used to construct/modify derivations and which aren't? `meta` is to some extent a special name, changing that would touch almost the entire nixpkgs tree (and introduce mismatches between derivations and modules). changing old names feels like asking for a disaster, and silently dropping assignments can also easily turn into a footgun. :confused:

which is not to say that guarding against accidental use of special names is bad! such a guard should (usually) fail very loudly though to catch mistakes early, behaving much like `assert ! elem name [ «specials» ] && match "__.*" name == null; name`. prefixing new special names also sounds very reasonable.

edolstra:
@fricklerhandwerk @roberth @Ma27 @pennae Are any of you interested in shepherding this RFC?

fricklerhandwerk:
@edolstra No. I‘m not qualified, both in terms of `nixpkgs` and RFC expertise, and also don’t want to dilute my focus on documentation work.

schuelermine:
A note: I am no longer necessarily in favor of this RFC, even though I authored it. . The comments have convinced me this is too big an undertaking. Maybe it should be closed? I’m sorry for cluttering up the RFC process in that case.

schuelermine:
I’m sorry, I pressed the wrong button on accident.

pennae:
@edolstra no, don't feel up to the task (and don't think the scope outlined in the RFC is good at all, if anything it should be cut down to "new attributes with special interpretation by nix should have a `__` prefix", which is common practice already)

roberth:
@edolstra The original idea behind this RFC doesn't work out in cost/benefit. I wouldn't mind doing a bit more guidance, but I don't think we need the full RFC process to carry this to a satisfactory conclusion.

@schuelermine Functions like you describe could be added to the Nixpkgs `lib` without controversy, when they don't impose your original breaking changes.
In my opinion, for this outcome, it's better _not_ to go through RFCs, because such an RFC would only be useful for describing the status quo, and it may go out of date when the implementation of the functions is improved. (For documentation, we have the manual)

lheckemann:
@schuelermine if you feel that there's anything useful this RFC could be rewritten to do instead, feel free; otherwise you're also welcome to withdraw it (just close it) :)

PR #131, opened 2022-07-26
[RFC 0131] Enforce the usage of SRI hashes in Nixpkgs

winterqt:


rfcs/0131-sri-hashes.md
```diff
@@ -0,0 +1,48 @@
...
+
+# Summary
+[summary]: #summary
+
+Use SRI hashes and the `hash` parameter for fetchers.
+
+# Motivation
+[motivation]: #motivation
+
+Currently in Nixpkgs, the hash formats that are used in parameters to fetchers range from SRI and base-32 (the most common ones) to hex. In addition to this, packages use the algorithm-specific arguments (`sha256`, `sha512`, etc.) or the generic `hash` argument at random. This creates inconsistencies that we'd rather not have.
```
8573, line 19: What is SRI in this context?

winterqt: [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity), hashes look like `sha256-2mE26ES+fYSWdfMr8uTsX2VVGTNMDQ9MXEk5E/L95UI=`. Maybe I should add at least a definition of SRI in the RFC?

rfcs/0131-sri-hashes.md
```diff
@@ -0,0 +1,48 @@
...
+# Detailed design
+[design]: #detailed-design
+
+1. Require all new packages to use SRI hashes + `hash` when possible
+2. Do a treewide migration for the majority of cases (single fetcher in `src`, using a fetcher that supports `hash` and SRI hashes (`fetchFromGitHub`, `fetchurl`, etc.)) to `hash` + SRI hashes
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Some fetchers still don't support SRI hashes (i.e. [`fetchgit`](https://github.com/NixOS/nixpkgs/pull/79987), which is blocked due to [this discussion](https://github.com/NixOS/nixpkgs/pull/79987#discussion_r378735698) not being resolved)
```
jtojnar, line 30: `fetchgit` actually supports it since https://github.com/NixOS/nixpkgs/pull/144693, there might be other fetchers that donʼt but I do not think it is an issue with any of the common ones. My problem with is that Eelco indicated `hash` is not a good fit and Nix builtins use different attribute name.

winterqt: I've removed the `fetchgit` drawback and have added the `hash` argument name to unresolved questions for further discussion.

rfcs/0131-sri-hashes.md
```diff
@@ -0,0 +1,50 @@
...
+# Detailed design
+[design]: #detailed-design
+
+1. Require all new packages to use SRI hashes + `hash` when possible
+2. Do a treewide migration for the majority of cases (single fetcher in `src`, using a fetcher that supports `hash` and SRI hashes (`fetchFromGitHub`, `fetchurl`, etc.)) to `hash` + SRI hashes
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None at this time.
```
sternenseemann, line 30: There's (anecdotically) still a significant number of Nix 2.3 users who contributing to nixpkgs would be made unnecessarily harder.

rfcs/0131-sri-hashes.md
sternenseemann: > This creates inconsistencies that we'd rather not have.

Why? This is not a motivation, just a claim.

Using the different hash encodings genuinely has value, for example I frequently just copy over the base16 hash from a `SHA256SUMS` file provided by upstream, requiring less button presses, round trips and most importantly allows others to quickly and easily match the hashes textually when reviewing.

ocharles: I think a link to what SRI is in the RFC would be good to have. I also didn't know what SRI was.

winterqt: > Using the different hash encodings genuinely has value...

I agree with this point, but at the same time most people use either the SRI hashes given by Nix's hash mismatch errors, or convert those to base-32.

sternenseemann: And this is the nice thing about the current system: You can just put in whatever your tool spits out into a fetcher and it'll be able to fetch and verify the result – it's not like hashes contain any information that's more comprehensible for humans in one format or the other…

winterqt: Sure, but most packages use whatever Nix gives them, which also leads to things like a lot of redundant usage of the `sha256` argument (when `hash` can/should be used):
```console
$ rg --files-with-matches "sha256 = \"sha256\-" | wc -l
    7388
```

sternenseemann: I don't see a problem with that, `hash = "sha256-…` is hardly better than `sha256 = "sha256-…`.

rfcs/0131-sri-hashes.md
AndersonTorres: Oh no, another ten years of deprecation window...

rfcs/0131-sri-hashes.md
AndersonTorres: "Less button presses"?
In Emacs it is just `shell-command` with `nix hash to-sri sha256:<...>` followed with `set-mark-command` and `kill-ring-save` the SRI output. With completions and everything else to help.
Seriously, less button presses is hardly a serious reason when we have shell and editor completions all around today.

And an unified name like `hash` is better, because, well, new and better hash technologies appear everywhere. E.g. NetBSD pkgsrc now uses BLAKE hashes as default.
If we choose to support it, or any other hash algorithm for that matter, it would be way easier to reuse `hash = blake-XXXXX...` instead of creating a new `blake = XXXXX...`.

rfcs/0131-sri-hashes.md
vojta001: > Why? This is not a motivation, just a claim.

One of the reasons might be `sha*` are semantically mutually exclusive, but not syntactically – you can specify both `sha256` and `sha512` at the same time by accident (pretty easy when using an override), but you cannot specify multiple different `hash` attributes.

rfcs/0131-sri-hashes.md
vojta001: >Why? This is not a motivation, just a claim.

One of the reasons might be sha* are semantically mutually exclusive, but not syntactically – you can specify both sha256 and sha512 at the same time by accident (pretty easy when using an override), but you cannot specify multiple different hash attributes.

rfcs/0131-sri-hashes.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- How can we do the treewide migration efficently?
+  - Text replacement for the majority of cases, with an AST-based solution for others?
+- Do we want to eventually deprecate the use of `sha256`, `sha512`, etc. in fetchers that support SRI hashes, leading to their eventual removal?
+  - This would require updating every script that uses these arguments, and would cause breakage for any usage of them out-of-tree; we'd need to weigh the benefits of this
+- Is `hash` the correct name to use for the argument?
+  - Nix builtins use `narHash`
```
piegamesde, line 45: I don't care about the name, but built-in fetchers and nixpkgs fetchers should be consistent.

rfcs/0131-sri-hashes.md
```diff
@@ -0,0 +1,50 @@
...
+[alternatives]: #alternatives
+
+- Do nothing: further inconsistencies will continue to be introduced
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- How can we do the treewide migration efficently?
+  - Text replacement for the majority of cases, with an AST-based solution for others?
+- Do we want to eventually deprecate the use of `sha256`, `sha512`, etc. in fetchers that support SRI hashes, leading to their eventual removal?
```
piegamesde, line 42: We can deprecate them without removing them.

piegamesde (changes requested):

rfcs/0131-sri-hashes.md
piegamesde: You joke, but I seriously think this would be a sensible thing to do.

sternenseemann: Overall we have developed a strange culture of removing stuff unnecessarily in nixpkgs, I feel like.

In this case, we shouldn't remove support for these formats from Nix anyways (to continue to be able to evaluate older nixpkgs revisions (the ability to reproduce *ancient* stuff truely is a strength of Nix)) and for existing APIs in nixpkgs it's laughably simple to keep supporting it (you don't need to change anything, in fact!). For all I care, stop supporting the `sha256` and `sha512` attributes for new APIs…

rfcs/0131-sri-hashes.md
sternenseemann: @vojta001 That's why these interfaces tend to have an `assert` or two behind the scenes…

> `nix hash to-sri`

Well first of all, this is actually a very good point: There is *no stable interface* for converting hashes into the SRI format, because `nix-hash(1)` only supports SRI as an input, not an output format.

rfcs/0131-sri-hashes.md
AndersonTorres: The text does not talk about removal, I suppose, but about enforcement. Older expressions can be rendered by newer versions, whereas new expressions should be written in newer versions of Nix.

sternenseemann: Nix 2.3 users ARE relevant because it makes contributing harder, since hashes need to be converted manually.

spacekookie:
Hello everyone, this RFC is now open for nominations. If you think of someone (including yourself) who you think would make a good shepherd on this RFC, please nominate them now

rfcs/0131-sri-hashes.md
milahu: > The text does not talk about removal, I suppose, but about enforcement.

you can only "enforce" your policy by "removing" the alternative

all software has [legacy parts](https://en.wikipedia.org/wiki/Legacy_system) for backward-compatibility.
with nix we will always have `sha256 = "..."`

related:

> the first rule of kernel maintenance:
> [WE DO NOT BREAK USERSPACE!](https://linuxreviews.org/WE_DO_NOT_BREAK_USERSPACE)

rfcs/0131-sri-hashes.md
milahu: > Using the different hash encodings genuinely has value, for example I frequently just copy over the base16 hash from a `SHA256SUMS` file provided by upstream, requiring less button presses, round trips and most importantly allows others to quickly and easily match the hashes textually when reviewing.

no need to convert the hash to base64

SRI hashes only require a prefix:

```nix
{
  sha256 = "1cc9110705165dc09dd09974dd7c0b6709c9351d6b6b1cef5a711055f891dd0f";
  sha256 = "sha256-1cc9110705165dc09dd09974dd7c0b6709c9351d6b6b1cef5a711055f891dd0f";
  hash = "sha256-1cc9110705165dc09dd09974dd7c0b6709c9351d6b6b1cef5a711055f891dd0f";
}
```

base16 hashes are found with
`grep -r -E 'sha256 = "[0-9a-f]{64}";'`

AndersonTorres: > sha256 is de-facto standard in nixpkgs,

MD5 was de facto standard no more than a decade ago.

Also, this was not about "good enough algorithms", but about a better format for the expression - namely, `hash = "..."` instead of `hashAlgorithm = "..."`.

rfcs/0131-sri-hashes.md
AndersonTorres: > Nix 2.3 users ARE relevant because it makes contributing harder, since hashes need to be converted manually.

Is this so hard to the point of being utterly impossible to copy the output of a command?

Okay then: we can relegate it to some form of merge bot.

AndersonTorres: > you can only "enforce" your policy by "removing" the alternative

Please elaborate more than scary quotes.

> with nix we will always have stuff like

Do not conflate nix with nixpkgs.

Anyone can change such hashes to SRI if they wish - both inside and outside Nixpkgs -, and this will not affect Nix 2.3 users.
Indeed no one of the anecdotal users of Nix 2.3 complained about being unable to run the newest KDE Plasma from Nixpkgs because of our new hash format.

> the first rule of kernel maintenance: WE DO NOT BREAK USERSPACE!

Which user space will be broken by this?

vcunat: Removal of support from fetchers would surely break various development environments (outside nixpkgs itself).  But that's undecided in the current proposal.

AndersonTorres: Because of it I am not for removal.
I just think that some things like "why fetchurl accepts hash but fetchgit does not accept" should be resolved and uniformized.

sternenseemann: > Is this so hard to the point of being utterly impossible to copy the output of a command?

It's not *hard*, just annoying – making the already annoying task of updating hashes even more annoying. Also it will surely mean more round trip in reviews when people unaware of this policy submit PRs with the wrong hash format.

I really don't see the benefit of such a change that justifies wasting everyone's time.

AndersonTorres: > making the already annoying task of updating hashes even more annoying

Annoying tasks should be amenable to automation - like the [r-ryantm](https://github.com/r-ryantm) bot that already uses SRI hashes.

> more round trip in reviews when people unaware of this policy

"People unaware of policies" is a common thing around Nixpkgs. Have you ever seen how many new people want to contribute with new packages? This is very annoying to fix their errors that could be easily avoided by reading the documentation.

Wrong `meta` attributes, wrong version strings, convoluted expressions trying to express unusual building instructions...

vojta001: >> "People unaware of policies" is a common thing around Nixpkgs.

> We should limit ourselves to policies that are actually worth enforcing to keep the friction caused by this problem in check.

Or we should make sure policies are very explicitly and clearly written and ideally automatically checked. The state of the art is that every reviewer has his own set of policies that are unknown beforehand and inconsistent. That's in my experience what causes most back-and-forth delay in the process. Although I accept the "per reviewer policies" are usually not about hash format

sternenseemann: Not even, there are a few clearly stated policies stated in CONTRIBUTING.md which are routinely ignored by contributors.

7c6f434c: > Indeed no one of the anecdotal users of Nix 2.3 complained about being unable to run the newest KDE Plasma from Nixpkgs because of our new hash format.

How is this relevant? The claim is that Nix 2.3 can evaluate SRI hashes, but makes updating them annoying; that Nix 2.3 can evaluate them does not refute the claim.

AndersonTorres: > How is this relevant?

To the point someone suggested that it will break the userspace, and, well, no userspace was broken by the new hash format in a long heck of time.

Follow the flow of argument being made.

> The claim is that Nix 2.3 can evaluate SRI hashes, but makes updating them annoying; 

I was not arguing this point, but the "muh userspace breaking" point.

Follow the flow of argument being made.

7c6f434c: This thread started with the claim that requiring SRI hash format makes _contributing_ while using Nix 2.3 harder (because it doesn't default to printing them), not about _evaluating_ Nixpkgs. The userspace-breaking argument was that we won't remove non-SRI hash support from _Nix_ because we need the ability to evaluate old versions of Nixpkgs, as well as out-of-Nixpkgs code, thus enforcement will be a mess.

Nix 2.3 evaluating SRI hashes fine refutes neither objection. _What_ is the flow of argument you want me to follow?

AndersonTorres: Is anyone here arguing for breaking non-SRI hashes?

vcunat: The current formulation sounds much more prescriptive than just "discouraging".  Point (1.) is a hard "require".  For (2.) I'm unsure about the meaning, maybe like excepting only fetchers that technically don't support.

winterqt: I'm open to changing it to "discourage", but... I don't know, if non-SRI hashes keep slipping in anyways because it's not required, what's the point of this? I guess that's why I chose that string wording.

AndersonTorres: The point can be social pressure.
Pointing an authoritative document that "allows, but strongly discourages" old formats is better than endless discussions about "taste preferences".

The idea would be similar to the patched GCC from OpenBSD, that warns the user when the code uses dangerous functions like `strcpy()` or `gets()`. It does not block you from it but warns you.

Also, in my way for finding old format hashes, I have found buggy expressions, from dead projects to plain wrong files. Old hashes hide more than mere "backwards compatibility".

winterqt: So how does something like this sound: we strongly encourage using SRI hashes and `hash` in new packages, if there is no justifiable reason to not use them. Justifiable reasons include, but are not limited to, taking hashes directly from upstream, and using automated tooling that does not use SRI hashes.

I'm not sure about that last reason, but I also want to provide more than one example of a reason.

I'm also not sure about how to handle converting old packages, if at all. Thoughts welcome.

AndersonTorres: Old packages can be converted by anyone interested ion converting them. I have started a preliminar work a long time ago :P

https://github.com/NixOS/nixpkgs/issues/176283

Further I believe this should be decided on a case by case basis. Maybe allowing old Nix tools is acceptable, whereas upstream hashes is not.

winterqt:
What do you mean by this (what "uses"), who would we be looking for?

rfcs/0131-sri-hashes.md
sternenseemann: > warns the user when the code uses dangerous functions

Not sure how apt this comparison is, as there's nothing fundamentally different about a sha256 hash in SRI format or one in base16 format.

> I'm open to changing it to "discourage", but... I don't know, if non-SRI hashes keep slipping in anyways because it's not required, what's the point of this? I guess that's why I chose that strong of a word.

Stale hashes is not something caused by a particular hash format. It happens because expressions bitrot and we'll have stale hashes in SRI format before long. Doing a mass conversion of hashes may turn up stale hashes, but doesn't have to since you would likely just convert them without refetching, since that is much quicker.

If you are concerned about stale hashes, I'd suggest trying to refetch fixed output derivations in nixpkgs with cache disabled. I believe this has been done quite a few times over the years and it should be easy to turn up useful tools for this.

> strongly encourage

Newer Nix versions already default to this, so there is already a lot of encouragement…

AndersonTorres:
Nominating myself

rfcs/0131-sri-hashes.md
AndersonTorres: "A lot of encouragement" that ends with "my version is too old" or "this is mere personal preferences, don't force them on me"? Wow cool.

7c6f434c: The version is probably too old for the format to be convenient to _produce_ but can evaluate the expressions with either hash format. For «mere personal preferences» argument new Nix making use of SRI format easier than non-SRI _is_ an encouragement, as submitters will typically use whatever is the easiest and changing the format is pointless churn so the initial hash format will stay.

7c6f434c:
I nominate @sternenseemann

rfcs/0131-sri-hashes.md
asymmetric: > This creates inconsistencies that we'd rather not have.

@winterqt could you add something more specific here, considering this is the whole motivation for the RFC?

The discussion above has brought up some more points:

- semantic vs syntactical mutual exclusion
- unique attribute name rather than multiple, algo-specific ones
- ...

edolstra:
We still need at least two more shepherds. @sternenseemann Are you willing to help shepherd this RFC?

sternenseemann:
> We still need at least two more shepherds. @sternenseemann Are you willing to help shepherd this RFC?

I won't have time until December sadly.

rfcs/0131-sri-hashes.md
SuperSandro2000, line 45: I don't see much value in renaming hash. A lot of work which we can safe ourselves.

winterqt: See https://github.com/NixOS/nixpkgs/pull/79987#discussion_r378735698 for context, I added it here just to include it since the question was raised (should probably cite that discussion, though).

SuperSandro2000: I think we are not talking about narHashes here, so this should be removed.

vcunat: NAR hashes are necessary for any cases where the result is more than a single file, e.g. `fetchFromGitHub`.

lheckemann:
@sternenseemann December would be fine IMHO, would you be willing to join the shepherd team then?

Alternatively, I nominate @vcunat -- would you be willing to shepherd?

rfcs/0131-sri-hashes.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None at this time.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Do nothing: further inconsistencies will continue to be introduced
```
bobvanderlinden, line 35: I think there are a number of alternatives.

Different rules can be considered. For each of the following formats it could be decided to allow/disallow/preferred:

* `hash = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";`
* `sha256 = "0000000000000000000000000000000000000000000000000000000000000000";`
* `sha256 = "sha256-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";`
* `sha256 = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=";`

I've gathered there are a number of ways these rules can be 'enforced':

* Throw evaluation errors upon not conforming to rule.
* Have a it be a potential future linter rule.
* Have PR reviewers nudge contributors to follow the rule
* Do nothing: keep things like the are right now. Reviewers will suggest their own opinion in PRs if they'd like.
* Allow any format: do not allow PR reviewers to suggest changing the format in PRs to their preferred format. If the format is allowed, just let it be. This avoids PRs being always potentially invalid due to reviewers having different opinions.

bobvanderlinden: > Allow anything: Do not allow PR reviewers to suggest changing the format in PRs. If the format is allowed, just let it be. Avoid conflicting reviewer opinions.

I'm mentioning this option, as at the moment PRs are being reviewed where the reviewers suggest their own 'opinion' on the matter; even though there is no real rule. One outcome of this RFC could be to just let it be.

vcunat:
Honestly, I see this discussion as rather hard to progress, so how would I shepherd it?  It seems mainly up to opinion.

I only have one compromise idea of allowing non-base64 in the SRI format.  That way we could unify to the `hash` attribute while keeping the ability to copy&paste hashes from upstreams.  PoC: https://github.com/NixOS/nixpkgs/pull/199660

rfcs/0131-sri-hashes.md
vcunat: @milahu: conversion would be needed because SRI is _currently_ only accepting base64.  However, it would be relatively easy to change: https://github.com/NixOS/nixpkgs/pull/199660

bobvanderlinden:
Maybe have the different options up to vote. Then @winterqt can see whether the higher voted options is worth integrating in the RFC?

I guess we can't make an RFC that aligns with all opinions.

7c6f434c:
@vcunat If you think that currently the discussion is just a statement of opinions, you could be a shepherd demanding a consistent standard of justification…

vcunat:
I meant that the final decision will be mainly based on "personal opinions" (I think, it seems rather hard to avoid in this case).

AndersonTorres:
>  mainly based on "personal opinions"

This is the least of problems. for all the "best" (in a purely technical sense) options, we can just throw a random number generator to choose one of them, and God Bless the RNG.

On the other hand, what is the technical advantage over SRI instead of older format? Maybe the files become smaller by a couple of bytes?

vcunat:
Yes.  And integration of the algorithm names, so that we _might_ use the `hash` attribute instead of mixing different ones.  Though for now the effect has been mostly reverse, as `sha256` algorithm is by far the most popular (I think).

AndersonTorres:
> Yes. And integration of the algorithm names, so that we _might_ use the `hash` attribute instead of mixing different ones. Though for now the effect has been mostly reverse, as `sha256` algorithm is by far the most popular (I think).

I am fine about the `hash = <algo>-<cksum>` format. As I have said before, it is future-proof.

The only other bikeshedding problem was if we call it hash or narHash.

rfcs/0131-sri-hashes.md
vcunat: Actually, as we have flat hashes for normal files and NAR hashes over arbitrary trees (or normal files), I'd say it makes sense to call them differently.  At least if I don't consider any backwards compatibility questions for now.  For example, keep `hash` for normal ones and rename NAR hashes to `narHash`.  It might even avoid some confusion, and functions could allow specifying either (though of course the value differs, as something different is hashed).

r-burns: Is this something that could be done under the hood in the fetchers? It seems confusing for new contributors to have to think about whether to use a flat hash or a NAR hash.

winterqt: Yes. We already do this with `outputHashMode`.

tomberek:
Currently nominated shepherds: 
@AndersonTorres (self-nominated), @sternenseemann (delay until December), @vcunat (nominated)

This would be enough if everyone accepted; of course another nomination would also help. If interested, please reach out to understand the role or to nominate anyone else.

sternenseemann:
I can be a sheperd if it's acceptable that the first meeting is after the 13th of December (or I don't attend an earlier one).

edolstra:
It appears that we need one more shepherd. Any takers?

rfcs/0131-sri-hashes.md
```diff
@@ -0,0 +1,50 @@
...
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Use SRI hashes and the `hash` parameter for fetchers.
```
Zimmi48, line 14: "SRI hash" needs a reference link with a definition and globally this RFC needs some more context to be explained.

FWIW, since I became aware that this kind of hash was preferred, I have tried to use them more systematically, but I'm annoyed by `nix-prefetch-url` not generating this type of hash (at least all the way to Nix 2.12) and not knowing how to convert a hash generated by `nix-prefetch-url` into an SRI hash.

vcunat: FYI, for conversion I'm using
```
nix hash to-sri --type sha256 theHash
```

Zimmi48: Thanks! But another thing that is becoming annoying to me lately is to be forced to activate the `nix-command` experimental feature for more and more basic things that I cannot avoid.

Another solution that I've found meanwhile was `openssl dgst -sha256 -binary /path/to/file | openssl base64 -A`, but it's not as convenient.

ShamrockLee: > But another thing that is becoming annoying to me lately is to be forced to activate the `nix-command` experimental feature for more and more basic things that I cannot avoid.

I would rather view it as a missing feature of `nix-hash` that needs to be implemented to get this RFC merged. In other words, we should have

```
nix-hash --sri --type sha256 path/to/file
```

and

```
nix-hash --to-sri --type sha256 theHash
```

tomberek:
From RFC Steering Committee: please nominate yourselves or someone you think would do a good job as a shepherd.

sternenseemann:
> From RFC Steering Committee: please nominate yourselves or someone you think would do a good job as a shepherd.

I think this should be closed due to a lack of interest according to RFC 130 – Search for shepherds started in [August](https://github.com/NixOS/rfcs/pull/131#issuecomment-1225565205) and the two month period has long passed.

winterqt:
> I think this should be closed due to a lack of interest according to RFC 130

I'm somewhat inclined to agree, though I suppose it wouldn't hurt to have a phase where we create a new RFC based on these ideas and reopen it later.

If anyone's interested in doing this, please reach out. I'd make a Matrix room, but that's proving harder than I'd expect... 😅

> Search for shepherds started in [August](https://github.com/NixOS/rfcs/pull/131#issuecomment-1225565205) and the two month period has long passed.

I don't think @vcunat responded to his nomination, though I don't think that would make that much of a difference.

ShamrockLee:
The followings sound more actionable:

1.  Ensure that SRI hashes can be computed without `experimental-features`.
2.  Ensure that all the in-tree fetchers / builders / lang-to-nix expression generators provide hash-algorithm-agnostic attributes (e.g. `hash`) to receive SRI hashes.
3.  Change the Nixpkgs documentation and recommend the use of hash-algorithm-agnostic attribute and SRI hashes.
4.  Upstream corresponding changes to out-of-tree lang-to-nix expression generators.

rfcs/0131-sri-hashes.md
Zimmi48: Thanks! And thanks to you, I also learned that, in the meantime, I can use `nix --extra-experimental-features nix-command file to-sri --type sha256 ...`. I didn't realize that I don't need to activate the experimental feature globally, and this is already better than nothing.

7c6f434c:
@ShamrockLee should there be the part where Nix never outputs a «new» non-SRI hash (new as in not the exact same meaningful characters as some non-SRI hash that was passed in)?

ShamrockLee:
> the part where Nix never outputs a «new» non-SRI hash (new as in not the exact same meaningful characters as some non-SRI hash that was passed in)

Sorry, but which hashes Nix would output are you referring to?

7c6f434c:
>> the part where Nix never outputs a «new» non-SRI hash (new as in not the exact same meaningful characters as some non-SRI hash that was passed in)
>
>Sorry, but which hashes Nix would output are you referring to?

Some package updates in reality are done by bumping the version and breaking the hash, then looking at the Nix error message and copying the «observed» hash.

ShamrockLee:
I see. IIUC, the hash is produced according to the `hashAlgo` of the source derivation set by the fetcher, and the fixes should go to each fetcher. Is that the case?

7c6f434c:
>I see. IIUC, the hash is produced according to the `hashAlgo` of the source derivation set by the fetcher, and the fixes should go to each fetcher. Is that the case?

Prefetchers might differ, but on the Nix level most (all?) hashAlgo's have matching SRI encodings thus the hash-mismatch messages are to be fixed once in Nix.

ShamrockLee:
> Ensure that the hash-mismatch error message produced by the Nix buder always show hashes in SRI format (Thanks @7c6f434c for pointing out this part. See [RFC 0131] Enforce the usage of SRI hashes in Nixpkgs #131 (comment))

Just change the content of the third point.

edolstra:
The RFC steering committee has decided to close this RFC due to a lack of shepherd nominations. However, it can be reopened if people are interested in shepherding.

PR #133, opened 2022-08-28
Integrate Nix with IPFS

Ericson2314:
Integrate Nix with IPFS, in phases of increasing sophistication.
This follows the work done and described in https://github.com/obsidiansystems/ipfs-nix-guide/.

[Rendered](https://github.com/obsidiansystems/nix-rfcs/blob/ipfs/rfcs/0000-ipfs.md)

### DIFF
--- a/rfcs/0000-ipfs.md
+++ b/rfcs/0000-ipfs.md
@@ -1,0 +1,58 @@
+---
+feature: (fill me in with a unique ident, my_awesome_feature)
+start-date: (fill me in with today's date, YYYY-MM-DD)
+author: (name of the main author)
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+One paragraph explanation of the feature.
+
+# Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected
+outcome?
+
+# Detailed design
+[design]: #detailed-design
+
+This is the core, normative part of the RFC. Explain the design in enough
+detail for somebody familiar with the ecosystem to understand, and implement.
+This should get into specifics and corner-cases. Yet, this section should also
+be terse, avoiding redundancy even at the cost of clarity.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This section illustrates the detailed design. This section should clarify all
+confusion the reader has from the previous sections. It is especially important
+to counterbalance the desired terseness of the detailed design; if you feel
+your detailed design is rudely short, consider making this section longer
+instead.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+# Alternatives
+[alternatives]: #alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What parts of the design are still TBD or unknowns?
+
+# Future work
+[future]: #future-work
+
+What future work, if any, would be implied or impacted by this feature
+without being directly part of the work?
--- a/rfcs/0000-ipfs.md
+++ b/rfcs/0000-ipfs.md
@@ -252,7 +252,7 @@ Even if we do end up adopting everything though, we think for the following two 
 
 2. Per the abstract vs concrete model of the nix store in https://github.com/NixOS/nix/pull/6877 , everything we are doing is simply flushing out alternative interpretations of the abstract model.
    This is the sense in which we are "removing the weaknesses and restrictions that make additional features appear necessary" per the Scheme mantra cited above:
-   Instead of extending the model wit new features, we are reflaxing concrete model assumptions (e.g. references are always opaque store paths) while keeping the abstract model the same.
+   Instead of extending the model with new features, we are relaxing concrete model assumptions (e.g. references are always opaque store paths) while keeping the abstract model the same.
 
 3. We also support plans to decouple the layers of Nix further, and update our educational and marketing material to reflect it.
    With Flakes and other post-2.0 features, the upper layer of Nix have gained an enormous amount of flexibility and sophistication.
--- a/rfcs/0000-ipfs.md
+++ b/rfcs/0000-ipfs.md
@@ -1,7 +1,7 @@
 ---
-feature: (fill me in with a unique ident, my_awesome_feature)
+feature: ipfs
 start-date: (fill me in with today's date, YYYY-MM-DD)
-author: (name of the main author)
+author: John Ericsion (@Ericson2314) on behalf of [Obsidian Systems](https://obsidian.systems)
 co-authors: (find a buddy later to help out with the RFC)
 shepherd-team: (names, to be nominated and accepted by RFC steering committee)
 shepherd-leader: (name to be appointed by RFC steering committee)
@@ -11,21 +11,149 @@ related-issues: (will contain links to implementation PRs)
 # Summary
 [summary]: #summary
 
-One paragraph explanation of the feature.
+Integrate Nix with IPFS, in phases of increasing sophistication.
+This follows the work done and described in https://github.com/obsidiansystems/ipfs-nix-guide/
 
 # Motivation
 [motivation]: #motivation
 
-Why are we doing this? What use cases does it support? What is the expected
-outcome?
+## Binary distribution
+
+Currently distributing Nix binaries takes a lot of bandwidth and storage.
+This is a barrier to being a Nix user in areas of slower internet --- which includes the vast majority of the world's population at this time.
+This is also a barrier to users running their own caches.
+
+Content-addressing opens up a *huge* design space of solutions to get around such problems.
+IPFS explores many of those solutions.
+
+## Source distribution and archival
+
+A goal of the Nix ecosystem is to package software in a way that never bitrots.
+Getting in the way of that, however, is the fact source code frequently goes off-line.
+The Software Heritage archive is the best in the world, and a natural partner in this effort.
+
+Unfortunately, as https://www.tweag.io/blog/2020-06-18-software-heritage/ describes at the end, a major challenge is the way nix content-addresses software.
+First of all, Nix hashes sources in bespoke ways that no other project will adopt.
+Second of all, tarballs instead of the underlying files leaking non-normative details (compression, odd perms, etc.).
+
+We should natively support git file hashing, which Git repos and Software Heritage both support.
+This will completely obliterate these issues.
+
+IPFS also supports git hashing, and so we also provide a good way for people and intuitions to "pin" the sources they need, especially if those sources include private ones SWH won't have.
+Finally, per (Obsidian's bridging work)https://github.com/obsidiansystems/go-ipfs-swh-plugin/ ,
+
+## Not just IPFS
+
+Many of the IPFS-specific logic could in fact live in a plugin if this is desired.
+However, we still need to adjust core abstractions of Nix store layer (as described below) to interface with IPFS in the best possible way.
+Those same adjustments would allow Nix to work better with *any* content-addressing system, so alternatives networks/projects to IPFS can also be just as easily experimented with.
+
+As always with my work, the manta (from Scheme) to follow is
+
+> *x* should be designed not by piling feature on top of feature, but by removing the weaknesses and restrictions that make additional features appear necessary.
+
+A ton of misc features have been added to Nix since 2.0, and we are very careful to not increase total ad-hoc complexity more than necessary.
+
+## Build adoption through seamless interop
+
+This last argument is more strategic than technical.
+
+A lot of people in this community would like to see Nix be used more widely, but as much as we all wish otherwise, the fact remains that there is some tension between making nix *better* and making it *more accessible*.
+
+Nix is very foreign from the "bad conventional" way things are done, and making Nix better can sometimes involve making it even more foreign.
+We don't want to steepen the learning curve or make it "seem more weird".
+
+On the other hand, making Nix more accessible by making it more like tools users are already use-to can obscure or chip-away at Nix's benefits.
+We don't want to "pander" in ways that will make Nix faddish but ultimately undermine it's popular over the long haul (see Docker the company's woes).
+
+One way to get around this tension to me is rather than pushing Nix towards the rest of the world, pushing the rest of the world towards us.
+Like-minded projects emphasizing content-addressing are our *natural* partners, and we should work with them to promote Nix-*agnostic* standards that further our values and mission.
 
 # Detailed design
 [design]: #detailed-design
 
-This is the core, normative part of the RFC. Explain the design in enough
-detail for somebody familiar with the ecosystem to understand, and implement.
-This should get into specifics and corner-cases. Yet, this section should also
-be terse, avoiding redundancy even at the cost of clarity.
+Each item can be done separately provided its dependent items are also done.
+
+## Augmented `narinfo`
+
+- **Purpose**: Binary distribution
+
+*This is taken from [RFC PR 122](https://github.com/lucasew/rfcs/blob/binary-cache-ipfs/rfcs/0122-binary-cache-ipfs.md), which was abandoned by its author.*
+
+The purpose of this is a "hybrid" store where the narinfo metadata is still severed via HTTPS, but the data itself is served via IPFS.
+
+Today, a narinfo looks like this:
+
+```
+StorePath: /nix/store/gdh8165b7rg4y53v64chjys7mbbw89f9-hello-2.10
+URL: nar/0i6ardx43rdg24ab1nc3mq7f5ykyiamymh1v37gxdv5xh5cm0cmb.nar.xz
+Compression: xz
+FileHash: sha256:0i6ardx43rdg24ab1nc3mq7f5ykyiamymh1v37gxdv5xh5cm0cmb
+FileSize: 40360
+NarHash: sha256:1ddv0iqq47j0awyw7a8dmm8bz71c6ifrliq53kmmsfzjxf3rwvb8
+NarSize: 197528
+References: 7gx4kiv5m0i7d7qkixq2cwzbr10lvxwc-glibc-2.27 gdh8165b7rg4y53v64chjys7mbbw89f9-hello-2.10
+Deriver: 5sj6fdfym58sdaf3r5p87v4l8sj2zlvn-hello-2.10.drv
+Sig: cache.nixos.org-1:K0thQEG60rzAK8ZS9f1whb7eRlIshlMDJAm7xvX1oF284H+PTqlicv/wGW6BIj+wWWONHvUZ2MYc+KDArekjDA==
+```
+
+This RFC proposes new key-value pairs that in this example would be:
+
+```
+IpfsCid: Qmf8NfV2hnq44RoQw9vxmSpGYTwAovA8FUCxeCJCqmXeNN
+IpfsEncoding: {"method":"wrapped-nar","chunking":{"leaf-format":"raw","strategy":"fixed-size"},"layout":"balanced","max-width":174}
+```
+
+Just as today, the `NarHash` and `NarSize` remain the *normative* way to lock down the store object the `narinfo` file describes.
+Conversely, The `URL`, `FileHash` and `FileSize` by contrast are *informational*, describing not what the store object *is*, but *how to get it*.
+
+The `IpfsCid` and `IpfsEncoding` are likewise informational, describing how to get the store object:
+
+- `IpfsCid`: Native content address for IPFS.
+
+- `IpfsEncoding`: Enough info to deterministically rebuild the IPFS representation from a non-IPFS copy of the store object.
+
+   For now, `IpfsEncoding` will only support `unixfs-nar`, which works as follows:
+
+   The NAR is itself wrapped in IPFS's [UnixFS](https://github.com/ipfs/specs/blob/main/UNIXFS.md).
+   This other format can be extracted from the CID (which is conceptually a pair of encoding metadata and a hash).
+   For now, only IPFS's "unixfs" is supported.
+   `chunking`, `layout`, and `max-size` are tuning parameters for unixfs [described in the UnixFS spec](https://github.com/ipfs/specs/blob/main/UNIXFS.md#importing).
+
+   "UNIXFS" is not used directly because it doesn't support the "executable bit** Nix does on files.
+   NAR archive are not used directly because IPFS doesn't support arbitrary large objects.
+
+## Git file hashing
+
+- **Purpose**: Source distribution and archival
+
+## Content address or store path in Store interface
+
+- **Purpose**: Source distribution and archival
+
+## Git fetching for `buitins.fetch`
+
+- **Purpose**: Source distribution and archival
+- **Depends on**: Git file hashing, Content address or store path in Store interface
+
+## NAR info or content address normative in `ValidPathInfo`
+
+- **Purpose**: Source distribution and archival
+
+## IPFS Narinfo
+
+- **Purpose**: Binary distribution
+- **Depends on**: Augmented `narinfo`
+
+## Wrapped git objects with references
+
+- **Purpose**: Binary distribution
+- **Depends on**: Git file hashing
+
+## IPLD Derivations
+
+- **Purpose**: Build plan distribution
+- **Depends on**: Wrapped git objects with references
 
 # Examples and Interactions
 [examples-and-interactions]: #examples-and-interactions
--- a/rfcs/0000-ipfs.md
+++ b/rfcs/0000-ipfs.md
@@ -12,7 +12,7 @@ related-issues: (will contain links to implementation PRs)
 [summary]: #summary
 
 Integrate Nix with IPFS, in phases of increasing sophistication.
-This follows the work done and described in https://github.com/obsidiansystems/ipfs-nix-guide/
+This follows the work done and described in https://github.com/obsidiansystems/ipfs-nix-guide/ .
 
 # Motivation
 [motivation]: #motivation
@@ -40,7 +40,7 @@ We should natively support git file hashing, which Git repos and Software Herita
 This will completely obliterate these issues.
 
 IPFS also supports git hashing, and so we also provide a good way for people and intuitions to "pin" the sources they need, especially if those sources include private ones SWH won't have.
-Finally, per (Obsidian's bridging work)https://github.com/obsidiansystems/go-ipfs-swh-plugin/ ,
+Finally, per [Obsidian's bridging work](https://github.com/obsidiansystems/go-ipfs-swh-plugin),
 
 ## Not just IPFS
 
@@ -123,64 +123,153 @@ The `IpfsCid` and `IpfsEncoding` are likewise informational, describing how to g
    "UNIXFS" is not used directly because it doesn't support the "executable bit** Nix does on files.
    NAR archive are not used directly because IPFS doesn't support arbitrary large objects.
 
+## IPFS Narinfo and "stateful" IPFS Store
+
+- **Purpose**: Binary distribution
+- **Depends on**: Augmented `narinfo`
+
+Instead of a "hybrid" store, where the narinfo index is served with HTTP but the data itself is served with IPFS, we can do an all-IPFS store with the data itself and mutable index stored in IPFS.
+The Narinfo instead of being encoded the legacy line-oriented text format can be IPFS's native DAG-CBOR IPLD codec, which is like JSON + content address links (but stored as CBOR).
+This allows Narinfos to reference each other and be nicely structured so the index is legible from Nix-agnostic IPFS tools and recursive pinning comes for free.
+
+Read-only is easier, since IPFS data is immutable but "writable" stores are supported by simple printing back a new CID for the new store root after some modifications, or modifying a mutable IPNS reference.
+IPNS is historically slow, but the update is automatic.
+Printing out a new CID for the index root allows the store administrator to update an out-of-bound mutable reference, but this cannot be automated because Nix doesn't know what the out-of-band method is.
+
 ## Git file hashing
 
 - **Purpose**: Source distribution and archival
 
+In addition to the various forms of content-addressing Nix supports today ("text", "fixed" with either "flat" or "nar" serialization of file system objects), Nix should support Git hashing.
+This support entails two basic things:
+
+ - Content addresses are used to compute store paths.
+ - Content addresses are used to verify store object integrity.
+
+Git hashing would not support references (since references in Nix's sense are not a Git concept), but that is not an issue for the intended use-case of exchanging source code.
+
+## Git file hashing for `buitins.fetch*`
+
+- **Purpose**: Source distribution and archival
+- **Depends on**: Git file hashing,
+
+The builtin fetchers can also be made to work with git file hashing just as they support the other types.
+In addition, Git repo fetching can leverage this better to than the other formats since the data in git repos is already content-addressed in this way.
+
 ## Content address or store path in Store interface
 
-- **Purpose**: Source distribution and archival
+- **Purpose**: All distribution
 
-## Git fetching for `buitins.fetch`
+Modify many store interface methods that today take store paths to instead accept *either* a store path or a content address.
 
-- **Purpose**: Source distribution and archival
-- **Depends on**: Git file hashing, Content address or store path in Store interface
+For stores that are purpose-built for Nix, like the ones we support today, all addressing can be done store paths, so the current interface is fine.
+But for Nix-agnostic stores, store paths are rather useless as a key type because Nix-agnostic tools don't know about them.
+They can, however, understand content addresses.
+And from such a content address we can always produce a store path again, so there is no loss of functionality with existing stores.
 
 ## NAR info or content address normative in `ValidPathInfo`
 
 - **Purpose**: Source distribution and archival
+- **Depends on**: Content address or store path in Store interface,
 
-## IPFS Narinfo
+As described in the first step, currently `NarHash` and `NarSize` are the *normative* fields which are used to verify a store object.
+But if the store object is content-addressed, we don't need these, because the content address (`CA` field) will also suffice, all by itself.
+Relax the abstract `ValidPathInfo` type to merely require that *one of* `NarHash` and `NarSize` or `CA` be defined.
 
-- **Purpose**: Binary distribution
-- **Depends on**: Augmented `narinfo`
+Existing Nix stores types are still required to contain a `NarHash` and `NarSize`, which is good for backwards compat and don't come with a cost.
+Only new nix-agnostic store types would take advantage of these new, relaxed rules.
+
+## "stateless" IPFS store
+
+- **Purpose**: Source distribution and archival
+- **Depends on**: NAR info or content address normative in `ValidPathInfo`
+
+Use the above functionality to create a "stateless" IPFS store.
+Opaque store path lookups always fail, but when the key is the new content address type, we can translate the key itself into a CID that we can look up.
+
+Unlike the previous two flavours of IPFS store, this one is stateless in that there is no need for an index at all.
+Only content-addressed data is looked up, and it doesn't need any nar-info metadata before the data is all there.
+
+We need the previous step for querying without fetching any data.
+In that case since there is no narinfo index we're looking up, we don't get any additional metadata back.
+But the content address key a successful query used is enough to create a bare-bones `ValidPathInfo` with a `CA` field, which with the enough step is valid.
+
+(A bare-bones `ValidPathInfo` might sound sub-par, but for plain old content-addressed data it is fine.
+Most of the other metadata in `ValidPathInfo` is really just for input-addressed derivation outputs, and is thus obviated by CA derivation trust maps which contain the same data but more naturally.)
 
 ## Wrapped git objects with references
 
 - **Purpose**: Binary distribution
 - **Depends on**: Git file hashing
 
+Merkelized formats like git file hashing are better than NAR because that allow for very natural deduplication and minimal transfers.
+This is the same benefit we get today with Nix within a closure of multiple store objects, now also *within a single store object*.
+But git has no notion of Nix-style references, so plain git hashing is only suitable for leaf store objects without references (like source code).
+
+However, we can use IPLD to wrap git-hashed data with a reference set, and "has self reference" bit.
+This easily creates a new content addressing scheme which handles all "shapes" of store objects.
+This gives is a nice way to thus share arbitrary nix store data (provided it is content-addressed) over IPFS.
+
+Like with "IPFS Narinfo", this format is also very easy to understand with nix-agnostic native IPFS tools.
+This is because, once again, the reference graph is made native to IPFS not done indirectly with store path strings which must be looked up.
+
+An interesting corollary to note:
+Content addressing today is "shallow", in that references are arbitrary store paths.
+With this form on content addressing, references are instead CIDs (native IPFS references) to other obligatorily content-addressed data.
+This means the content addressing is "deep", such that any such content-addressed store object always has a content-addressed closure.
+At the cost of interop with existing derivation outputs, this make such data easier to manage because there are fewer trust issues and degrees of freedom in general for something to go wrong.
+
 ## IPLD Derivations
 
 - **Purpose**: Build plan distribution
-- **Depends on**: Wrapped git objects with references
+- **Depends on**: Wrapped git objects with references,
+                  IPFS as substitutor
+
+Natively represent derivations in IPFS, again with the same benefits of leverage the native graph representations.
+
+This is a culmination of all the futures so far.
+The derivations must be CA derivations (floating or fixed).
+They must also produce wrapped git objects with references, though they can also depend on regular unwrapped git file hashed store objects.
+
+The derivations and their outputs are thus all fully IPFS native, leveraging the IPFS graph and trust vs plain old data separation for the high standard of interoperability.
 
 # Examples and Interactions
 [examples-and-interactions]: #examples-and-interactions
 
-This section illustrates the detailed design. This section should clarify all
-confusion the reader has from the previous sections. It is especially important
-to counterbalance the desired terseness of the detailed design; if you feel
-your detailed design is rudely short, consider making this section longer
-instead.
+We encourage anyone interested to check our tutorial in https://github.com/obsidiansystems/ipfs-nix-guide/ which demonstrates the above functionality.
+Note at the time of writing this guide uses our original 2020 fork of Nix.
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
-Why should we *not* do this?
+The main cost is more complexity to the store layer.
+For two reason we think this is not so bad:
+
+1. Per the abstract vs concrete model of the nix store in https://github.com/NixOS/nix/pull/6877 , everything we are doing is simply flushing out alternative interpretations of the abstract model.
+   This is the sense in which we are "removing the weaknesses and restrictions that make additional features appear necessary" per the Scheme mantra cited above:
+   Instead of extending the model wit new features, we are reflaxing concrete model assumptions (e.g. references are always opaque store paths) while keeping the abstract model the same.
+
+2. We also support plans to decouple the layers of Nix further, and update our educational and marketing material to reflect it.
+   With Flakes and other post-2.0 features, the upper layer of Nix have gained an enormous amount of flexibility and sophistication.
+   RFCs like this show that the so-far more sleepy lower layers also have plenty of potential to gain sophistication too.
+
+   Embracing layering on technical, educational, communications, and managerial levels can scale our capacity to manage complexity and sophistication without the project growing out of control.
+   It will "divide and conquer" the project so the interfaces between each layer are still rigorously enforced preventing a combinatorial explosion in complexity.
+
+   We plan on more formally proposing this next.
 
 # Alternatives
 [alternatives]: #alternatives
 
-What other designs have been considered? What is the impact of not doing this?
+The dependency graph of steps can be sliced to save some for future work.
+For now they are all written together, but during the RFC meetings we will decide which steps (if any) to ratify now, and which steps to save for later.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-What parts of the design are still TBD or unknowns?
+Per the above, deciding which steps to leave as future work.
 
 # Future work
 [future]: #future-work
 
-What future work, if any, would be implied or impacted by this feature
-without being directly part of the work?
+Chiefly, any steps which we don't wish to commit to initially; to be decided as described above.
--- a/rfcs/0000-ipfs.md
+++ b/rfcs/0000-ipfs.md
@@ -39,7 +39,7 @@ Second of all, tarballs instead of the underlying files leaking non-normative de
 We should natively support git file hashing, which Git repos and Software Heritage both support.
 This will completely obliterate these issues.
 
-IPFS also supports git hashing, and so we also provide a good way for people and intuitions to "pin" the sources they need, especially if those sources include private ones SWH won't have.
+IPFS also supports git hashing, and so we also provide a good way for people and institutions to "pin" the sources they need, especially if those sources include private ones SWH won't have.
 Finally, per [Obsidian's bridging work](https://github.com/obsidiansystems/go-ipfs-swh-plugin),
 
 ## Not just IPFS
@@ -48,7 +48,7 @@ Many of the IPFS-specific logic could in fact live in a plugin if this is desire
 However, we still need to adjust core abstractions of Nix store layer (as described below) to interface with IPFS in the best possible way.
 Those same adjustments would allow Nix to work better with *any* content-addressing system, so alternatives networks/projects to IPFS can also be just as easily experimented with.
 
-As always with my work, the manta (from Scheme) to follow is
+As always with my work, the mantra (from Scheme) to follow is
 
 > *x* should be designed not by piling feature on top of feature, but by removing the weaknesses and restrictions that make additional features appear necessary.
 
@@ -64,7 +64,7 @@ Nix is very foreign from the "bad conventional" way things are done, and making 
 We don't want to steepen the learning curve or make it "seem more weird".
 
 On the other hand, making Nix more accessible by making it more like tools users are already use-to can obscure or chip-away at Nix's benefits.
-We don't want to "pander" in ways that will make Nix faddish but ultimately undermine it's popular over the long haul (see Docker the company's woes).
+We don't want to "pander" in ways that will make Nix faddish but ultimately undermine it's popularity over the long haul (see Docker the company's woes).
 
 One way to get around this tension to me is rather than pushing Nix towards the rest of the world, pushing the rest of the world towards us.
 Like-minded projects emphasizing content-addressing are our *natural* partners, and we should work with them to promote Nix-*agnostic* standards that further our values and mission.
@@ -227,7 +227,7 @@ At the cost of interop with existing derivation outputs, this make such data eas
 
 Natively represent derivations in IPFS, again with the same benefits of leverage the native graph representations.
 
-This is a culmination of all the futures so far.
+This is a culmination of all the features so far.
 The derivations must be CA derivations (floating or fixed).
 They must also produce wrapped git objects with references, though they can also depend on regular unwrapped git file hashed store objects.
 
@@ -248,7 +248,7 @@ For a few reason we think this is not so bad.
 Most importantly is the division of the work into a dependency graph of steps.
 This allows us to slowly try IPFS out and not commit to more change than we want to up front.
 
-Even if we do end up adopting everything though, we thing for the following two reasons the complexity can still be kept manageable:
+Even if we do end up adopting everything though, we think for the following two reasons the complexity can still be kept manageable:
 
 2. Per the abstract vs concrete model of the nix store in https://github.com/NixOS/nix/pull/6877 , everything we are doing is simply flushing out alternative interpretations of the abstract model.
    This is the sense in which we are "removing the weaknesses and restrictions that make additional features appear necessary" per the Scheme mantra cited above:
--- a/rfcs/0000-ipfs.md
+++ b/rfcs/0000-ipfs.md
@@ -243,18 +243,24 @@ Note at the time of writing this guide uses our original 2020 fork of Nix.
 [drawbacks]: #drawbacks
 
 The main cost is more complexity to the store layer.
-For two reason we think this is not so bad:
+For a few reason we think this is not so bad.
 
-1. Per the abstract vs concrete model of the nix store in https://github.com/NixOS/nix/pull/6877 , everything we are doing is simply flushing out alternative interpretations of the abstract model.
+Most importantly is the division of the work into a dependency graph of steps.
+This allows us to slowly try IPFS out and not commit to more change than we want to up front.
+
+Even if we do end up adopting everything though, we thing for the following two reasons the complexity can still be kept manageable:
+
+2. Per the abstract vs concrete model of the nix store in https://github.com/NixOS/nix/pull/6877 , everything we are doing is simply flushing out alternative interpretations of the abstract model.
    This is the sense in which we are "removing the weaknesses and restrictions that make additional features appear necessary" per the Scheme mantra cited above:
    Instead of extending the model wit new features, we are reflaxing concrete model assumptions (e.g. references are always opaque store paths) while keeping the abstract model the same.
 
-2. We also support plans to decouple the layers of Nix further, and update our educational and marketing material to reflect it.
+3. We also support plans to decouple the layers of Nix further, and update our educational and marketing material to reflect it.
    With Flakes and other post-2.0 features, the upper layer of Nix have gained an enormous amount of flexibility and sophistication.
    RFCs like this show that the so-far more sleepy lower layers also have plenty of potential to gain sophistication too.
 
    Embracing layering on technical, educational, communications, and managerial levels can scale our capacity to manage complexity and sophistication without the project growing out of control.
    It will "divide and conquer" the project so the interfaces between each layer are still rigorously enforced preventing a combinatorial explosion in complexity.
+   That frees up "complexity budget" for project like this.
 
    We plan on more formally proposing this next.
 
--- a/rfcs/0000-ipfs.md
+++ b/rfcs/0000-ipfs.md
@@ -40,7 +40,10 @@ We should natively support git file hashing, which Git repos and Software Herita
 This will completely obliterate these issues.
 
 IPFS also supports git hashing, and so we also provide a good way for people and institutions to "pin" the sources they need, especially if those sources include private ones SWH won't have.
-Finally, per [Obsidian's bridging work](https://github.com/obsidiansystems/go-ipfs-swh-plugin),
+Finally, per [Obsidian's bridging work](https://github.com/obsidiansystems/go-ipfs-swh-plugin), we have a 3 way integration between IPFS, SWH, and Nix. 
+Data can be directly downloaded from SWH via HTTPS, or indirectly via IPFS, which can act as a CDN to not put as much load on SWH's servers.
+
+Overall, are building out a uniform way to work with source code, regardless of its origins or the exact tools involved.
 
 ## Not just IPFS
 


rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+
+   For now, `IpfsEncoding` will only support `unixfs-nar`, which works as follows:
+
+   The NAR is itself wrapped in IPFS's [UnixFS](https://github.com/ipfs/specs/blob/main/UNIXFS.md).
+   This other format can be extracted from the CID (which is conceptually a pair of encoding metadata and a hash).
+   For now, only IPFS's "unixfs" is supported.
+   `chunking`, `layout`, and `max-size` are tuning parameters for unixfs [described in the UnixFS spec](https://github.com/ipfs/specs/blob/main/UNIXFS.md#importing).
+
+   "UNIXFS" is not used directly because it doesn't support the "executable bit** Nix does on files.
+   NAR archive are not used directly because IPFS doesn't support arbitrary large objects.
```
kevincox, line 127:
```suggestion
   NAR archive are not used directly because IPFS doesn't support arbitrarily large objects.
```

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+- **Purpose**: Binary distribution
+- **Depends on**: Augmented `narinfo`
+
+Instead of a "hybrid" store, where the narinfo index is served with HTTP but the data itself is served with IPFS, we can do an all-IPFS store with the data itself and mutable index stored in IPFS.
+The Narinfo instead of being encoded the legacy line-oriented text format can be IPFS's native DAG-CBOR IPLD codec, which is like JSON + content address links (but stored as CBOR).
+This allows Narinfos to reference each other and be nicely structured so the index is legible from Nix-agnostic IPFS tools and recursive pinning comes for free.
+
+Read-only is easier, since IPFS data is immutable but "writable" stores are supported by simple printing back a new CID for the new store root after some modifications, or modifying a mutable IPNS reference.
+IPNS is historically slow, but the update is automatic.
+Printing out a new CID for the index root allows the store administrator to update an out-of-bound mutable reference, but this cannot be automated because Nix doesn't know what the out-of-band method is.
```
kevincox, line 140: > but this cannot be automated because Nix doesn't know what the out-of-band method is.

I'm not sure the point here. Just as updating IPNS can be automated we can automate publishing DNSLink entries for popular providers or any number of other methods. Maybe you trying to point out that this would be a protocol outside of IPFS?

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+And from such a content address we can always produce a store path again, so there is no loss of functionality with existing stores.
+
+## NAR info or content address normative in `ValidPathInfo`
+
+- **Purpose**: Source distribution and archival
+- **Depends on**: Content address or store path in Store interface,
+
+As described in the first step, currently `NarHash` and `NarSize` are the *normative* fields which are used to verify a store object.
+But if the store object is content-addressed, we don't need these, because the content address (`CA` field) will also suffice, all by itself.
+Relax the abstract `ValidPathInfo` type to merely require that *one of* `NarHash` and `NarSize` or `CA` be defined.
```
kevincox, line 180:
```suggestion
Relax the abstract `ValidPathInfo` type to merely require that *one of* (`NarHash` and `NarSize`) or `CA` be defined.
```

I am mixing English and math syntax but I think this makes it more clear. Or maybe something like "both NarHash and NarSize or just CA be defined".

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Integrate Nix with IPFS, in phases of increasing sophistication.
```
kevincox, line 14: While I think most of the proposals in this RFC make sense, I think that proposing them all together makes discussion and acceptance difficult. I would recommend that each proposal is split into its own RFC to keep each change atomic. The whole roadmap is very useful to see however I think this roadmap can be maintained outside of the RFC process or as a bunch of draft RFCs that can then be opened and progressed once the dependencies have been accepted.

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+A goal of the Nix ecosystem is to package software in a way that never bitrots.
+Getting in the way of that, however, is the fact source code frequently goes off-line.
+The Software Heritage archive is the best in the world, and a natural partner in this effort.
+
+Unfortunately, as https://www.tweag.io/blog/2020-06-18-software-heritage/ describes at the end, a major challenge is the way nix content-addresses software.
+First of all, Nix hashes sources in bespoke ways that no other project will adopt.
+Second of all, tarballs instead of the underlying files leaking non-normative details (compression, odd perms, etc.).
+
+We should natively support git file hashing, which Git repos and Software Heritage both support.
+This will completely obliterate these issues.
```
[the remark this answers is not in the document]
Ericson2314: > Or does the lack of adoption of SHA2 in Git pose issues.

I think it does. I am happy to prioritize adding SHA-2 Git (and SWHID) content-addressing ahead of wide adoption, and indeed try to lead that change, but I am hesitant to get rid of SHA-1 support until that change happens.

Put another way, as long as Git SHA-1 is omnipresent, our our additional hashes adding much value? Whoever wrote down the additional hashes was verifying something by SHA-1, so a hash collision attack could sneak in then.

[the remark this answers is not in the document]
Ericson2314: > Or does the lack of adoption of SHA2 in Git pose issues.

I think it does. I am happy to prioritize adding SHA-2 Git (and SWHID) content-addressing ahead of wide adoption, and indeed try to lead that change, but I am hesitant to get rid of SHA-1 support until that change happens.

Put another way, as long as Git SHA-1 is omnipresent, our our additional hashes adding much value? Whoever wrote down the additional hashes was verifying something by SHA-1, so a hash collision attack could sneak in then.

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+References: 7gx4kiv5m0i7d7qkixq2cwzbr10lvxwc-glibc-2.27 gdh8165b7rg4y53v64chjys7mbbw89f9-hello-2.10
+Deriver: 5sj6fdfym58sdaf3r5p87v4l8sj2zlvn-hello-2.10.drv
+Sig: cache.nixos.org-1:K0thQEG60rzAK8ZS9f1whb7eRlIshlMDJAm7xvX1oF284H+PTqlicv/wGW6BIj+wWWONHvUZ2MYc+KDArekjDA==
+```
+
+This RFC proposes new key-value pairs that in this example would be:
+
+```
+IpfsCid: Qmf8NfV2hnq44RoQw9vxmSpGYTwAovA8FUCxeCJCqmXeNN
+IpfsEncoding: {"method":"wrapped-nar","chunking":{"leaf-format":"raw","strategy":"fixed-size"},"layout":"balanced","max-width":174}
```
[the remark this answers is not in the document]
Ericson2314: Oh oops, I tried to write down things per the spec.

This is the one part that *isn't* already implemented. And so me writing that has some guesswork.

TBH I feel somewhat demotivated about this field because I am not sure how much UnixFS intends to make writing and verifying it anyways, and unixfs-wrapped NAR is somewhat a stop-gap anyways.

rfcs/0133-ipfs.md
Ericson2314:
```suggestion
Relax the abstract `ValidPathInfo` type to merely require that *either* the pair of `NarHash` and `NarSize` or `CA` alone be defined.
```

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
kevincox: I think I am partly confused because it seems like there may be multiple uses of the hash, both for the native importing of sources that use Git (which the sha1 support is definitely a huge benefit) and use for non-git sources and other artifacts (where the benefits are less clear). I think if we can make it clear which hash is used for which things and where parity matters then I think you are right and we will need to support sha1-git for the time being.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
kevincox: I think the important thing is without this field someone who has a nar can't (without some guesswork and trial-and-error) share it with the network. This means that if we want the official cache.nixos.org to support IPFS it will need to do so natively. By adding this field we add a transition option where cache.nixos.org doesn't need to provide IPFS resources directly but interested community members can "rehost" the nars and others can find them.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: We could I suppose *only* allow git sha-1 fetching with `builtins.fetchgit` and friends, but what about fixed-output derivations like Nixpkgs uses that also fetch git repos?

One thing I am hoping we can get to is just writing down *one* hash for git repo fetching, and if it is a commit hash or tag object instead of git hash, using a [Merkle inclusion proof](https://computersciencewiki.org/index.php/Merkle_proof) to connect it to the tree hash we'd ultimately use.

That would be a *huge* ergonomics improvement for Nix.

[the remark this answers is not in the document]
kevincox: I'm trying to comprehend the security implications of sha1 hashes for store content and I think they aren't as bad as I thought. Since I think that sha1 hashes in the store will look different (different length) than other hashes you won't be able to replace a non-sha1 path with a colliding sha1 path. So I think it will only reduce security for places where it is used which is probably not that much of a problem. With little thought the worst attack I can think of is forcing someone to pull your source for a build. That is obviously bad but as sha1 is (hopefully) phased out less and less packages will be affected by this, just having sha1 as a supported hash doesn't reduce the integrity of derivations that don't use it.

That being said I wonder if it still makes sense to have `builtins.fetchgit` still support providing an additional hash. This would be less ergonomic than just providing one hash for both commit ID + fixed output verification but would provide higher confidence in the integrity of the downloaded result. So this would basically look like downloading from IPFS via the sha1 git hash and then verifying the result again with the other hash.

rfcs/0133-ipfs.md
Ericson2314: Well if every step gets its own RFC it will take quite a while!

I am hoping in the first meeting we can chop this into two parts parts, move the second to future work and circle back to it later. Maybe we do that a few times, but I hope the resulting partition ends up being more than one step per RFC :)

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: I definitely agree strongly with that on principle. I just hope I don't get very bogged down trying to extra all this metadata and test it. Perhaps the thing to do is only support very few combinations at first.

rfcs/0133-ipfs.md
kevincox: Yeah, maybe coarse chops are the right approach.

It is important to remember:

1. The time to review an RFC is super-linear with the size.
2. The number of smaller discussions to be resolved is higher for a large RFC.

So while smaller RFCs may have more bureaucratic overhead and organizational overhead it may help actually solidify each component and get it approved. Because trying to understand and discuss the later changes when the assumptions is is based on are mutable provide valuable flexibility but this flexibility also makes it hard to keep discussions concrete and productive.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
kevincox: +1 If we say that the only supported method to start is fixed raw-leaves with 4k chunks or whatever I think that is a good start. I still think the JSON format is nice for future extensibility even if we only currently support one option.

rfcs/0133-ipfs.md
Ericson2314: Yes exactly. It is quite possible to get Nix to call some hook, but there are perhaps details to figure out there (though I suppose we've rather wantonly added hooks so far...) and I didn't do it yet.

There is an interesting thing going on where I am very keen on the source code use-case, which due to it's statelessness is just much easier / more ergonomic on both the "provider" and "consumer" sides, but it also requires more up-front changes to Nix which could be controversial.

I look forward to RFC meetings to sort out all that.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: Yeah we can and should have full JSON metadata if possible even if we only support one case to begin.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: > So I think it will only reduce security for places where it is used which is probably not that much of a problem.

Yup. And the name of the hash is always included in the store path hash computation too. And the *second* round of hashing used for that is SHA-256.

I should add some text here emphasizing this property.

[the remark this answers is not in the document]
Ericson2314: > That being said I wonder if it still makes sense to have `builtins.fetchgit` still support providing an additional hash. This would be less ergonomic than just providing one hash for both commit ID + fixed output verification but would provide higher confidence in the integrity of the downloaded result.

Yeah that is just what we have today, more or less.

> So this would basically look like downloading from IPFS via the sha1 git hash and then verifying the result again with the other hash.

This is a little tricky because right now one *just* fetches from IPFS by using it as a substitutor, this is no `builtins.fetchipfs` because no libfetchers additional logic is needed! But there are still options we can discuss.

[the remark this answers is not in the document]
kevincox: A simple option to work around this would be to keep `fetchgit` as a sha1 derivation but then wrap it in another derivation that just checks the hash and copies the result. You wouldn't be able to download this derivation from IPFS but it is likely cheap enough that it isn't a huge issue to rebuild it locally.

rfcs/0133-ipfs.md
Ericson2314: Yeah right it's an optimization problem with overhead on balkanized discussion on one hand vs superlinearity of review cost w.r.t RFC on the other.

Moving things on first meeting into the "future work" section is a sort of non-balanced bisection algorithm to approximate an optimal solution :).

Ericson2314:
I would like to nominate @kevincox based on his feedback already (yay!) and his participation in the previous RFC #122 meetings.

I also suspect we'll need one or both of @edolstra and @thufschmitt for changes to Nix like this, so I nominate them both too.

I also suspect @fricklerhandwerk and @tomberek might be interested also, time permitting.

kevincox:
I love the possibility of distributed content-addressed caches. I accept the nomination.

rfcs/0133-ipfs.md
asymmetric, line 107: Should these perhabs be capitalized as

```suggestion
IPFSCID: Qmf8NfV2hnq44RoQw9vxmSpGYTwAovA8FUCxeCJCqmXeNN
IPFSEncoding: {"method":"wrapped-nar","chunking":{"leaf-format":"raw","strategy":"fixed-size"},"layout":"balanced","max-width":174}
```

following the convention that acronym initials are capitalized?

Ericson2314: I think that is hard to read, so I did the rust-style "treat acronyms as words for better camel-casing" thing, but I am fine with other options like snake case or kebob case too.

edolstra:

fricklerhandwerk:
@Ericson2314 Sorry, I can't commit any time to this. I like the idea a lot, but would have to go into more depth to be of any help, and this would be too much at the moment.

Ericson2314:
@fricklerhandwerk No problem at all :). I figured might be the case, but wanted to extend the nomination anyways based on our prior conversation.

rfcs/0133-ipfs.md
amjoseph-nixpkgs: It isn't as trendy, but Bittorrent's [ed25519-signed DHT entries](https://www.bittorrent.org/beps/bep_0044.html) can be used for this.  The protocol is very lightweight and easy to understand -- not to mention battle-tested.  Getting block-level content addressing requires [v2 torrents](http://www.bittorrent.org/beps/bep_0052.html) which are still not as popular as the v1 torrents -- but they share the same DHT network and uTP protocol.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
amjoseph-nixpkgs: > Git hashing is currently on SHA1 which is showing weaknesses.

There is some useful and recent information on the git-sha256 situation [in this LWN article](https://lwn.net/Articles/898522/).

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+# Summary
+[summary]: #summary
+
+Integrate Nix with IPFS, in phases of increasing sophistication.
+This follows the work done and described in https://github.com/obsidiansystems/ipfs-nix-guide/ .
+
+# Motivation
+[motivation]: #motivation
+
+## Binary distribution
```
rht, line 20: IIUC, the phase one of this RFC is to use IPFS as a decentralized CDN for binary/source distributions. But I do think there still needs to be a benchmark on whether the transport is more efficient than `git fetch` of Git packfiles.

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+
+Currently distributing Nix binaries takes a lot of bandwidth and storage.
+This is a barrier to being a Nix user in areas of slower internet --- which includes the vast majority of the world's population at this time.
+This is also a barrier to users running their own caches.
```
rht, line 24: How is it addressed in Obsidian that IPFS (by default, I think) duplicates the storage requirement of a package? Both in the output directory `/nix/store/...` and in IPFS own blocks store.

rfcs/0133-ipfs.md
abueide: i like the rule of acronyms larger than 3 get camel-cased

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+- `IpfsEncoding`: Enough info to deterministically rebuild the IPFS representation from a non-IPFS copy of the store object.
+
+   For now, `IpfsEncoding` will only support `unixfs-nar`, which works as follows:
+
+   The NAR is itself wrapped in IPFS's [UnixFS](https://github.com/ipfs/specs/blob/main/UNIXFS.md).
+   This other format can be extracted from the CID (which is conceptually a pair of encoding metadata and a hash).
+   For now, only IPFS's "unixfs" is supported.
+   `chunking`, `layout`, and `max-size` are tuning parameters for unixfs [described in the UnixFS spec](https://github.com/ipfs/specs/blob/main/UNIXFS.md#importing).
+
+   "UNIXFS" is not used directly because it doesn't support the "executable bit** Nix does on files.
```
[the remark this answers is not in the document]
Ericson2314: Oh hmm perhaps I am mistaken. @warpfork do you know about this?

rfcs/0133-ipfs.md
Ericson2314: This is not yet addressed. The focus of this RFC is interfaces not implementation details. I think it is fine that is out of scope for now because we can transparently improve it later.

Also Obsidian Systems is just the name of the company I work at. This project doesn't have a "branded" name, because it is meant to become a seamless part of upstream Nix.

rfcs/0133-ipfs.md
Ericson2314: I agree, but they also are arguably more "invasive" depending on exactly where we draw the line.

rfcs/0133-ipfs.md
Ericson2314: `git fetch` and packfiles can do very wonderful and clever things, but acting as a CDN is not among them. I think this is an apples-and-oranges comparison without a clear answer benchmarking could resolve.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: Thanks! My take away is:

1. Make sure we also support [SHA-1DC](https://github.com/cr-marcstevens/sha1collisiondetection) not vanilla SHA-1.
2. Try out supporting greenfield SHA-256 git repos.
3. Do not worry about migrating code as upstream git doesn't have a plan to migrate / dual-sign repos either.

We didn't do (2) in 2020 when the support in git itself was new, but we could do (2) now.

Ultimately crypto-agility is an interesting and important problem, and I think there is a great opportunity to lead the way here *eventually*, but lets worry about trying to catch up to the state of the art before we push it forward!

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+```
+
+Just as today, the `NarHash` and `NarSize` remain the *normative* way to lock down the store object the `narinfo` file describes.
+Conversely, The `URL`, `FileHash` and `FileSize` by contrast are *informational*, describing not what the store object *is*, but *how to get it*.
+
+The `IpfsCid` and `IpfsEncoding` are likewise informational, describing how to get the store object:
+
+- `IpfsCid`: Native content address for IPFS.
+
+- `IpfsEncoding`: Enough info to deterministically rebuild the IPFS representation from a non-IPFS copy of the store object.
```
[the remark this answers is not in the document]
Ericson2314: Yes I agree we should prohibit not ignore extra freedom on the axes we don't care about. No Postel's law permissiveness for the first version!

rfcs/0133-ipfs.md
Ericson2314: Yeah I am down to experiment the other protocols here. As I've long maintain, very little of the work here is *actually* specific to IPFS.

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+The builtin fetchers can also be made to work with git file hashing just as they support the other types.
+In addition, Git repo fetching can leverage this better to than the other formats since the data in git repos is already content-addressed in this way.
+
+## Content address or store path in Store interface
+
+- **Purpose**: All distribution
+
+Modify many store interface methods that today take store paths to instead accept *either* a store path or a content address.
+
+For stores that are purpose-built for Nix, like the ones we support today, all addressing can be done store paths, so the current interface is fine.
```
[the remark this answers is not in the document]
Ericson2314: Thanks! My thinking on what the store is / ought to be in general was definitely informed by working on this stuff :)

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
...
+Most of the other metadata in `ValidPathInfo` is really just for input-addressed derivation outputs, and is thus obviated by CA derivation trust maps which contain the same data but more naturally.)
+
+## Wrapped git objects with references
+
+- **Purpose**: Binary distribution
+- **Depends on**: Git file hashing
+
+Merkelized formats like git file hashing are better than NAR because that allow for very natural deduplication and minimal transfers.
+This is the same benefit we get today with Nix within a closure of multiple store objects, now also *within a single store object*.
+But git has no notion of Nix-style references, so plain git hashing is only suitable for leaf store objects without references (like source code).
```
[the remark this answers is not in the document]
Ericson2314: Yes, but I wouldn't want to "hack in" our references as invalid submodules.

Our references are also not associated with a single "place" in the file system tree. They are really store object wide, or if one thinks of scanning associated with hashes in specific files. Neither of those corresponds to where submodules are referenced.

rfcs/0133-ipfs.md
rht: SGTM

rfcs/0133-ipfs.md
rht: One could have the mirrors in the form of Git repos, though this is not as automatic as running the IPFS daemon.

At the very least, for low-resource/remote systems, in addition to improving the download speed, I think we wouldn't want the package download to consume a lot of bandwidth overall? But then again, this optimization issue is implementation detail, and so is out of scope for this RFC.

Ericson2314:
Yay! Thank you so much, @gador.

lheckemann:
Thanks @gador! We do unfortunately need one more shepherd, since the current accepted nominations (@edolstra, @kevincox, @gador) result in a team where more than half of the members are Steering Committee members.

I'll nominate @amjoseph-nixpkgs since it looks like you've engaged with the contents in detail :)

amjoseph-nixpkgs:
> I'll nominate @amjoseph-nixpkgs since it looks like you've engaged with the contents in detail :)

I'd be happy to.

Ericson2314:
Yay! Thanks @amjoseph-nixpkgs, I very much appreciate it.

@lheckemann is it possible for the shepherd team to pick a lead shepherd so we can finalize the shepherd team before the next meeting in two weeks?

Ericson2314:
I retconned the room for now-closed RFC #122 as one for this one: [#nix-rfc-133:matrix.org](https://matrix.to/#/#nix-rfc-133:matrix.org).

amjoseph-nixpkgs:
> I retconned the room for now-closed RFC #122 as one for this one: [#nix-rfc-133:matrix.org](https://matrix.to/#/#nix-rfc-133:matrix.org).

Would it be possible to use a venue that supports *both* IRC and Matrix connections?  That way everybody can use their preferred client.  On both hackint and libera.chat I'm `amjoseph` -- unlike github, nobody over there has squatted on my preferred username.

I know hackint in particular has a lot of Matrix users, including some of the people in this PR thread.  As a test, I've created `#nix-rfc-133` on hackint, which I believe should be [accessible to Matrix users](https://www.hackint.org/transport/matrix) as `#nix-rfc-133:hackint.org`.  If people would prefer to use another IRC server, or create the channel themselves, just let me know and I'll delete it.

gador:
@amjoseph-nixpkgs I can't seem to find the matrix room `#nix-rfc-133` on `hackint.org`..

amjoseph-nixpkgs:
> @amjoseph-nixpkgs I can't seem to find the matrix room `#nix-rfc-133` on `hackint.org`..

So I asked in #hackint and a random benevolent matrix user joined the channel to make sure it was possible, and it was.  I asked if they had any instructions I could relay to you, and they said "open a direct message with @appservice-irc:hackint.org and send `!join #nix-rfc-133`."

gador:
thanks for the tips. It still doesn't work though.
I tried with both desktop and mobile client. I opened a DM with `@appservice-irc:hackint.org` on the matrix server `hackint.org` -> No reply.
The room still does not exist. Either on `matrix.org` nor on `hackint.org`. Did I misunderstood the way this is supposed to work? I thought there will be a room on the matrix server `hackint.org` which will be a bridge to the IRC server on `hackint.org`?

amjoseph-nixpkgs:
@Ericson2314 should be awake in a few hours.  He's on several hackint channels via matrix.  I'm sure he knows the necessary trick.

andir:
> thanks for the tips. It still doesn't work though. I tried with both desktop and mobile client. I opened a DM with `@appservice-irc:hackint.org` on the matrix server `hackint.org` -> No reply. The room still does not exist. Either on `matrix.org` nor on `hackint.org`. Did I misunderstood the way this is supposed to work? I thought there will be a room on the matrix server `hackint.org` which will be a bridge to the IRC server on `hackint.org`?

Don't mess around with the appservice, just join ["#nix-rfc-133:hackint.org"](https://matrix.to/#/#nix-rfc-133:hackint.org) like any other matrix channel.

gador:
for some mysterious reason, my client now found the room..thanks for the help

Ericson2314:
I don't know any secrets but it worked for me.

tomberek:
RFC Steering: shepherds selected: @edolstra, @kevincox, @gador, @amjoseph-nixpkgs 

@amjoseph-nixpkgs, would you be willing to be shepherd lead?

amjoseph-nixpkgs:
> @amjoseph-nixpkgs, would you be willing to be shepherd lead?

Oh my.  Well, um, this is my first experience with the RFC process.  I've read RFC 36 (which has a *state transition diagram*! :smile:).  I get the general idea of the shepherd committee's role (sort of like a dissertation committee?)... but I imagine the specific details are the sort of thing you learn from experience; I was hoping this would be my chance.  I guess the best way to learn is to jump in the deep end. 

If the rest of the shepherds and @Ericson2314 are willing to meta-shepherd me, and let me know what I'm supposed to be doing, I can take on the shepherd lead responsibility.

PS, the next four days I will be flying a lot and may only be online ~once per day.  I have no other travel planned for the rest of the year.  The email address I use on my git commits is the best way to reach me; IRC works too.  I'll give @Ericson2314 a telephone number for use as a last resort.

edolstra:
@Ericson2314 Can you make this PR editable? Then I can update the shepherds.

rfcs/0133-ipfs.md
```diff
@@ -0,0 +1,284 @@
+---
+feature: ipfs
+start-date: (fill me in with today's date, YYYY-MM-DD)
+author: John Ericsion (@Ericson2314) on behalf of [Obsidian Systems](https://obsidian.systems)
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
```
edolstra, line 7:
```suggestion
shepherd-team: edolstra, kevincox, gador, @mjoseph-nixpkgs
shepherd-leader: mjoseph-nixpkgs
```

Ericson2314: Woo!

0zAND1z:
IPFS user since 2017 here :wave: 

Just started using `nix` today and understood the value of using IPFS as an alternative for github public archival. We had something similar to this called `gx` back in the old days.

Really appreciate the work here and love the energy. Happy to help in any way possible!

lheckemann:
@amjoseph-nixpkgs maybe organising a meeting with the shepherd team and author would make sense at this stage to work out what needs to be done to get this one step closer to being decided on :)

amjoseph-nixpkgs:
@lheckemann okay, I will do that.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
SuperSandro2000: So the git upload is signed by the signing key to make sure that it is trusted?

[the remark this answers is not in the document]
amjoseph-nixpkgs: > 3\. dual-sign

I think Ericson2314 meant "dual-hash".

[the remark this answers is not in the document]
SuperSandro2000: How would dual hashing prevent the following scenario:
- a malicious commit is produced with a sha1 collision
- it gets uploaded to ipfs before the nix CI can process it

Also how would it be prevented that wrongly signed nix store entries are uploaded to ipfs? Would the hash then be different or could a collision be produced?

SuperSandro2000:
I have some questions and problems with this:
- IPFS cannot be free. The data must be stored somewhere, so it either relies on someones goodwill, donations or the users are the product with their data. Protocol Research Inc., the company behind ipfs.tech, is not a non-profit like eg archive.org and I also could not find a way to donate money to them. So how do they keep things running? How is the massive amount of data in ipfs stored?
- > Currently distributing Nix binaries takes a lot of bandwidth and storage.

  Isn't that solved by ca-derivatons? Are they a prequedite for this?
- Why not torrent? Why not old school style mirrors? Both technologies are well established and are definitely past their hype cycles. 

- I am fully for making nix easier to use for plugins and to better distribute the cache but I am really sceptical about ipfs, the web3 hype and all to hood to be true claims around them.

gador:
>   IPFS cannot be free. The data must be stored somewhere, so it either relies on someones goodwill, donations or the users are the product with their data. Protocol Research Inc., the company behind ipfs.tech, is not a non-profit like eg archive.org and I also could not find a way to donate money to them. So how do they keep things running? How is the massive amount of data in ipfs stored?

The data is not stored with the company. IPFS isn't a datastore on some server, but a protocol. The way it works: 
-  Request some data over IPFS
- The local (or remote, whatever you setup) IPFS daemon is looking for the CID (the content identifier, basically the hash of the data)
- Through the IPFS network, the daemon gets a list of peers who have that hash
- The download from those peers beginns
- Now, you also have that file. When someone else requests it, you are now one of the peers who provide it (until garbage collection starts on your daemon, or you pin it)

So the data is stored, based on demand, on everyones computer running IPFS.  Servers for example could pin their data to keep them available over a longer time. Private persons or non-profit organizations could do the same for their data, or for data they think is relevant.  
> 
>     > Currently distributing Nix binaries takes a lot of bandwidth and storage.
>       
>       
>       Isn't that solved by ca-derivatons? Are they a prequedite for this?
> 

The way I understand it: CA-derivations require less bandwidth and storage, just as IPFS. CA-derivations are a per-requisite to work effectively with IPFS. But CA-derivations still need to be downloaded from one central server (or different substituters). With IPFS, they are downloaded potentially from thousand other nix users and therefore require less bandwidth and storage of a central server.

>   Why not torrent? Why not old school style mirrors? Both technologies are well established and are definitely past their hype cycles.

Mirrors still take time to setup and a specific for the usecase of mirroring the existing data. IPFS distributes the load of storage and bandwidth among the users. 
Torrent OTOH could probably solve similar usecases as IPFS, but I'm not too involved with them, so I can't answer that.  
> 
>   I am fully for making nix easier to use for plugins and to better distribute the cache but I am really sceptical about ipfs, the web3 hype and all to hood to be true claims around them.

I understand. So am I. But I still think IPFS (the protocol) could be beneficial to the nix ecosystem.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: A proper dual hashing solution would be something like `sha256 || sha1` hashes so you have ways to verify not only the root object but every object in it's closure.

shazow:
Here's one simple example why I'm excited about having my Nix store on IPFS: I have several Nix devices. I'd have them peer with each other on IPFS, so anytime a chunk is available on my LAN then it would automagically grab it from a local peer.

As @gador mentioned, IPFS is just a protocol. There's no seeding/leeching "asshole" ratio like there is in BitTorrent culture. There is no obligation to be a public mirror.

It's really a generalization of the BitTorrent protocol, it uses a similar DHT to BitTorrent's Mainline DHT, except there is no concept of individual "trackers" or silo'd bundles. Everything is content-addressable in a similar way that Nix or Git is, which leads to the network topology to dictate what is available.

There's lots of other useful infrastructure around IPFS, like IPNS which allows for mutable pins backed by a private key -- this could map nicely to the equivalent of flake pins.

7c6f434c:
> When someone is on a limited internet connection that is not even enough for their own needs then I don'
t think we can expect from them to spend a good chunk or even all of their connection to upload nix store entries to other people.

I remember that a pretty realistic setting is tiered. With good service you are likely to have unthrottled unmetered connection within your ISP's local network and maybe its good local peers (thousands to tens of thousands of end-users). Then there might be local-ish tier, national-ish tier. And a slow and/or metered international connection.

In that case of course external uploading will be blocked, and on the other hand there is no real slowdown from sharing into the ISP LAN whatever you have already managed to pull from outside.

(I have in my life used an unmetered connection with significant throughput difference between local-federation-of-ISPs and international traffic)

So this usecase is not impossible.

Ericson2314:
@SuperSandro2000 Virtually all of the IPFS stuff can be kept in a plugin if we like. (And we could just not do the parts that cannot). This does build on CA derivations --- in fact my work on  CA derivations was largely done as part of implementing this. It has nothing to do with filecoin or any other monetization strategy.

amjoseph-nixpkgs:
## Nix RFC 133 meeting notes 01-Feb-2023

All mistakes are @amjoseph-nixpkgs' fault.  Please feel free to edit to make corrections.

- @ericson2314
  - it's big
  - possible connections to: lazy trees, libgit2, grahamc's extract-before-verify (?)
  - desires that the RFC be as non-normative as possible
  - two application areas
    - fetchers
    - substituters

- @ericson2314
  - existing hash types: flat, recursive, text
    - add a fourth option for git
      - gador: so this would extend the existing ca-hashing
  - git as substituter would require further changes
    - maybe that would be later (after this rfc?)
  - generalize `StorePath`

- @edolstra
  - thinks git hashes instead of narhash would be interesting
  - migration path is important
  - what is the connection between git-hashes and ipfs?

- @ericson2314
  - ipfs tries to be maximally-general, so it does support git-hashing as one option
    - notable limitation: wants everything to be chunked
  - git has 1MByte max chunk size ("git MTU")
    - this is why git transport doesn't work well for big binaries

- @kevincox: so isn't 1MB a non-starter?

- @ericson2314: for substituters, yes
  - git does have some kind of fragmentation layer strategy in the work
    - software heirtage
    - "advanced data layouts" (?)
  - everything works fine as long as no single file is >1MByte
  - software heritage has an IPFS bridge

- @edolstra
  - so (for git?) the narinfo file would just have another `Hash:` line
  - not clear how this maps to IPFS

- @ericson2314
  - Every hash function is, in some sense, a (Merkle) tree-hash.  But
    conventional hashes (like sha1 sha2) use a maximally-unbalanced trees.
    Tree-hashes (like Blake tree-mode) use a balanced (or maximally-balanced)
    tree instead.

- @ericson2314
  - wants dynamically-loaded plugins (shared objects) for nix

*There was some discussion of dynamically loaded shared-object `nix` plugins
here, which I (@amjoseph-nixpkgs) had a bit of trouble following since this is
the first time I'd heard of it*

- @edolstra
  - What does this mean for `.narinfo`?  Do we have to record something other
    than the git-hash?
  - When you say "plugin", is this like an experimental feature-flag?
    - what *couldn't* you add as a plugin?
      - the CA field
      - what else?
  - Sounds not too had to implement, and easy to revert
    - Could pave the way for fetching via ipfs with a plugin

- @ericson2314
  - wants to move away from `ValidPathInfo` entirely
  - git references to other git objects
    - this is more ipfs-specific

- @edolstra
  - RFC is pretty big
  - what is the plan?
  - git-tree-hashing is easier to get accepted than the more ambitious stuff
    - separate rfc for that?
    - or some of it just an issue in the issue tracker?

- @ericson2314
  - could just rename the RFC, move parts to "future work"
  - the more we could break up the `Store` class the better
  - discussion of the "three-store": substituter, builder store, destination store
    - build-hook protocol is crufty
      - was motivated by github actions?

- @edolstra
  - There is refactoring that is a good idea regardless of this rfc.
  - There are other things that are only because of this rfc.
  - Reluctant to make giant changes to nix for something that might not pan out.
  - Experience with IPFS: somewhat slow.
  - It would be annoying if we ended up making huge changes to nix and ended up
    not using IPFS.
    - Would prefer if things were done in a way that didn't require that
    - Can we have an IPFS substituter that doesn't require changes to nix?

- @ericson2314
  - Agree: would want refactors to not be IPFS-specific.
  - Could imagine a software heritage store, using their API, as a fallback.
    - Would benefit from the same refactors.
    - Software heritage doesn't understand storepaths/`NarHash`es.
  - bittorrent, others don't support git-hashing at all
    - so you have to deal with hash-translation issues

- @edolstra
  - We should decide on the minimal set of changes to the core nix api to
    support IPFS but not create too much work if it doesn't pan out.

- @ericson2314
  - ok; that set is (approximately):
    - git file-hashing for build+fetch
    - content-addressed storepath
    - store interface git-addressed substituter, exact method TBD
  - Can be done via code review rather than rfc meetings.

- @edolstra
  - As a first step: flesh out git file-hashing.
  - (Subsequent) obstacles:
    - In git the top level is always a directory
      - Top level is never a file or symlink
    - Executable bits are a property of the directory rather than the file
  - These obstacles are surmountable.

- @ericson2314
  - two separate problems:
    - nix importing git stuff
    - nix exporting storepaths/nars as git

- @edolstra
  - If we want to replace `NarHash`es with githashes we need to adress this problem
  - Performance: if you package a nix source tree with thousands of directories,
    does that result in thousands of IPFS calls?
    - @ericson2314 no, IPFS has batching mechanisms
    - most problems with IPFS emanate from its DHT design

tennox:
Thanks a lot for keeping us in the loop, and thanks for pushing this forward! 🤘
(And especially @Ericson2314 for your initiative and capacity!)

Really curious and happy about each step that brings two of my favorite technologies (Nix and IPFS) closer together 🥳

(Fun fact & to show why I care: [I hacked on a PoC](https://gitlab.com/txlab/ipnix) to build Nix derivations in an IPFS-based "compute over data" environment, where each flake is built & cached deterministically and is automatically verifiable {which I know is also already possible with current tools} - which would enable a decentralized source of truth/consensus for derivation-/outputs)

Ericson2314:
I made one more part of the detailed design normative (`ValidPathInfo` relaxing, so we can have remote git stores).

I also finished shuffling the motivation now so the IPFS stuff is all confined to future work, and the RFC once again reads well.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+We encourage anyone interested to check our tutorial in https://github.com/obsidiansystems/ipfs-nix-guide/ which demonstrates the above functionality.
+Note at the time of writing this guide uses our original 2020 fork of Nix.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Complexity
+
+The main cost is more complexity to the store layer.
+For a few reason we think this is not so bad.
```
kevincox, line 125:
```suggestion
For a few reasons we think this is not so bad.
```

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+
+Most importantly is the division of the work into a dependency graph of steps.
+This allows us to slowly try out things like IPFS that leverage git hashing, and not commit to more change than we want to up front.
```
kevincox, line 128:
```suggestion
This allows us to slowly try out things like IPFS that leverage Git hashing, and not commit to more change than we want to up front.
```

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+IPFS is a potential solution leveraging content addressing for a P2P CDN.
+It out-of-the-box match's Nix's data model extremely well.
+
+### Source distribution and archival
+
+IPFS also supports git hashing, and so we also provide a good way for people and institutions to "pin" the sources they need, especially if those sources include private ones SWH won't have.
+Finally, per [Obsidian's bridging work](https://github.com/obsidiansystems/go-ipfs-swh-plugin), we have a 3 way integration between IPFS, SWH, and Nix.
+Data can be directly downloaded from SWH via HTTPS, or indirectly via IPFS, which can act as a CDN to not put as much load on SWH's servers.
+
+### Not just IPFS
```
kevincox, line 185: This may be too far off topic but since IPFS was mentioned I recall that we discussed that IPFS has a max block size whereas Git doesn't. Can you elaborate how this will affect our Git hashing implementation and eventually our IPFS implementation? It seems that IPFS is one of the major motivators of this work so it would be nice to have a rough idea of how this will work in the end.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Source distribution and archival
+
+A goal of the Nix ecosystem is to package software in a way that never bitrots.
+Getting in the way of that, however, is the fact source code frequently goes off-line.
+The Software Heritage archive is the best in the world, and a natural partner in this effort.
+
+Unfortunately, as https://www.tweag.io/blog/2020-06-18-software-heritage/ describes at the end, a major challenge is the way nix content-addresses software.
+First of all, Nix hashes sources in bespoke ways that no other project will adopt.
+Second of all, tarballs instead of the underlying files leaking non-normative details (compression, odd perms, etc.).
```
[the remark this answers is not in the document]
lheckemann:
```suggestion
Second of all, using tarballs instead of the underlying files leaks non-normative details (compression, odd perms, etc.).
```
otherwise it should be "leak" ;)

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+
+- **Purpose**: Source distribution and archival
+
+In addition to the various forms of content-addressing Nix supports today ("text", "fixed" with either "flat" or "nar" serialization of file system objects), Nix should support Git hashing.
+This support entails two basic things:
+
+ - Content addresses are used to compute store paths.
+ - Content addresses are used to verify store object integrity.
+
+Git hashing would not support references (since references in Nix's sense are not a Git concept), but that is not an issue for the intended use-case of exchanging source code.
```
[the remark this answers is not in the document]
lheckemann: I don't think supporting fewer formats overall is desirable, and find it hard to imagine a future where NAR goes away -- Nix's astounding capabilities for running software from the past would suffer significantly from such a deprecation. I'm more in favour of supporting _more_ formats :)

As for extending Git trees, I think representing it in-band as you suggest would open a lot of potential for breakage, and a new format that builds on the strengths of Git's representation while avoiding its weaknesses -- as [Tvix is doing, with a conscious and justified rejection of Git trees](https://code.tvl.fyi/tree/tvix%2Fstore%2Fdocs%2Fwhy-not-git-trees.md) -- is probably a better path forwards.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+
+   We plan on more formally proposing this next.
+
+## Git and Nix's file system data models do not entirely coincide
+
+Nix puts the permission info of a file (executable bit for now) with that file, whereas Git puts it with the name and hash in the directory.
+The practical effect of this discrepancy is that a root file (as opposed to directory) in Nix has permission info, but does not in Git.
+
+If we are trying to convert existing Nix data into Git, this is a problem.
+Assuming we treat "no permission bits" as meaning "non-executable", we will have a partial conversion that will fail on executable bare files.
```
lheckemann, line 152:
```suggestion
Assuming we treat "no permission bits" as meaning "non-executable", we will have a partial conversion that will fail on executable files without a parent directory.
```

But does the Git model even need to support files without a parent directory? I can see some value in allowing content-addressing models that don't support all things that can live in a store.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+
+ - Content addresses are used to compute store paths.
+ - Content addresses are used to verify store object integrity.
+
+Git hashing would not support references (since references in Nix's sense are not a Git concept), but that is not an issue for the intended use-case of exchanging source code.
+
+## Git file hashing for `buitins.fetch*`
+
+- **Purpose**: Source distribution and archival
+- **Depends on**: Git file hashing,
```
lheckemann, line 85: A visual representation of this graph would be very helpful for understanding the way that the parts interact with each other :)

Ericson2314:
I need to catch back up on this completely, but I wanted to say right now that our work bridging SWH and IPFS now has a blog post. https://blog.obsidian.systems/software-heritage-bridge/ This does not relate to the RFC properly, but should hope provide context for the motivation section.

kevincox:
It looks like this RFC is mostly good to go. There are a bunch of open threads but they appear to be effectively resolved or trivial. I think it makes sense to clean up these threads and then we are ready for FCP? Do we want to have one last shepherd meeting to close everything off or just move to FCP?

rnhmjoj:
Given the recent [events](https://discourse.nixos.org/t/the-nixos-foundations-call-to-action-s3-costs-require-community-support) regarding the binary cache, it would probably be a good time to discuss this.

asymmetric:
@amjoseph-nixpkgs as the shepherd leader, do you think this RFC should advance to FCP?

Ericson2314:
I was supposed to push one more thing, and then FCP.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,346 @@
+---
+feature: git-hashing
+start-date: (fill me in with today's date, YYYY-MM-DD)
```
lucasew, line 3: Missing date

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,346 @@
...
+Most importantly is the division of the work into a dependency graph of steps.
+This allows us to slowly try out things like IPFS that leverage git hashing, and not commit to more change than we want to up front.
+
+Even if we do end up adopting everything though, we think for the following two reasons the complexity can still be kept manageable:
+
+2. Per the abstract vs concrete model of the nix store in https://github.com/NixOS/nix/pull/6877 , everything we are doing is simply flushing out alternative interpretations of the abstract model.
+   This is the sense in which we are "removing the weaknesses and restrictions that make additional features appear necessary" per the Scheme mantra cited above:
+   Instead of extending the model with new features, we are relaxing concrete model assumptions (e.g. references are always opaque store paths) while keeping the abstract model the same.
+
+3. We also support plans to decouple the layers of Nix further, and update our educational and marketing material to reflect it.
```
amjoseph-nixpkgs, line 145:
```suggestion
2. We also support plans to decouple the layers of Nix further, and update our educational and marketing material to reflect it.
```

Github's markdown renderer doesn't seem to be renumbering these for some reason.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,346 @@
...
+   With Flakes and other post-2.0 features, the upper layer of Nix have gained an enormous amount of flexibility and sophistication.
```
amjoseph-nixpkgs, line 146:
```suggestion
   With Flakes and other post-2.0 features, the upper layers of Nix have gained an enormous amount of flexibility and sophistication.
```

amjoseph-nixpkgs (commented):
Just a few typos.

amjoseph-nixpkgs:
> @amjoseph-nixpkgs as the shepherd leader, do you think this RFC should advance to FCP?

After consultation with the committee: the RFC's content is in essentially FCP-ready format, but we're taking one last quick pass for grammar/typos/etc.  Anybody who doesn't mind overlooking those is welcome to register their final comments early; there won't be any more major revisions.

FCP is imminent.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+
+Content-addressing opens up a *huge* design space of solutions to get around such problems.
+
+The first steps proposed below do *not* tackle this problem directly, but it lays the ground-work for future experiments in this direction.
+
+## Source distribution and archival
+
+A goal of the Nix ecosystem is to package software in a way that never bitrots.
+Getting in the way of that, however, is the fact source code frequently goes off-line.
+The Software Heritage archive is the best in the world, and a natural partner in this effort.
```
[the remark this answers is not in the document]
Ericson2314: Sure, sounds good

rfcs/0133-git-hashing.md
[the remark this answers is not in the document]
Ericson2314:
```suggestion
Second of all, hashing tarballs instead of the underlying files leads non-normative details (compression, odd perms, etc.).
```

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+
+Unfortunately, as https://www.tweag.io/blog/2020-06-18-software-heritage/ describes at the end, a major challenge is the way nix content-addresses software.
+First of all, Nix hashes sources in bespoke ways that no other project will adopt.
+Second of all, tarballs instead of the underlying files leaking non-normative details (compression, odd perms, etc.).
+
+We should natively support git file hashing, which Git repos and Software Heritage both support.
+This will completely obliterate these issues.
+
+Overall, are building out a uniform way to work with source code, regardless of its origins or the exact tools involved.
```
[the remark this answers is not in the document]
Ericson2314:
```suggestion
Overall, we are building out a uniform way to work with source code, regardless of its origins or the exact tools involved.
```

rfcs/0133-git-hashing.md
Ericson2314:
```suggestion
start-date: 2022-08-27
```

Ericson2314: Thanks

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: Collapsing IPFS threads. Since the RFC is now about Git hashing, those drawbacks or more intrinsic to the goal. (Whereas with IPFS we could hash in other way.s)

rfcs/0133-ipfs.md
Ericson2314: Collapsing IPFS threads.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: Collapsing IPFS threads.

rfcs/0133-ipfs.md
Ericson2314: Collapsing IPFS threads.

rfcs/0133-ipfs.md
Ericson2314: Collapsing IPFS threads.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: Collapsing IPFS threads.

rfcs/0133-ipfs.md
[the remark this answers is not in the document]
Ericson2314: Collapsing IPFS threads.

rfcs/0133-git-hashing.md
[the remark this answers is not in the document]
Ericson2314: Yes!

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,317 @@
...
+
+## Build adoption through seamless interop
```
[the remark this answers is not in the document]
Ericson2314: I am fine just cutting it entirely if people think the other motivation is sufficient.

rfcs/0133-git-hashing.md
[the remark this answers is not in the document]
Ericson2314:
```suggestion
Git hashing would not (in this first proposed version) support references, since references in Nix's sense are not part of Git's data model.
This is OK for now; encoding references is not needed for the intended initial use-case of exchanging source code.
```

rfcs/0133-git-hashing.md
Ericson2314: It won't work for larger files until that restriction is lifted, but there is a few ideas on how to do that.

Ericson2314: Collapsing IPFS threads. I hope even if we just were to do git hashing and not IPFS the normative part would still stand on it's own so don't need to resolve this completely now?

rfcs/0133-git-hashing.md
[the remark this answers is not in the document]
kevincox: I don't mind the paragraph but I don't think it is critical to the proposal either.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,346 @@
...
+
+# Future work --- IPFS
+[future]: #future-work-ipfs
+
+## Motivation
+
+### Binary Distribution
+
+IPFS is a potential solution leveraging content addressing for a P2P CDN.
+It out-of-the-box match's Nix's data model extremely well.
```
kevincox, line 202: This sentence doesn't read properly for me. Did you mean to say "It's out-of-the box data model matches Nix's data model very well."?

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,346 @@
...
+- `IpfsCid`: Native content address for IPFS.
+
+- `IpfsEncoding`: Enough info to deterministically rebuild the IPFS representation from a non-IPFS copy of the store object.
+
+   For now, `IpfsEncoding` will only support `unixfs-nar`, which works as follows:
+
+   The NAR is itself wrapped in IPFS's [UnixFS](https://github.com/ipfs/specs/blob/main/UNIXFS.md).
+   This other format can be extracted from the CID (which is conceptually a pair of encoding metadata and a hash).
+   For now, only IPFS's "unixfs" is supported.
+   `chunking`, `layout`, and `max-width` are tuning parameters for unixfs [described in the UnixFS spec](https://github.com/ipfs/specs/blob/main/UNIXFS.md#importing).
```
kevincox, line 276: For now do we want to fix the supported parameters to simplify implementation? I think it may make sense to keep the format as a JSON blob as above but state that at this time only exactly that JSON string is supported. (Although readers are encouraged to support more flexibility in parameters).

This way we have a simple and well defined set of parameters to start. We can always expand and add more as we are sure that relevant clients support them.

kevincox (approved):

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,346 @@
...
+NarSize: 197528
+References: 7gx4kiv5m0i7d7qkixq2cwzbr10lvxwc-glibc-2.27 gdh8165b7rg4y53v64chjys7mbbw89f9-hello-2.10
+Deriver: 5sj6fdfym58sdaf3r5p87v4l8sj2zlvn-hello-2.10.drv
+Sig: cache.nixos.org-1:K0thQEG60rzAK8ZS9f1whb7eRlIshlMDJAm7xvX1oF284H+PTqlicv/wGW6BIj+wWWONHvUZ2MYc+KDArekjDA==
+```
+
+This RFC proposes new key-value pairs that in this example would be:
+
+```
+IpfsCid: Qmf8NfV2hnq44RoQw9vxmSpGYTwAovA8FUCxeCJCqmXeNN
```
flokli, line 258: It's not entirely clear to me why we're using IPFS CIDs (and unixfs) here, instad of let's say, pointing to a raw git (blob or tree) object. This feels like a more logical choice, considering the first part starts with git tree hashing.

Can you make it more clear why unixfs was chosen?

In addition to that, I'd personally be in favor having `IpfsCid / IpfsEncoding` be one single field, and having specified in that RFC already how additional representations are handled. (Can these fields be repeated, or is this a ?-separated string of multiple possible representations?)

This allows us gradually introducing more representations, and pointing to more than one representation, without worrying about older Nix versions not supporting these protocols to not be able to parse it at all).

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,346 @@
...
+
+IPFS also supports git hashing, and so we also provide a good way for people and institutions to "pin" the sources they need, especially if those sources include private ones Software Heritage won't have.
+Finally, per [this IPFS plugin](https://blog.obsidian.systems/software-heritage-bridge/), we can have a 3 way integration between IPFS, Software Heritage, and Nix.
+Data can be directly downloaded from Software Heritage via HTTPS, or indirectly via IPFS.
+
+This would hopefully supercede the direct Software Heritage store outlined above.
+This is because Software Heritage is primarily an archive, and we should be careful not to waste their limitted resources with copious egress bandwidth.
+IPFS can act as a CDN to not put as much load on Software Heritage's servers.
+
+### Not just IPFS
```
flokli, line 214: I initially didn't see the other marked-as-resolved threads, but I think this should be reworded to focus on the content-adressability, and marking IPFS as one possible vehicle/transport for that data.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,346 @@
...
+The `IpfsCid` and `IpfsEncoding` are likewise informational, describing how to get the store object:
+
+- `IpfsCid`: Native content address for IPFS.
+
+- `IpfsEncoding`: Enough info to deterministically rebuild the IPFS representation from a non-IPFS copy of the store object.
+
+   For now, `IpfsEncoding` will only support `unixfs-nar`, which works as follows:
+
+   The NAR is itself wrapped in IPFS's [UnixFS](https://github.com/ipfs/specs/blob/main/UNIXFS.md).
+   This other format can be extracted from the CID (which is conceptually a pair of encoding metadata and a hash).
```
flokli, line 274: It's not clear to me what "wrapping" means here. Do you /decompose/ the NAR file, and use unixfs for its "seekability", or do you put the NAR file as a single file inside a unixfs? If the latter is the case, why bother wrapping unixfs around it at all (just to workaround the blocksize problem)?, and not just querying the NAR as a sha256 blob?

rfcs/0133-git-hashing.md
flokli: I added another comment above - this RFC should at least specify how alternative/additional "not-yet-understood" representations would look like, to ensure we can add more of them without breaking code only implementing this RFC.

rfcs/0133-git-hashing.md
kevincox: >  I'd personally be in favor having IpfsCid / IpfsEncoding be one single field

I think it is best to keep it separated. For the common operation of retrieving the file you only need the CID. The encoding settings are only needed if you have the store path and want to add it to IPFS to share with others which is probably a relatively rare operation.

rfcs/0133-git-hashing.md
kevincox: I believe that this is answered here: https://github.com/NixOS/rfcs/pull/133/files/165979cbedf6bff63dbf2379d61c8b38b631d63e#diff-4d468043c922b901f9d6c43b21703e397e19abdca0cb219ffb9e9005d9b9e5c2R278-R279

rfcs/0133-git-hashing.md
kevincox: Actually on further review it is somewhat mixed. `"method":"wrapped-nar"` is needed for both reading and resharing. Maybe it makes sense to move that property to `IpfsCid` (which will then need to be renamed) so that all reading properties are in the same place.

amjoseph-nixpkgs:
On behalf of the shepherd committee, I declare this RFC to be in the Final Comment Period!

rfcs/0133-git-hashing.md
Ericson2314: I think I meant
```suggestion
It, out-of-the-box, matches Nix's data model extremely well.
```

kevincox: Ok that works for me. However I think that it is easier to read like this

```suggestion
Out-of-the-box it matches Nix's data model extremely well.
```

rfcs/0133-git-hashing.md
flokli: As written above, we might need to specify how multiple different encodings are specified, and thinking about "describing the NAR file twice, but with different chunking parameters" would be one usecase where it'd be quite complicated if cid and "generation method/interpretation" were two separate fields.

kevincox: Would we ever want to describe the same NAR file twice with different chunking parameters? It seems that in general the creator should pick the parameters and to ensure maximum sharing everyone should re-use them.

In theory we don't ever need chunking parameters. Ideally the producer would share on IPFS and everyone would just download that CID. The main reason that we provide them is to help bootstrapping. For example right now https://cache.nixos.org isn't going to publish via IPFS, so the standardized chunking parameters will allow the community to do that. If the producer does publish the CID themselves then everyone would just download that and now worry about how it was chunked.

However interpretation is important as once you download the CID you need to know how to use it. It is needed for both reading and writing.

I agree that **if** we want to support multiple CIDs then it probably makes sense to group everything together to avoid parallel corresponding lists. However if we don't want to support mutliple CIDs then I think it makes sense to separate "things required to download and use" from "things required if you want to start seeding the CID".

rfcs/0133-git-hashing.md
Ericson2314: Sure, though I think I'll put back in the single comma :)

rfcs/0133-git-hashing.md
Ericson2314: This is all non-normative future work right? At least I thought it was.

rfcs/0133-git-hashing.md
Ericson2314: I think "This RFC" is misleading, because this is the the non-normative future work section, no?

rfcs/0133-git-hashing.md
Ericson2314: We perhaps should just delete all this text as move it to a draft new RFC :)

rfcs/0133-git-hashing.md
Ericson2314: The no-yet-understood content addressing method boils down to unrecognized CA fields `ValidPathInfo` (narinfo, DB, etc.). But we can call that out explicitly.

rfcs/0133-git-hashing.md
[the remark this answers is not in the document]
Ericson2314: It's gone. I can just blog about it :D

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,170 @@
...
+## Source distribution and archival
+
+Source code used by Nix expressions frequently goes off-line. It would be beneficial if there was some resistance to this form of bitrot.
+The Software Heritage archive stores much of the source code that Nix expressions use. They would be a natural partner in this effort.
+
+Unfortunately, as https://www.tweag.io/blog/2020-06-18-software-heritage/ describes at the end, a major challenge is the way Nix content-addresses software.
+First of all, Nix hashes sources in bespoke ways that no other project will adopt.
+Second of all, hashing tarballs instead of the underlying files leads non-normative details (compression, odd perms, etc.).
+
+We should natively support git file hashing, which Git repos and Software Heritage both support.
```
fricklerhandwerk, line 42:
```suggestion
We should natively support Git file hashing, which is supported both by Git repos and Software Heritage.
```

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,170 @@
...
+   Those store can, however, understand content addresses.
+   And from such a name + content address, we can always produce a store path again, so there is no loss of functionality with existing stores.
+
+- Relax `ValidPathInfo` to merely require that *either* the pair of `NarHash` and `NarSize` or just `CA` alone be defined.
+
+  As described in the first step, currently `NarHash` and `NarSize` are the *normative* fields which are used to verify a store object.
+  But if the store object is content-addressed, we don't need these, because the content address (`CA` field) will also suffice, all by itself.
+  
+  Existing Nix stores types are still required to contain a `NarHash` and `NarSize`, which is good for backwards compat and don't come with a cost.
+  Only new nix-agnostic store types would take advantage of these new, relaxed rules.
```
fricklerhandwerk, line 105:
```suggestion
  Only new Nix-agnostic store types would take advantage of these new, relaxed rules.
```

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,170 @@
...
+
+The main cost is more complexity to the store layer.
+For a few reasons we think this is not so bad.
+
+Most importantly is the division of the work into a dependency graph of steps.
+This allows us to slowly try out things like IPFS that leverage Git hashing, and not commit to more change than we want to up front.
+
+Even if we do end up adopting everything though, we think for the following two reasons the complexity can still be kept manageable:
+
+1. Per the abstract vs concrete model of the nix store in https://github.com/NixOS/nix/pull/6877 , everything we are doing is simply flushing out alternative interpretations of the abstract model.
```
fricklerhandwerk, line 126:
```suggestion
1. Per the abstract vs concrete model of the Nix store in https://github.com/NixOS/nix/pull/6877, everything we are doing is simply flushing out alternative interpretations of the abstract model.
```

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,170 @@
...
+   This is the sense in which we are "removing the weaknesses and restrictions that make additional features appear necessary" per the Scheme mantra cited above:
```
fricklerhandwerk, line 127:
```suggestion
   This is the sense in which we are, per the Scheme mantra, "removing the weaknesses and restrictions that make additional features appear necessary":
```

It doesn't seem to be mentioned anywhere else.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,170 @@
...
+
+Nix puts the permission info of a file (executable bit for now) with that file, whereas Git puts it with the name and hash in the directory.
+The practical effect of this discrepancy is that a root file (as opposed to directory) in Nix has permission info, but does not in Git.
+
+If we are trying to convert existing Nix data into Git, this is a problem.
+Assuming we treat "no permission bits" as meaning "non-executable", we will have a partial conversion that will fail on executable files without a parent directory.
+Tricks like always wrapping everything in a directory get around this, but then we have to be careful the directory is exactly as expected when "unwrapping" in the other direction.
+
+For now, we only focus on ingesting data *from* Git *to* Nix, and this side-steps the issue.
+That conversation is total (though not surjective), and so there is no problem for now.
```
RaitoBezarius, line 150:
```suggestion
That mapping is total, i.e. all Git data can be mapped, and injective, i.e. each Git data has a unique Nix data representative (though not surjective, i.e. not all Nix data can be represented as a piece of Git data), and so there is no problem for now.
```

(saying total and "not surjective" is confusing to me, I have no strong opinion on keeping conversion, but using function vocabulary with conversion is a bit strange.)

feel free to drop the `i.e.` clarifications

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,170 @@
...
+
+1. Per the abstract vs concrete model of the nix store in https://github.com/NixOS/nix/pull/6877 , everything we are doing is simply flushing out alternative interpretations of the abstract model.
+   This is the sense in which we are "removing the weaknesses and restrictions that make additional features appear necessary" per the Scheme mantra cited above:
+   Instead of extending the model with new features, we are relaxing concrete model assumptions (e.g. references are always opaque store paths) while keeping the abstract model the same.
+
+2. We also support plans to decouple the layers of Nix further, and update our educational and marketing material to reflect it.
+   With Flakes and other post-2.0 features, the upper layers of Nix have gained an enormous amount of flexibility and sophistication.
+   RFCs like this show that the so-far more sleepy lower layers also have plenty of potential to gain sophistication too.
+
+   Embracing layering on technical, educational, communications, and managerial levels can scale our capacity to manage complexity and sophistication without the project growing out of control.
```
fricklerhandwerk, line 134:
```suggestion
```

This doesn't seem helpful to convey the argument.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,170 @@
...
+   It will "divide and conquer" the project so the interfaces between each layer are still rigorously enforced preventing a combinatorial explosion in complexity.
+   That frees up "complexity budget" for project like this.
```
fricklerhandwerk, line 136:
```suggestion
   That frees up "complexity budget" for projects like this.
```

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,170 @@
...
+
+   We plan on more formally proposing this next.
```
fricklerhandwerk, line 138:
```suggestion
```

This seems unrelated and is unspecific.

fricklerhandwerk (approved):

sambacha:
different versions of IPFS have different default chunking parameters so ipfs add gives different CIDs. 

you have two options:
1. generate a .car file (archive of IPLD blocks) and send that to the pinning service.
2. Send raw IPLD blocks through bitswap, for example with a tool like auspinner 

guess what, There is defaults in the reference implementation of ipfs (balanced dag, 256KiB dag, protobuf wrapping) but pinning services don't follow them necessarily.

Additional info:
https://stackoverflow.com/questions/59183147/are-cid-different-for-the-same-file-but-uploaded-by-2-users-in-ipfs
https://discuss.ipfs.tech/t/cid-concept-is-broken/9733

Here is a google 'version' of IPFS, note the segmentation of the hash in the URI format:

https://github.com/google/ent

rfcs/0133-git-hashing.md
Ericson2314: Ah yes this a great, and I'll keep the "i.e." for those that aren't familiar with the math.

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,170 @@
...
+This will completely obliterate these issues.
+
+Overall, we are building out a uniform way to work with source code, regardless of its origins or the exact tools involved.
+
+# Detailed design
+[design]: #detailed-design
+
+Each item can be done separately provided its dependent items are also done.
+These are the items we wish to commit to at this time.
+(The goals mentioned future work are, in a separate document, also broken down into a dependency graph of smaller steps.)
```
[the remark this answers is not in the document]
Ericson2314:
```suggestion
(The goals mentioned under [future work](#future-work) are, in a separate document, also broken down into a dependency graph of smaller steps.)
```

rfcs/0133-git-hashing.md
Ericson2314: Guess it got removed!

rfcs/0133-git-hashing.md
```diff
@@ -0,0 +1,168 @@
...
+   This is the sense in which we are, per the Scheme mantra, "removing the weaknesses and restrictions that make additional features appear necessary":
+   Instead of extending the model with new features, we are relaxing concrete model assumptions (e.g. references are always opaque store paths) while keeping the abstract model the same.
+
+2. We also support plans to decouple the layers of Nix further, and update our educational and marketing material to reflect it.
+   With Flakes and other post-2.0 features, the upper layers of Nix have gained an enormous amount of flexibility and sophistication.
+   RFCs like this show that the so-far more sleepy lower layers also have plenty of potential to gain sophistication too.
+
+   Embracing layering on technical, educational, communications, and managerial levels can scale our capacity to manage complexity and sophistication without the project growing out of control.
+   It will "divide and conquer" the project so the interfaces between each layer are still rigorously enforced preventing a combinatorial explosion in complexity.
```
Ericson2314, line 135:
```suggestion
   Layering will "divide and conquer" the project so the interfaces between each layer are still rigorously enforced preventing a combinatorial explosion in complexity.
```

Ericson2314:
Excited to report we now have the first part implemented: the [`git-hashing`](https://nixos.org/manual/nix/unstable/contributing/experimental-features#xp-feature-git-hashing) experimental feature exists, and can use git hashing for adding objects to the store!

For how to use this, one can consult https://github.com/NixOS/nix/blob/master/tests/functional/git-hashing/simple.sh

In particular,
```bash
nix store add --mode git --hash-algo sha1 <file>
```
should work!

PR #132, opened 2022-08-25
[RFC 0132] Meson Builds Nix

AndersonTorres:
This RFC proposes to use Meson description files to deploy Nix.

[Rendered!](https://github.com/AndersonTorres/rfcs/blob/meson-builds-nix/rfcs/0132-meson-builds-nix.md)

PR #134, opened 2022-09-15
[RFC 0134] Carve out a store-only Nix

Ericson2314:
[Rendered](https://github.com/nixos/rfcs/blob/master/rfcs/0134-nix-store-layer.md)

[Tracking issue](https://github.com/NixOS/nix/issues/7868)


rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+We should instead fully embrace it:
+
+- Docs
+
+  - More advanced documentation can explain layering for those that want a deeper understanding of Nix.
+
+  - Even more basic documentation can still benefit from separate terminology before the layering is fully explained.
+    See https://www.haskellforall.com/2022/08/stop-calling-everything-nix.html for a phenomenal take-down of how calling everything "Nix" today confuses users and leaves them unable to articulate what parts of the ecosystem are frustrating then.
+
+- Separate executable ensure lower layers build in isolation, new integration tests those executable without aid of high-layer info.
```
lheckemann, line 41: I'm having difficulty parsing this, maybe this sentence could do with a rewrite.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+As Nix grows more popularity, it will be inevitable that different groups want to explore in different directions.
+This is the *pluralism* of a larger community, and we should embrace it as a strength.
+We do that be fostering a *marketplace of ideas*.
+
+There are many possible ways in which to write down packages, The Nix language and Nixpkgs idioms, and Guix, for example, are just two points in a much larger space.
+There are also many possible ways set up build farms.
+Our current central dispatcher, many remote-builder agents model, point-to-point protocol model is also just point in a much larger design space.
+
+The "derivation language" and store *interface* however, seems to me at least to be a very natural design.
```
lheckemann, line 81: By derivation language, you mean the ATerm-based language that drv files are written in?

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+# Future work
+[future]: #future-work
+
+## Nix Store Team
+
+Now that we have this division in the implementation, we also have the opportunity to leverage it for governance purposes.
+An official, NixOS-foundation-authorized team could be set up to manage store layer design decisions (below the threshold of needing an RFC) that don't effect that the rest of Nix.
+\[Some sort of decision that affect all layers is out of scope, must be deliberated with stakeholders from other layers too, probably should be RFC due to such large scope.\]
+
+To be clear, this is to say we should abandon the idea of Nix as a whole.
```
lheckemann, line 236: Missing negation?

lheckemann (commented):
I like the idea and hope we can get this going!

rfcs/0134-nix-store-layer.md
Ericson2314: Haha, yes.

mweinelt: https://en.wikipedia.org/wiki/Freudian_slip :thinking:

rfcs/0134-nix-store-layer.md
Ericson2314: How's this?
```suggestion
- Separate executables to ensure lower layers build in isolation without higher layers,

   - Integration tests for those executables that also don't use the higher layers to ensure those executables work in isolation. 
```
Frankly this is somewhat a restatement of the detailed design so I would be fine cutting it too,

rfcs/0134-nix-store-layer.md
Ericson2314: The abstract syntax of that. Basically "a derivation is an execve system call arguments, input black box data, and input other drvs' outputs". I think that right there is a very natural design.

rfcs/0134-nix-store-layer.md
Ericson2314, line 236:
```suggestion
To be clear, this is *not* to say we should abandon the idea of Nix as a whole.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+- [*Tvix*](https://cs.tvl.fyi/depot/-/tree/tvix) is a new implementation of the Nix language evaluator,
+
+- [*go-nix*](https://github.com/nix-community/go-nix) is a new implementation of the store layer.
+
+First of all, the fact that they are planning on two completely separate implementation oriented around this same "narrow waist" is testament to the appeal of the design.
+
+Second of all, note that per they have separate, orthogonal experiments they wish to run on both sides of the store interface divide.
+Above, they want to experiment with radically different evaluation strategies, especially to speed up Nixpkgs evaluation.
+Below, they want to experiment with the standardized containerization technologies that already exist for new ways of sandboxing and distributing builds with less bespoke Nix-specific code.
+They also want to apply the layering paradigm *within* go-nix, fostering even more modularity.
```
asymmetric, line 109: Where do you get all of this info about go-nix (e.g. what they want to experiment with, and where)? I can't find it in the links you mention.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+\[Some sort of decision that affect all layers is out of scope, must be deliberated with stakeholders from other layers too, probably should be RFC due to such large scope.\]
+
+To be clear, this is *not* to say we should abandon the idea of Nix as a whole.
+There can still be governance of Nix as a whole; this team, and similar hypothetical, say, Flakes, Nix language, or User Experience teams would ultimately need to report to.
+The goal is not to overreact, but strike a balance between:
+
+1. Making sure Nix as a whole continues to make sense
+2. Make sure layers make sense in isolation not just in the context of the way they are currently used.
+
+## Standardization across projects
```
patricksjackson, line 243: Have you reached out to Guix yet? Or if not, when do you think it would be valuable to get their input?

rfcs/0134-nix-store-layer.md
Ericson2314: Taking on https://matrix.to/#/!YUnRYAzgytLSZbBhbx:hackint.org more information should be published soon after which I'll update the RFC to link it.

rfcs/0134-nix-store-layer.md
Ericson2314: I have! But I think for this RFC the ball is purely in our court. This RFC is not about making technical decisions that would impact Guix, but about ratifying the idea of a store layer that is usable in arbitrary ways including the Nix language and Scheme via Guix.

I'll continue talking to them, and I encourage anyone else that wants to reach and say hi too — let's build more bridges between our communities — but if we don't do something like this I don't see why they should take us very seriously about store layer interop.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+The division between `libnixstore`, `libfetchers`, `libexpr`, etc. is not yet exposed to users, emphasized in documentation (though this is changing thanks to @Fricklerhandwerk's efforts with new [architecture documentation](https://github.com/NixOS/nix/pull/6888)!).
+
+We should instead fully embrace it:
+
+- Docs
+
+  - More advanced documentation can explain layering for those that want a deeper understanding of Nix.
+
+  - Even more basic documentation can still benefit from separate terminology before the layering is fully explained.
+    See https://www.haskellforall.com/2022/08/stop-calling-everything-nix.html for a phenomenal take-down of how calling everything "Nix" today confuses users and leaves them unable to articulate what parts of the ecosystem are frustrating then.
```
7c6f434c, line 39: Is it «frustrating then» or «frustrating when» or «frustrating them»?

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+[examples-and-interactions]: #examples-and-interactions
+
+## Maximal Nix unlike today
+
+I cannot emphasize this enough, but the interface of maximal store+exprs+flakes Nix remains *exactly* like today.
+In particular, lower level commands can be used with higher level "installables" (arguments), so e.g.
+```
+nix show-derivation flake#bar
+```
+will still work.
```
7c6f434c, line 188: Should implementation of this be migrated to a model where no executable links `libstore` and `libexpr` at once, and the CLI executable invokes the store-only backend (possibly for multiple stores as in case of `nix copy`)?

AndersonTorres:
In a preliminar look, this can affect #132 somehow.

rfcs/0134-nix-store-layer.md
bjornrud: I would just use frustrating,
```suggestion
    See https://www.haskellforall.com/2022/08/stop-calling-everything-nix.html for a phenomenal take-down of how calling everything "Nix" today confuses users and leaves them unable to articulate what parts of the ecosystem are frustrating.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+Yes, strictly speaking, we only need a stable daemon protocol to accommodate that goal, which we have.
+But ensuring the C++ Nix components can be built separately for use in isolation further send the message that such experimentation is *good* and *welcomed*.
+Messaging matters, and making our layered architecture "official" as this RFC proposes I think sends a better message.
+
+### Guix
+
+Guix is more diverged from Nix than Tvix + go-nix, and thus hints more at the end breadth of the design space yet to be explored.
+
+The store layer is the same, but the layers above, instead of being a implementing of the Nix language, is a completely different design with Guile Scheme.
```
amjoseph-nixpkgs, line 122: I heard that Guix started out using a store format which was compatible with Nix's `libstore` (presumably so they could use the Nix tools to bootstrap their development process) but that their store is no longer perfectly compatible.

If this is true, does anybody know what motivated them to diverge?  Or did they simply not adopt some forward-incompatible change that Nix made to its store format?

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+I think these are great goals, and for the sake of the ecosystem as a whole, it should be as easy as possible to run such new experiments.
+In particular, a novel evaluator should be usable with the standard C++ Nix store layer, and a novel store layer should be reusable with the standard C++ Nix evaluator.
+
+Yes, strictly speaking, we only need a stable daemon protocol to accommodate that goal, which we have.
+But ensuring the C++ Nix components can be built separately for use in isolation further send the message that such experimentation is *good* and *welcomed*.
+Messaging matters, and making our layered architecture "official" as this RFC proposes I think sends a better message.
+
+### Guix
+
+Guix is more diverged from Nix than Tvix + go-nix, and thus hints more at the end breadth of the design space yet to be explored.
```
amjoseph-nixpkgs, line 120: TVL has taught me an excellent software engineering trick: if you want to ensure a clean separation between two parts of your codebase, write the two parts in two different languages whose communities see each other as competitors.

Absolutely brilliant; I tip my hat to these folks for figuring that out. :)

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+What should the store-only Nix be called?
+
+# Future work
+[future]: #future-work
+
+## Nix Store Team
+
+Now that we have this division in the implementation, we also have the opportunity to leverage it for governance purposes.
+An official, NixOS-foundation-authorized team could be set up to manage store layer design decisions (below the threshold of needing an RFC) that don't effect that the rest of Nix.
```
amjoseph-nixpkgs, line 233:
```suggestion
An official, NixOS-foundation-authorized team could be set up to manage store layer design decisions (below the threshold of needing an RFC) that don't affect that the rest of Nix.
```

rfcs/0134-nix-store-layer.md
Ericson2314: https://code.tvl.fyi/about/tvix/docs/components.md I'll update this once I can get permalink.

rfcs/0134-nix-store-layer.md
Ericson2314: I think I lean against that, but regardless I would defer such questions until after this RFC and this first step. If we can be sure full Nix stays the same in the short run, that makes this a lot less risky — hardly any risk at all, I think.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+Allow building a store-only version of Nix.
+This is a Nix executable that links `libstore` but not `libfetchers`, or `libexpr`.
+Plenty of commands like `nix daemon`, `nix log`, and the `nix store` sub-commands don't care about evaluation, fetching, or flakes at all.
+https://github.com/NixOS/nix/issues/6182 is a draft PR implementing this, splitting `libcmd` into two parts so the CLI code reused.
+We will finish it off with as many commands as are reasonable to include, and merge it.
+
+Initially, we can test this store-only version of Nix with no changes to the test suite, by running full Nix with the store-only Nix's daemon.
+Support for testing Nix against a separately-built daemon already exists and is in use today.
+
+The store-only Nix and its tests should be built as part of CI, as "first class" as our existing CI jobs.
```
fricklerhandwerk, line 156: I see no reason to emphasize this.
```suggestion
The store-only Nix and its tests should be built as part of CI.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+That means both in the channel-blocking Hydra eval, and per PR.
```
fricklerhandwerk, line 157:
```suggestion
That means both in the channel-blocking Hydra evaluation, and for each pull request.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+If we hit the limits of Github Actions in per-PR CI, we should consider using Hydra instead / in addition, something that has already been discussed.
```
fricklerhandwerk, line 158: Please link where it has been discussed.

```suggestion
If we hit the limits of Github Actions in per-pull-request CI, we should consider using Hydra instead or in addition, something that has already been discussed.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+## 2. Manual
+
+It should be possible to build a store-only manual without information on the other layers too.
```
fricklerhandwerk, line 162:
```suggestion
It should be possible to build a store-only manual without information on the other layers, too.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+This would be the manual that is distributed with the store-only Nix.
+Of course, store-only and full  can share sections, so we aren't duplicating work.
```
fricklerhandwerk, line 164:
```suggestion
Of course, store-only and full Nix can share sections, so we aren't duplicating work.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+## 3. Website
+
+The fully and store-only version of Nix should both be presented for download on the website.
```
fricklerhandwerk, line 168:
```suggestion
The full and store-only version of Nix should both be presented for download on the website.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+We should write new tests that don't use the Nix language.
+E.g. we might create a `read-derivation` complement of `show-derivation` that accepts a nicer JSON representation of a derivation as input.
+This will allow the store-only Nix to be tested in isolation, but these tests can also be used with full Nix.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+## Maximal Nix unlike today
+
+I cannot emphasize this enough, but the interface of maximal store+exprs+flakes Nix remains *exactly* like today.
```
fricklerhandwerk, line 183:
```suggestion
The interface of maximal Nix, including the store, the Nix language, and flakes, remains *exactly* like today.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+In particular, lower level commands can be used with higher level "installables" (arguments), so e.g.
+```
+nix show-derivation flake#bar
+```
+will still work.
+
+## Good tasks anyways!
+
+Lots of the plan above I think is good work we should be doing anyways, regardless of whether we expose a store-only Nix.
+If you believe this, then the "cost" of this RFC is a lot less.
```
fricklerhandwerk, line 193: I don't think that's relevant.

```suggestion
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Tests
+
+Splitting the test suite per natural layer of the implementation is good work because it combines the specificity of unit tests with the real-world-ness of integration tests.
+"entire kitchen sink" tests make it harder to narrow down root causes of failures.
```
fricklerhandwerk, line 198:
```suggestion
"Entire kitchen sink" tests make it harder to narrow down root causes of failures.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Manual
+
+The documentation team is already working to clean up the manual, and this effort already involves emphasizing layering.
```
fricklerhandwerk, line 202:
```suggestion
The documentation team is already working to clean up the manual, and this effort already involves emphasizing the architecture.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+Do programming parts without the documentation and website parts.
+But that feels to me like turning an ongoing shift in focus to a one-off change that is likely to bit-rot.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What should the store-only Nix be called?
```
fricklerhandwerk, line 225: `nix-store`

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+# Future work
+[future]: #future-work
+
+## Nix Store Team
+
+Now that we have this division in the implementation, we also have the opportunity to leverage it for governance purposes.
+An official, NixOS-foundation-authorized team could be set up to manage store layer design decisions (below the threshold of needing an RFC) that don't affect that the rest of Nix.
+\[Some sort of decision that affect all layers is out of scope, must be deliberated with stakeholders from other layers too, probably should be RFC due to such large scope.\]
```
fricklerhandwerk, line 234:
```suggestion
Decisions that would affect all layers is out of scope for that team, must be deliberated with stakeholders from other layers too, and probably should be RFCs.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+To be clear, this is *not* to say we should abandon the idea of Nix as a whole.
+There can still be governance of Nix as a whole; this team, and similar hypothetical, say, Flakes, Nix language, or User Experience teams would ultimately need to report to.
+The goal is not to overreact, but strike a balance between:
```
fricklerhandwerk, line 238:
```suggestion
The goal is to strike a balance between:
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+1. Making sure Nix as a whole continues to make sense
+2. Make sure layers make sense in isolation not just in the context of the way they are currently used.
```
fricklerhandwerk, line 241:
```suggestion
2. Make sure layers make sense in isolation, and not just in the context of the way they are currently used.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Standardization across projects
+
+If we establish informal interop across store-layer implementations with Guix, a next step would be establish some sort of living standard that both communities have equal say in.
```
fricklerhandwerk, line 245:
```suggestion
If we establish informal interoperability across store-layer implementations with Guix, a next step would be establish some sort of living standard that both communities have equal say in.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+- logging
+- store paths on `stdout` at end of build?
+- Should commands like `show-derivation` should use `--json` by default
+- Flat vs hierarchical commands
+- is `--derivation` a good flag? (I think not!)
+
+are all in-scope.
+
+Having a conversation just on this narrow first batch of stabilization both builds trust, and ensures these still-important issues they aren't lost in flame wars over more divisive topics.
```
fricklerhandwerk, line 271: I'm not sure how this part relates to the rest. I suppose the topics listed would be interesting for a potential Nix Store Team to take care of, and maybe you can salvage just that list and an introductory note.
```suggestion
```

fricklerhandwerk (commented):
I had a look at two previous drafts in the past months, and like the idea and the detailed elaboration very much.

If that RFC is to be accepted, I would really like to have it at most 2/3 of the length just to save future readers the time to work through it. This is of course only about writing style, but more brevity and focus would have helped me evaluate the proposal.

PS: Suggestion for title: Nix store as separate executable.

thufschmitt:
I'm sympathetic with the idea, with a few caveats:
- As stated, the motivations feel a bit off to me. I fully agree with the complexity part (though that's mostly an internal implementation problem), but the part about the pluralism not so much: 
    - The store protocol is a very bad one to build on because It's very painful to implement and debug, and way too rigid for several implementations to cohabit peacefully. In particular I suspect that the Guix protocol is now irreconcilably incompatible with the Nix one.
    - Being able to mix and match different client and daemons doesn't strictly require the code to be splitted, as long as there's a well-defined protocol. Which is anyways the case (I said _well defined_, not _well documented_ mind you) because of Nix being a distributed system. And there's no clear explanation of why just building the store part could be beneficial.
- I might be convinced otherwise, but _a priori_ distributing a store-only Nix with its own store-only documentation seems like a fairly bad idea – unless that's explicitly stated as a reference document pointed at people who want to try nasty things in their basement. There's already enough important concepts for people to grasp without having to care about the fact that they are running a client-server thing
- The testing part is in my opinion the biggest gain and would deserve more than just a mention (although that also is mostly geared towards the internal implementation rather than the external interface)

But again, I agree with the broad idea, even if only from an implementation point of view

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+So focusing on the lowest layer first, we get the most "bang for buck" in terms of managing extremely different sorts of work separately.
+
+## Marketplace of Ideas
+
+As Nix grows more popularity, it will be inevitable that different groups want to explore in different directions.
+This is the *pluralism* of a larger community, and we should embrace it as a strength.
+We do that be fostering a *marketplace of ideas*.
+
+There are many possible ways in which to write down packages, The Nix language and Nixpkgs idioms, and Guix, for example, are just two points in a much larger space.
```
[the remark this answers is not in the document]
Ericson2314:
```suggestion
There are many possible ways in which to declare packages: the Nix language and Nixpkgs idioms, and Guix, for example, are just two examples of what is possible in principle.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+There are also many possible ways set up build farms.
+Our current central dispatcher, many remote-builder agents model, point-to-point protocol model is also just point in a much larger design space.
```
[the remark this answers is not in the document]
Ericson2314:
```suggestion
Our current model of a central dispatcher, many remote-builder agents, and a client–server where only one side initiates, is also just one point in a much larger design space of possible solutions.
```

tomberek:
> * I might be convinced otherwise, but _a priori_ distributing a store-only Nix with its own store-only documentation seems like a fairly bad idea – unless that's explicitly stated as a reference document pointed at people who want to try nasty things in their basement. There's already enough important concepts for people to grasp without having to care about the fact that they are running a client-server thing

My interpretation of this is that is should be possible to distribute such a thing, not that it will be prominent or even recommended. `Allow building a store-only version of Nix.` and `It should be possible to build a store-only manual without...` seem to be deliberately intending that the possibility to create such a limited binary. And the ability to do so is itself the proof that the store layer is independent. It is also an artifact with a much smaller surface for testing, fuzzing, documentation, and potential re-use in unexpected places.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+## Managing complexity
+
+We have recently added a ton of functionality to Nix.
+Most of this is in Flakes, which just has enormous surface area, but things like floating content-addressed derivations and other RFCs add complexity to the core store layer too.
+
+If we view Nix as one monolithic whole, it will grow too complex and unwieldy, and we will be unable to manage it as we the complexity bogs us down.
+However, if we embrace layering we can "divide and conquer" the project, and manage that complexity.
+This will ensure the continued sustainability of Nix.
+
+We currently embrace layering somewhat as an implementation detail, but only as an implementation detail.
```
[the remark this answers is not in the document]
Ericson2314:
```suggestion
We currently care about layering enough to partition the implementation into separate `libnix*` libraries, but this is only an implementation detail.
```
how this?

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+- Separate executable ensure lower layers build in isolation, new integration tests those executable without aid of high-layer info.
+
+### Starting with the store layer
+
+Ultimately, I would like to take this approach to all the layers.
+But I want to constrain the scope of this RFC to keep it tight an actionable.
+
+Layering between e.g. Flakes and the Nix Language doesn't yet exist in the implementation in the form of a library separation.
+I don't want to be "blocked" on major new development work, and anything involving Flakes is also far more controversial, and best avoided at first.
```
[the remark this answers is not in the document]
Ericson2314: Oh i mean blocked on *self-imposed* work to separate these parts, not blocked on other stuff going on. (This wasn't meant to be a snarky comment Flakes or anything.)

```suggestion
Separating Flakes into its own library would require major new development work, and I don't want to "block" exposing the layering for the first time to users on such work when there is an easier way to try this idea how.
Flakes also has more development going on due to its unstable nature, so it is nice to get out of the way of that.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+
+Conversely, the store layer is already quite well separated.
+The orthogonality between it and other layers "proven" in the wild by projects like Guix (more on that latter part in the next section).
```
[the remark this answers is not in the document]
Ericson2314:
```suggestion
Its conceptual independence from Nix's other layers is "proven" in the wild by projects like Guix.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,271 @@
...
+All functionality the store-only Nix offers factors through this interface.
+The upper half of Nix likewise uses the lower half through this interface.
+The daemon protocol represents this interface for IPC, and allows either half to be swapped for a different implementation.
+
+To help explain the community-building benefits, it might help to go over some specific examples.
+
+### Tvix and go-nix
+
+In https://tvl.fyi/blog/rewriting-nix, TVL announced that, frustrated in trying to refactor Nix into something more modular and flexible, they were aiming to make a new implementation from scratch.
+More recently, in https://tvl.fyi/blog/tvix-status-september-22 they lay out a basic approach of two projects:
```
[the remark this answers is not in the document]
Ericson2314:
```suggestion
[TVL announced](https://tvl.fyi/blog/rewriting-nix) that, frustrated in trying to refactor Nix into something more modular and flexible, they were aiming to make a new implementation from scratch.
More recently, [they lay out a basic approach](https://tvl.fyi/blog/tvix-status-september-22) of two projects:
```

rfcs/0134-nix-store-layer.md
Ericson2314:
```suggestion
If we hit the limits of GitHub Actions in per-PR CI, we should consider using Hydra instead / in addition to GitHub Actions, something that has already been discussed.
```

Ericson2314:
@tomberek Agree completely. I'll try to make that clearer in the RFC proper, but suggestions on how to do so welcome!

@thufschmitt I very much agree the current protocol is bad. What I like is the *abstract* store interface, like the C++ class or its "Platonic form", not the way we communicate between processes today.

I think some rephrasing I can do to emphasis this and in fact use the things you are saying as motivation:

- If we always distribute the sole client and server, then it is easy to have a "closed world assumption" where the protocol being bad doesn't matter because both end are Nix itself, and they already understand each other, and there isn't any reason to improve the protocol.

- if we distribute just the store side, we are very strongly implying "bring your own frontend" is a kosher idea, and this shatters the "closed world assumption" above, thereby justifying making a much better protocol.

You are right that all the end-user benefits do not inherently depend on this; using and "overkill" binary indeed doesn't prevent use-cases. But the combination of concrete implementation-side benefits with more nebulous more more far reaching "messaging" benefits user-side feels good to me. If you want to see the implementation-side concrete benefits like testing giving more weight in the motivation I am happy to do that.

rfcs/0134-nix-store-layer.md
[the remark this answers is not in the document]
fricklerhandwerk: Yes. (Although I would prefer to phrase it as separation of concerns instead of layering, but that may be me having Heard- and read "layering" way too often.)

fricklerhandwerk:
Maybe the following is a slight tangent, so excuse the noise: What the success of such RFCs (where @Ericson2314 shows a prototype and has a track record of following up and implementing the whole thing) seem to boil down to is the commitment and capacity of maintainers to review and merge the proposed changes, which will be nontrivial, over a longer period. So the question to @thufschmitt and ultimately @edolstra is if they care enough to provide the necessary resources to support such an endeavour once it's conceptually matured enough to be accepted as RFC.

Ericson2314:
Yes that is true. In this case the draft PR is very much incomplete (a lot of things crash), but I think getting it "pre-approved" and then I go finish it is going to feel better than other ones where I am farther along but blocked on approval.

7c6f434c:
I, myself, am sometimes annoyed when I cannot reshuffle the UI by wrapping the low-level operations as a high-level operation is the only simple way. I prefer instantiate+realise for some of the scripting instead of calling build operations… but I still don't know how to get a single output path by output name that way.

Hopefully with official splits it would be easier to understand the exact scope of each layer!

edolstra:

rfcs/0134-nix-store-layer.md
roberth: I don't think we _need_ to decide on a new name. It could be just a build configuration of the already named components. 
For an analogy, in Nixpkgs both `docker` and `docker-client` packages provide the `docker` command, both of which are produced by the same package function.

On the other hand, renaming would be a nice gesture, if we drop `nix` from the names: `libstore`, `store add-file FILE`, etc.

L-as:
I nominate myself.

tomberek:

Ericson2314:
I suspect the issue is that the most exciting motivation for this is #136, which has gotten lots of attention, but that is a draft RFC and this is a non-draft one (that #136 depends on).

I'll see what I can do.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
...
+
+Second of all, note that they have separate, orthogonal experiments they wish to run on both sides of the store interface divide.
+Above, they want to experiment with radically different evaluation strategies, especially to speed up Nixpkgs evaluation.
+Below, they want to experiment with the standardized containerization technologies that already exist for new ways of sandboxing and distributing builds with less bespoke Nix-specific code.
+They also want to apply the layering paradigm *within* go-nix, fostering even more modularity.
+
+I think these are great goals, and for the sake of the ecosystem as a whole, it should be as easy as possible to run such new experiments.
+In particular, a novel evaluator should be usable with the standard C++ Nix store layer, and a novel store layer should be reusable with the standard C++ Nix evaluator.
+
+Yes, strictly speaking, we only need a stable daemon protocol to accommodate that goal, which we have.
```
lheckemann, line 117: I don't think this is a given. I seem to recall @edolstra stating that he does not want to guarantee interoperability between Nix versions[citation needed]. I think people disagreed on this and I'm not sure if they changed his mind, but it would be good to have a clear statement on this -- perhaps we could _make it true_ as part of the RFC?

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
...
+More profoundly, they also have a more "library" than "interpreter" model where packages depend on Guix as a library, which talks to a small rump daemon.
+Guile sits far lower in their stack than the Nix language interpreter does; it is as if we rewrote some of our C++ into Nix language code, and Nix language code could do enough side effects to make that possible.
+
+The point of this discursion is to show that not only are radically different implementation of the same specification possible on either side of the store interface (what Tvix + go-nix aim for), but also radically different designs not at all trying to be compatible.
+
+All that said, below the store layer there is no difference in vision.
+Because our communities are so separate, it would be easy to come up with diverging versions of how derivations, store objects, etc. should work.
+That we have not done so I think is testament to the broad applicability of the Nix store design to many diverse groups of people with diverse goals.
+
+What I hope to do with Guix, then, is convene both projects to make their store store layers interoperate.
```
lheckemann, line 136: nit
```suggestion
What I hope to do with Guix, then, is convene both projects to make their store layers interoperate.
```

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
...
+Complementing the idea of a "marketplace of ideas" is when there is a certain design (like the Nix store layer), that is so broadly popular as to be a sort of "natural monopoly", that we should foster the most expansive and general idea of it as an exercise in coalition building and outreach.
```
lheckemann, line 137: I find this sentence hard to understand, I think it tries to make too many points at once.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
...
+This is a Nix executable that links `libstore` but not `libfetchers`, or `libexpr`.
+Plenty of commands like `nix daemon`, `nix log`, and the `nix store` sub-commands don't care about evaluation, fetching, or Flakes at all.
+[NixOS/nix#6182](https://github.com/NixOS/nix/issues/6182) implements this, splitting `libcmd` into two parts so the CLI code is reused.
+We will finish it off with as many commands as are reasonable to include, and merge it.
+
+Initially, we can test this store-only version of Nix with no changes to the test suite, by running full Nix with the store-only Nix's daemon.
+Support for testing Nix against a separately-built daemon already exists and is in use today.
+
+The store-only Nix and its tests should be built as part of CI, as "first class" as our existing CI jobs.
+That means both in the channel-blocking Hydra evaluation, and for each pull request.
```
lheckemann, line 160: Can we use the store-only Nix as the for NixOS's Nix daemon, and run Nix's daemon tests against a store-only-built daemon? Once we have the layering, I see no reason not to use only the necessary parts wherever possible.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
...
+The goals motivated above a broken down into small steps that we can execute in isolation.
+This keeps the cost of this work initially lower, and generally reduces risk.
+
+## 1. The split itself
+
+Allow building a store-only version of Nix.
+This is a Nix executable that links `libstore` but not `libfetchers`, or `libexpr`.
+Plenty of commands like `nix daemon`, `nix log`, and the `nix store` sub-commands don't care about evaluation, fetching, or Flakes at all.
+[NixOS/nix#6182](https://github.com/NixOS/nix/issues/6182) implements this, splitting `libcmd` into two parts so the CLI code is reused.
+We will finish it off with as many commands as are reasonable to include, and merge it.
```
lheckemann, line 154: Can we move the store-only parts of Nix into a separate drv output where non-dependency on the "full" output is enforced?

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
...
+### Manual
+
+The documentation team is already working to clean up the manual, and this effort already involves emphasizing layering.
+So whether we formally make a store-only Nix or not, I suspect the overhauled manual will natural have easy boundaries from which to "extract" the store-only manual.
+
+## Security
+
+The daemon is a privileged process.
+Even if with upcoming changes it shouldn't need root, it does tasks like administrating OS sandboxes correctly which still are security-critical.
+Having less code in the story-only Nix daemon, even if we think the removed code was "dead anyways" is always good.
```
lheckemann, line 212: nit
```suggestion
Having less code in the store-only Nix daemon, even if we think the removed code was "dead anyways" is always good.
```

rfcs/0134-nix-store-layer.md
Ericson2314: How's this?
```suggestion
Complementing the idea of a "marketplace of ideas" is when there is a certain design (like the Nix store layer), that is so broadly popular as to be a sort of "natural monopoly".
Rather than keep that designed "bundled" with more opinionated ideas (like the rest of Nix), we should foster the most expansive and general idea of the broadly popular idea as an exercise in outreach to other communities and coalition building with them.
```

rfcs/0134-nix-store-layer.md
Ericson2314: Right now it would just be a separate derivation, but it would be nice to try to build both where possible.

Separate outputs might play funny with NixOS but I think we can. We can also just build libraries and exes in separate derivations.

thufschmitt:
> that is a draft RFC and this is a non-draft one (that https://github.com/NixOS/rfcs/pull/136 depends on).

It seems to me that the dependency is very shallow. The only think from this RFC that we really need for #136 is a set of CLI commands that only concern the store part of Nix. Which is _also_ a prerequisite for this RFC, but a very small one compared to the rest of it. And not even something that warrants an RFC since it's just observing the current state of things.

thufschmitt:
Unless you mean that this split would require changing the CLI to get a clear split between the commands that are store-only and the ones that aren't?

Ericson2314:
I like the split because it forces us to be correct in our differentiating between layers. If we just say "hmm these commands seem to be libexpr-agnostic" we are just sipping at the hopium, there is no factual basis that our list of commands is correct. 

We can still incrementally stabilize without that, but that I really would rather not. Stabilizing is contentious, and the correctness of construction of actually doing the split is what gives me confidence the portion of the CLI I'd like to stabilize first is actually good; the enforcement of layers *disciplines* the design progress and keeps us on track.

garbas:
:heart: Could another outcome also be the possibility to gradually rewrite Nix in Rust (or whatever the language)?

Ericson2314:
@garbas I do think once we have "proven" layer boundaries, it makes incrementally rewriting Nix much less risky! We know exactly how and where a mini-nix-in-Rust can fit in.

rfcs/0134-nix-store-layer.md
andir: Just out of curiosity: What is the general perception of what the CI of Nix is? Can you point at it? I've heard different answers to that question over the years.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
...
+
+## 2. Manual
+
+It should be possible to build a store-only manual without information on the other layers, too.
+This would be the manual that is distributed with the store-only Nix.
+Of course, store-only and full Nix can share sections, so we aren't duplicating work.
+
+## 3. Website
+
+The full and store-only version of Nix should both be presented for download on the website.
```
andir, line 171: What exactly do you mean by presented? Provided as binaries or discussed/explained/promoted?

andir:
> This RFC has not acquired enough shepherds. This typically shows lack of interest from the community. In order to progress [a full shepherd team is required](https://github.com/NixOS/rfcs/#shepherd-team). Consider trying to raise interest by [posting in Discourse](https://discourse.nixos.org/), [talking in Matrix](https://matrix.to/#/#community:nixos.org) or reaching out to people that you know.
> 
> If not enough shepherds can be found in the next month we will close this RFC until we can find enough interested participants. The PR can be reopened at any time if more shepherd nominations are made.

I don't mind being a shepherd on this. This is a very good idea and I'd really like to see it succeed. Nix needs this!

rfcs/0134-nix-store-layer.md
Ericson2314: Yeah, provided downloads and discuss/explain what they are, but not necessary promote.

I am not going as far as saying they should be given equal billing. It's OK if users just download the entire Nix and aren't forced to learn about the layering. We can revisit that later if this is a success :).

infinisil:
I'm nominating myself to be a shepherd for this RFC

Ericson2314:
Yay! Shall we have @thufschmitt and/or @edolstra as the last shepherd?

Ericson2314:
We don't officially have an shepherd team yet, but it seems like we will reach quorum now so I went ahead and made a channel [#nix-rfc-134:matrix.org](https://matrix.to/#/#nix-rfc-134:matrix.org).

edolstra:
Yeah I can shepherd as well.

Ericson2314:
I rebased https://github.com/NixOS/nix/pull/6225, unfortunately the very OOP the command line works currently is making it hard to get this right. The way it works is that *self-contained* objects for each command are globally registered. The problem is that we would like into a downstream library augment the store-only commands so they can also take more complex installations (ones that involve the evaluator, not just store paths).

This will be something to discuss in the RFC meeting.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
+---
+feature: nix-store-layer
+start-date: 2022-09-06
+author: John Ericson
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
```
lheckemann, line 7: Thanks to the nominees for accepting! Here's the team the Steering Committee has decided on :)
```suggestion
shepherd-team: edolstra, infinisil, tomberek, L-as, andir
shepherd-leader: edolstra
```

Ericson2314: Hooray!

Ericson2314:
@roberth and I thinking of branding names for the store layer:

- trolley, tram
- honeycomb, or just comb
- batchgraph

roberth:
> thinking

Honeycomb was ChatGPT's, although maybe it was not "thinking". And do we credit AIs?
Interesting times.

amjoseph-nixpkgs:
> And do we credit AIs?

I expect the legal status of perceptron-generated works to be finalized shortly after the court system realizes they can't be called to the stand in copyright infringement trials.

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
...
+There are a few tweaks and generalizations we can make, but I struggle to envision what wildly different alternatives one might want instead.
+
+A small, stable interface that allows for design exploration above and below it is known as a [narrow waist](https://www.oilshell.org/cross-ref.html?tag=narrow-waist#narrow-waist).
+It's not every day that a project happens upon a great narrow waist design.
+I believe we've discovered a very good one with Nix, and that should be seen as a *key asset*, even if it is not how we recruit "regular users".
+
+By making a store-only Nix, we put more emphasis on this key interface.
+All functionality the store-only Nix offers factors through this interface.
+The upper half of Nix likewise uses the lower half through this interface.
+The daemon protocol represents this interface for IPC, and allows either half to be swapped for a different implementation.
```
andir, line 94: So will the daemon protocol need to be standardized as part of this work? it seems to me that it wouldn't make sense to have a defined external executable without a defined way to talk to it (in the long run). The answer could be to link against `libnixstore` and use that to communicate with the (same version of) the store daemon. Is that so?

Ericson2314: @andir We do have a *defined* deamon protocol, in that you could implement it yourself instead of using `libnixstore` and it is versioned. Yes, it is a not nice, but I don't think improving the situation should be a blocker on this PR.

Rather, I think this PR will help make this sort of thing more noticeable, allowing us to build momentum / community interest about improving the daemon protocol. I think that's great!

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,257 @@
...
+Initially, we can test this store-only version of Nix with no changes to the test suite, by running full Nix with the store-only Nix's daemon.
+Support for testing Nix against a separately-built daemon already exists and is in use today.
+
+The store-only Nix and its tests should be built as part of CI, as "first class" as our existing CI jobs.
+That means both in the channel-blocking Hydra evaluation, and for each pull request.
+If we hit the limits of GitHub Actions in per-pull-request CI, we should consider using Hydra either instead or in addition to GitHub Actions, something that has already been discussed.
+
+## 2. Manual
+
+It should be possible to build a store-only manual without information on the other layers, too.
```
tomberek, line 165:
```suggestion
It should be possible to build a store-only manual without information on the other layers, too.
```
It should be possible, but not necessary for us to implement, build, host, etc.

rfcs/0134-nix-store-layer.md
tomberek: I've been liking:
- store: only the store layer
- nix: includes nix language
- flake: include flakes concepts

rfcs/0134-nix-store-layer.md
andir: Yeah, I don't mean this PR should implement it. It is just that the protocol might currently be defined but not documented or carries any guarantees. It has been subtly broken several times and (as you know) it isn't currently being tested across a broad set of nix versions.

I already implemted it (several times in fact) but the protocol isn't fun to reverse engineer from the current code base. My big hope is that with this RFC passing we could move to another RFC that defines a (new) interface that doesn't just allow exchanging implementations but embraces it.

infinisil:
# 2022-01-23 Meeting #1 Notes

Present: @infinisil @edolstra @Ericson2314 @andir @L-as

- @edolstra: Historically there was a nix daemon that only linked to libstore
- Intentionally moved away for smaller footprint, only a single nix binary, nix-* binaries deprecated
- Having a separate binary would put a strain on maintainers, adds another case to the test matrix
  - @L-as: Could it be a library, like store-command-library that the main binary calls into?
    - @edolstra @Ericson2314: Basically the case right now with the daemon
- @edolstra: Definitely not a separate name for the store layer
- @Ericson2314: Draft PR https://github.com/NixOS/nix/pull/6225
- Hardest parts:
  - Making store-only command accept fancier installables
  - --help is written using expression language
- @edolstra: Why is it so big
  - @Ericson2314: Moves things around
- @edolstra: What is the use case for a binary like that?
  - Guix wouldn't use this
  - @Ericson2314: Carves out this lower level infrastructure, allows enticing people to do it in different ways, not involving the language, nixpkgs, etc.
  - @edolstra: E.g. use case of having a separate cabal CLI and using Nix in the background. CLI or only library?
    - @Ericson2314: Both, e.g. ninja has both CLI and library too, easier to get started with CLI
    - @L-as: Hard to link to C library from different languages
  - @tomberek: Other use cases: Data processing, store layer that handles annoying details of e.g. hashing, allows pushing things around
    - @Ericson2314: Data science use cases are not tapped
  - @Ericson2314: More flexible store layer over time, e.g. CA, more use cases
  - @tomberek: Binary serves as a witness that the layering is there
    - @edolstra: Don't think it's a good argument. Layers always serve something else
    - @tomberek: E.g. nixel (using Nickel as the language). Abstracting away build layer, alternative Hydra, build orchestration using the same underlying store
  - @andir: Not very interested in binary, more in the layer that we can replace in either direction. Making sure the coupling is not too tight.
  - @L-as: Attempt to improve code quality of Nix, that's what layering is about, removing dependencies that aren't needed. Hope for replacing Nix with something better (Nickel, Guile)
    - @edolstra: YAGNI, refactoring only out of theoretical belief that it's beneficial in e.g. 5 years. Needs a lot of effort for something that might not pay off
  - @Ericson2314: Tom and me know a lot of people interested in this. Hard to build trust in layering from third-parties. E.g. Guix
    - @edolstra: Can forget about them, they want scheme
    - @Ericson2314: We can tell them we're serious about layering, might standardize the daemon protocol together, combining of components should work
  - @L-as: Separation of concern is good because in 100 years everybody should use Nix, but it needs to not have security issues. By separation of concern we reduce security problems. With more implementations of parts it won't compromise everybody.
    - @edolstra: Not sure about this, everybody runs the Linux kernel. Fractures community. Most popular implementation will exist
      - @Ericson2314: Clang and GCC still exist today. If we want to take over the world we will rise to that level
      - @L-as: Effort will be split anyways, implementations will be written in different languages. With a common interface they won't compete in a destructive way but a constructive one.
      - @Ericson2314: Would rather have competition with learning from each other

- @tomberek: Is the effort to do this worth the time? Is the direction good? Then is the cost worth it?
  - @edolstra: Ability to build mini-Nix is worthwhile, depending on the implementation cost. The PR is less scary than it looks
    - @Ericson2314: Biggest churn is splitting CLI
    - @edolstra: Have a bunch of `#if` flags?
- @Ericson2314: Prefers doing more work now than people having trouble later
- @Ericson2314: Procedural: Workaround for virtual classes GCC bug, should have Nix require a compiler that doesn't have the bug. Fix in nixpkgs is backported: https://github.com/NixOS/nixpkgs/pull/210816. Makes the PR for this RFC much easier. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80431 and https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=de31f5445b12fd9ab9969dc536d821fe6f0edad0
- @infinisil: Effort seems doable then
  - @edolstra: Another cost is it causes a lot of merge conflicts
    - @Ericson2314: I can fix those merge conflicts, assign me :)
      - @tomberek: Can we split that effort so it's not just you?
      - @Ericson2314: Should be able to finish it
- @edolstra: Splitting installables is a good thing anyways. Do we need libstore-cmd? Once installables is split, should be easy.
  - @infinisil: Implementation detail
    - @edolstra: +1
- @infinisil: Standards and protocol is the way forward, internet standard, Email standard existed for a long time, not individual implementations
  - @edolstra: Internet is moving more towards single implementation, allows faster iteration
    - @tomberek: Periods of more and less standards
  - @Ericson2314: Bazel remote execution protocol has gotten popular. Big corps will have in-house teams that appreciate being able to do their own thing too.

- @tomberek: In summary: Direction desirable, cost isn't too high. We can make non-regrettable steps to get to the end.
- @infinisil: Do we need anything else? Can we FCP? Does the text look good?
  - @edolstra: What about the name?
  - @Ericson2314: We can also leave the RFC open to naming, branding, etc.
  - @edolstra: Everybody should have a read through the RFC
  - @infinisil: Don't need to find a decision for every small detail, we can just remove those
- @tomberek: Benefit to mention: Nix store team, iteration on code base because easier, trust between layers, easier organization

- @edolstra: Regarding building store-only manual, I think that's too much effort
  - @Ericson2314: Can move to future work
- @edolstra: Regarding website, shouldn't have a separate page for that
- @edolstra: Regarding tests
  - @Ericson2314: Current PR doesn't test mini nix
  - @edolstra: Tests would be a big change, most of them require the Nix language
  - @infinisil: Just a possibility
    - @tomberek: What possibility? Of somebody doing so or that it could be done?
  - @Ericson2314: Let's move to future work
  - @edolstra: Testing needs a way to get things into the store, currently nix language is used for that. Can push that to the future
  - @Ericson2314: Can run store only daemon only

- @infinisil: Move detailed points 2-4 to future work, doesn't need to be in the RFC
- @infinisil: Let's give reviews for the RFC, accepting if agreed
  - @edolstra: Let's read through it this week, maybe don't need a meeting
  - @infinisil: Let's keep the next meeting for in 2 weeks so that we don't stall the RFC, FCP next time if everybody agrees with it

Ericson2314:
> Move detailed points 2-4 to future work, doesn't need to be in the RFC

This is now done.

rfcs/0134-nix-store-layer.md
amjoseph-nixpkgs: > I already implemted it (several times in fact) but the protocol isn't fun to reverse engineer from the current code base.

Might you be interested in documenting these fruits of your protocol-reverse-engineering labors?

rfcs/0134-nix-store-layer.md
infinisil: I agree with @roberth's idea of not needing a rename. For the purposes of somebody wanting to make sure they only depend on the store-only Nix, we can have `nix --version` indicate that it's a store-only Nix.

infinisil (approved):
Due to GitHub it's very hard to find, but [this thread](https://github.com/NixOS/rfcs/pull/134#discussion_r974688598) is being discussed.

Other than that I think this RFC looks good. I'd also encourage the other shepherds (@edolstra, @tomberek, @L-as, @andir) to either approve it or request changes before the meeting next week (2023-02-06).

rfcs/0134-nix-store-layer.md
Ericson2314: Moved to future work in 3194ad43a5c302b5b261eafb25c0f1426203dbee. If everyone is good with that we can mark this thread resolved.

edolstra:
We just a shepherd meeting and decided to call FCP for this PR!

fgaz:
Happy to see this moving forward!
I have a bunch of notes about adding a nix backend to the Haskell [cabal](https://www.haskell.org/cabal/). Due to its [nix-like architecture](https://cabal.readthedocs.io/en/3.8/nix-local-build-overview.html) it should be relatively simple and would solve some issues such as garbage collection and deployment. A stable and polished store interface is a prerequisite (NixOS/nix#7271 would help too).

I expect similar tools to benefit as well. Nix makes for a great backend for the part of build tools that concerns dependencies, just like meson/ninja/... for individual packages. If I had to write a language build tool today, I'd base it off both.

Ericson2314:
@fgaz I am beyond thrilled to hear you, a Cabal developer, say so. First party Nix integration in Cabal and similar tools (just as CMake uses Ninja) is the dream I've had for years, and seeing in finally happen would be absolutely revolutionary. :tada: :tada: :tada:

Ericson2314:
(Also, stay tuned for https://github.com/NixOS/rfcs/pull/92 so Nix can do the Nix part *and* the Ninja part :))

Ericson2314:
I think it has been 10 days, so FCP is over?!

edolstra:
Yes indeed!

Ericson2314:
Thanks everyone for participating with this! I have made a tracking issue, https://github.com/NixOS/nix/issues/7868, for this.

Ericson2314:
https://github.com/NixOS/nix/pull/9063 general layering update — here's a PR for Flakes moved to a new library!

I can't believe it took me so long to get around to this fairy easy thing 😅

rfcs/0134-nix-store-layer.md
```diff
@@ -0,0 +1,260 @@
...
+They also want to apply the layering paradigm *within* go-nix, fostering even more modularity.
+
+I think these are great goals, and for the sake of the ecosystem as a whole, it should be as easy as possible to run such new experiments.
+In particular, a novel evaluator should be usable with the standard C++ Nix store layer, and a novel store layer should be reusable with the standard C++ Nix evaluator.
+
+Yes, strictly speaking, we only need a stable daemon protocol to accommodate that goal, which we have.
+But ensuring the C++ Nix components can be built separately for use in isolation further send the message that such experimentation is *good* and *welcomed*.
+Messaging matters, and making our layered architecture "official" as this RFC proposes I think sends a better message.
+
+### Guix
```
Mic92, line 121: Guix is not conceptional compatible on the store layer level with Nix. They support having scheme expression that are executed at build-time in their derivation files.

roberth: Depends how they've integrated that, how that scheme is defined and whether that's inconsistent with any `system` value we've used in Nix.
I couldn't find anything obvious about this in the Guix manual. Do you have a link to this info?

Mic92: https://guix.gnu.org/manual/en/html_node/G_002dExpressions.html

Ericson2314: @Mic92 No that is just them swapping out scheme for bash.

https://guix.gnu.org/en/blog/2023/dissecting-guix-part-1-derivations/ is not entirely clear, but shows that the drv format is not affected by g-expressions.

PR #135, opened 2022-09-20
[RFC 0135] Custom asserts

schuelermine:
[Rendered](https://github.com/schuelermine/nix-rfcs/blob/rfc-0135/rfcs/0135-custom-asserts.md)

### DIFF
--- a/rfcs/0135-custom-asserts.md
+++ b/rfcs/0135-custom-asserts.md
@@ -19,9 +19,8 @@ Allow users to use attribute sets with a boolean attribute `success` and a strin
 Since Nix is an untyped language, asserts are often needed to ensure that a function or program works correctly. However, the current assert system is unsuitable for more sophisticated error reporting that aims to inform the user as to what has happened.
 
 Consider a nixpkgs package expression that wants to validate its arguments. Currently, the best way to
-provide a custom error message is to use `assert … -> throw …; …`.
-This method has several disadvantages: Since an implication from a falsehood is always true, you are required
-to invert the condition. Additionally, since the assertion itself is not triggered by the error,
+provide a custom error message is to use `assert … || throw …; …`.
+This method has several disadvantages: Since the assertion itself is not triggered by the error,
 the function of the `assert` keyword is reduced to providing an imperative shorthand for `seq`. This also means that by default,
 the error location is not printed, and there is no mention of an assert in the error message.
 Instead, expressions could use this more natural syntax:
--- a/rfcs/0135-custom-asserts.md
+++ b/rfcs/0135-custom-asserts.md
@@ -17,7 +17,25 @@ Allow users to use attribute sets with a boolean attribute `success` and a strin
 [motivation]: #motivation
 
 Since Nix is an untyped language, asserts are often needed to ensure that a function or program works correctly. However, the current assert system is unsuitable for more sophisticated error reporting that aims to inform the user as to what has happened.
-Consider, for instance, this simple type system:
+
+Consider a nixpkgs package expression that wants to validate its arguments. Currently, the best way to
+provide a custom error message is to use `assert … -> throw …; …`.
+This method has several disadvantages: Since an implication from a falsehood is always true, you are required
+to invert the condition. Additionally, since the assertion itself is not triggered by the error,
+the function of the `assert` keyword is reduced to providing an imperative shorthand for `seq`. This also means that by default,
+the error location is not printed, and there is no mention of an assert in the error message.
+Instead, expressions could use this more natural syntax:
+
+```nix
+{ foo, bar }:
+assert {
+  success = foo || bar;
+  message = "At least one of foo or bar must be set";
+};
+[ foo bar ]
+```
+
+Also consider, for instance, this simple type system:
 
 ```nix
 rec {
@@ -70,23 +88,6 @@ with types;
 ```
 
 While these implementations get more and more verbose, they also get more and more idiomatic and flexible.
-
-Consider also a nixpkgs package expression that wants to validate its arguments. Currently, the best way to
-provide a custom error message is to use `assert … -> throw …; …`.
-This method has several disadvantages: Since an implication from a falsehood is always true, you are required
-to invert the condition. Additionally, since the assertion itself is not triggered by the error,
-the function of the `assert` keyword is reduced to providing an imperative shorthand for `seq`. This also means that by default,
-the error location is not printed, and there is no mention of an assert in the error message.
-Instead, expressions could now use this more natural syntax:
-
-```nix
-{ foo, bar }:
-assert {
-  success = foo || bar;
-  message = "At least one of Foo or Bar must be set";
-};
-[ foo bar ]
-```
 
 # Detailed design
 [design]: #detailed-design
--- a/rfcs/0135-custom-asserts.md
+++ b/rfcs/0135-custom-asserts.md
@@ -112,3 +112,4 @@ This change requires no change to the language grammar.
 [questions]: #unresolved-questions
 
 - What should the exact name of `message` and `success` be?
+- What should the error messsage be?
--- a/rfcs/0135-custom-asserts.md
+++ b/rfcs/0135-custom-asserts.md
@@ -1,0 +1,99 @@
+---
+feature: custom-asserts
+start-date: 2022-09-21
+author: Anselm Schüler
+co-authors: None
+shepherd-team: None
+shepherd-leader: None
+related-issues: None
+---
+
+# Summary
+[summary]: #summary
+
+Allow users to use attribute sets with a boolean attribute `success` and a string attribute `message` instead of a boolean in `assert …; …` expressions.
+
+# Motivation
+[motivation]: #motivation
+
+Since Nix is an untyped language, asserts are often needed to ensure that a function or program works correctly. However, the current assert system is unsuitable for more sophisticated error reporting that aims to inform the user as to what has happened.
+Consider, for instance, this simple type system:
+
+```nix
+rec {
+  assertType = type: locDescr: value:
+    if ! type.check value
+    then throw "Value at ${locDescr} was not of type ${type.name}"
+    else builtins.traceVerbose "Successfully checked type ${type.name} at ${locDescr}" value;
+  assertTypeSeq = type: locDescr: value1: value2:
+    builtins.seq (assertType type locDescr value1) value2;
+  intType = {
+    check = builtins.isInt;
+    name = "integer";
+  };
+  attrsOfType = subType: {
+    check = value:
+      builtins.isAttrs value
+      && builtins.all subType.check (builtins.attrValues value);
+    name = "attribute set of ${subType.name}";
+  };
+}
+```
+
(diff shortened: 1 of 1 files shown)


rfcs/0135-custom-asserts.md
```diff
@@ -0,0 +1,99 @@
...
+
+- Implementation clutter
+- This could end up as a confusing and underutilized feature that hardly anybody knows about
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Doing nothing and continuing to use whichever assertion method is most appropriate for a given use case
+- Deprecating `assert …; …` expressions in favour of user-made type systems
+
```
amjoseph-nixpkgs, line 94:
```suggestion
- continue using the `assert !condition -> throw "message"` idiom
```

Isn't `assert !condition -> throw "message"` the current way of doing this?  Shouldn't the RFC be comparing primarily against that?

Example:

https://github.com/NixOS/nixpkgs/blob/0fdc7224a24203d9489bc52892e3d6121cacb110/pkgs/applications/networking/browsers/firefox/common.nix#L141

schuelermine: Yeah, I completely fixated on my fictional use case. I will fix it tomorrow (in <10h or so)

schuelermine: I have amended the rationale to include this practice

schuelermine:
I have removed the `--trace-verbose` feature from the proposal since it jeopardizes what is likely the most common use case.

rfcs/0135-custom-asserts.md
```diff
@@ -0,0 +1,116 @@
...
+This change requires no change to the language grammar.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Implementation clutter
+- This could end up as a confusing and underutilized feature that hardly anybody knows about
+
+# Alternatives
```
sternenseemann, line 106: I think you should first consider the options we have *today* without language changes which don't seem to be mentioned in your RFC at all:

* `lib.assertMsg` which outputs a message via `builtins.trace` if it is `false`: `assert lib.assertMsg condition "message"; …`
* An idiom of using `|| throw ` to get an even nicer error message at the expense of it reading a bit strange to the uninitiated reader `assert condition || throw "message"; …`. This will also have correct location information etc. while hiding the often ugly and uninformative AST dump `assert` does by default.

rfcs/0135-custom-asserts.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Since Nix is an untyped language, asserts are often needed to ensure that a function or program works correctly. However, the current assert system is unsuitable for more sophisticated error reporting that aims to inform the user as to what has happened.
+
+Consider a nixpkgs package expression that wants to validate its arguments. Currently, the best way to
+provide a custom error message is to use `assert … -> throw …; …`.
+This method has several disadvantages: Since an implication from a falsehood is always true, you are required
+to invert the condition. Additionally, since the assertion itself is not triggered by the error,
```
sternenseemann, line 24: Missed this bit on my first read through. Note that this drawback is not an issue if you do `assert condition || throw "…"; …`. No need to invert, quite straightforward.

rfcs/0135-custom-asserts.md
```diff
@@ -0,0 +1,116 @@
...
+the function of the `assert` keyword is reduced to providing an imperative shorthand for `seq`. This also means that by default,
+the error location is not printed, and there is no mention of an assert in the error message.
```
sternenseemann, line 26: This is not true. `throw` also retains the location where it is called and it is displayed correctly to the user.

schuelermine: hm, I thought it was.

rfcs/0135-custom-asserts.md
schuelermine: OK, I removed that.

rfcs/0135-custom-asserts.md
```diff
@@ -0,0 +1,116 @@
...
+  attrsOfType = subType: {
+    check = value:
+      builtins.isAttrs value
+      && builtins.all subType.check (builtins.attrValues value);
+    name = "attribute set of ${subType.name}";
+  };
+}
+```
+
+Users of this system would be forced to forgo the convenience of imperative-style `assert …; …` in favor of `seq`-like syntax in order to benefit from improved type errors. With this change, no longer! A variant `isType` function could be declared:
```
sternenseemann, line 59: You can kind of have your cake and eat it, too, with a custom system. Such one has actually been added to nixpkgs by @roberth:

```
lib.throwIfNot cond1 "msg1"
lib.throwIfNot cond2 "msg2"
stdenv.mkDerivation { /* … */ }
```

This relies on a neat trick that you can return the identity function from a function, causing anything further in the source file to act as if that `lib.throwIfNot` and its first two arguments hadn't been there.

schuelermine: oh that’s neat! I still think `assert …;` is more idiomatic

roberth: See this prior discussion https://github.com/NixOS/nixpkgs/issues/154292

Nitpick: Idioms evolve from a language and its use, not the other way around. We're designing a language feature, so existing idioms are relevant but of secondary importance.

schuelermine: that’s a nice way to view it! I rephrase: I think my suggestion would make it more natural, easier to understand and more distinctive. It uses the syntax specifically for asserts and only for that, instead of co-opting returning identity sometimes.

schuelermine: In my opinion, these other idioms do not demonstrate that this change is not needed, but demonstrate that there is a lack of unified assertion functionality, which has left the conciseness of the dedicated `assert` syntax behind.

Ericson2314: If we had `$` like Haskell, we wouldn't miss the `assert ..;` syntax so much because we wouldn't need more parentheses either way.

sternenseemann: `$` keeps getting brought up, *but* it is an even more invasive change and would introduce more unfamiliar syntax to a language non FP programmers already struggle with conceptually…

schuelermine: Yeah, IMO `assert` is a more user-friendly idiom. And it already exists in the language!

amjoseph-nixpkgs: That's little a bit unfair: for anybody with no lazy FP experience, syntax will be the least of their problems.  No debugger, no `printf()`, and brain-meltingly weird stack traces.

But yeah, language stability is precious.

edolstra:

kevincox:

edolstra:
Since there are not enough shepherds yet, we have moved this RFC to draft state for now. It can be reopened at any time if more shepherd nominations are made.

infinisil:
Here's an interesting [comment](https://github.com/NixOS/nixpkgs/pull/209099#discussion_r1104310984) by @roberth:

> Assertions are for checking the invariants of self-contained software units, and not for input validation across a public interface.

Thinking about this I'm coming to these conclusions:
- The benefit of `assert` being a builtin language feature is that it can **print the failing assertion expression**, which is an easy way for programmers to know what went wrong without having to provide and maintain error messages for users.
  - Notably, to do this one doesn't need the `assert` _syntax_, a `builtins.assert` should theoretically also work. The main benefit with the `assert` syntax however is that you don't need to wrap the assertion expression in parenthesis, and that the `;` clearly separates the assertion expression from the result expression:
    ```nix
    # assert syntax
    assert foo == bar;
    null
    ```
    ```nix
    # Potential assert builtin
    builtins.assert (foo == bar)
    null
    ```
- Contrary to that, user-facing messages need to be manually written and maintained, so this use-case doesn't need any special builtin language feature, it can just use `throw`

Considering this, what's the benefit of this RFC then?
- This RFC doesn't need the assertion expression printing ability of `assert`, since the programmer providers their own message to print.
- This RFC also doesn't benefit from not having to wrap the assertion expression in parenthesis, because the assertion expression is an attribute value:
  ```nix
  assert {
    success = foo == bar;
    message = "failed";
  };
  null
  ```

This means that it's very much possible to implement _almost_ exactly this using either a builtin function or a library function. And yes indeed, similar to the [previously-mentioned `throwIfNot`](https://github.com/NixOS/rfcs/pull/135#discussion_r976508676):
```nix
# Could also be `builtins.assert`
lib.assert {
  success = foo == bar;
  message = "failed";
}
null
```

The only difference is that there's no `;` indicating the end of the assertion expression, and from what I can tell this would therefore be the only advantage of making this a language builtin over just a `builtins` or library function.

I did think about how formatters might struggle formatting it correctly if there's no `;`, but alejandra handles it [just fine](https://kamadorueda.com/alejandra/?before=lib.assert%27+%7B+success+%3D+true%3B+message+%3D+%22failed%22%3B+%7D%0Alib.assert%27+%7B%0A++success+%3D+true%3B%0A++message+%3D+%22failed%22%3B%0A%7D%0Anull).

infinisil:
In addition, I want to point out that there's more than just boolean-style checks one might want to check. Notably there's also [`lib.asserts.assertOneOf`](function-library-lib.asserts.assertOneOf), which wouldn't fit into this RFC's model. In addition I just opened [a PR to introduce `lib.asserts.assertAll`](https://github.com/NixOS/nixpkgs/pull/215166). Undoubtedly there's other useful utilities that could be added. In comparison the proposal from this RFC is fairly limited and couldn't be used idiomatically for such cases.

infinisil:
So here's an alternate proposal:

- Let's reinforce that the `assert` syntax is for internal invariant validation, with the benefit of always printing the assertion expression, it should not be used for user-facing errors.
  - In fact I think when an assertion expression fails to evaluate, it should not only throw that error, but also print the containing assertion expression
- Let's introduce nixpkgs library functions in the style of this RFC. But we shouldn't use the name "assertions" because that would be confusing with `assert`. Instead I'm proposing ["Guard"](https://en.wikipedia.org/wiki/Guard_%28computer_science%29). Here's some examples:
  ```nix
  lib.guard.success {
    value = foo;
    message = "something failed";
  }
  null
  ```
  ```nix
  lib.guard.oneOf {
    value = foo;
    valids = [ "foo" "bar" ];
    message = "something failed";
  }
  null
  ```

schuelermine:
That sounds good, I will close this RFC. If you want you can create an RFC for your idea.

infinisil:
Just realized a mistake I made above: Library functions like that only work for a subset of expressions of course, e.g.

```
lib.guard.success { ... }
if ... then ... else ...
```

wouldn't work of course.

While I'm not a huge fan of the original idea in the RFC here, how about [this other idea I proposed](https://github.com/NixOS/nix/issues/1845#issuecomment-1520798023), to have an `apply` expression like this:

```nix
apply foo;
apply bar;
baz

# Equivalent to
foo (bar baz)
```

This would then be general enough to be also usable for

```nix
apply lib.guard { ... };
if ... then ... else ...
```

but also much more, see https://github.com/NixOS/nix/issues/1845.

8573:
> how about [this other idea I proposed](https://github.com/NixOS/nix/issues/1845#issuecomment-1520798023), to have an `apply` expression like this

For these guards, would it be sufficient to use the existing `with` keyword, with `lib.guard` returning an empty set?
```nix
with lib.guard { ... };
if x then y else z
```

PR #141, opened 2023-02-03
[RFC 0141] Private Derivations

poscat0x04:
# Summary
This RFC proposes to add a special type of derivation called private derivation, which, upon being built, will have their file permissions set to 000 instead of the usual 444.

[Rendered](https://github.com/poscat0x04/nix-rfcs/blob/private-derivations/rfcs/0141-private-derivations.md)

### DIFF
--- a/rfcs/0141-private-derivations.md
+++ b/rfcs/0141-private-derivations.md
@@ -18,7 +18,7 @@ This RFC proposes to add a special type of derivation called private derivation,
 
 In short: This RFC mainly concerns with how to safely store credentials on NixOS.
 
-The world readability of nix store means that, to safely store credentials, they must be first somehow be encrypted before written into the store. They also need to be decrypted before the services are started.
+The world readability of nix store means that, to safely store credentials, they must first somehow be encrypted before written into the store. They also need to be decrypted before the services are started.
 
 This is less than ideal because one needs to setup a key (which is stored as plaintext on disk) on every machine just to prevent unauthorized users from seeting the credentials.
 
--- a/rfcs/0141-private-derivations.md
+++ b/rfcs/0141-private-derivations.md
@@ -49,6 +49,14 @@ An alternative would be to support more complicated ACLs as described in [this](
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
+## Binary caches and copying
+How do we prevent the attacker from using `nix copy` to simply copy out the
+private derivation to another machine?
+
+What changes are needed in binary cache providers such as `nix-serve` to handle
+private derivations?
+
+## Content-Addressed paths
 It is not yet known how this might interact with content addressed paths.
 
 # Future work
--- a/rfcs/0141-private-derivations.md
+++ b/rfcs/0141-private-derivations.md
@@ -61,7 +61,7 @@ An alternative would be to support more complicated ACLs as described in [this](
 How do we prevent the attacker from using `nix copy` to simply copy out the
 private derivation to another machine?
 
-What changes are needed in binary cache providers such as `nix-serve` to handle
+What changes are needed for binary cache providers such as `nix-serve` to handle
 private derivations?
 
 ## Content-Addressed paths
--- a/rfcs/0141-private-derivations.md
+++ b/rfcs/0141-private-derivations.md
@@ -11,7 +11,7 @@ related-issues: (will contain links to implementation PRs)
 # Summary
 [summary]: #summary
 
-This RFC proposes to add a special type of derivation called private derivation, which, upon being built, will have their file permissions set to 000 instead of the usual 444.
+This RFC proposes to add a special type of derivation called private derivation, which, upon being built, will have their file permissions set to 000/111 instead of the usual 444/555.
 
 # Motivation
 [motivation]: #motivation
@@ -24,7 +24,7 @@ This is less than ideal because one needs to setup a key (which is stored as pla
 
 Furthermore, if encryption is done before the evaluation of the system configuration (as is the case with [agenix](https://github.com/ryantm/agenix) and [sops-nix](https://github.com/Mic92/sops-nix)), then the nixos module system cannot be utilized to generate configs that contain credentials and one must write them manually.
 
-All of this can be prevented if we added the ability to make derivation outputs as not readable by anyone other than root, by setting the file mode to 111 (directories) or 000 (files). We can then use a trustworthy credential manager, for example systemd with its `LoadCredential=`, to distribute these derivations to the consumers safely.
+All of this can be prevented if we added the ability to make derivation outputs as not readable by anyone other than root, by setting the file mode to 111 (directories and executables) or 000 (regular files). We can then use a trustworthy credential manager, for example systemd with its `LoadCredential=`, to distribute these derivations to the consumers safely.
 
 # Detailed design
 [design]: #detailed-design
--- a/rfcs/0141-private-derivations.md
+++ b/rfcs/0141-private-derivations.md
@@ -29,7 +29,11 @@ All of this can be prevented if we added the ability to make derivation outputs 
 # Detailed design
 [design]: #detailed-design
 
-We propose adding a `noReadAccess` option to `builtins.derivation`, which, when set to true, makes this derivation a private derivation. Relevant changes should also be made in nix-instantiate and nix-daemon to understand this attribute.
+We propose adding a `noReadAccess` option to `builtins.derivation`, which, when set to true, makes this derivation a private derivation.
+
+The only difference between a private derivation and a normal derivation, apart
+from the hash, is that upon instantiation and after building, the read bit of
+the `.drv` file and the output path will be removed.
 
 # Examples and Interactions
 [examples-and-interactions]: #examples-and-interactions
--- a/rfcs/0141-private-derivations.md
+++ b/rfcs/0141-private-derivations.md
@@ -34,7 +34,15 @@ We propose adding a `noReadAccess` option to `builtins.derivation`, which, when 
 # Examples and Interactions
 [examples-and-interactions]: #examples-and-interactions
 
-TBA
+## Intended usage in NixOS modules
+On NixOS there are many modules where configurations that might contain sensitive
+information get written into nix store in plaintext (for example the wpa_supplicant module). Private derivations can solve this issue by
+
+1. Writing a helper function `writeTextPrivate` that functions similar to
+`writeText`, but instead outputs a private derivation
+2. Replace the `writeText` function with `writeTextPrivate` inside the module
+3. Use `LoadCredential=` to load the private derivation
+4. Replace the derivation output path with `%d/<credName>` (see systemd.exec)
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/rfcs/0141-private-derivations.md
+++ b/rfcs/0141-private-derivations.md
@@ -64,6 +64,10 @@ private derivation to another machine?
 What changes are needed for binary cache providers such as `nix-serve` to handle
 private derivations?
 
+## Leaking metadata
+The hash is still exposed to the attacker, which opens up some possible attacks.
+How does this impact the security?
+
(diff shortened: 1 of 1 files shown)


rfcs/0141-private-derivations.md
```diff
@@ -0,0 +1,85 @@
...
+This RFC proposes to add a special type of derivation called private derivation, which, upon being built, will have their file permissions set to 000/111 instead of the usual 444/555.
+
+# Motivation
+[motivation]: #motivation
+
+In short: This RFC mainly concerns with how to safely store credentials on NixOS.
+
+The world readability of nix store means that, to safely store credentials, they must be first somehow be encrypted before written into the store. They also need to be decrypted before the services are started.
+
+This is less than ideal because one needs to setup a key (which is stored as plaintext on disk) on every machine just to prevent unauthorized users from seeting the credentials.
```
SuperSandro2000, line 23: As an alternative you could imperatively copy credentials over or use a credential service like vault.

rfcs/0141-private-derivations.md
```diff
@@ -0,0 +1,85 @@
...
+
+Furthermore, if encryption is done before the evaluation of the system configuration (as is the case with [agenix](https://github.com/ryantm/agenix) and [sops-nix](https://github.com/Mic92/sops-nix)), then the nixos module system cannot be utilized to generate configs that contain credentials and one must write them manually.
```
SuperSandro2000, line 25: I think you mean the right thing but the module system can handle that the credentials are in files but most config files do not handle this to well.

rfcs/0141-private-derivations.md
```diff
@@ -0,0 +1,85 @@
...
+
+All of this can be prevented if we added the ability to make derivation outputs as not readable by anyone other than root, by setting the file mode to 111 (directories and executables) or 000 (regular files). We can then use a trustworthy credential manager, for example systemd with its `LoadCredential=`, to distribute these derivations to the consumers safely.
```
SuperSandro2000, line 27: What about substituter and caches?

rfcs/0141-private-derivations.md
```diff
@@ -0,0 +1,85 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+We propose adding a `noReadAccess` option to `builtins.derivation`, which, when set to true, makes this derivation a private derivation.
+
+The only difference between a private derivation and a normal derivation, apart
```
SuperSandro2000, line 34: How to handle backwards compatibility?

rfcs/0141-private-derivations.md
```diff
@@ -0,0 +1,85 @@
...
+- Supporting more complicated ACLs as described in [this](https://github.com/NixOS/nix/issues/8) Nix issue.
+- Storing private derivations in a separate store path, for example
+  `/nix/private-store` that have its executable bit removed so that the hashes
+  are not visible to non-root users.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+## Binary caches and copying
+How do we prevent the attacker from using `nix copy` to simply copy out the
```
SuperSandro2000, line 68: This is more general: how to prevent copying derivations onto machines with older nix versions or caches?

rfcs/0141-private-derivations.md
```diff
@@ -0,0 +1,85 @@
...
+private derivation to another machine?
+
+What changes are needed for binary cache providers such as `nix-serve` to handle
```
SuperSandro2000, line 71: and hydra, harmonia, etc

rfcs/0141-private-derivations.md
```diff
@@ -0,0 +1,85 @@
...
+private derivations?
+
+## Leaking metadata
+The hash is still exposed to the attacker, which opens up some possible attacks.
```
SuperSandro2000, line 75: how to prevent leaking credentials in build logs?

rfcs/0141-private-derivations.md
poscat0x04: Hmm, what is the main challenge of backwards compatibility? A machine with outdated nix versions trying to build private derivations?

rfcs/0141-private-derivations.md
poscat0x04: Hmm, I'm not very familiar with how store and copying works. Do we need assistance on the receiving end or is it just like `scp`-ing some files?

rfcs/0141-private-derivations.md
poscat0x04: They definitely might leak, but for the usual use case, which is to write some string into a file akin to `writeText`, this shouldn't be a problem. Maybe we can add this to the future works section?

rfcs/0141-private-derivations.md
poscat0x04: We should look into how content-addressed paths are implemented, since they also change the store model.

Ericson2314:
See also https://gist.github.com/edolstra/afa5a41d4acbc0d6c8cccfede7fd4792

rfcs/0141-private-derivations.md
poscat0x04: Mentioned

poscat0x04:
I don't quite like the design of store ACLs presented in the gist.

1. It is very reasonable for a public derivation to reference a private derivation. For example, systemd unit files (which should be public) can reference private configurations in their `ExecStart=` option and a public configuration file (itself a derivation) might also reference credentials.
2. Building a derivation is no longer a simple process of converting a build specification to a build product, it now involves filesystem permissions. The end product also differs depending on the user issuing the build command, which IMO breaks purity.

rfcs/0141-private-derivations.md
```diff
@@ -0,0 +1,85 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+- Supporting more complicated ACLs as described in [this](https://github.com/NixOS/nix/issues/8) Nix issue.
+- Storing private derivations in a separate store path, for example
+  `/nix/private-store` that have its executable bit removed so that the hashes
+  are not visible to non-root users.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
```
kevincox, line 65: What about starting a build to access this path? It could just copy out the value. Or would private derivations not be able to be used as input to other derivations? But in that case how would you use these in NixOS? (Maybe the path can be referenced in a build but the actual contents isn't available? That would probably require sandboxing.)

poscat0x04: Yes, the path can be referenced but the contents are not accessible during build. The build is carried out by `nixbld*` users which do not have read access of private derivations.

kevincox: I think it would be good to explicitly list this in the RFC. Along with other security properties such as `nix copy`, and interfacing with binary caches.

edolstra:
I think this needs a lot more thought. To mention a couple of issues:

* This proposal talks about changing the permission on .drv files, but doesn't say much about the output paths. Should those also be private?
* If it's only about encryption of secrets to be decrypted at runtime, then it's probably easier to add some encryption primitive to the Nix evaluator (see #5).
* "Private" is a misnomer. What is really meant is "only readable by root". (The ACL proposal has truly private, i.e. user-specific, closures.)
* An important property of Nix is the closure property, i.e. if a store path is valid and readable, its references should be valid and readable as well. This proposal breaks that, i.e. you can have "holes" in closures that you can't access. This means e.g. that you can't do a `nix copy` to build a .drv on a remote builder, or to copy a NixOS system closure to a remote machine.
* It's not clear what it means for `LoadCredential=` to "load the private derivation".  

BTW I don't understand the objections to the ACL proposal:

> Building a derivation is no longer a simple process of converting a build specification to a build product, it now involves filesystem permissions.

... but this RFC *also* involves filesystem permissions.

>  The end product also differs depending on the user issuing the build command, which IMO breaks purity.

Quite the opposite: two users who instantiate the same derivation can both build it and get the same output closure, which they can both access.

poscat0x04:
> This proposal talks about changing the permission on .drv files, but doesn't say much about the output paths. Should those also be private?

This is mentioned in the design details section. To quote, "The only difference between a private derivation and a normal derivation, apart from the hash, is that upon instantiation and after building, the read bit of the `.drv` file and the output path will be removed."

> If it's only about encryption of secrets to be decrypted at runtime, then it's probably easier to add some encryption primitive to the Nix evaluator

This RFC is about storing plaintext data in nix store, maybe I should've made it more clear in the RFC?

> "Private" is a misnomer. What is really meant is "only readable by root". (The ACL proposal has truly private, i.e. user-specific, closures.)

Yeah. Maybe a better name would be "non-readable derivations"?

> An important property of Nix is the closure property, i.e. if a store path is valid and readable, its references should be valid and readable as well. This proposal breaks that, i.e. you can have "holes" in closures that you can't access.

Again, I really don't see how the closure property is desirable. To quote my previous response, "It is very reasonable for a public derivation to reference a private derivation. For example, systemd unit files (which should be public) can reference private configurations in their ExecStart= option and a public configuration file (itself a derivation) might also reference credentials.".

If you're still not convinced, consider the following scenario: suppose I'm using NixOS and I want to store a configuration file for service A that contain sensitive information in store without it being able to be readable by everyone. Obviously, the system derivation is world readable. If we have the closure property, then since the configuration file is in the closure of the system derivation, it must also be world readable.

>  This means e.g. that you can't do a nix copy to build a .drv on a remote builder, or to copy a NixOS system closure to a remote machine.

Yes, that is by design. The user is encouraged to build the private derivations locally. Also, you can copy it if you have root access of the current machine.

> It's not clear what it means for LoadCredential= to "load the private derivation".

I meant adding a `LoadCredential = <derivation-path>` attribute to the systemd service config.

> Quite the opposite: two users who instantiate the same derivation can both build it and get the same output closure, which they can both access.

Well. I think we a different view of what does it mean for two outputs to be equal. IMO filesystem metadata such as permissions, ACL, atime and mtime are all part of the derivation. For two derivation outputs to be equal, not only do they need to have the same content, but also the same metadata. This is also what nix currently enforces: it always sets the mtime to unix time 0, the permissions to 444 or 555, the ownership to root:root. Under this view, the store ACL proposal produces different outputs when different users invoke the nix build command.

Kiskae:
I've previously tinkered with a solution that solves a lot of the same issues as this PR, so I'd like to offer it as an alternative:

### Ephemeral derivations
- derivations that produce a "private" output when it is requested as input, this output never becomes part of the nix store
- this output is never allowed to be referenced, making it safe regarding future nix closures.
- since it can never be referenced it does not need to exist in the nix store and can exist temporarily elsewhere in the file system, which means it can be implemented on macos and linux without mount namespaces.
- since outputs are generated each time they are used and don't need to be persistent, it could be directly integrated with a secrets provider like vault or systemd-creds.

#### Comparison to this RFC
- As long as there is something providing the required credentials, secret-to-secret transformations can be done as part of the nix build.
- This would not allow configuration with embedded secrets, but that is probably a sign of bad configuration and can be solved by performing runtime substitution of secrets rather than using the copy in the store. This would essentially result in the same results this RFC would accomplish while being compatible with all ways a secret can be provisioned.
- Rather than requiring a change to this nix store this probably requires quite a comprehensive change to the nix builder.

In my honest opinion any secrets that exist in plaintext while data is at rest are an issue, but with this RFC this includes all referenced secrets, while existing solutions would be restricted to only the secrets that are compromised by the plaintext decryption key.

infinisil:
Please also check out #143!

poscat0x04:
Again, my objection to the ACL model still applies:
1. The permission of a derivation and its dependencies should be entirely independent
2. ACLs are part of the file content, and hence any changes to it should also change the hash

infinisil:
RFCSC:

nyabinary:
This would be nice with eventual systemd-creds (maybe it should rebrand as an RFC focused on secrets management?)

GetPsyched:
RFCSC:

Since there has not been any significant activity, we will close this RFC. The PR can be reopened at any time if more shepherd nominations are made.

PR #143, opened 2023-03-17
[RFC 0143] Nix Store ACLs

balsoft:
# Summary

This RFC proposes a system of Access Control Lists in the Nix Store, allowing multiple users to share software in a single store without exposing it to each other.

[Rendered](./rfcs/nix-store-acls.md)

### DIFF
--- a/rfcs/0143-nix-store-acls.md
+++ b/rfcs/0143-nix-store-acls.md
@@ -1,0 +1,195 @@
+---
+feature: nix-store-acls
+start-date: 2023-01-16
+author: Alexander Bantyev
+co-authors: Silvan Mosberger, Théophane Hufschmitt
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Implement a way to only allow user access to a store path if they provide proof that they have all the necessary sources available, or had the access permission explicitly granted to them.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, the Nix Store on a local machine is world-listable and world-readable.
+
+This means that it's not possible to share a machine between multiple users who want to build proprietary software they don't want other people to look at.
+
+Also, it makes storing secrets in the Nix store even more dangerous then it could be.
+
+Finally, it means that substituter access is all-or-nothing: either a user can access the cache and download everything there is in there just by knowing store paths (even without having the source code or Nix expressions available), or they can't download anything.
+
+# Detailed design
+[design]: #detailed-design
+
+Change the implementation of the Nix daemon (and, potentially, nix-serve, depending on the chosen "Remote Store" implementation) to store access control lists as metadata, automatically update them when users provide proof that they have the necessary source, and allow `trusted-user`s to manipulate those ACLs manually.
+
+This should ensure that this change is as seamless as possible for the users: they will still always be able to execute `nix build` or similar for derivations where they have access to all the sources, substituting as much as possible, as though nothing had changed.
+The only major difference would be that `/nix/store` itself is now not readable.
+If a user needs to be able to access some store paths without having access to their sources (e.g. for proprietary software where sharing the artifacts is ok but sharing the sources isn't), such access can be granted explicitly by the administrators (`trusted-user`'s).
+
+## Local store
+
+For a local store, we keep a list of local (POSIX) users who have access to each store path as path metadata.
+When the relevant config option (perhaps `acl`) is enabled, this ACL (access control list) is enforced, meaning that only users on that list can read the store path, execute files from it, or use it as a dependency in their derivations.
+We also add a less strict mode (`selective-acl`) in which derivations are only protected if they are marked as such.
+
+For paths external to the Nix sandbox (added via `nix store add-{file,path}`, paths in the Nix language, `builtins.fetch*`, or flake sources), we add the user to the list when they send the path to the daemon.
+We might need to add a flag (like `--protect`) for the selective ACL mode.
+
+For derivations themselves (.drv files), we add the user to the list when they send the derivation to the daemon.
+
+For derivation outputs, we add user to the list when the user requests to realise the derivation and has access to the transitive dependencies, including the sources.
+
+When `selective-acl` is enabled, protected paths should not be readable by anyone during the build.
+Necessary permissions are granted after the build.
+
+There also should be a way to enable the protection for selective ACLs (perhaps `nix store access protect`), and explicitly grant (`nix store access grant`) or revoke (`nix store access revoke`) access of certain user to each individual path.
+Naturally, this should only be available to `trusted-user`s and users who have access to this path already.
+
+### `db.sqlite`
+
+The `ValidPaths` table should get a new field, something like `allowed-users`, which should list the users who have either provided proof of source for this derivation or had been explicitly granted access.
+When `selective-acl` is enabled, there should be a field like `protected` to mark a derivation as protected.
+
+### Changes to the system
+
+We should implement a way to restrict access to all the store paths for users.
+A first "line of defense" could be something like [RFC 97], which makes the store non-world-listable.
+However, it is separate from this RFC, and is not required.
+On top of that, we must enforce a stricter access control, using [POSIX ACLs](https://man7.org/linux/man-pages/man5/acl.5.html) to only allow users access to store paths if they are part of the ACL for that path.
+Nix daemon (or other local store implementations) should execute the appropriate `setfacl` calls whenever a path is added to store or gets different permissions in the db.
+
+[RFC 97]: https://github.com/NixOS/rfcs/pull/97
+
+### Nix language
+
+We change the `derivation` and `path` builtins, such that a `__permissions` attribute, when set, is expected to be a list of strings, representing the users who should be granted access to the store path (or the derivation outputs, in the case of `derivation`).
+
+If this list is set, the Nix daemon (or other store implementation) should be notified that the path/derivation is supposed to be protected (perhaps using a new worker protocol command), and after the build is completed, the ACLs should be set appropriately (only the users specified in the `__permissions` list should have access to the path).
+
+### Nix Daemon/local store implementation
+
+The Nix daemon (or the local store implementation) should, if necessary, update, and then check whether the user has permission to access all dependencies before accepting a derivation realization request from any client.
+After the build is performed, the user should be granted access to all the derivation outputs.
+Also, paths dumped to store by users should automatically be accessible by the user, and the access list should be updated should any other user dump the same path in the future.
+
+This should work as follows: when `acl` are enabled, we recursively set permissions on all the existing store paths to `500`, so that only the store owner can access it by default.
+When `selective-acl` are enabled, we only set this permission on the `protected` store paths.
+
+Whenever the user adds a path to the store (`wopAdd*ToStore`, `wopImportPaths`), we add them to the ACL in `db.sqlite` for that path, and also set the ACL in the filesystem accordingly, like this (but in C++, of course):
+
+```shell
+setfacl -R -m user:$UID:rx /nix/store/...
+```
+
+Whenever a user tries to build a store path (`wopBuildPaths*`, `wopBuildDerivation`), we check if they have access to all dependencies or they are granted permission explicilty, then we build the path if necessary, and then recursively add an entry for them to the ACL of the store path both in the DB and in the FS.
+
+All these operations should support a flag to mark the path as protected, so that it is not exposed during building or adding.
+
+There should also be a couple of new operations (perhaps `wopSetACL`/ `wopGetACL`) which allows to get and set the list of users/groups with access to the path, both in the DB and FS simultaneously.
+This should be emitted by Nix clients after the path is added to the store or a build is completed, and by `nix store access grant`/`nix store access revoke`.
+
+For all other operations working on paths, we check if the user has access before doing anything.
+
+When the access is revoked explicilty, we remove the user from the ACL:
+
+```shell
+setfacl -R -x user:$UID:rx /nix/store/...
+```
+
+Also, depending on the design of substituters, we might need to handle the new proof-of-source protocol, to prove that the necessary sources are present in the store.
+As mentioned, this should only happen if the user requesting realization has the relevant access to those paths.
+
+## Remote stores
+
+For remote stores, this problem is a bit more difficult since sending all the inputs to the remote store every time a dependency has to be downloaded is really expensive.
+There should be a protocol to establish proof-of-source, as in proof that the client has access to all the sources of a derivation, before that derivation output is provided.
+
+Simply providing the hash of the inputs as the proof would be one possible (but problematic) implementation.
+It is succeptible to replay attacks, furthermore typically obtaining the hash of the inputs is easier then obtaining the inputs themselves.
+
+Another example of such a protocol would be a challenge-response protocol, where the substituter sends a challenge salt to the client, the client then hashes all the NARs of inputs of the derivation, adding this salt, and finally sends the hash back to the substituter, which, if the hash is correct, provides the path.
+
+Yet another example would be a time-based protocol, where the salt is the current POSIX timestamp.
+The substituter then checks that the timestamp is recent enough (say, 5 seconds to allow for discrepancies between clocks and the network delays) and then validates the hash.
+If all is good, the path is returned.
+This has the advantage that it does not need two-way interaction, so it can easily work with e.g. HTTPS.
+But it's somewhat problematic since a replay attack is possible if executed quickly.
+
+Another solution (which can be implemented together with the previous one) is to keep an access list as metadata in the cache, and keep a mapping between local users with access to that list and simple credentials (e.g. simple HTTP auth).
+This has the benefit of being the easiest one to implement, but is a hassle to use (requires granting the permissions and setting up credentials externally), and also is succeptible to credential leaks.
+
+Finally, in a situation where local daemons have full access to the cache but restrict local user access, it is possible to leave the substituter logic as is, and offload all the checks to the local daemons.
+It should be rather easy to implement, since all the same checks would have to be done for substitution as for simply reusing the locally built outputs.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+## Local example
+
+Alice, Bob, Carol and Eve share a build-machine with a single nix store.
+The daemon on this machine has `acl` enabled.
+
+Alice builds some proprietary software from source.
+She can do this as usual, by just running `nix build`.
+She now can access all the relevant store paths, including the source (which has been copied to the store for the build), the build dependencies, the runtime dependencies, and the resulting derivation output.
+
+She wants to collaborate on this software with Bob.
+She shares the source with Bob, he executes `nix build` as well, which is really fast -- the only thing that happens is that he is granted access to all the same paths as Alice.
+No actual building occurs.
+
+Now, Alice and Bob want to share the resulting binaries (but not the sources) with Carol.
+They can add a `__permissions = [ "alice" "bob" "carol" ];` and re-build the derivation (nothing will get re-built, only the permissions will be updated).
+Alternatively, either of them can issue a `nix store access grant --recursive --user carol /nix/store/...-software` command, granting Carol access to software and all its runtime dependencies, but none of the sources.
+Carol can inspect and run the binary version of the software, and even (theoretically) build other software on top of it.
+
+Finally, an evil Eve wants to steal the software.
+She has multiple obstacles:
+
+1. The nix store is not readable, hence she can't easily figure out the store path of the software
+2. If she somehow figures out the store path of the software or the sources, those paths are not readable to her.
+
+The only two ways for her to get access to the software would be either to obtain the sources via some other means, at which point she doesn't really need anything in the store anyways since it would be trivial to produce the binary artifacts on her personal machine, or trick the machine's administrators into explicitly granting her access.
+
+## Remote example
+
+Alice sets up a binary cache with `acl` enabled.
+
+She uploads some proprietary software (both the source and the realised derivation output) to the store.
+
+Alice wants to collaborate on this software with Bob.
+She shares the source with him, and he can now add this substituter to his `substituters` and fetch both the sources and the binary products for local development.
+
+Alice wants to share the resulting binary with Carol.
+She grants explicit access to `carol` on the substituter, generates some credentials and adds them to the "password file" for the substituter.
+Carol then can add the substituter with these credentials to her `substituters`, and fetch the artifact (but not the source) to run it locally.
+
+Evil Eve still can't steal the software, since to download the path she once again either needs to obtain the source or the credentials to download it from the cache.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This change requires significant refactoring of many Nix components;
+- Implementing ACLs will impose a performance penalty. Hopefully this penalty will not be too bad, and also since ACLs are entirely optional this shouldn't affect users who don't need it;
+- Short secrets in the store can be brute-forced, since the hash of the content is known, unless [RFC 97] is also implemented.
+- Security achieved by ACLs on a multi-user system will depend on the Nix daemon implementation, which has quite a large attack surface;
+- Syncing ACLs between machines can be difficult to do properly.
+
+# Alternatives
+[alternatives]: #alternatives
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Does the nix store permission setup belong as a NixOS module in nixpkgs, or as part of the Nix installation, or both? More investigation is needed.
+- How to handle the remote store case.
+
+# Future work
+[future]: #future-work
+


bew:
'rendered' link is wrong, should be https://github.com/tweag/rfcs/blob/acls/rfcs/0143-nix-store-acls.md

(will delete this message once you edited)

edolstra:
This RFC is open for shepherd nominations! (It already was but we never asked for nominations explicitly...)

Ericson2314:
I am not sure whether this is a good thing to prioritize at this moment, but I do want to get here eventually. I nominate myself,

thufschmitt:
I'd gladly nominate myself as a shepherd (with a conflict of interest warning as this comes from a client project for Tweag, so there shouldn't be only Tweagers in the shepherd team).

@edolstra, I'd also nominate yourself if you're willing since you're both knowledgeable and interested in that problem space

rfcs/0143-nix-store-acls.md
```diff
@@ -0,0 +1,195 @@
+---
+feature: nix-store-acls
+start-date: 2023-01-16
+author: Alexander Bantyev
+co-authors: Silvan Mosberger, Théophane Hufschmitt
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
```
lheckemann, line 7:
```suggestion
shepherd-team: John Ericson, Théophane Hufschmitt, Eelco Dolstra
shepherd-leader: Théophane Hufschmitt
```
@thufschmitt are you willing to take the lead on this RFC?

thufschmitt: Gladly :)

kevincox:
@thufschmitt what is the status of this RFC? Is there anything that is needed to move it towards FCP?

thufschmitt:
> @thufschmitt what is the status of this RFC? Is there anything that is needed to move it towards FCP?

Sorry for the long silence. It's ongoing, although for $REASONS it's hard to find a time to gather the team.

rfcs/0143-nix-store-acls.md
```diff
@@ -0,0 +1,195 @@
...
+setfacl -R -x user:$UID:rx /nix/store/...
+```
+
+Also, depending on the design of substituters, we might need to handle the new proof-of-source protocol, to prove that the necessary sources are present in the store.
+As mentioned, this should only happen if the user requesting realization has the relevant access to those paths.
+
+## Remote stores
+
+For remote stores, this problem is a bit more difficult since sending all the inputs to the remote store every time a dependency has to be downloaded is really expensive.
+There should be a protocol to establish proof-of-source, as in proof that the client has access to all the sources of a derivation, before that derivation output is provided.
```
thufschmitt, line 112: IIRC [attic](https://github.com/zhaofengli/attic) has a similar problem, and some solution to it. It would be nice to see how they handle it (maybe Cachix has it too)

rfcs/0143-nix-store-acls.md
```diff
@@ -0,0 +1,195 @@
...
+Alice builds some proprietary software from source.
+She can do this as usual, by just running `nix build`.
+She now can access all the relevant store paths, including the source (which has been copied to the store for the build), the build dependencies, the runtime dependencies, and the resulting derivation output.
+
+She wants to collaborate on this software with Bob.
+She shares the source with Bob, he executes `nix build` as well, which is really fast -- the only thing that happens is that he is granted access to all the same paths as Alice.
+No actual building occurs.
+
+Now, Alice and Bob want to share the resulting binaries (but not the sources) with Carol.
+They can add a `__permissions = [ "alice" "bob" "carol" ];` and re-build the derivation (nothing will get re-built, only the permissions will be updated).
```
thufschmitt, line 148: > nothing will get re-built, only the permissions will be updated

Re. my comment about whether `__permissions` should affect the hash: This means that it doesn't. If you go that way it should be spelled out explicitly

rfcs/0143-nix-store-acls.md
```diff
@@ -0,0 +1,195 @@
...
+Carol then can add the substituter with these credentials to her `substituters`, and fetch the artifact (but not the source) to run it locally.
+
+Evil Eve still can't steal the software, since to download the path she once again either needs to obtain the source or the credentials to download it from the cache.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This change requires significant refactoring of many Nix components;
+- Implementing ACLs will impose a performance penalty. Hopefully this penalty will not be too bad, and also since ACLs are entirely optional this shouldn't affect users who don't need it;
+- Short secrets in the store can be brute-forced, since the hash of the content is known, unless [RFC 97] is also implemented.
```
thufschmitt, line 180: I would be a bit stronger here (or somewhere else), and highlight that this isn't a good mechanism for storing secrets

thufschmitt (commented):
Left an army of comments :)

Overall: I like the broad direction quite a lot, but there is a handful of things to iron out and details to clarify

rfcs/0143-nix-store-acls.md
```diff
@@ -0,0 +1,195 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+Change the implementation of the Nix daemon (and, potentially, nix-serve, depending on the chosen "Remote Store" implementation) to store access control lists as metadata, automatically update them when users provide proof that they have the necessary source, and allow `trusted-user`s to manipulate those ACLs manually.
+
+This should ensure that this change is as seamless as possible for the users: they will still always be able to execute `nix build` or similar for derivations where they have access to all the sources, substituting as much as possible, as though nothing had changed.
+The only major difference would be that `/nix/store` itself is now not readable.
+If a user needs to be able to access some store paths without having access to their sources (e.g. for proprietary software where sharing the artifacts is ok but sharing the sources isn't), such access can be granted explicitly by the administrators (`trusted-user`'s).
```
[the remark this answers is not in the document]
balsoft: > This might not matter too much, but shouldn't anyone with access to the path be allowed to share it further?

Yes, indeed

rfcs/0143-nix-store-acls.md
```diff
@@ -0,0 +1,195 @@
...
+
+The `ValidPaths` table should get a new field, something like `allowed-users`, which should list the users and groups who have either provided proof of source for this derivation or had been explicitly granted access.
+When `selective-acl` is enabled, there should be a field like `protected` to mark a derivation as protected.
+
+### Changes to the system
+
+We should implement a way to restrict access to all the store paths for users.
+A first "line of defense" could be something like [RFC 97], which makes the store non-world-listable.
+However, it is separate from this RFC, and is not required.
+On top of that, we must enforce a stricter access control, using [POSIX ACLs](https://man7.org/linux/man-pages/man5/acl.5.html) to only allow users access to store paths if they are part of the ACL (or belong to a group on the ACL) for that path.
```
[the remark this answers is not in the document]
balsoft: Yes, they are available but the API differs a lot.

rfcs/0143-nix-store-acls.md
```diff
@@ -0,0 +1,195 @@
...
+Nix daemon (or other local store implementations) should execute the appropriate `setfacl` calls whenever a path is added to store or gets different permissions in the db.
+
+[RFC 97]: https://github.com/NixOS/rfcs/pull/97
+
+### Nix language
+
+We change the `derivation` and `path` builtins, such that a `__permissions` attribute, when set, is expected to be a list of strings, representing the users and groups who should be granted access to the store path (or the derivation outputs, in the case of `derivation`).
+
+If this list is set, the Nix daemon (or other store implementation) should be notified that the path/derivation is supposed to be protected (perhaps using a new worker protocol command), and after the build is completed, the ACLs should be set appropriately (only the users specified in the `__permissions` list should have access to the path).
```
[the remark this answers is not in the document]
balsoft: > Should that attribute have an impact on the derivation hash? It feels like it shouldn't (as it only touches the metadata), but at the same time not doing so wouldn't be properly backwards-compatible

It probably shouldn't. The lack of backwards-compatibility is annoying, but since `__permissions` wouldn't be `toString`-able, it would result in a clear failure on older Nix, which is a fine behaviour IMHO.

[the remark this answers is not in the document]
balsoft: > should bob be given access to "${./somePath}" since it's in the closure of that derivation output?

I don't think so; IMHO, Permissions are applied only to the object on which they are specified.

> And what if I don't add myself to the __permissions field?

Perhaps there should be a check to automatically add the evaluating user to the __permissions.

rfcs/0143-nix-store-acls.md
balsoft: I'd say it depends on what is meant by "secrets"; after all, the only reason to have ACLs in the first place is to have some information which is only accessible to a select group, and thus "secret" in some sense. Of course, it should be highlighted that this is intended for "secret" (proprietary or otherwise protected) source-code and build artifacts, rather than runtime secrets such as credentials.

infinisil:
## 2023-10-10 Meeting 1
Attendees: @balsoft @edolstra @thufschmitt @infinisil 
Notes by: @infinisil 

* Eelco: Closure invariant: Need to have access to the whole closure of a path
   * Alex: Sometimes doesn't make sense, you might want to have a library be proprietary and link it to, but have the result be public.
   * Theophane: Generally inputs may be private, output may be public
* Eelco: User need to exist before they can be given access
   * Alex: Indeed, users need to exist at eval time
* Silvan: How about just agreeing on design constraints in the RFC, not go into a specific implementation
   * Alex: Probably better that way
* Theophane:
   * Closure invariant, very fundamental to Nix, probably don't want to break that
   * How to specify permissions, currently a derivation builtin attribute
* Eelco: Should take a closer look at [my design](https://gist.github.com/edolstra/afa5a41d4acbc0d6c8cccfede7fd4792)
   * Privacy is not a property of a derivation, but rather how the user builds it with the CLI
   * Alex: Mostly the case as well. It's convenient to specify which users get access to paths in a company. The __permissions is like syntactic sugar for a similar CLI interface.
* Alex: Regarding runtime closure requirement, doesn't feel secure, might leak details you don't want to have exposed
   * Theophane: Should err on the side of being cautious, deny access by default.
   * Reminds of requiredReferences/allowedReferences, something similar could work
   * Alex: Have a constraint when setting acls on a store path, the entities must also be able to access the runtime closure. Set permissions recursively.
* Alex: Another thought: If a user has access to the build-time closure, should they get access to the resulting path as well?
   * Theophane: E.g. remote builders have extra sandbox paths that could provide extra secret stuff
   * Alex: Non-reproducible derivations don't work since the hash doesn't work.
   * Theophane: Poison the closure to indicate impurities in the build graph. Having to rebuild to give more people access is expensive
   * Alex: How would you poison the closure? Not sure it could work
   * Theophane: We have a somewhat weak security model
   * Alex: Impure derivations probably make this work
   * Eelco: You need to prove that you can instantiate the derivation, then you get added to the ACLs. If you have access to a .drv you get access to the output paths.
      * Alex: Something in the build inputs might be protected.
      * Eelco: Combining the drv closure invariant and drv giving you access to the outputs, then you get access the entire graph
      * Alex: What about non-derivation paths?
      * Eelco: Also would be in the drv inputs, included in drv closure
      * Alex: Need the hash to instantiate the derivation
      * Theophane: To make it private you need to prove you have the content
      * Eelco: If you try to create a drv but some references aren't readable it would be rejected. Need to provide proof to the daemon that you have access to them.
      * Alex: Yeah that would also work. One of these might be more efficient
* Eelco: Fear with ACLs: Requires auditing of Nix to ensure no leaks. E.g. daemon operations that return a NAR for a store path
   * Theophane: Only a fixed number of daemon protocol operations, the boundary would be there
   * Alex: Not many places where you read from the store.
   * Daemon vs store boundary, maybe both
* Eelco: How does the RFC handle building, temporary build user access?
   * Alex: check if the user has access to the inputs; grant access to the build user, revoke access afterwards
   * Eelco: Risk of killing the build not revoking access. Write a sort of recovery log that shows what access was granted so it can be cleaned up
   * Alex: Maybe better, use extended file system acls to mark the build user that has access
   * Eelco: Doesn't work, would require scanning the store
   * Eelco: Not a problem with the sandbox
   * Theophane: Could require the sandbox for private paths
   * Silvan: Safe default to require it
   * Eelco: See https://github.com/edolstra/nix/blob/e00259c7c83abc62c9221b065a636edac2e55beb/src/libstore/build/local-derivation-goal.cc#L665-L687 and https://github.com/edolstra/nix/blob/e00259c7c83abc62c9221b065a636edac2e55beb/src/libstore/build/local-derivation-goal.cc#L993C27-L1006 for how to handle this
* Alex: Consider dropping ACLs in the build, would be an impurity
   * Might require a fuse layer, expensive
   * Silvan: Probably not a big deal to have this impurity
* Eelco: What are the use cases?
   * E.g. teams not being able to access each others closures
   * E.g. NixOS use case: Files with passwords in them
      * E.g. user would build it as a private closure, at boot time the activation script would make paths accessible somehow, haven't gotten to implementing that
      * Alex: Might not be a good idea because everything is hashed
      * Silvan: E.g. imagine having a pin, could be brute-forced checking against the hash
      * Eelco: Could have a salt
      * Eelco: Theoretical concern, would require a lot of brute forcing to exploit
      * Theophane: People could put password as a standalone file in the store
      * Alex: Should be able to store small secrets as well
      * Silvan: Trivial example of storing a single bit in the store, could be easily brute-forced
   * Eelco: Need to have good use cases to justify the complexity
* Theophane: Should also go through the [encryption RFC](https://github.com/NixOS/rfcs/pull/5)
   * Deterministic encryption
* Alex: Current loc: 2.5k lines total, 300 lines test, maybe 200 docs
* Silvan: We can all agree that we want something like ACLs (only need to make sure it's secure, worth it, doable, etc.)
* Alex: Will look at Eelco's [previous design](https://gist.github.com/edolstra/afa5a41d4acbc0d6c8cccfede7fd4792)

infinisil:
# 2023-11-07 Meeting 2
Attendees: @infinisil @edolstra @Ericson2314 @thufschmitt @balsoft 
Notes by: @infinisil 


Notes
* PR open: https://github.com/NixOS/nix/pull/9287
   * Needs more tests
* Updated RFC after the last meeting
* Looked at Eelco's previous ACL proposal
   * Remove some bits from the RFC so it could be closer to that
* Don't have the variant "If you have access to a path, have access to the closure"
   * Should list arguments for and against that in the RFC
   * Eelco: Very critical imo
   * Alex: Use case: Give access to the output but not to the input, e.g. not giving access to the source code
   * Alex: What if it's only the default but can be changed?
      * Eelco: Doesn't sound reasonable, kind of like a dangling reference
      * Alex: Good point, will rethink
* Add use cases in the RFC
   * Include non-supported use cases, e.g. like storing small use cases
* No need to pin down details in the RFC
   * Best remove everything that's not necessary
   * RFC should specify the expected semantics and use cases
* Eelco: Would like secrets to be supported
   * Alex: Does work, but can be attempted to brute-force using the store hash
      * E.g. storing a secret 4-digit pin in the store
   * Lots of discussion in the previous RFC 005
      * Use salt, but where does it come from?
   * Make the store not readable by everybody?
      * Mentioned in the RFC
   * Only a problem with fixed-output/CA derivations?
      * Such secrets would often be added from local paths -> fixed-output/CA
   * Attack vector of being able to test whether specific secrets are in the store
   * Alternatively just make it clear that people shouldn't use brute-forceable secrets
* John: Consider fuse filesystem whose interface really hides things, provide separate view of the store to every user
   * Could make adding features much easier and more flexible
   * Comparatively, using unix features to implement all of these features is very tricky
   * Eelco: Ideally ACL-based and fuse-based store interface would be the same
   * John: Soon in linux: fuse passthru, makes things a lot faster
   * Eelco: Not sure if fuse is easier or harder, requires a daemon to be running
      * Might be hard for NixOS, since it would need the daemon running perhaps already in the initrd
* John: ACLs as a security tool seems kind of wrong
   * Don't want to write down a username in Nix code
   * Eelco: How does that work if users don't exist yet?
      * Alex: Fails, mainly for convenience when on the local system
   * Doesn't really make sense on NixOS where users are only created during activation
      * Alex: Would give access to the user with the Nix command in the activation script, no need to specify the user in Nix code
   * Alternatively could support that in some post build hook
      * Alex: Doesn't work, it's only metadata and doesn't end up in the store, wouldn't want to rebuild to change it
   * John: In capability systems, it's not an intrinsic property who has access to what, you always need a reference to give you access. In the very end of e.g. a NixOS system you have some export list about access
      * Map from user to path rather than path to user?
   * Is it really a good use case to explicitly give somebody else access to a store path, instead of everybody just building and getting access themselves?
      * Useful in teams
* What if two users have the same expression but with different access control? Would access be revoked when the other builds it?
   * Alex: Yes, makes sense to me
      * Silvan: Please put it as an example into the RFC
   * John: Revoking permissions is hard to do modularity
      * Initially had a grant tree but didn't end up working out well
   * Maybe don't support declarative revoking of permissions
* John: Coming back to capability systems, what if we didn't have to worry about using ACLs underneath
   * Maybe ACLs is a mismatch with Nix
   * Alex: I don't think it's a mismatch, just an implementation detail
   * How would a capability system look like?
   * John: Might be able to write down some thoughts on that
* John: Can the rootless daemon be finished soon? Having that and the no o-r on the /nix/store (RFC 97) would help a lot
   * Alex: These seem orthogonal
   * John: Implementation-wise orthogonal, but conceptually very related, first step to not think of it as an open world


Action items
* Alex: Write down use cases
* Alex: Add runtime closure invariant and implement it
* Alex: Write advantages/disadvantages about __permissions

YorikSar:
I see FUSE mentioned in the discussion here. I'd like to note that on macOS FUSE is hard to set up (you have to mangle with System Integrity Protection stuff), so it would not work for people who just need Nix. ACLs on the other hand are supported pretty much everywhere.

Ericson2314:
I would rather not support macOS than support I feature that I feel is in the wrong direction. macOS Nix talking to Linux Nix still can get benefits.

----------

(Adding more detail/polish to a few things I said in the meeting)

I am a proponent of Capability-based security as generally better, and in particular "the Nix way" giving the shared overlaps with functional programming, etc. While it is possible have some capability system and then set and unset file ACLs based on that, the result amplifies reads into writes (O(1) share capability becomes O(files in closure) `setfacl` calls), which is bad for performance, and bad for introducing more state. I am thus very skeptical that good things can come from file ACLs.

The order I am thinking of is:

1. Rootless daemon, and using rootless daemon with NixOS. Results:
     - The bulk of Nix is out of the trusted compute base. This is sort of table stakes about thinking about security at all.
     - Store objects that NixOS needs should be root owned. Either that means different children of  `/nix/store` have different owners/groups --- making Nix cope with that is IMO "MVP ACL store", or another solution such as "baking" separate stores for NixOS.

2. RFC-97. Exactly what this looks like is different based on the paths taken above (e.g. is there one store or multiple). Results:
  - No more unprivileged figuring out what all the store paths in the store are
  - Potential problem: store paths granting permissions may not be good enough if store paths are too easy to forge (e.g. because of various side channels)

3. Non-forgable capabilities: Guarding access to the store with stronger capabilities than store paths. Examples:
- Special daemon will hand out directory or mount namespace file descriptors with just the closure of root path(s) available.
- Systemd services sandboxed in various ways (dynamic UIDs, Linux namespeaces, etc.) only get access to their closures, regardless of what store paths will be guessed.

A basic theme here is that with the exception of the rootless daemon, which is almost finished, I think most of the work *doesn't* need to happen with Nix itself. I think its is more of an integration-y / dev-ops exercise in making various bits of software continue working as we take away more and more permissions. And dev work probably goes into separate little high-privilege daemons (like the GC roots scanner that comes with the rootless daemon), not the "main bulk" of Nix.

thufschmitt:
@Ericson2314 this is giving up on an existing batteries-included cross-platform implementation in favour of a hypothetical build-your-own-cake Linux-only one. As much as I like your final vision, I don't see it happening any time soon (if ever).

Ericson2314:
As always, we want to make sure Nix is going in the right direction. Some experimental features are alright, but also nicely self contained. https://github.com/NixOS/nix/pull/9287 looks pretty deeply involved some some core interfaces, making a *lot* more code away of Unix users. For me, the more deeply entwined something is, the higher the bar is that it be for sure in the "right" direction.

roberth:
I'm sorry to have to say this, but I feel strongly that the Nix team does not have the bandwidth to work on this concept. It vastly complicates the Nix model, which we simply can't afford, considering the state of the other unfinished projects.

For what it's worth, the existence of a proof of concept PR doesn't change anything about that, because usually the responsibility to *actually* complete the project falls on the Nix team, which has to respond to issues, fix bugs and generally put in a lot of effort to make sure invariants are upheld. In this case those a security properties. Nothing to sneeze at.

I very much appreciate @Ericson2314's suggestion to explore what can be done by simplifying and decoupling *what we already have*, which is actually feasible now, because it does not require any changes to the model.

@thufschmitt No such implementation exists, because we can not support it.

**I urge the Nix team to focus on finalizing existing projects first and postpone this RFC until we have time to consider a significant feature like this**

infinisil:
## 2023-11-21 Meeting 3
Attendees: @infinisil, @thufschmitt, @edolstra, @balsoft
Notes by: @infinisil 

* @balsoft: Update: Added closure invariant to the RFC
* Regarding @Ericson2314's ideas in https://github.com/NixOS/rfcs/pull/143#issuecomment-1799381398
   * @balsoft: Should be a separate RFC
* Regarding @roberth's concerns in https://github.com/NixOS/rfcs/pull/143#issuecomment-1802072148
   * Can it be separated more from core Nix? Like a separate store implementation
   * @balsoft: Partially done already, but could be extended
   * Eventually upstream into local store
   * Problem with local derivation goal, no abstraction layer currently, needed to set ACLs on the fly
   * @infinisil: Would this be in-tree or a plugin?
      * Just a single separate file, but in-tree
   * @balsoft: Can we merge the RFC even if we aren't implementing it yet?
      * @thufschmitt: It's going to be a maintenance burden, RFC's shouldn't promise what can't currently be done
      * @balsoft: But it's gonna be unblocked in the future
      * @infinisil: RFC acts as a blocker against the Nix team having to care about it
   * @thufschmitt: Change the RFC to enable Nix to have out-of-tree store implementations?
      * @balsoft: Is that going to help with with @roberth's concerns? Invariants are the tricky bit, that would have to be in-tree
      * @thufschmitt: Store interface is reasonably stable, might work
   * @balsoft: Can try implementing ACLs as a subclass of local store. Trying to move as much functionality outside of local derivation goal
      * @thufschmitt: LocalDerivationGoal changes needed?
      * @balsoft: Yes for getting access to paths
      * @thufschmitt: What if paths are accessible by default to all build users?
      * @balsoft: Sounds problematic, builders change, requires changes on the store
      * @balsoft: Use case of enabling ACLs, requires going through the entire store
         * @thufschmitt: Not a problem, only needed for ACLs paths, store is world-readable by default
      * @balsoft: Have to figure out which paths have ACLs, require expensive traversing
      * @balsoft: Feels hacky generally, how to detect change in build users?
         * @thufschmitt: Build group might work
         * @balsoft: What if build user gets removed?
         * @edolstra: What about auto uid allocation
         * Would not work, could be a follow-up
      * @infinisil: Sounds like more experimentation needed to see if it works with little maintenance burden
   * @edolstra: I thinks it's unavoidable to touch the core of Nix
      * @thufschmitt: Question is how much it can be avoided in a first version
      * @balsoft: Looks very difficult, would probably have to drop the primop
      * @thufschmitt: Best way to succeed: Shrink it to its minimum, add things on top later
      * @infinisil: Still requires the same maintenance amount in the end
      * @thufschmitt: But more manageable with smaller PRs
      * @infinisil: Should always have small PRs irrespective of the RFC
      * @balsoft: Reasonable small commits already, can be split up
      * @infinisil: Not worth splitting up if it's not gonna get merged. Should talk to Nix team about whether to accept this or not.
      * @infinisil: Already half the Nix team generally against
      * @thufschmitt: Room for negotiation if it's smaller and more split up

infinisil:
2023-12-19 Meeting 4

Attendees: @infinisil, @thufschmitt, @edolstra, @balsoft
Notes by: @infinisil


* @edolstra: Seems hard to do without deeply involving the store
   * @balsoft: Either extending plugins to have a local-store subclass, changing a couple things and adding functions, though wouldn't be composable
   * Or integrate into Nix, but keep diff very small by creating separate files for everything
* @edolstra: Users need to be passed around in the daemon?
   * @balsoft: There could be a way to hook plugin data into the worker protocol
* @edolstra: Mark it as experimental to future-proof?
   * @infinisil: Probably good to break it regularly, otherwise users rely on it
   * @edolstra: Doesn't mean people will use it, encourage more testing, set expectations
   * @infinisil: Trial period of e.g. 1 year. If success -> stable, otherwise remove again.
   * @thufschmitt: What if people don't try it out?
   * @infinisil: Have a warning after half a year that it will get removed again unless people give positive feedback in a linked issue
   * @balsoft: What is the criteria?
   * @infinisil: Has to be decided on a case-by-case basis
* @thufschmitt: Would the new plugin infra for worker protocol be useful for non-ACL use cases?
   * @edolstra: How would that worker protocol plugin look like?
   * @balsoft: Register plugin as a worker protocol plugin, allow operations that pass data between client/daemon based on plugin name
   * @thufschmitt: JSON?
   * @balsoft: Any serialisation works
* @infinisil: Nix team has little resources but the RFC is generally a good idea, go towards accepting it?
   * @thufschmitt/@balsoft: Important to make it as easy as possible on the Nix team
   * @balsoft: Maybe RFC could be simplified to just "We want ACLs and they should look something like this", probably not controversial at all
   * @edolstra: Semantics is important, more like my original proposal, focusing more on use cases
   * @edolstra: Prototype could be okay if minimal, plugins seem overcomplicated for this
   * Make it as minimal as possible? But how minimal can it be made?
   * @balsoft: Could drop some parts of the RFC
   * @balsoft: Drop CLI commands? Probably doesn't work
      * @edolstra: CLI command can be in a separate file, marked as experimental
   * @thufschmitt: Daemon changes could be dropped by having ACL stuff be run as root
* @infinisil: Sounds fine to merge if we properly figure out how to merge it as experimental
   * @thufschmitt: John and Robert made good points against merging it. Would be hard to undo.
   * @infinisil: Combine with the idea of making the diff minimal
   * @thufschmitt: Not easy
* @thufschmitt: In favor of merging it as minimally as possible, expand upon that, discussing on the changes as we go
   * @balsoft: Need to prevent an eternal experimental feature situation
   * @infinisil: Combine with an experimental feature timeline
      * @thufschmitt: Not sure if it needs to be in the RFC
      * @infinisil: This RFC could lay the groundwork
      * @thufschmitt: +1
   * @thufschmitt: Doesn't solve the problem of limited maintainer time, but might be more okay very incrementally
   * @infinisil: Justify maintainer cost by generalising the work to as much of the codebase as possible, e.g. a general implementation of experimental feature handling
   * General agreement on this
   * Should ask Robert on that, but after the holidays

balsoft:
Drafted due to no progress being possible at the moment, particularly in light of @roberth 's comment and steering committee discussions.

PR #144, opened 2023-03-19
[RFC 0144] Versioned Flake References

figsoda:
# Summary
[summary]: #summary

Flake references can have a special placeholder that will specify the version
requirement of the flake, which `nix flake update` can use to update the pin to
the latest version compatible with the version specification. A new Nix
command, `nix flake upgrade` will upgrade the version requirement in
`flake.nix` to the latest possible version without regard to compatibility.

[Rendered](https://github.com/figsoda/rfcs/blob/version/rfcs/0144-versioned-flake-references.md)

### DIFF
--- a/rfcs/0144-versioned-flake-references.md
+++ b/rfcs/0144-versioned-flake-references.md
@@ -102,6 +102,7 @@ Flake reference | The tag it points to | The `upgrade`ed flake reference
 
 - Using a different syntax for the placeholders
 - Using a version scheme other than semantic versioning
+- Adding flags to `nix flake update` instead of creating a new `upgrade` command
 - Not adding the feature, manually update the flake references
 
 # Unresolved questions
--- a/rfcs/0144-versioned-flake-references.md
+++ b/rfcs/0144-versioned-flake-references.md
@@ -1,0 +1,117 @@
+---
+feature: versioned-flake-references
+start-date: 2023-03-19
+author: figsoda
+co-authors: None
+shepherd-team: None
+shepherd-leader: None
+related-issues: None
+---
+
+# Summary
+[summary]: #summary
+
+Flake references can have a special placeholder that will specify the version
+requirement of the flake, which `nix flake update` can use to update the pin to
+the latest version compatible with the version specification. A new Nix
+command, `nix flake upgrade` will upgrade the version requirement in
+`flake.nix` to the latest possible version without regard to compatibility.
+
+# Motivation
+[motivation]: #motivation
+
+This will allow Nix libraries to be versioned without requiring their users
+to manually update them. Some package managers (e.g. cargo) for more
+conventional programming languages have this functionality built-in, allowing
+library authors to introduce breaking changes in a communicatable way that will
+not break their downstream dependents's code.
+
+# Detailed design
+[design]: #detailed-design
+
+## Syntax
+Version placeholders will be in `{}`, with a list of version requirements
+separated by `,` (commas), e.g. `{^1.0,<3}` will expand to a version that
+specifies the version requirements `^1.0` and `<3`
+
+Version requirements consists of a comparator and version. The version does not
+have to be a valid version defiined by the versioning scheme, `1` and `1.0` are
+also valid versions for version requirements. The comparator has to be one of
+the following options:
+- `^` compatible (as defined by the versioning scheme)
+- `<` less than
+- `<=` less than or equal to
+- `>` greater than
+- `>=` greater than or equal to
+- `=` equal to
+
+A version placeholder with no version requirements will match all valid
+versions defined by the versioning scheme.
+
+## Versioning scheme
+The versions will follow [semantic versioning] (semver). The tags have to
+follow semver, `1.0` and `1.0.0.0` will not match the placeholder `{=1.0}` as
+it is not a valid version defined by semver. Build identifiers will not be
+taken into consideration when calculating version requirements, and the latest
+tag will be selected if multiple tags have the same precedence defined by
+semver.
+
+## Upgrading
+A new Nix command, `nix flake upgrade`, will edit `flake.nix` and upgrade all
+the version placeholders in the inputs. A new flag, `--upgrade-input` will
+upgrade only the specified flake input. Only version placeholders that contain
+exactly one version requirement which the comparator is either `^` or `=` will
+be upgraded. All other version placeholders will be left unchanged.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+Here are the git tags of a hypothetical Nix flake "github:foo/bar":
+- v0.1.0
+- v0.1.1
+- v0.2.0
+- v1.0.0
+- v1.0.1
+- v1.1.0
+- v2.0.0
+- unrelated
+
+Flake reference | The tag it points to | The `upgrade`ed flake reference
+-|-|-
+`github:foo/bar/v{^1}` | `v1.1.0` | `github:foo/bar/v{^2}`
+`github:foo/bar/v{=1.0}` | `v1.0.1` | `github:foo/bar/v{=2.0}`
+`github:foo/bar/v{^1.0}` | `v1.1.0` | `github:foo/bar/v{^2.0}`
+`github:foo/bar/v{=1.0.0}` | `v1.0.0` | `github:foo/bar/v{=2.0.0}`
+`github:foo/bar/v{^1.0.0}` | `v1.1.0` | `github:foo/bar/v{^2.0.0}`
+`github:foo/bar/v{^0.1.0}` | `v0.1.1` | `github:foo/bar/v{^2.0.0}`
+`github:foo/bar/v{<2}` | `v1.1.0` | `github:foo/bar/v{<2}` (no change)
+`github:foo/bar/v{>1.0}` | `v2.0.0` | `github:foo/bar/v{>1.0}` (no change)
+`github:foo/bar/v{^1,<1.1}` | `v1.0.1` | `github:foo/bar/v{^1,<1.1}` (no change)
+`github:foo/bar/v{}` | `v2.0.0` | `github:foo/bar/v{}` (no change)
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This is impossible with some types of flake references, such as `path`s and
+  `git+` URLs, which will make the flake interface less consistent.
+- This adds extra complexity to `nix flake update`.
+- The difference between `update` and `upgrade` might cause confusion.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Using a different syntax for the placeholders
+- Using a version scheme other than semantic versioning
+- Not adding the feature, manually update the flake references
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- How will this work in query parameters?
+- Should only tags be used, or should branches also be considered?
+- Should whitespace be allowed in the version placeholders?
+
+# Future work
+[future]: #future-work
+
+[semantic versioning]: https://semver.org/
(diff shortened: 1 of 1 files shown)


alyssais:
I'm not sure it makes sense to have RFCs for modifications to flakes, when flakes itself is still experimental and pending an RFC?  ([Previously](https://github.com/NixOS/rfcs/pull/105#issuecomment-1033765525))  It might be better to just discuss with the Nix developers.

Profpatsch:
It was only a matter of time.

figsoda:
@edolstra @roberth What do you think?

I feel that this proposal would benefit from being an RFC, but if it fundamentally shouldn't be, feel free to close this and I can turn this into an issue in the nix repo.

roberth:
> would benefit from being an RFC

Unlike the first flakes RFC, this one has a reasonable scope. I feel like this RFC could provide a good centerpoint for the discussion of such a feature.

@figsoda do you plan to implement this feature yourself? An issue in the repo could serve a complementary role, such as for discussion of implementation work for example. We have an RFC label :)

@Profpatsch it's not clear to me what you are referring to. Could you elaborate?

7c6f434c:
> Unlike the first flakes RFC, this one has a reasonable scope. I feel like this RFC could provide a good centerpoint for the discussion of such a feature.

Well, another question is which of the assumptions about flakes might become complicated issues in the next attempt to perform stabilisation through RFC. (The current plan might avoid all of them; but RFCs often get large rewrites during the discussion)

vcunat:
I'd rather add some `--flag`(s) to the `update` sub-command instead of adding `upgrade`?

thufschmitt:
I don't really have an opinion whether that change could be a good one for Nix or not. However, since this looks like it could be easily implemented outside of Nix, I'd strongly suggest doing so first so that everyone can judge the usefulness. It would be much faster for a first version, and that would make the case much stronger. Especially since the main concern right now with Flakes is getting them stabilized, which goes against adding extra features on top of them in the short term.

shlevy:
How would this be implemented outside of nix? Doesn't it change the schema of the `inputs` of `flake.nix`?

shlevy:
I like the idea, but think versions should be in `flake.nix` and not use SCM tags.

rfcs/0144-versioned-flake-references.md
```diff
@@ -0,0 +1,118 @@
...
+`github:foo/bar/v{<2}` | `v1.1.0` | `github:foo/bar/v{<2}` (no change)
+`github:foo/bar/v{>1.0}` | `v2.0.0` | `github:foo/bar/v{>1.0}` (no change)
+`github:foo/bar/v{^1,<1.1}` | `v1.0.1` | `github:foo/bar/v{^1,<1.1}` (no change)
+`github:foo/bar/v{}` | `v2.0.0` | `github:foo/bar/v{}` (no change)
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This is impossible with some types of flake references, such as `path`s and
+  `git+` URLs, which will make the flake interface less consistent.
```
edolstra, line 96: Why is it impossible with `git+` URLs? Seems like the prerequisite is an input type that supports enumerating a set of versions (e.g. by getting the tags of a repo).

rfcs/0144-versioned-flake-references.md
```diff
@@ -0,0 +1,118 @@
...
+The versions will follow [semantic versioning] (semver). The tags have to
+follow semver, `1.0` and `1.0.0.0` will not match the placeholder `{=1.0}` as
+it is not a valid version defined by semver. Build identifiers will not be
+taken into consideration when calculating version requirements, and the latest
+tag will be selected if multiple tags have the same precedence defined by
+semver.
+
+## Upgrading
+A new Nix command, `nix flake upgrade`, will edit `flake.nix` and upgrade all
+the version placeholders in the inputs. A new flag, `--upgrade-input` will
```
edolstra, line 61: Upgrading `flake.nix` in-place is not really an option since we don't have the ability to do that. But I also don't see why it's necessary. `nix flake update` can simply update to the latest versions that are consistent with the version placeholders. At most, we want a flag to `nix flake update` to allow/disallow major upgrades that wouldn't be semantically compatible.

rfcs/0144-versioned-flake-references.md
```diff
@@ -0,0 +1,118 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+- Using a different syntax for the placeholders
+- Using a version scheme other than semantic versioning
+- Adding flags to `nix flake update` instead of creating a new `upgrade` command
+- Not adding the feature, manually update the flake references
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
```
edolstra, line 109: Not discussed here is the algorithm for selecting versions. In conjunction with follows/overrides, finding a feasible solution could become expensive, e.g. in
```
inputs.foo.url = "github:bla/foo/{>1.2,<1.5}";
inputs.bar.url = "github:bla/bar/{^1}";
inputs.bar.inputs.foo.follows = "foo";
```
Nix would need to find a version of `bar` whose `foo` input has constraints compatible with the ones given here.

I wouldn't want to have a SAT solver in Nix...

rfcs/0144-versioned-flake-references.md
```diff
@@ -0,0 +1,120 @@
...
+requirement of the flake, which `nix flake update` can use to update the pin to
+the latest version compatible with the version specification. A new Nix
+command, `nix flake upgrade` will upgrade the version requirement in
+`flake.nix` to the latest possible version without regard to compatibility.
+
+# Motivation
+[motivation]: #motivation
+
+This will allow Nix libraries to be versioned without requiring their users
+to manually update them. Some package managers (e.g. cargo) for more
```
AndersonTorres, line 24: Here you cite cargo, but the specification they employ to do those machine-driven updates is very dumb - by design. Cargo can update patch version, but minor version has restrictions and major is almost impossible.

It looks so full of restrictions that it is easier to edit by hand.

AndersonTorres:
Overall, to me this RFC looks like a solution to a non-existent problem.

rfcs/0144-versioned-flake-references.md
figsoda: Could you elaborate on what "full of restrictions" means? I don't think attack my references is a good explanation.

AndersonTorres: The example tool you gave, Cargo, is purposefully very limited. This design decision conveys the idea that major and minor versions should not be upgraded by a machine, but by a human being.

But your specification is too complex (maybe a consequence of Nix flakes being too complex), and it feels problematic to me.

figsoda: > This design decision conveys the idea that major and minor versions should not be upgraded by a machine, but by a human being.

The whole point of defining a set of rules for the version specification is to make them upgradable by a machine, hence the strict rules on what version bumps can introduce breaking changes. This is so libraries can follow a predictable pattern, instead of their own preference, so updating can be automated.

Your response still doesn't explain what is limited, and I don't see why Cargo being supposedly limited has anything to do with my proposal being problematic. The purpose of this proposal is to give library authors the freedom to version their library, instead of the current situation of users following the main branch and library authors not having a good communicable way to introduce breaking changes.

AndersonTorres: > Your response still doesn't explain what is limited

I said, and I will say again: Cargo (or more precisely `cargo.toml`) is limited. Being limited, it is easy to understand and implement.

Your proposal is too wide. It introduces more syntatical and semantical rules - implying more burden for the compiler writer (of a language with no formal specification yet, by the way).

Further,

> give library authors the *freedom* to version their library

Goes against

> *strict rules* on what version bumps can introduce breaking changes.

The design decision should find the equilibrium between these two points.

figsoda: > Cargo (or more precisely cargo.toml) is limited.

If you are talking about the expressiveness of `Cargo.toml` vs `flake.nix`, I agree. However, I don't think the difference applies to this specific feature.

> It introduces more syntatical and semantical rules.

Cargo defines [much](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) [more](https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility) [rules](https://doc.rust-lang.org/cargo/reference/semver.html) in this aspect and also more strictly enforces it, as it will fail if the version is absent or not a valid semver.

> Implying more burden for the compiler writer

I agree.

> **freedom** goes against **strict rules**

It doesn't. They are talking about completely different things. This feature will always be opt-in. Library authors can still not version their libraries, or define their own versioning scheme - they just wouldn't be benefited by the change. The only way tooling like this can happen is if people following a predictable set of rules when versioning their libraries, which this proposal will allow by defining strict rules.

roberth:
> solution to a non-existent problem.

It comes down to how we define the scope of Flakes. Is it a small extension to a functional build system, or is it a package manager for a highly distributed ecosystem of packages similar to language-specific package managers?
Both are valid perspectives.

If you use Nix for a traditional snapshot-style distribution like Nixpkgs or NixOS, your needs are covered by the functionality already provided.

If you use many flakes, to the point where either the libraries used in your expressions, or the many semi-independent components that constitute your build, all coming from independently versioned flakes, then the problem domain starts to look a lot like that of the language-specific package managers. Bumping to latest doesn't work, and tracking all constraints manually becomes a pain.

rfcs/0144-versioned-flake-references.md
roberth: > I wouldn't want to have a SAT solver in Nix...

We should consider having one outside of Nix, making lockfile generation pluggable.
Perhaps the version constraints should be in a separate attribute that's open for extension?

In fact, we could consider the lock reading and fetching to be part of Nix, and the Flake CLI to be one implementation of the broader Nix package management idea. Dream2nix is also in the business of lock file generation. Perhaps the role of Nix should be limited to the *integration* of lock file generators, and not actually implementing them. We don't even need a dumb lock file generator like we currently have because that one could just as well live in a flake that's already been locked.

Nix should focus on its core competencies that it does very well.
 - building immutable store paths
 - bootstrapping other software

*Anything* that's built on top of that functionality is going to feel as good as native, plus have the benefit of pinning, patching, and decentralized development.

Flakes is oddly centralized for a package manager built on Nix...

rfcs/0144-versioned-flake-references.md
Lord-Valen: If a flag were to temporarily ignore semantic compatibility, wouldn't that introduce some weird behaviour? Wouldn't the `update` behaviour on subsequent uses (without the flag) be to *downgrade* to a compatible version, thus defeating the purpose of initially using the flag? The idea here is to permanently alter the semver to accept the latest version as a convenience (so the user doesn't need to track down the latest tags of *every* input). I can't see any way of doing that (without potentially confusing behaviour) but inplace changes to `flake.nix`.

Alternatively, we could have a command/flag which would report the latest tags of the inputs (similar to how `update` reports lock file changes). This way there's no need to edit `flake.nix` inplace and the user gets the necessary information at once instead of playing tag whack-a-mole. It's less powerful but still works to make version tracking much less annoying which, I think, is the ulterior motive of this RFC.

edolstra:

rfcs/0144-versioned-flake-references.md
matklad: Late to the party, but let me just write this down just in case. One way to avoid the need for SAT-solving is to allow only semver open (`^1.2.3`) version requirements. That way, a greedy algorithm which just always picks the latest semver-compatible version works. 

Moreover, in practice, in Cargo:

* 99% of constraints are in fact just `^1.2.3` (it was _genius_ design decision to make `version="1.2.3"` be a syntactic sugar for `version="^1.2.3"`)
* people often try to use non-semver open constraints, most of the time that's a wrong thing to do, and leads to problem down the line (eg, their build works, but builds of folks depending on them gridlocks due to unsatisfiable constraints)
* at some point we even considered deprecating non-`^` constraints, but didn't do that, because there are occasionally good reasons for them

So, perhaps a safe move wold be to allow only `^1.2.3` constraints, (and using only implicit `1.2.3` syntax). If that turns out to be expressive enough, that would be a massive simplification. If not, fancier constrains can be added later.

domenkozar:
https://github.com/NixOS/rfcs/pull/184

PR #149, opened 2023-05-26
[RFC 0149]: Cache key rotation

vcunat:
[Rendered](https://github.com/vcunat/rfcs/blob/p/cache-key-rotation/rfcs/0149-cache-key-rotation.md)

### DIFF
--- a/rfcs/0149-cache-key-rotation.md
+++ b/rfcs/0149-cache-key-rotation.md
@@ -1,0 +1,73 @@
+---
+feature: (fill me in with a unique ident, my_awesome_feature)
+start-date: (fill me in with today's date, YYYY-MM-DD)
+author: (name of the main author)
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Rotate cache.nixos.org signing key.
+
+If we tolerate that people using stuff older than *X* could get a validation failure by default, fully rotating the key in 2*X* time seems relatively straightforward.
+
+# Motivation
+[motivation]: #motivation
+
+Never rotating a key is bad security practice.
+The current one has been in use at least since 2015.
+
+# Detailed design
+[design]: #detailed-design
+
+Preliminary action plan:
+- generate a new key
+- make it trusted by default (nix+nixpkgs, perhaps with backports to some branches)
+- wait until enough people trust the new key (at least one year, probably)
+- switch to signing with the new key
+- wait - until paths not signed by new key aren't commonly needed anymore
+- make Nix not need signatures for fixed-output derivations
+  (this step could be completed anytime earlier, too)
+  FIXME: maybe this holds already:
+    https://nixos.org/manual/nix/unstable/command-ref/conf-file.html#conf-trusted-public-keys
+- stop trusting the old key (nix+nixpkgs)
+
+
+
+- - -
+
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This section illustrates the detailed design. This section should clarify all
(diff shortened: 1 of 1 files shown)


vcunat:
This is very preliminary for now, but let me open it as a PR already.

evrim:
subscribe

rfcs/0149-cache-key-rotation.md
```diff
@@ -0,0 +1,73 @@
...
+The current one has been in use at least since 2015.
+
+# Detailed design
+[design]: #detailed-design
+
+Preliminary action plan:
+- generate a new key
+- make it trusted by default (nix+nixpkgs, perhaps with backports to some branches)
+- wait until enough people trust the new key (at least one year, probably)
+- switch to signing with the new key
```
lheckemann, line 31: Can we start signing with _both_ keys as soon as it's generated?

vcunat: I had that under preliminary consideration, even with some text.  Well, let me really push that text now 🤷🏽

vcunat: For me the important milestone isn't signing with the new key but the ability to remove trust for the old key (by default).

evrim:
It would be wise to attach keys to releases. This way there will be no need track key start/expiration dates separately, the release number will tell. If a key is compromised, a new release can be published with an incremented minor number. It may be the case where the key is required be published/in place before the release.

vcunat:
I wouldn't do that.  Our infrastructure is concurrently building various stuff: multiple NixOS/NixPkgs release branches and also some that are not tied to such a number at all.  Similarly from the other side &ndash; when Nix looks for some store path in a binary cache, it doesn't care about release numbers.

evrim:
If that's the case, then there should be mechanism to track key start/expiration dates the keys are bare not certs. I would be glad to help you in the matter but no free time Im afraid. Even if I have the time, I can't get past the marketing team, they say they have rules. Good luck, hope this problem is gone forever without much trouble ;)

vcunat:
Once a workflow for regular key rotation is finalized, I suppose it might still be useful to tie its schedule to the schedule of NixOS releases.

puckipedia:
Note that there's a bunch of older CA store paths that are not marked as such on `cache.nixos.org` (and, in the case of older Nix at least, not verified on CA-ness) due to predating the CA attribute in binary caches in Nix. These probably need to be fixed if re-signing.

picnoir:
What's blocking this RFC from moving forward?

Provided somebody put the work into it, would it be possible to generate a new key and start signing NARs generated by the new infra?

vcunat:
It would be trivial to start.  Note that signing in this scheme comes much later.  (And some people said they prefer a double-signing scheme.)

picnoir:
Are there any downside to start double signing the NARs with the new key and gradually introduce the new signature in the new Nix release?

vcunat:
- No implementation of that so far, I believe: https://github.com/NixOS/hydra/issues/1366
- As long as you keep trust in the old key, you don't really get extra security in my eyes anyway.

picnoir:
The Nix PR to support multiple keys landed: https://github.com/NixOS/nix/pull/12976

vcunat:
Do we have any confirmation/link on this point?
> I don't know if **consumers** support multiple signatures.

Because if the current/older Nix versions had issues with multiple signatures where e.g. one isn't trusted (yet), we'd be in for a delay necessarily.

picnoir:
It does work and it is tested in the functional test suite: https://github.com/NixOS/nix/blob/b509502bd51dac581294ed809fe05666b02b758e/tests/functional/signing.sh#L97

Here, `outPath` has been signed by [two keys](https://github.com/NixOS/nix/blob/b509502bd51dac581294ed809fe05666b02b758e/tests/functional/signing.sh#L14) and is successfully verified provided one of the two keys (`pk1`).

After a bit of git spelunking, it seems like the multiple signature support for a store path has been initially introduced in March 2016: https://github.com/NixOS/nix/commit/712b616a8443b940e94ac443499246f4de6ee5cd

PR #152, opened 2023-06-14
`local-overlay` store

Ericson2314:
Add a new `local-overlay` store implementation to Nix. This will be a local store that is layered upon another local filesystem store (local store or daemon). This allows locally extending a shared store that is periodically updated with additional store objects.

- [Rendered](https://github.com/NixLayeredStore/rfcs/blob/local-overlay-store/rfcs/0000-local-overlay-store.md)
- [Blog post](https://blog.replit.com/super-colliding-nix-stores)
- [WIP Implementation](https://github.com/NixOS/nix/pull/8397)

_This work is sponsored by [Replit](https://replit.com/) :sparkles:_

### DIFF
--- a/rfcs/0000-local-overlay-store.md
+++ b/rfcs/0000-local-overlay-store.md
@@ -1,0 +1,58 @@
+---
+feature: (fill me in with a unique ident, my_awesome_feature)
+start-date: (fill me in with today's date, YYYY-MM-DD)
+author: (name of the main author)
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+One paragraph explanation of the feature.
+
+# Motivation
+[motivation]: #motivation
+
+Why are we doing this? What use cases does it support? What is the expected
+outcome?
+
+# Detailed design
+[design]: #detailed-design
+
+This is the core, normative part of the RFC. Explain the design in enough
+detail for somebody familiar with the ecosystem to understand, and implement.
+This should get into specifics and corner-cases. Yet, this section should also
+be terse, avoiding redundancy even at the cost of clarity.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This section illustrates the detailed design. This section should clarify all
+confusion the reader has from the previous sections. It is especially important
+to counterbalance the desired terseness of the detailed design; if you feel
+your detailed design is rudely short, consider making this section longer
+instead.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we *not* do this?
+
+# Alternatives
+[alternatives]: #alternatives
+
+What other designs have been considered? What is the impact of not doing this?
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+What parts of the design are still TBD or unknowns?
+
+# Future work
+[future]: #future-work
+
+What future work, if any, would be implied or impacted by this feature
+without being directly part of the work?
--- a/rfcs/0000-local-overlay-store.md
+++ b/rfcs/0000-local-overlay-store.md
@@ -276,6 +276,9 @@ We can have a "fsck" operation however that manually scans for missing / duplica
 
 ## Read-only `local` Store
 
+*This was [already approved](https://github.com/NixOS/nix/pull/8356#event-9483342493) by the Nix team on an experimental basis, as an experimental feature that is trivial enough to be approved with out requiring an RFC.
+It is still included here just to provide context.*
+
 In order to facilitate using `local-overlay` where the lower store is entirely read only (read only SQLite files too, not just store directory), it is useful to also implement a new "read-only" setting on the `local` store.
 The main thing this does is use SQLite's [immutable mode](https://www.sqlite.org/c3ref/open.html).
 
--- a/rfcs/0000-local-overlay-store.md
+++ b/rfcs/0000-local-overlay-store.md
@@ -1,5 +1,5 @@
 ---
-feature: (fill me in with a unique ident, my_awesome_feature)
+feature: local-overlay-store
 start-date: (fill me in with today's date, YYYY-MM-DD)
 author: (name of the main author)
 co-authors: (find a buddy later to help out with the RFC)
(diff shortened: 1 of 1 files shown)


rfcs/0000-local-overlay-store.md
```diff
@@ -0,0 +1,408 @@
...
+It is quite common for organizations using Nix to first adopt it behind the scenes.
+That is to say, Nix is used to prepare some artifacts which are then presented to a consumer that need not be aware they were made with Nix.
+Later though, because of Nix's gaining popularity, there may be a desire to reveal its usage so consumers can use Nix themselves.
+Rather than Nix being a controversial tool worth hiding, it can be a popular tool worth exposing.
+Nix-unware usage can still work, but Nix-aware usage can do additional things.
+
+The `local-overlay` store can serve as a crucial tool to bridge these two modes of using Nix.
+The lower store can be as before
+--- however the artifacts were disseminated in the "hidden Nix" first phase of adoption
+--- perhaps with only a small tweak to expose the DB / daemon socket if it wasn't before.
```
7c6f434c, line 69: Is exposing the socket used anywhere in the proposal, or is it just mentioned as a separate possibility (with relevant metadata sharing done via reading the underlying SQLite DB)?

rhendric:
What if, instead of requiring that the lower store grow monotonically, the overlay store maintained GC roots in the lower for any lower paths it references? I haven't thought this through to the level of detail in the spec, but it strikes me as an alternative worth considering; not being able to run garbage collection on the lower store would certainly be a barrier for personal use (perhaps not a central goal for this RFC) and possibly for org use as well—storage is cheap but it doesn't round down to free in every context.

ryantm:
I work for Replit (who sponsored work on this RFC) and helped out with the creation of this RFC. I am happy to serve as a Shepherd on this RFC if needed, but also happy to cheer from the sidelines if people (or the Steering Committee) see this as a conflict of interest.

baloo:
I'm generally excited about this. I have a slightly different target than Replit has.
I'm shipping firmware image (Multiple GB -large firmware images) with a nix/store that is immutable and signed (via dm-verity and secureboot). The lower overlay ships with a document that provides a way to rebuild the db with everything that was put in the store, at runtime the db is rebuilt. Configuration is then evaluated and a new system is built in the upper store.

This would solve the initial import of the DB which is still not that fast (although probably way faster than importing 16TB worth of derivations!).
For our use-case, I still believe we'll trash the upper layer at each boot because it's easier to reason about.

rfcs/0000-local-overlay-store.md
Ericson2314: The local overlay store is potential a *consumer* of the socket provided by another Nix daemon. A Nix daemon can also be spun up using the local overlay store instead of the local store.

Basically, no new socket code is needed for this. As far as I can tell, everything one would want with sockets already works without limitation.

Ericson2314:
@rhendric That is useful for some things, but probably not the use-case of large numbers of consumers all sharing the same underlying store --- it is pretty important the underlying store be truely read only in that case, including any GC roots.

Ericson2314:
@baloo We have separately thought about those sorts of issues, including a persistent upper store that then "pivots" onto a new lower store when one does can upgrade of NixOS (and I suppose GC of the old generations). I think the pivoting feature is a nice future work item.

rfcs/0000-local-overlay-store.md
7c6f434c: A consumer — doing what with the socket?

kevincox:

Ericson2314:
I nominate @roberth

7c6f434c:
> That is useful for some things, but probably not the use-case of large numbers of consumers all sharing the same underlying store

Now this makes me wonder if `local-overlay` sounds a bit more local than this use-case (from a naming point of view)…

Ericson2314:
I suppose would deprecate all "local" and "remote" as not being misleading names. I just picked `local-overlay` on moments notice to match the existing pattern.

lheckemann:

ryantm:
@baloo, @rickynils, @zhaofengli, @arianvp, @edolstra  would any of you be open to shepherding this along? 

I don't see much controversy or drama here and, in my opinion, the document is in good shape too, so overall I expect the shepherd work to be low-commitment.

If you've never done it before, you can look at https://github.com/NixOS/rfcs/blob/master/rfcs/0036-rfc-process-team-amendment.md#shepherd-team for more information about being a Shepherd.

arianvp:
I can Shepard

rfcs/0000-local-overlay-store.md
Ericson2314: Doing regular client, like anything else using the daemon. It will in fact only use it to read metadata; the lower store can be read-only.

arianvp:
I also poked the #nixos-systemd channel. As some people are looking at Appliance images / immutable `nix-store` partitions and it seems to have a lot overlap with this RFC

rfcs/0000-local-overlay-store.md
```diff
@@ -0,0 +1,408 @@
...
+
+This gives us a "best of all three worlds" solution.
+
+## Marketing motivation
+
+It is quite common for organizations using Nix to first adopt it behind the scenes.
+That is to say, Nix is used to prepare some artifacts which are then presented to a consumer that need not be aware they were made with Nix.
+Later though, because of Nix's gaining popularity, there may be a desire to reveal its usage so consumers can use Nix themselves.
+Rather than Nix being a controversial tool worth hiding, it can be a popular tool worth exposing.
+Nix-unware usage can still work, but Nix-aware usage can do additional things.
```
7c6f434c, line 64:
```suggestion
Nix-unaware usage can still work, but Nix-aware usage can do additional things.
```

rfcs/0000-local-overlay-store.md
7c6f434c: I have reread the paragraph, and even with your explanations I am not sure what process the paragraph as written describes.

rfcs/0000-local-overlay-store.md
```diff
@@ -0,0 +1,408 @@
...
+There is no "have store object, don't yet have DB entry" middle state to worry about.
+
+The downside of this is that Nix needs elevate permissions in order to create those bind mounts, and the impact of having arbitrarily many bind mounts is unknown.
+Even if this design works fine once set up, the imposition of an O(n) initialization setting up each bind mount is prohibitive for many use-cases.
+
+## Store implementations using FUSE
+
+We could have a single FUSE mount that could manually implement the "bind on demand" semantics described above without cluttering the mount namespace with an entry per each shared store object.
+FUSE however is quite primitive, in that every read need to be shuffled via the FUSE server.
+There is nothing like a "control plane vs data plane" separation where Nix could tell the OS "this directory is that other directory", and the OS can do the rest without involving the FUSE server.
```
RaitoBezarius, line 372: This is false, FUSE passthrough can lead to bypass the FUSE server.
There are multiple folks in the Linux filesystem ecosystem that are working on bringing FUSE ecosystems with a relatively on-par performance with in-kernel filesystem or reuse the existing filesystems.

https://lwn.net/Articles/932060/
https://github.com/extfuse/extfuse
https://lpc.events/event/16/contributions/1339/

I am not really convinced of not pursuing this alternative as I feel like this bring the maximum flexibility and compatibility for all the usecases instead of making it a very limited thing based on OverlayFS.

If you are interested into chatting more on how to make this alternative possible, feel free to ping me, I know quite about FUSE filesystems and I am planning to write a nixstorefs at some point, which will rely on FUSE semantics first.

I also challenge the "worse performance" than in-kernel mounting solutions, it would be good to bring data on that, if you only pay the `open` cost, this is quite cheap.

RaitoBezarius: Anyway, I have mixed feelings about the RFC because I feel like the FUSE approach is a much better route than this one, I can say that I am not satisfied and feel like it should be more ambitious if it is going to take a RFC route.

rfcs/0000-local-overlay-store.md
```diff
@@ -0,0 +1,408 @@
...
+## Marketing motivation
+
+It is quite common for organizations using Nix to first adopt it behind the scenes.
+That is to say, Nix is used to prepare some artifacts which are then presented to a consumer that need not be aware they were made with Nix.
+Later though, because of Nix's gaining popularity, there may be a desire to reveal its usage so consumers can use Nix themselves.
+Rather than Nix being a controversial tool worth hiding, it can be a popular tool worth exposing.
+Nix-unware usage can still work, but Nix-aware usage can do additional things.
+
+The `local-overlay` store can serve as a crucial tool to bridge these two modes of using Nix.
+The lower store can be as before
```
Ericson2314, line 67:
```suggestion
The lower store can be served as before
```

rfcs/0000-local-overlay-store.md
Ericson2314: 1. Server Nix store dir, clients don't use Nix but do use Nix-built things
2. Serve either socket or SQLite database, clients can use Nix with that store but with restrictions (e.g. perhaps only read-only)
3. Use local overlay store (the writable upper layer makes the read-only lower layer less of an issue)

Doe that make sense?

rfcs/0000-local-overlay-store.md
Ericson2314: Oh that is fantastic! That is getting at the heart of the problems with FUSE. I wish I had known about this earlier.

However, even if I did, I would have advocated starting with the OverlayFS approach. That is because all these things are not yet mainline, and kernel development / trying out bleeding edge features is vastly more expensive.

IMO the right thing to do is

1. Accept what we have on an experimental basis, using it to drum up interests and move us towards being able to pool resources on this.
2. Get in communication with these Kernel devs; indeed I was already emailing back and forth with Amir Goldstein about some restrictions in OverlayFS.
3. Try to be an early adopter of this stuff as it matures; nudge its development so it better meets our needs.

Also CC @flokli, because Tvix may be better positioned to be at the vanguard of trying this stuff out, as they are already exploring FUSE.

rfcs/0000-local-overlay-store.md
7c6f434c: So in the step 1 each snapshot gets its own store path, and the rest of the store is visible? (I guess one could also share a fixed-path directory with links to all the currently-relevant snapshots?)

rfcs/0000-local-overlay-store.md
RaitoBezarius: Disclosure: I am also a tvix-store developer ;)

RaitoBezarius: Alright, your argument is sound and convinced me :).

Ericson2314: Oh my bad I didn't realize that. @RaitoBezarius you should shepherd this :).

RaitoBezarius: Aaaaaaaaaaa, OK for the nerdsnipe.

RaitoBezarius:
I am auto-nominating myself as a shepherd, to be completely clear, my availability for August is limited as I will be on vacations, but I will have time back on September, though I may have other responsibilities like Release Manager that have a higher priority, if you will have me, that's fine. :)

tomberek:
Self-nominate

rfcs/0000-local-overlay-store.md
```diff
@@ -0,0 +1,408 @@
+---
+feature: local-overlay-store
+start-date: 2023-06-14
+author: John Ericson (@Ericson2314)
+co-authors: Ben Radford (@benradf)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
```
tomberek, line 7: Thanks to the nominees for accepting! Here's the team the Steering Committee has decided on :)

```suggestion
shepherd-team: @RaitoBezarius @tomberek @arianvp 
shepherd-leader: @tomberek
```

tomberek (changes requested):

rfcs/0000-local-overlay-store.md
```diff
@@ -0,0 +1,408 @@
...
+[alternatives]: #alternatives
+
+## Stock Nix with OverlayFS ("overlay everything" from motivation)
+
+It is possible to use OverlayFS on the entire Nix store with stock Nix.
+Just as added store objects would appear in the upper layer, so would the SQLite DB after any modification.
+
+There are a few problems with this:
+
+1. Once the lower store is "forked" in this way, there is no "merge".
```
arianvp, line 335: This is not how overlay `/nix/store` works today in e.g. the ISO.    There's a `/nix/store/.nix-path-registration` file that registers paths on startup with a `nix-store --load-db`.   That file can grow and will add new paths on restart.  The combined store _will_ learn about new paths

rfcs/0000-local-overlay-store.md
```diff
@@ -0,0 +1,408 @@
...
+
+Because the `local-overlay` store is a completely separate store implementation, the interactions with the rest of Nix are fairly minimal and well-defined.
+In particular, users of other stores and not the `local-overlay` store will not be impacted at all.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Read-only local store is delicate
+
+SQLite's immutuable mode doesn't mean "we promise not to change the database".
```
arianvp, line 300: I found this paragraph very surprising. The rest of the RFC is written as if  the lower store can monotonically grow without being a problem for the overlays and that sounded really cool! And then suddenly this paragraph says that none of that is possible due to technical issues.

With this technical issue in place I am not so sure if a `local-overlay` store is that attractive or useful. It sounds like a deal-breaker that makes the abstraction kind of useless. If the sqlite database in the lower store can't be updated online, then what is the point?

How is it different from a lower store  containing `nix-registration-paths` text file with all valid paths that we `nix-store --load-db` on startup into the overlay sqlite database?   This is something we have already today and is well-tested (the ISO depend on it). And it seems to give the same benefits to me.     The lower store can be updated offline, and if you remount it then new paths can be added to the overlay sqlite database.

arianvp: Ah we address this later in https://github.com/NixOS/rfcs/pull/152/files#diff-36321d79e9a5897b43eaf0f9c622b4a14978b61641f934bc4d33cdc9a49f6669R345

I think we could put way more focus on  "This is an optimisation for `--load-db`".  As that is the main selling point here.

tomberek:
Initial meeting took place on 2023-09-24

## Notes

[ericson]: RyanTM discussed review+merge of PR in current form as experimental feature

[RaitoBezarius]: go with FUSE proposal
    - [ericson]: interaction with ACL proposal?
    - https://cl.tvl.fyi/c/depot/+/9449/4/tvix/tests/README.md
    
[ericson]: delete has similar problem as whiteout/tombstones, I do agree FUSE is probably necessary to "truly" solve that.

[RaitoBezarius]: sqlite is not longer applicable
[arian]: also store nar+narinfo?
    - [ericson]: can layer with sqlite, would need initial load to build indexes
[RaitoBezarius]: would like DB to be impl detail
    - NixStoreFS
    - [ericson]: allow building with `LocalFSStore` (abstract superclass), then rename to `LocalSqliteStore`
    - [raito]: blobStore, directoryStore, metadata, etc?
        - too much separation of processes?
        - current store has issues when used as a build store, leaking FDs
           - `dmesg` for [the win?](https://github.com/NixOS/nix/issues/8802)!
           - WIP to make everything in builds a bind mount, not hardlinks

[arian]: concern about what is used as the lowest layer in a standard setup. possibility to use the narinfo layer
[RaitoBezarius]: deduplication in store layer?
[ericson]: this is a very stateful feature, requires migration?
[RaitoBezarius]: proper migration provides much more peace of mind, expectation to have store->store migration.

[ericson]: would be good to have Postgres store
    - s3fs, goofys?
    - bcachefs - potential?
    - TODO: get contacts from RaitoBezarius about kernel developments

[tomberek]: most of this is to help decide if current proposal is worth working on. 
    - [arian]: local-overlay-store is somewhat of a stopgap, should not be blocked by potential of future work 
    - code and implementation exists

[RaitoBezarius]: propose merge as experimental, get feedback
    - more ambitious: migration stories, encourage store layer experimentation, another RFC
    - proliferation of store layers that are not "friendly" wrt migrations, version, etc.
    - can accept this, but need to be careful
    - this will be used by larger organizations, thus cementing the xp features
    - 
[tomberek]: interaction with scheduling of remote builds? removes deadlock?

## Experimental?

Team is concerned that such an experimental feature will be used and quickly become a non-experiment and relied upon, making breaking changes more difficult, requiring migration path. Afterward people will likely complain. We must set the expectation that this is not a stable feature.

- TODO: discuss with ReplIt, ask community for who is interested in this as a production feature. Inivited ryantm to join and answer a few questions.
- [ryantm]: we plan to use the PR as-is
- [ryantm]: have been on 2.3, planning to jump forward to use this PR and pinning if necessary
- [ryantm]: will need to support better infra in the future and support by Nix team
- [ryantm]: local overlay store is a nice way to integrate with other underlying store implementation (tvix-store for example) 

# Summary
- Merging of code is agreed upon
- Decision: merge PR as xp, get feedback, RFC discussion continues, stabalize + accept
- TODO: get "hands-dirty" [arian,tomberek]
- TODO: need feedback from experimental usage
- expectation: possible to merge within 2 months
- reconvene after experience with merged xp feature
           
Misc Links:

 - https://aws.amazon.com/about-aws/whats-new/2023/03/mountpoint-amazon-s3/
 - https://github.com/project-machine/puzzlefs
 - [FUSE passthrough]()

infinisil:
> Team is concerned that such an experimental feature will be used and quickly become a non-experiment and relied upon, making breaking changes more difficult, requiring migration path.

Here's a general idea: Make experimental feature flags be versioned to the current Nix version. E.g. in order to use a local overlay store with Nix 2.15.x you need this in `nix.conf`:

```ini
experimental-features = local-overlay-2.15
```

Only features that match the current Nix version get enabled, do not provide a way to enable it for future Nix versions.

RaitoBezarius:
> > Team is concerned that such an experimental feature will be used and quickly become a non-experiment and relied upon, making breaking changes more difficult, requiring migration path.
> 
> Here's a general idea: Make experimental feature flags be versioned to the current Nix version. E.g. in order to use a local overlay store with Nix 2.15.x you need this in `nix.conf`:
> 
> ```ini
> experimental-features = local-overlay-2.15
> ```
> 
> Only features that match the current Nix version get enabled, do not provide a way to enable it for future Nix versions.

I think it still doesn't address migration paths for multiple store backends, alas.

flokli:
> I think it still doesn't address migration paths for multiple store backends, alas.

Not quite sure I understand the probIem statement, but 'm not sure migration of data from one layer to another *needs* to be part of this initial RFC. Nix is not doing the overlayfs mounts on its own either, but relies on the user of the feature to take care of this, and know the surrounding context - so it seems prudent to rely on that same user with the context to also migrate/move data around as it makes sense for their context.

flokli:
> > Team is concerned that such an experimental feature will be used and quickly become a non-experiment and relied upon, making breaking changes more difficult, requiring migration path.
> 
> Here's a general idea: Make experimental feature flags be versioned to the current Nix version. E.g. in order to use a local overlay store with Nix 2.15.x you need this in `nix.conf`:
> 
> ```ini
> experimental-features = local-overlay-2.15
> ```
> 
> Only features that match the current Nix version get enabled, do not provide a way to enable it for future Nix versions.

This seems to be orthogonal to this RFC, I'm not sure it should be discussed here.

infinisil:
> I think it still doesn't address migration paths for multiple store backends, alas.

I'm also not entirely sure if I get it, but it could be required that experimental features cannot share their Nix store with stable Nix versions, so e.g. it would have to be `/nix/experimental/store` or so. And only provide non-destructive ways to migrate stores while it's still experimental, so e.g. "copy + migrate the store" but no "migrate the store in place".

> This seems to be orthogonal to this RFC, I'm not sure it should be discussed here.

Mostly yeah. I think it's appropriate to bring up since it was mentioned as a concern in the meeting notes, but it might not have to be a concern with this idea.

alyssais:
> Here's a general idea: Make experimental feature flags be versioned to the current Nix version. E.g. in order to use a local overlay store with Nix 2.15.x you need this in `nix.conf`:
>
> ```ini
> experimental-features = local-overlay-2.15
> ```
>
> Only features that match the current Nix version get enabled, do not provide a way to enable it for future Nix versions.

This is how you end up with people advising others to do

nix.settings.experimental-features = [ "local-overlay-${lib.versions.majorMinor config.nix.package.version}" ];

It's not that people don't know experimental features might change, it's that they don't care until it happens, or don't believe it really will.

RaitoBezarius:
> > I think it still doesn't address migration paths for multiple store backends, alas.
> 
> I'm also not entirely sure if I get it, but it could be required that experimental features cannot share their Nix store with stable Nix versions, so e.g. it would have to be `/nix/experimental/store` or so. And only provide non-destructive ways to migrate stores while it's still experimental, so e.g. "copy + migrate the store" but no "migrate the store in place".
> 
> > This seems to be orthogonal to this RFC, I'm not sure it should be discussed here.
> 
> Mostly yeah. I think it's appropriate to bring up since it was mentioned as a concern in the meeting notes, but it might not have to be a concern with this idea.

> > I think it still doesn't address migration paths for multiple store backends, alas.
> 
> Not quite sure I understand the probIem statement, but 'm not sure migration of data from one layer to another _needs_ to be part of this initial RFC. Nix is not doing the overlayfs mounts on its own either, but relies on the user of the feature to take care of this, and know the surrounding context - so it seems prudent to rely on that same user with the context to also migrate/move data around as it makes sense for their context.

My point was about that we don't have the right primitives to ingest store paths from a backend to move away from this backend, except serializing the store and deserializing it, which cannot always work in certain scenarios of large stores.

I am okay with experimental features not having migration knobs, I am meh (or even against) about multiplying the amount of store backends and having no migration stories for them, which will hamper our ability to make them evolve.

There are a lot of things in Nix that just perform automatic migration, I think it's important we take the time to formalize this, I am not saying we need to have a good answer as part of this RFC, but someone has to care about this.

rfcs/0000-local-overlay-store.md
```diff
@@ -0,0 +1,408 @@
...
+#### Edges
+
+- Solid edge: "adding" or "deduplicating" direction
+- Dotted edge: "deleting" direction
+
+The graph when reduced to just one type of edge is always acyclic, and thus represents a partial order.
+
+### Lower Store
+
+While in use as a lower store by one or more overlay stores, a store most only grow "monotonically".
```
ballit6782, line 171: In Linux, [changes to the underlying storage of overlay devices has undefined behaviour](https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html#changes-to-underlying-filesystems), so a store used as a lower store cannot grow at all.

ballit6782 (changes requested):
I have a feeling this RFC does not take into account the fact that changes to the underlying storage of overlay filesystems are not allowed in Linux :

> Changes to the underlying filesystems while part of a mounted overlay filesystem are not allowed. If the underlying filesystem is changed, the behavior of the overlay is undefined, though it will not result in a crash or deadlock.

See [filesystems/overlayfs.html#changes-to-underlying-filesystems](https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html#changes-to-underlying-filesystems)

I think that to avoid confusing users of this RFC, this should be acknowledged, as it is a serious limitation of the feature (it means that any host sharing its store with guests would be immutable as long as a single guest would be running). Also, if this was not taken into account when comparing alternatives, it's possible that they need to be revisited. Namely, I think it makes the fuse-binding-on-demand alternative much more interesting.

It seems to me that overlayed stores have a very limited usefulness given this limitation, except for very specific cases like the NixOS live CD, where the base store is guaranteed not to change, and this use case seems properly handled with the existing tooling.

arianvp:
There's a new method that landed in the kernel that allows you to "tuck"
the lower store.
 Does that address these things? https://lwn.net/Articles/927491/

rfcs/0000-local-overlay-store.md
ballit6782: There may be another approach with FUSE : using [MergerFS](https://github.com/trapexit/mergerfs)  with `symlinkify=true` and `cache.symlinks=true`. This way, FUSE is only used to resolve the symlink:
 - for a lot of applications, the opened directory fd will be reused, so most operations will run directly on the underlying store
 - later uses of the symlink would be cached by the kernel, so even without fuse-ebpf, we'd need the jump to userspace only once per symlink

lheckemann:
@tomberek any chance the shepherd team could have a meeting sometime soon? Some of @ballit6782's concerns don't seem to have been responded to and it would be good to have these addressed in some way so the RFC can move forwards :)

infinisil:
Another ping, please give an update on the state of this RFC, if any. If no progress is happening, it would be best if RFC is marked as a draft.

ballit6782:
I think that for this RFC to progress, one of these things need to happen :

- 1 some clarification from fsdevel as to what kind of changes, if any, are we allowed to do on the overlay's upper filesystem
- 2 add the restriction that the upper store should always be frozen if the lower store is used
- 3 choosing to experimentally rely on things that "seem to work" but that are not part of the kernel's guarantees (as I understand them)
- 4 use a FUSE based filesystem to make this feature work, as proposed in the alternatives

I think that options 2 and 3 are not viable (frozen upper stores would be largely useless).

On the subject of FUSE alternatives, I think that this paragraph in the RFC does not take into account the caching mechanisms of Linux's VFS :
> FUSE however is quite primitive, in that every read need to be shuffled via the FUSE server. There is nothing like a "control plane vs data plane" separation where Nix could tell the OS "this directory is that other directory", and the OS can do the rest without involving the FUSE server. That means the performance of FUSE are potentially worse than these in-kernel mounting solutions.

For example, if using MergerFS with the options `symlinkify=true` and `cache.symlinks=true`, MergerFS would in effect be telling the OS "this directory is that other directory". Furthermore, using `cache.symlinks`, only the first read of a root-level symlink in the store would involve userland code. While this might involve a bit more memory usage as there would be more dentries to cache, I think that they are largely offset by the memory usage that overlayed stores can bring.

MergerFS also opens up other possibilities that are very interesting for my use-case of the local-overlay store, namely providing "late persistence", by adding a tmpfs branch that can be dynamically shrunk to 0. And being able to set `fsname=nix` would also be quite neat.

I have planned since some time to make some complete tests of this in practice, but had not found the time nor motivation recently. However, if the shepherds think that this is a viable path for this RFC, I can document some experiments, run some benchmarks (especially measuring the additional context switches and how they scale) and try to provide an updated version of the RFC that would rely on MergerFS.

I'd also like to add : I think it would be great if the RFC could express more clearly if this feature could be used for multiple layered stores. I know this would be a very niche feature, but there are cases where it would be very useful (for integration with the Shufflecake layered plausible-deniability storage system). It would be helpful to know if it would be a supported or not use case.

MMesch:
Since there was no activity for a while we decided to mark this as a draft. Feel free to undraft it any time once activity picks up again.

jonringer:
I think this is super cool usage of Nix, and would hate for it to fall by the wayside. I'll happily be a shepherd.

tomberek:
Looks like the notes never made it here (my mistake): from Feb 01

# Quick sync : local-overlay store

Present: raitobezarius, tomberek, John Ericson, ryantm.

**Question** : Update on where we are?

From an RFC perspective, it seems like there's stalling, but there have been progress and we should write that down.

## `local-overlay` store usage at replit

Started the use few months ago. Objective: reduce the copying of the lower store database (gigabytes in size). In the past, every time a user has to do an action, they had to copy it before doing any Nix action. With the local overlay store, this is eliminated now.

It seems to be working fine, we have no bug report regarding Nix so far, though we may not have the most advanced usecase of Nix workloads.

We just finished the work for persisting the upper store and we will start testing that internally.

## Where do we want to drive this?

John is trying to merge the PR, he worked with Ben from Tweag on this.

## Tvix

flokli has been working on store compositional, Tvix has no database like Nix's SQLite, we are looking at building more compatibility and getting a theory of composition of stores. We will need at least six more months to have interesting results.

replit has contracted flokli to work on plumbing the blob store / CA store for their main package storage.

**Conclusion** : This group has been working more as a special interest group / working group than a purely RFC group, we are happy of the results it brought, ryantm mentions that he's happy of going through the RFC because it crystallized a lot of important information and the RFCSC has been helpful to keep us able to sync everyone else on what work is going on.

tomberek:
## Update

https://github.com/NixOS/nix/pull/8397 is now merged into Nix master branch.

kevinh-canva:
There's mentions both [in this RFC](https://github.com/NixOS/rfcs/pull/152#issuecomment-1595797428) and in [discourse](https://discourse.nixos.org/t/super-colliding-nix-stores/28462/15) that the `local-overlay` store would break if the lower store is garbage-collected. 

Can I get a clarification on this: does this mean if user runs `nix-store --gc` that targets an already existing local-overlay store, it will be a `faux delete` (but the store object will still exist in the lower store), and there will be no corruption?

And the garbage collection here only applies to cases when the store objects is somehow deleted from the lower store prior to mounting? And in that case, would running a `nix-store --verify --repair` after mounting fixes it?

PR #136, opened 2022-10-03
[RFC 0136] A plan to stabilize the new CLI and Flakes incrementally

Ericson2314:
Ever since the closing of [RFC 49](https://github.com/NixOS/rfcs/pull/49), we've had the new CLI and Flakes marked as experimental, with no clear plan forward.

With the goal of ending this current limbo and soothe longstanding tensions in the Nix community, this RFC does two things:

1. Establish general principles about Nix's architecture and evolution in order to ensure we do not get in this situation again.

   Notably we are allowed to make breaking changes to experimental features, which includes both the new CLI and Flakes, until they are stable.

2. Establish an incremental plan adhering to the principles deciding on the order and priority in which to stabilize these features:

   - First, the non-Flakes CLI wll be stabilized, in phases.

   - Afterwards, Flakes itself and its CLI components can be stabilized. The final design of Flakes will also require another RFC.

[Rendered](https://github.com/Ericson2314/nix-rfcs/blob/stabilize-incrementally/rfcs/0136-stabilize-incrementally.md)

---------

_A note to anyone leaving comments, remember the point of this RFC is *not* to debate the **content** of what is being stabilized, but debate the **process** of stabilization. To my fellow Flakes skeptics in particular, please hold back on the minutiae what you think is wrong with the design of Flakes, the new CLI or other features, or we will end up with another monster RFC thread that goes nowhere._

_The current trajectory as far as I can tell, prior to writing this. is that we all of the sudden get a Nix 3.0 where a huge swath of stuff without review becomes stable at once. This is because many people in charge feel exhausted, and fear properly audit and debate the new features will devolve into yet another messy useless flame war._

_Only by keeping the conversation narrowly on topic can demonstrate that a cordial compromise is possible_

PR #157, opened 2023-08-02
[RFC 0157] More granular control over the system

Animeshz:
## Summary

RFC proposes to grant more granular low-level control over the system (reduction of unnecessary abstractions). This may results in finer, transparent and more accessible setups.

[Rendered](https://github.com/Animeshz/nix-rfcs/blob/more-granular-control/rfcs/0157-more-granular-control.md)

### DIFF
--- a/rfcs/0157-more-granular-control.md
+++ b/rfcs/0157-more-granular-control.md
@@ -11,19 +11,9 @@ related-issues: (will contain links to implementation PRs)
 # Summary
 [summary]: #summary
 
-To grant more granular low-level control over the system. Majorly:
-
-* `nix-cli`: Introduce direct access to the store and binary cache.
-  - This improvement will facilitate easier binary cache setup, upstream caching, and shallow (non-closure) copies.
-  - Users will be able to pull specific paths directly from the cache without relying on derivation realisation as an abstraction.
-
-* `nixos`: Decouple various low-level components of the operating system and make everything "opt-in."
+To grant more granular low-level control over the system. Majorly, to decouple various low-level components of the operating system and make everything "opt-in."
   - This change aims to reduce the disparity observed when using nix inside NixOS and outside NixOS.
   - By making the infrastructure flexible and allowing components to be explicitly chosen, users can adapt to future breaking changes, such as the introduction of pipewire and wayland, even for init systems and bootloaders.
-
-* `flakes`: ([on hold](#flake-transit)) Enforce the use of flakes for scripting operations.
-  - Currently, script infrastructure often involves a `script -> nix language -> script` abstraction, which can make it challenging to modify, reuse, or apply fixes to the underlying script. Examples of this include nixos-rebuild, nixos-generate, and home-manager.
-  - With the enforcement of flakes, the former `script ->` abstraction will be reduced, and users can directly use a `nix` command to build the script. The activation script `./result/bin/activation-script` can then be executed to complete the task.
 
 
 # Motivation
@@ -38,19 +28,6 @@ The problem becomes significant when working with (or swapping) the lower-level 
 
 And the list goes on...
 
-Another observation is that there are scripts (`nixos-generate`) that act as an indirection to nix derivation (`nixosConfiguration.config.system.build.isoImage`), which is, in turn, an indirection to yet another shell script. This abstraction can be reduced by utilizing flake outputs.
-
-This also unintentionally hinders the inspectibility and scriptability of the system. For instance, [Void Linux has highly scriptable frontends](https://animeshz.github.io/site/blogs/void-linux.html#why-is-void-so-cool) to illustrate this.
-
-Although most things are straightforward in nix, such as changing a desktop environment with a one-liner, setting up remote builds and upstream cache can be quite challenging. There are several design problems, such as:
-
-* Dependencies are first downloaded on the host, copied over to the remote builder via ssh, and everything is built there. Then the primary object is copied back to the host.
-* There's a `builders-use-substitutes` option on both nix commands (which generally doesn't work without being a `trusted-user`) and the config option at global config `/etc/nix/nix.conf`, which somewhat fixes this problem. However, after the build, the builder still needs to serve the whole closure of the build with the host.
-* The output from `nix-store --export /nix/store/{drv-or-output-path} > output` is not importable using `nix-store --import < output` unless the dependencies were already present on the host. You can't ask --import to look for dependencies in the binary cache. One could get a list of dependency paths using `nix-store --query --requisites /nix/store/{drv-or-output-path}`, but there's no exposed way to effectively "pull" those paths from the binary cache before running --import. The only way is to copy the whole closure for every derivation or its output (from the remote builder to the host).
-* Upstream caching (not caching the packages available in upstream cache such as `https://cache.nixos.org`) requires extra specific 3rd party infrastructure to set up, such as [cachix](https://www.cachix.org) (service) or [attic](https://docs.attic.rs/tutorial.html#pushing) (self-host).
-
-In void (just as a reference), a package that ignores already upstream-dependency is built (in a chroot by default). This output folder can then act as a simple ad-hoc repository mirror with any http server (including `python3 -m http.server -d hostdir/binpkgs` on the host-machine or even github pages after a build from GitHub actions).
-
 While the current configuration does work for most people, and many are content with the existing solutions, a more granular control system would offer tremendous possibilities for customization and make the infrastructure flexible for future shifts.
 
 Instead of deprecating pulseaudio's `sound.*` in favor of `services.pipewire.*`, we should aim to build a flexible API that persists even when new standards emerge, reducing confusion caused by generic names like "sound" getting deprecated.
@@ -59,7 +36,7 @@ Instead of deprecating pulseaudio's `sound.*` in favor of `services.pipewire.*`,
 # Detailed design
 [design]: #detailed-design
 
-### Everything is opt-in!
+**Everything is opt-in!**
 
 First and foremost, during image building (whether it's nixos-rebuild or vm/iso generation), the filesystem should be empty, with no packages or files.
 
@@ -69,29 +46,16 @@ This approach also facilitates the use of NixOS-modules on non-NixOS nix install
 
 <sub>[1]: Recently [dracut merged](https://github.com/NixOS/nixpkgs/issues/77858), but there are no clear instructions on how to use that.</sub>
 
-### (on hold) Emphasis on flake over shell script wrappers
-
-With the introduction of `nix run` and `nix build`, all current implementations should transition to using these first-party nix commands instead of wrappers like nixos-rebuild, nixos-generate, and home-manager.
-
-Operating everything under flake also opens up various opportunities, such as building the latest nixos iso image with or without plasma/gnome directly from the nixpkgs repository if its derivation is exposed in flake.nix (for example, `nix build nixpkgs#isoInstallerImageWithGnome`).
-
-### CLI improvements
-
-As explained earlier in the motivation section, the binary cache has no direct access, and using `nix-store --import {exported}` does not automatically pull missing dependencies from the binary cache in case the export does not include the whole closure.
-
-Allowing direct binary-cache access, such as querying information and pulling specific paths into the local `/nix/store`, would be a significant enhancement.
-
 
 # Examples and Interactions
 [examples-and-interactions]: #examples-and-interactions
-
-### Everything is opt-in!
 
 Decoupling the system from fixed parts can be done by availing configuration options under the NixOS modules:
 
 ```nix
 {
   # bootloader.grub.enable = false;
+  # bootloader.zfsbootmenu.enable = false;
   bootloader.gummiboot.enable = true;
   init.runit.enable = true;
   boot.kernelPackages = pkgs.linuxPackages_latest;
@@ -104,62 +68,11 @@ This is an example constituting how one can easily replace grub with gummiboot.
 
 By providing this level of granularity, users can precisely control their system components, opting for their preferred choices while excluding unnecessary defaults. This empowers users to create tailored configurations and enhances the flexibility of their NixOS setups.
 
-### Emphasis on flake over shell script wrappers
-
-([on hold](#flake-transit)).
-
-[nixos-rebuild](https://github.com/NixOS/nixpkgs/blob/master/pkgs/os-specific/linux/nixos-rebuild/nixos-rebuild.sh) is nothing but a complex shell script, that has to be present in the system beforehand, which basically calls `nix-build` under the hood, building a derivation, and then running the built derivation. All the options can be instead defined in nix language, in the flake definition itself, and the output derivation can be built using the nix command directly.
-
-[nixos-generators](https://github.com/nix-community/nixos-generators#using-in-a-flake) has quite recently started adding flake support through which there's no need to install or do anything of it, a simple `nix build` will build the iso/vm images. Making it transparent as well as reducing the extra script installation setup.
-
-### CLI improvements
-
-This is majorly for improving the binary cache user interaction, remote-builders setup, and cache serving.
-
-This is not the exact design required, but something like this will suffice:
-
-```bash
-nix-store --export /nix/store/{somepath} > file
-
-## This is the main point, pull paths from binary cache, if we exactly know it
-nix store pull $(nix-store --query --requisites /nix/store/{some.drv})
-nix-store --import < file
-```
-
-<sub><b>Note:</b> `nix copy` is for copying "closure" between two nix stores and cannot facilitate upstream caching, which substituter setup do; to fallback to next substituter if path doesn't exist in one. There isn't a way to pull paths from configured caches even if we know the paths beforehand.</sub>
-
-The purpose of remote builder shouldn't be exactly to serve the built file immediately, maybe to store it in some static-storage, and shutdown, the cache serving should be made accessible using stateless http server, which can happen if we can serve a folder through http.
-
-Articles confuse the binary cache with closure copying, such as [this one](https://nix-tutorial.gitlabpages.inria.fr/nix-tutorial/deploying.html), one of top results in Google with "setting up a binary cache nix" query.
-
-Hence there shouldn't be need of third party package in order to serve something, unnecessary complexity & abstractions should be reduced to fit a simple custom http file server to act as binary cache.
-
-There should be a way to build isolated directory with the filtered store path references only, or even an option to have upstream filtered paths (that are not in https://cache.nixos.org). Something like:
-
-```nix
-{
-  # ...
-  outputs = args: {
-    packages.x86_64-linux.my-binary-cache = some-binary-cache-setup-function {
-      packages = [ custom-package-1 custom-package-2 ];
-      # ...
-    };
-  };
-}
-
-# nix build '.#my-binary-cache'
-# python3 -m http.server -d result
-```
-
 
 # Drawbacks
 [drawbacks]: #drawbacks
 
 If everything is made opt-in, it may result in a slightly longer initial configuration file, although way more explicit.
-
-There are no perceived drawbacks in nix-cli improvements.
-
-<p id="flake-transit">Flake enforcement is on hold, and if the necessary requirements are fulfilled by default.nix (typical, oldschool interfaces), it will get removed from RFC after checking all edge cases.</p>
 
 # Alternatives
 [alternatives]: #alternatives
@@ -173,13 +86,6 @@ If left unimplemented however, abstraction will keep growing and complexity will
 
 # Prior art
 [prior-art]: #prior-art
-
-### Remote Builds
-
-* [reference #1 (reddit)](https://www.reddit.com/r/NixOS/comments/l5221x/comment/gm9xre8).
-* [reference #2 (gist)](https://gist.github.com/danbst/09c3f6cd235ae11ccd03215d4542f7e7).
-* [reference #3 (LnL7/nix-docker) - Setting up a docker container as a remote builder](https://github.com/LnL7/nix-docker/tree/master#running-as-a-remote-builder).
-* [reference #4 (DeterminateSystems/magic-nix-cache#24)](https://github.com/DeterminateSystems/magic-nix-cache/issues/24).
 
 ### Standalone service / process management
 
@@ -202,6 +108,4 @@ If left unimplemented however, abstraction will keep growing and complexity will
 # Future work
 [future]: #future-work
 
-1. Document the changes introduced in cli.
-1. In case stdenv with minimal requisites is introduced (opt-in C compiler setup), do necessary changes to shell-script or non-C builds from stdenv.
-1. Making low-level primitives openly accessible will expand the range of choices available for enhancing nix, for instance on some points of Hackernews criticisms, and make it easy to create remote builders and substituters directly through the nix command-line interface. Third-party implementations like [nixbuild.net](https://nixbuild.net) and [attic](https://github.com/zhaofengli/attic) will no longer be required to reconstruct all specifications from the ground up.
+1. Service unit creation may be majorly affected by the change, may need a module that generalize the api and expands to each of possible service manager types so whatever is current init can work with that.
--- a/rfcs/0157-more-granular-control.md
+++ b/rfcs/0157-more-granular-control.md
@@ -21,7 +21,7 @@ To grant more granular low-level control over the system. Majorly:
   - This change aims to reduce the disparity observed when using nix inside NixOS and outside NixOS.
   - By making the infrastructure flexible and allowing components to be explicitly chosen, users can adapt to future breaking changes, such as the introduction of pipewire and wayland, even for init systems and bootloaders.
 
-* `flakes`: Enforce the use of flakes for scripting operations.
+* `flakes`: ([on hold](#flake-transit)) Enforce the use of flakes for scripting operations.
   - Currently, script infrastructure often involves a `script -> nix language -> script` abstraction, which can make it challenging to modify, reuse, or apply fixes to the underlying script. Examples of this include nixos-rebuild, nixos-generate, and home-manager.
   - With the enforcement of flakes, the former `script ->` abstraction will be reduced, and users can directly use a `nix` command to build the script. The activation script `./result/bin/activation-script` can then be executed to complete the task.
 
@@ -51,13 +51,13 @@ Although most things are straightforward in nix, such as changing a desktop envi
 
 In void (just as a reference), a package that ignores already upstream-dependency is built (in a chroot by default). This output folder can then act as a simple ad-hoc repository mirror with any http server (including `python3 -m http.server -d hostdir/binpkgs` on the host-machine or even github pages after a build from GitHub actions).
 
+While the current configuration does work for most people, and many are content with the existing solutions, a more granular control system would offer tremendous possibilities for customization and make the infrastructure flexible for future shifts.
+
+Instead of deprecating pulseaudio's `sound.*` in favor of `services.pipewire.*`, we should aim to build a flexible API that persists even when new standards emerge, reducing confusion caused by generic names like "sound" getting deprecated.
+
 
 # Detailed design
 [design]: #detailed-design
-
-While the current configuration does work for most people, and many are content with the existing solutions, a more granular control system would offer tremendous possibilities for customization and make the infrastructure flexible for future shifts.
-
-Instead of deprecating pulseaudio's `sound.*` in favor of `services.pipewire.*`, we should aim to build a flexible API that persists even when new standards emerge, reducing confusion caused by generic names like "sound" being deprecated.
 
 ### Everything is opt-in!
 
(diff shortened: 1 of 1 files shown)


rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+
+```bash
+nix-store -r /nix/store/{some.drv}        ## OR: nix build /nix/store/{some.drv}
+# pulls/builds whatever is in $(nix show-derivation /nix/store/{some.drv} | jq '.[] | .inputSrcs, .inputDrvs')
+# builds and produces output
+
+nix-store --export /nix/store/{somepath} > file
+
+## This is the main point, pull paths from binary cache, if we exactly know it
+nix store pull $(nix-store --query --requisites /nix/store/{some.drv})
```
Ma27, line 135: Can't this be done with `nix copy`?

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+
+The purpose of remote builder shouldn't be exactly to serve the built file immediately, maybe to store it in some static-storage, and shutdown, the cache serving should be made accessible using stateless http server, which can happen if we can serve a folder through http.
+
+Articles confuse the binary cache with closure copying, such as [this one](https://nix-tutorial.gitlabpages.inria.fr/nix-tutorial/deploying.html), one of top results in Google with "setting up a binary cache nix" query.
+
+Hence there shouldn't be need of third party package in order to serve something, unnecessary complexity & abstractions should be reduced to fit a simple custom http file server to act as binary cache.
+
+```bash
+# Something like:
+nix store setup > /nix/store/nix-cache-info
```
Ma27, line 147: fwiw setting up a binary cache somewhere can be done with `nix copy --to file:///path/to/binary/cache/on/disk` and then let nginx serve that.

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+
+python3 -m http.server -d /nix/store
```
Ma27, line 149: This is probably not a good idea because a binary cache has a different format (i.e. NAR files) rather than verbatim store-paths on disk and now you have both in the same directory.

If you really want to serve your entire `/nix/store` (and I'm not sure if you want to do that in most of the cases), there's still https://github.com/nix-community/harmonia.

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+  outputs = args: {
+    packages.x86_64-linux.my-binary-cache = some-binary-cache-setup-function {
+      packages = [ custom-package-1 custom-package-2 ];
+      # ...
+    };
+  };
+}
+
+# nix build '.#my-binary-cache'
+# python3 -m http.server -d result
```
Ma27, line 166: `nix build .# --store file:///$(pwd)/result`? Or if you don't want the build closure in there, `nix build .#; nix copy ./result --to file:///$(pwd)/cache`? (haven't tested the exact commands though).

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+}
+```
+
+This is an example constituting how one can easily replace grub with gummiboot.
+
+By providing this level of granularity, users can precisely control their system components, opting for their preferred choices while excluding unnecessary defaults. This empowers users to create tailored configurations and enhances the flexibility of their NixOS setups.
+
+### Emphasis on flake over shell script wrappers
+
+[nixos-rebuild](https://github.com/NixOS/nixpkgs/blob/master/pkgs/os-specific/linux/nixos-rebuild/nixos-rebuild.sh) is nothing but a complex shell script, that has to be present in the system beforehand, which basically calls `nix-build` under the hood, building a derivation, and then running the built derivation. All the options can be instead defined in nix language, in the flake definition itself, and the output derivation can be built using the nix command directly.
```
Ma27, line 117: This would effectively lock out everyone who doesn't want to use experimental features. Also, the scripts have an existing shell completion whereas there's no proper completion system out there that completes nix code for me.

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+
+As explained earlier in the motivation section, the binary cache has no direct access, and using `nix-store --import {exported}` does not automatically pull missing dependencies from the binary cache in case the export does not include the whole closure.
+
+Allowing direct binary-cache access, such as querying information and pulling specific paths into the local `/nix/store`, would be a significant enhancement.
+
+### (optional) stdenv improvements
+
+Optionally, we could accompany this with a simpler, low-abstraction alternative to stdenv, or an opt-in C compiler setup.
+
+It is worth noting that `stdenv.mkDerivation` pulls gcc and autotools, but not all builds require these dependencies. Some simple shell scripts and builds in other languages redundantly pull these dependencies as part of the build, even when builtins.derivation provided by nix language could be sufficient.
```
Ma27, line 88: `stdenvNoCC` is what you looking for I guess?

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+
+And most of the time usage of the `builtins.derivation` is far from practical, examples shown in Prior Arts section below.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+If everything is made opt-in, it may result in a slightly longer initial configuration file, although way more explicit.
+
+There are no perceived drawbacks in flake enforcement / nix-cli improvements.
```
Ma27, line 191: Even though I'm a heavy flake user, I strongly disagree with this sentence.

rfcs/0157-more-granular-control.md
Animeshz: Wait, is that so? I didn't knew a binary cache can be used as `--from` on nix copy, all the examples in the documentation https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-copy.html seems to show `file://` or `ssh://` (machine-to-machine copy, not involving binary cache). Since this works, I think this is redundant to look for then...

rfcs/0157-more-granular-control.md
Animeshz: From the description https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-copy.html:

`nix copy` copies store path closures between two Nix stores.

I don't think there's a ingenious way to do this in the new cli at the moment, the closest is the `nix-store --export`...

rfcs/0157-more-granular-control.md
Animeshz: The configuration options (including personal comments) are more flexible than remembering / checking history.

Surely there's no alternative than shell completion at the moment, but you can look up for `nix repl`, all the auto-completion work there, you can import your flake `:lf .` and then you can query the nixos modules using tab completion. As documentation improves, flake usage improves as well.

And as said this will be more flexible too in the future because flake outputs are accessible easily, remotely:

> Operating everything under flake also opens up various opportunities, such as building the latest nixos iso image with or without plasma/gnome directly from the nixpkgs repository if its derivation is exposed in flake.nix (for example, `nix build nixpkgs#isoInstallerImageWithGnome`).

And its not to remove the script wrappers altogether, but just like nixos-generators add an alternative way to configure with flakes which could in future stablize and script wrappers be deprecated.

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+
+This approach also facilitates the use of NixOS-modules on non-NixOS nix installations.
+
+<sub>[1]: Recently [dracut merged](https://github.com/NixOS/nixpkgs/issues/77858), but there are no clear instructions on how to use that.</sub>
+
+### Emphasis on flake over shell script wrappers
+
+With the introduction of `nix run` and `nix build`, all current implementations should transition to using these first-party nix commands instead of wrappers like nixos-rebuild, nixos-generate, and home-manager.
+
+Operating everything under flake also opens up various opportunities, such as building the latest nixos iso image with or without plasma/gnome directly from the nixpkgs repository if its derivation is exposed in flake.nix (for example, `nix build nixpkgs#isoInstallerImageWithGnome`).
```
7c6f434c, line 76: This is confusing. _If_ a version of ISO is exposed (like some are in `release.nix`), you can just `nix-build -A` it already. If it is not… then Flakes make things harder because you probably want `nix-build -E ` overriding something from the command line.

rfcs/0157-more-granular-control.md
Animeshz: If that is the case, then most of the packages and script wrappers unknowingly just use the `stdenv`, for instance this [iso generation shell script](https://github.com/NixOS/nixpkgs/blob/master/nixos/lib/make-iso9660-image.nix#L48) has nothing to do with c compiler yet still use stdenv. And there are hundred of other packages that do this too...

I think the use of `stdenvNoCC` when possible can be effectively enforced (for most of non-C derivations) with this RFC? If that is ok?

Ma27: > I think the use of stdenvNoCC when possible can be effectively enforced (for most of non-C derivations) with this RFC. If that is ok?

If no c compiler is needed, then switching `stdenv` to `stdenvNoCC` is absolutely fine IMHO. This is probably not so much of an issue to discuss, but somethign to be documented more prominently and in the end somebody has to do the work (i.e. "fixing" up the packages).

rfcs/0157-more-granular-control.md
Ma27: > The configuration options (including personal comments) are more flexible than remembering / checking history.

Can you give an example perhaps about which options you'd like to see in Nix rather than as CLI options? I wouldn't want to e.g. change the flake whenever I exchange `switch`/`test`/etc whenever needed for instance?

rfcs/0157-more-granular-control.md
7c6f434c: > Flakes were introduced for more flexibility in pinning, and managing the derivations with more isolation (from NIX_PATH and rest of environment). 

This is not relevant to the alleged features coming from flakes.

> They also allow to be called remotely using github:xxx/yyy and more of such syntax.

This is just telling Nix to download the expressions and evaluate them… which was actually already in use before flakes (at least with full HTTPS urls)

> You can in theory do the same with flakes, and in addition, you can take flake input and write a definition in your flake output which overrides something from original derivation easily, whereas without flakes, how would you import nixos/release.nix in your config and override something without command line option -E through a nix file (to keep the config in vcs).

This is contradictory. Either you can do the same thing with flakes (doubt it, flakes are currently extremely bad at flexibility), or flakes force you to actually create a file and check it in for trying out a small override.

rfcs/0157-more-granular-control.md
Animeshz: I'm not suggesting to put these in flakes

```
      switch|boot|test|build|edit|dry-build|dry-run|dry-activate|build-vm|build-vm-with-bootloader)
```
these are the options that are freqently changed, and can be added as different outputs (`.dry`|`.test`) suffixes.

but rather options such as

```
      --install-grub)
      --install-bootloader)
      --no-build-nix)
      --use-substitutes|--substitute-on-destination|-s)
      -I|--max-jobs|-j|--cores|--builders|--log-format)
      -j*|--quiet|--print-build-logs|-L|--no-build-output|-Q| --show-trace|--keep-going|-k|--keep-failed|-K|--fallback|--refresh|--repair|--impure|--offline|--no-net)
      --option)
      --fast)
      --profile-name|-p)
# and many more...
```
are generally fixed, always, in every switch/build, these can be specified as a nix configuration option instead in the flake output.

rfcs/0157-more-granular-control.md
Animeshz: What I'm trying to say is that, if it was exposed in flake, we can do something like:

```
outputs = args: {
  my-custom-override = inputs.nixpkgs.nixos.release.imagePath.overrideAttrs (...);
}
```

whereas currently there's no way to access it in nix expressions, as release.nix is not exposed in any way, not even through [default.nix in nixpkgs/nixos](https://github.com/NixOS/nixpkgs/tree/master/nixos).

I'm not sure how would you be editing it via `-E` either, since its not in default.nix, you can't import it as `let a = import <nixpkgs/nixos/release> in a.{path}`...

7c6f434c: > you can't import it as `let release = import <nixpkgs/nixos/release.nix> {}; in `

Yes you can (with minor fixes to syntax)

> And I don't think flakes are by far any less flexible to do the same thing, rather the enforcement makes them more standardized to work with consistent api.

The mere appearance of `nixpkgs-unfree` shows that flakes are more restrictive (without flakes that single flag is passed via command line or maybe single-command env var; the fact that people do find a ready-made flake that does all the necessary dance useful shows that flake interfaces are not good enough).

Another example is cross-compilation.

Animeshz: > Yes you can (with minor fixes to syntax)

Would be great if you could put a complete example, as I'm not familiar enough with old syntax very much :)

rfcs/0157-more-granular-control.md
Animeshz: `nix copy` is for copying closure between two nix stores and cannot facilitate upstream caching, which substituter setup do; fallback to next substituter if path doesn't exist in one. So its not exactly the most generic solution (similar situation also mentioned in one of the [below comment](https://github.com/NixOS/rfcs/pull/157#discussion_r1281835622)).

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+* `nixos`: Decouple various low-level components of the operating system and make everything "opt-in."
+  - This change aims to reduce the disparity observed when using nix inside NixOS and outside NixOS.
+  - By making the infrastructure flexible and allowing components to be explicitly chosen, users can adapt to future breaking changes, such as the introduction of pipewire and wayland, even for init systems and bootloaders.
+
+* `flakes`: Enforce the use of flakes for scripting operations.
+  - Currently, script infrastructure often involves a `script -> nix language -> script` abstraction, which can make it challenging to modify, reuse, or apply fixes to the underlying script. Examples of this include nixos-rebuild, nixos-generate, and home-manager.
+  - With the enforcement of flakes, the former `script ->` abstraction will be reduced, and users can directly use a `nix` command to build the script. The activation script `./result/bin/activation-script` can then be executed to complete the task.
+
+
+# Motivation
```
AndersonTorres, line 29: Your motivation is composed of at least three ideas.
They look too abstract and too decoupled. "Provide direct access to the Nix Store" has nothing to do with "Enforce the use of flakes", and "Enforce flakes" has nothing to do with "decouple various" (various? Not all?) "low-level components".

It looks like scrambling three RFCs into only one. And it makes the thing at least four times harder to understand and reason about.

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+
+In void (just as a reference), a package that ignores already upstream-dependency is built (in a chroot by default). This output folder can then act as a simple ad-hoc repository mirror with any http server (including `python3 -m http.server -d hostdir/binpkgs` on the host-machine or even github pages after a build from GitHub actions).
+
+
+# Detailed design
+[design]: #detailed-design
+
+While the current configuration does work for most people, and many are content with the existing solutions, a more granular control system would offer tremendous possibilities for customization and make the infrastructure flexible for future shifts.
+
+Instead of deprecating pulseaudio's `sound.*` in favor of `services.pipewire.*`, we should aim to build a flexible API that persists even when new standards emerge, reducing confusion caused by generic names like "sound" being deprecated.
```
AndersonTorres, line 60: These two paragraphs are not a "detailed design", but a "motivation".

rfcs/0157-more-granular-control.md
AndersonTorres: > And there are hundred of other packages that do this too...

This is just a matter of pull-requesting a treewide modification. No need for an RFC.

Also, and again, it adds four completely decoupled components to this RFC.

> I think the use of `stdenvNoCC` when possible can be effectively enforced (for most of non-C derivations) with this RFC. If that is ok?

It is already enforced when code is reviewed.

You can basically remove this section, it is 100% redundant.

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+[prior-art]: #prior-art
+
+### Remote Builds
+
+* [reference #1 (reddit)](https://www.reddit.com/r/NixOS/comments/l5221x/comment/gm9xre8).
+* [reference #2 (gist)](https://gist.github.com/danbst/09c3f6cd235ae11ccd03215d4542f7e7).
+* [reference #3 (LnL7/nix-docker) - Setting up a docker container as a remote builder](https://github.com/LnL7/nix-docker/tree/master#running-as-a-remote-builder).
+* [reference #4 (DeterminateSystems/magic-nix-cache#24)](https://github.com/DeterminateSystems/magic-nix-cache/issues/24).
+
+### Directly using builtins.derivation
```
AndersonTorres, line 214: IMnsHO, `builtins.derivation` is similar to `ninja.build`: too low-level to be used directly by human beings.

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+* `nix-cli`: Introduce direct access to the store and binary cache.
+  - This improvement will facilitate easier binary cache setup, upstream caching, and shallow (non-closure) copies.
+  - Users will be able to pull specific paths directly from the cache without relying on derivation realisation as an abstraction.
+
+* `nixos`: Decouple various low-level components of the operating system and make everything "opt-in."
+  - This change aims to reduce the disparity observed when using nix inside NixOS and outside NixOS.
+  - By making the infrastructure flexible and allowing components to be explicitly chosen, users can adapt to future breaking changes, such as the introduction of pipewire and wayland, even for init systems and bootloaders.
+
+* `flakes`: Enforce the use of flakes for scripting operations.
+  - Currently, script infrastructure often involves a `script -> nix language -> script` abstraction, which can make it challenging to modify, reuse, or apply fixes to the underlying script. Examples of this include nixos-rebuild, nixos-generate, and home-manager.
```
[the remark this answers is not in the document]
Animeshz: That was there just to serve as a general example :)

rfcs/0157-more-granular-control.md
7c6f434c: Uhm,  on my system I sometimes change the state of  `-j` and `-k` depending on wokflow context of preparing the update. It's not mainline NixOS but I would expect _some_ users to do the same on NixOS.

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Everything is opt-in!
+
+First and foremost, during image building (whether it's nixos-rebuild or vm/iso generation), the filesystem should be empty, with no packages or files.
+
+We can then assemble basic components by defining choices such as the bootloader, kernel, initramfs generator [1], libc, init, and then the choice of sound & de/wm.
+
+This approach also facilitates the use of NixOS-modules on non-NixOS nix installations.
+
+<sub>[1]: Recently [dracut merged](https://github.com/NixOS/nixpkgs/issues/77858), but there are no clear instructions on how to use that.</sub>
```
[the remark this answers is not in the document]
Animeshz: Not sure why?

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,241 @@
...
+  - Users will be able to pull specific paths directly from the cache without relying on derivation realisation as an abstraction.
+
+* `nixos`: Decouple various low-level components of the operating system and make everything "opt-in."
+  - This change aims to reduce the disparity observed when using nix inside NixOS and outside NixOS.
+  - By making the infrastructure flexible and allowing components to be explicitly chosen, users can adapt to future breaking changes, such as the introduction of pipewire and wayland, even for init systems and bootloaders.
+
+* `flakes`: Enforce the use of flakes for scripting operations.
+  - Currently, script infrastructure often involves a `script -> nix language -> script` abstraction, which can make it challenging to modify, reuse, or apply fixes to the underlying script. Examples of this include nixos-rebuild, nixos-generate, and home-manager.
+  - With the enforcement of flakes, the former `script ->` abstraction will be reduced, and users can directly use a `nix` command to build the script. The activation script `./result/bin/activation-script` can then be executed to complete the task.
```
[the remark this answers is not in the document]
Animeshz: I'm not really comfortable with oldschool syntax, I'll carefully read them up tomorrow and see if prime purpose - indirection removal be done for derivations that give rise to yet another shell script. If it can be served using that, I'll see how RFC has to be modified that way.

rfcs/0157-more-granular-control.md
Animeshz: In the past I thought they all belong to same idea so I've initially grouped them together in single RFC. Any suggestions (if its not too late)?

rfcs/0157-more-granular-control.md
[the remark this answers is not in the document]
AndersonTorres: It is not a useful information. There are many initramfs generators out there, there is no reason to cite an example.

Or, if you want to use this as an example, put this in the text as a real example, not as a footnote.

rfcs/0157-more-granular-control.md
AndersonTorres: Reduce the scope, and/or open other RFCs.

You can write a whole RFC just for the most sensitive issue (IMHO): provide direct access to Nix Store.

Decouple low-level components can be done in another RFC, since it deals with NixOS.

Flake enforcement is already being discussed by other RFC, as far as I remember: https://github.com/NixOS/rfcs/pull/136

amjoseph-nixpkgs:
Could we have more granular control over this RFC?

This seems like (at least) three separate RFCs that have been lumped together.

Animeshz:
> This seems like (at least) three separate RFCs that have been lumped together.

Sure, the decoupling of nixos module is the major one here, that'll enhance the usage of nix in and outside NixOS, I'll try to split rest of them into separate RFC (or remove the parts unneeded, if any) by today's evening.

rfcs/0157-more-granular-control.md
lucasew: > This would effectively lock out everyone who doesn't want to use experimental features. Also, the scripts have an existing shell completion whereas there's no proper completion system out there that completes nix code for me.

I think it's time to not feature flag nix-command and flakes anymore. Everyone will use these features anyway, this extra step only annoys everyone.

AndersonTorres: Flakes are still not stable enough.

Animeshz:
@Nyabinary Sorry I've been really busy with other things, I've reduced the scope of the RFC as others pointed out, now to only aim at allowing to expose more fundamental parts of the system from modules, and make it more flexible to swap components in future.

CC @amjoseph-nixpkgs @AndersonTorres @Ma27 @7c6f434c let me know if this shortened RFC scope is relevant or not :) Marking it open for review!

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,111 @@
...
+
+**Everything is opt-in!**
+
+First and foremost, during image building (whether it's nixos-rebuild or vm/iso generation), the filesystem should be empty, with no packages or files.
+
+We can then assemble basic components by defining choices such as the bootloader, kernel, initramfs generator [1], libc, init, and then the choice of sound & de/wm.
+
+This approach also facilitates the use of NixOS-modules on non-NixOS nix installations.
+
+<sub>[1]: Recently [dracut merged](https://github.com/NixOS/nixpkgs/issues/77858), but there are no clear instructions on how to use that.</sub>
```
AndersonTorres, line 47: This is not a paper of philosophy of mathematics. Get rid of this footnote - include it as an example for your text or remove it entirely.

rfcs/0157-more-granular-control.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+Decoupling the system from fixed parts can be done by availing configuration options under the NixOS modules:
+
+```nix
```
AndersonTorres, line 55: I  have a feeling it needs a PoC. Maybe creating a small system [from power up to  bash prompt](https://tldp.org/HOWTO/From-PowerUp-To-Bash-Prompt-HOWTO.html).

AndersonTorres (approved):

kevincox:
This RFC is now open for shepherd nominations.

rfcs/0157-more-granular-control.md
Animeshz: @AndersonTorres I had started working on the PoC right away that day, although soon I realized it doesn't really relates much to this.

I believe a better idea would be to shrinking down
* the activation script (`rg activationScripts\.` on root of nixpkgs clone).
* [the main derivation's buildCommand](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/activation/top-level.nix#L53) that populates the nixos-rebuild output directory.
* and [the etc derivation's buildCommand](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/etc/etc.nix#L11) that populates etc/ subdirectory in the nixos-rebuild output directory.

behind a dedicated options for each of them in the configuration.

That will slowly create options, as well as allocate multiple choices for that options. This can also be slowly integrated into the nixpkgs, as we could do it first for bootloader, then for init, and so on in separate PRs, without touching everything at once.

Let me know what do you think, making things from groundup or shrinking down what's already in there. Because latter one looks much better to do.

infinisil:

Aleksanaa:
But how do you support multiple init/service supervisor without adding an abstraction layer? Requiring every service maintainer to write service configuration (that is already very low level btw) in a couple of other formats again and again? Or letting the users write services themselves?

For other parts (boot method, libc, kernel and so on) the abstraction layer shouldn't be the blocking issue. The real problem is we lack people actively creating, maintaining and using different implementations with great interest. without a lot of continuous investment, these implementations will quickly become unusable.

aanderse:
I'm interested in being a shepherd for this RFC. Hopefully some other interested parties will show up as well.

AndersonTorres:
I personally believe this RFC will hugely benefit from https://github.com/NixOS/rfcs/pull/163.

Animeshz:
> I personally believe this RFC will hugely benefit from #163.

Yes, since the generalization of service management is handled by that RFC in detail, it'll remove some unanswered questions about service creations and reduce a huge load on this.

---

Also, on the https://github.com/NixOS/rfcs/pull/163#issuecomment-1763238786,

> I have a plan to assimilate [Home Manager](https://github.com/nix-community/home-manager) to Nixpkgs repo.
> Can your RFC help this transition, since it would provide a portable service layer to HM too?

If we eliminate the assumptions NixOS makes, about the system, a part of this RFC, the home-manager can also be easily merged into nixpkgs as _nixos module itself_, and then nixos modules can be used inside/outside NixOS.

Nyabinary:
> > I personally believe this RFC will hugely benefit from #163.
> 
> Yes, since the generalization of service management is handled by that RFC in detail, it'll remove some unanswered questions about service creations and reduce a huge load on this.
> 
> Also, on the [#163 (comment)](https://github.com/NixOS/rfcs/pull/163#issuecomment-1763238786),
> 
> > I have a plan to assimilate [Home Manager](https://github.com/nix-community/home-manager) to Nixpkgs repo.
> > Can your RFC help this transition, since it would provide a portable service layer to HM too?
> 
> If we eliminate the assumptions NixOS makes, about the system, a part of this RFC, the home-manager can also be easily merged into nixpkgs as _nixos module itself_ (theoretically), and then nixos modules can be used inside/outside NixOS.

Wouldn't including home-manaager into NixOS/nixpkgs include a lot of refactorings?

Animeshz:
> > > I personally believe this RFC will hugely benefit from #163.
> > 
> > 
> > Yes, since the generalization of service management is handled by that RFC in detail, it'll remove some unanswered questions about service creations and reduce a huge load on this.
> > Also, on the [#163 (comment)](https://github.com/NixOS/rfcs/pull/163#issuecomment-1763238786),
> > > I have a plan to assimilate [Home Manager](https://github.com/nix-community/home-manager) to Nixpkgs repo.
> > > Can your RFC help this transition, since it would provide a portable service layer to HM too?
> > 
> > 
> > If we eliminate the assumptions NixOS makes, about the system, a part of this RFC, the home-manager can also be easily merged into nixpkgs as _nixos module itself_ (theoretically), and then nixos modules can be used inside/outside NixOS.
> 
> Wouldn't including home-manaager into NixOS/nixpkgs include a lot of refactorings?

Technically, the types in home-manager such as dag & gvariant may benefit nixpkgs modules. And regarding refactorings, yes, it might incur a few, but since home-manager is primarily focused on files on $HOME, I don't think there'd be a lot of overlap and hence easily integratable. If however, some package can be configured both ways (system and home level) then probably something has to be done to deduplicate and reduce the confusion among the config options.

edolstra:

Nyabinary:
> I'm interested in being a shepherd for this RFC. Hopefully some other interested parties will show up as well.

Wasn't this party interested in being a shepherd?

aanderse:
No one else expressed interest though, right? 1 person isn't enough.

PR #158, opened 2023-08-11
[RFC 0158] Flake version 

lucasew:
# Summary

This RFC proposes the addition of a `version` option for `flake.nix` to define which flake spec version one flake is using, like how it's done with `docker-compose`.

This is for flake stabilization but without sacrificing future revisions.

[Rendered](https://github.com/lucasew/rfcs/blob/flake-version/rfcs/0158-flake-version.md)

### DIFF
--- a/rfcs/0158-flake-version.md
+++ b/rfcs/0158-flake-version.md
@@ -1,0 +1,82 @@
+---
+feature: flake-version
+start-date: 2023-08-11
+author: lucasew
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Add a `version` option for a flake to define which flake spec version one flake is using.
+
+# Motivation
+[motivation]: #motivation
+
+Flakes are marked as unstable but are so widespread in the ecosystem that the
+feature flag is more like an annoyance than something really good, pragmatically speaking.
+
+# Detailed design
+[design]: #detailed-design
+
+In Docker, there is a tool called `docker-compose`, the reason why this tool exists
+is not very relevant to this document but it's basically a tool that the user
+uses some kind of data definition language to define a set of entities that
+Docker will setup and spawn. This data definition language structure is as
+stable as flake is today but they apply an extra concept to have more flexibility
+if they need to change something: the `version` attribute.
+
+Right now we have basically two flake specifications that became minimally widespread.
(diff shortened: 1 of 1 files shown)


Ma27:
This existed in the past in form of an `edition` field, similar to the field in `Cargo.toml`. However it was removed later because parsing an edition field from a `flake.nix` means that the Nix version you're using must be able to parse the entire file. This is not the case if e.g. a very old Nix is used to read a flake.nix using newer language features (or vice versa). See also https://github.com/nixos/nix/commit/e5ea01c1a8bbd328dcc576928bf3e4271cb55399.

This RFC has the exact same issue and doesn't mention this at all.

kevincox:
This RFC is now open for shepherd nominations.

lucasew:
> This existed in the past in form of an `edition` field, similar to the field in `Cargo.toml`. However it was removed later because parsing an edition field from a `flake.nix` means that the Nix version you're using must be able to parse the entire file. This is not the case if e.g. a very old Nix is used to read a flake.nix using newer language features (or vice versa). See also [NixOS/nix@e5ea01c](https://github.com/NixOS/nix/commit/e5ea01c1a8bbd328dcc576928bf3e4271cb55399).
> 
> This RFC has the exact same issue and doesn't mention this at all.

I think this breakage shouldn't be so unwanted as flakes are still not marked as stable. The disruption would be much smaller if something like this would actually be introduced sooner.

It's unstable for a reason I guess

And this compatibility system can be backported to old versions if that's the case.

Ma27:
> I think this breakage shouldn't be so unwanted as flakes are still not marked as stable. The disruption would be much smaller if something like this would actually be introduced sooner.

This is not about a break to the current flake format (I agree that it's fine especially as long as flakes are unstable to change stuff), but about future language changes that may be used in flake.nix and as a result it's not possible for older Nix versions to parse the flake.nix (and the version field inside) to find out what language level is supposed to be used.

> And this compatibility system can be backported to old versions if that's the case.

If the suggestion is about backporting new language features to older Nix versions to still be able to parse newer flake.nix files, then I'd like to hear from the Nix (and probably also Tvix) developers if they consider that feasible. I'm not too involved in either, but I highly doubt that this is sustainable and something time should be invested in.

justinlovinger:
What about storing the version in `flake.lock`? Presumably, `flake.lock` is less likely to need future Nix features. It can be included when the lock file is generated.

If there are still concerns about `flake.lock` compatibility with Nix versions, the version could be stored in a separate plaintext file, like `.flake-version`.

Additionally, being automatically added, the version could be fully automated. Instead of manually adding a `version` field, the version can be added based on the Nix version used to run the Flake file.

tomberek:

PR #160, opened 2023-08-11
Update 92: `^` instead of `!`

Ericson2314:
The syntax we actually got is different than what was in the RFC. Updating the RFC to reflect this later decision.


Ma27:
I guess it would make sense to elaborate here on why this has happened (or link the discussion around this) :)

Ericson2314:
https://github.com/NixOS/nix/pull/6449#issuecomment-1114599074 did this in a tagentially related feature, which set the precedent. That comment was in response to a thread elsewhere which I have yet to dig up.

PR #140, opened 2023-01-30
[RFC 0140] Simple Package Paths

infinisil:
### Summary
Auto-generate trivial top-level attribute definitions in Nixpkgs' `pkgs/top-level/all-packages.nix` from a Nixpkgs-internally standardised directory structure that matches the attribute name.
This makes it much easier to contribute new packages, since there's no more guessing needed as to where the package should go, both in the ad-hoc directory categories and in `pkgs/top-level/all-packages.nix`.

[Rendered](https://github.com/nixpkgs-architecture/rfcs/blob/master/rfcs/0140-simple-package-paths.md)

Implementation of part 1: https://github.com/NixOS/nixpkgs/pull/237439
Outdated implementation of part 2: https://github.com/NixOS/nixpkgs/pull/211832

### Nixpkgs Architecture Team
This RFC has been worked on openly in [a repository](https://github.com/nixpkgs-architecture/simple-package-paths) by the new [Nixpkgs Architecture Team](https://nixos.org/community/teams/nixpkgs-architecture.html) in the last couple months. It is the first of hopefully many quality of life improvements to nixpkgs.

The team will continue meeting regularly ([calendar](https://github.com/nixpkgs-architecture#meetings)), see [here](https://github.com/nixpkgs-architecture#meetings) for more details, anybody is free to join and listen in or ask questions. Also feel free to join [the Matrix room](https://matrix.to/#/#nixpkgs-architecture:nixos.org) to discuss this RFC.

### Shepherd team
Here are the shepherds with their nominations, all of which are [now accepted](https://github.com/NixOS/rfc-steering-committee/issues/113).
- @phaer [nominated](https://github.com/NixOS/rfcs/pull/140#issuecomment-1466548091)
- @06kellyjac [nominated](https://github.com/NixOS/rfcs/pull/140#issuecomment-1468408943)
- @aakropotkin [nominated](https://github.com/NixOS/rfcs/pull/140#issuecomment-1494488288), but he's part of the NAT team that created the RFC together, see [this discussion](https://github.com/NixOS/rfcs/pull/140#discussion_r1149458540)
- @piegamesde [nominated](https://github.com/NixOS/rfcs/pull/140#issuecomment-1512080692)

### Major RFC updates

- [Shepherd Meeting #1](https://github.com/NixOS/rfcs/pull/140#issuecomment-1545956282)
- [FCP initiated](https://github.com/NixOS/rfcs/pull/140#issuecomment-1567229950)
[...]

PR #163, opened 2023-10-05
[RFC 0163] Portable Service Layer

svanderburg:
This RFC proposes a generic, portable service layer that makes it possible to manage services on top of (in theory) any operating system that is capable of running the Nix package manager.

[Rendered](https://github.com/svanderburg/rfcs/blob/portable-service-layer/rfcs/0163-portable-service-layer.md)

### DIFF
--- a/rfcs/0163-portable-service-layer.md
+++ b/rfcs/0163-portable-service-layer.md
@@ -851,12 +851,6 @@ to be operating system agnostic.
 This RFC is not the first approach that combines Nix with a process management
 solution:
 
-* The service deployment chapter (chapter 9) of
-  [Eelco Dolstra's PhD thesis](https://github.com/edolstra/edolstra.github.io/blob/master/pubs/phd-thesis.pdf)
-  describes a very early service deployment approach that uses generated
-  start/stop scripts implemented in `bash`. Because shell scripts are so common,
-  they can be used on a variety of operating systems -- in addition to Linux, it
-  can also be used on Darwin and FreeBSD.
 * [NixOS](https://nixos.org) uses `systemd` as its process management solution.
   Services can be deployed by generating systemd unit configuration files in a
   Nix expression. NixOS deploys `systemd` units globally and requires super
@@ -872,20 +866,48 @@ solution:
 * [system-manager](https://github.com/numtide/system-manager) addresses the
   limitation of NixOS to use systemd units independently and can deploy many
   services in the NixOS repository on any Ubuntu-based distribution.
-* [devenv](https://devenv.sh) also makes it possible to run multiple processes
-  (of different services) in a Nix-based development environment. It works a
-  with a variety container-friendly/non-PID-1 based process managers, such as
-  process-compose, overmind, and honcho.
+* [nixos-init-freedom](https://git.sr.ht/~guido/nixos-init-freedom) puts some
+  efforts on replacing systemd-as-pid-1 with s6. Instead of implementing a
+  generic service layer, it translates parts of systemd module configuration
+  into s6 configuration.
+
+The biggest drawback is that these solutions are not generic and there is only
+little reuse possible between service configurations.
+
+There are also solutions that introduce a portable approach for managing
+processes in combination Nix:
+
+* The service deployment chapter (chapter 9) of
+  [Eelco Dolstra's PhD thesis](https://github.com/edolstra/edolstra.github.io/blob/master/pubs/phd-thesis.pdf)
+  describes a very early service deployment approach that uses generated
+  start/stop scripts implemented in `bash`. Because shell scripts are so common,
+  they can be used on a variety of operating systems -- in addition to Linux, it
+  can also be used on Darwin and FreeBSD.
 * [Dysnomia](https://github.com/svanderburg/dysnomia) is a system to manage the
   life-cycle of services deployed by
   [Disnix](https://github.com/svanderburg/disnix) in a generic way. It has its
   own `process` module that deploys services as daemons running the background.
   It can also use the Nix process management framework to deploy services that
   are managed by any kind process manager supported by the framework.
+* [devenv](https://devenv.sh) also makes it possible to run multiple processes
+  (of different services) in a Nix-based development environment. It works a
+  with a variety container-friendly/non-PID-1 based process managers, such as
+  `process-compose`, `overmind`, and `honcho`.
+* [NixNG](https://github.com/nix-community/NixNG) is considered a late sibling
+  to NixOS and shares many designs of it. It also implements the function of
+  generating services for several kinds of container-friendly init systems from
+  the same Nix expressions. NixNG is primarily used to run in containers.
+* [dream2nix](https://github.com/nix-community/dream2nix) has a work-in-progress
+  implementation for a high-level specification that can be translated to
+  configurations for multiple process managers.
 
-Aside from the first and last option, the biggest drawback is that these
-solutions are not generic and there is only little reuse possible between service
-configurations.
+The drawback of some of these solutions is that they rely on services that
+daemonize on their own. Although they are portable, not all services have
+such functionality. Moreover, foreground processes can be more reliably managed
+because they do not rely on PID files -- PID files may get inconsistent.
+
+The other solutions only work with container-friendly supervisors and not with
+commonly used process managers, such as `systemd`.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/rfcs/0163-portable-service-layer.md
+++ b/rfcs/0163-portable-service-layer.md
@@ -1,0 +1,893 @@
+---
+feature: portable_service_layer
+start-date: 2023-09-10
+author: Sander van der Burg
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+This proposal introduces a generic, portable, high-level process management
+framework that makes it possible to manage a variety of commonly used
+application services on preferably all operating systems/environments that the
+Nix package manager supports.
+
+# Motivation
+[motivation]: #motivation
+
+The Nix package manager is a deployment solution that offers many powerful
+features. For example, it is a very convenient solution to deploy
+*multiple versions* and *variants* of packages in isolation so that they do not
+conflict with each other and packages of an existing system installation.
+
+The `nix-shell` command is particularly useful to safely experiment with
+packages -- you can conveniently spawn a shell session in which you can use
+different versions of packages than the host system without changing the host
+system's package configuration.
+
+For Nix, it is *not* even required to have *super-user privileges* to deploy
+packages -- thanks to its isolation properties this can be done without
+interfering with the packages of another user or the system-wide set of packages.
+
+Moreover, the Nix package manager is also a *portable* solution. In addition
+to Linux, Nix also has first-class support for macOS/Darwin. With some effort,
+it can also be used on other UNIX-like operating systems, such as FreeBSD and
+Cygwin.
+
+Unfortunately, Nix does not solve all application deployment problems -- Nix's
+purpose is to deploy packages (e.g. delivering the required artifacts, such as
+executables and configuration files) but not to manage the life-cycle of
+applications, such as long running services (the only exception is the `nix run`
+command that allows you to conveniently start an executable).
+
+For these kinds of applications, Nix is typically combined with a *process
+manager*, such as `systemd`, that is responsible for managing the life-cycle of
+a service. Configuration files for these process managers are also deployed as
+Nix packages.
+
+There are a variety of Nix-based solutions that combine Nix with a process
+management solution (see: [prior art](#prior-art)). These solutions are quite
+useful, but have a number of drawbacks. The most notable drawback is that there
+is almost no reuse between the configurations of the services that are managed by
+these solutions. In practice, I have noticed that quite a few properties remain
+the same in spite of the process management solution that was selected.
+
+Moreover, there are also a number of interesting Nix properties for which we also
+want to provide a process-management equivalent:
+
+* Nix makes it possible to have *multiple versions* and variants of the same
+  package to safely co-exist on the same system. We also want to bring this
+  property to process management -- this becomes possible if we make sure that
+  multiple instances do not share the same state (e.g. state directories, port
+  numbers, UIDs, GIDs).
+* Nix supports *unprivileged* user package installations. It is also possible for
+  many kinds services to run as an unprivileged user as long as we make sure
+  that they use state directories have write permissions (e.g. using a state
+  directory in the user's home directory) and that these services use TCP/UDP
+  port numbers that are greater than 1024
+
+The ideas described in this RFC have been explored in the experimental
+[Nix process management framework](https://github.com/svanderburg/nix-processmgmt).
+There is also an
+[example services repository](https://github.com/svanderburg/nix-processmgmt-services)
+repository that demonstrates how it can be used with a number of commonly used
+services.
+
+# Detailed design
+[design]: #detailed-design
+
+This RFC proposes a number of feature additions to the Nixpkgs ecosystem. These
+features correspond to the concepts implemented in the experimental Nix process
+management framework.
+
+There are variety of ways to implement these concepts. Moreover, the idea is that
+these concepts can be implemented one-by-one and in an iterative fashion.
+
+The following sections explain its concepts one by one and provide possible
+implementation strategies.
+
+## A service layer that can be used independently of NixOS
+
+Currently, the Nixpkgs repository only contains systemd service configurations.
+These systemd configurations cannot be used separately from NixOS (as a
+sidenote: there is project called
+[system-manager](https://github.com/numtide/system-manager)
+that makes it possible to only generate systemd units that can be used outside
+NixOS, but it can only be used on Ubuntu and is still under heavy development).
+
+Another project that makes services management available in combination with
+Nix is [nix-darwin](https://github.com/LnL7/nix-darwin). This project uses
+launchd (on macOS) in combination with Nix to support process management.
+
+To make service deployment accessible to a broader audience, it would be nice to
+create a separation in the Nixpkgs tree for portable systemd services (e.g. a
+`services/` directory in the root folder) and to develop a derivation and
+command-line tool (e.g. `nix-services-rebuild`) that allows you to deploy (at
+least a reasonable sub set) of systemd units separately on a conventional Linux
+distribution.
+
+Low-level system services, such as `cups`) do not need to be separated from
+NixOS, but high-level services should, such as `Apache`, `Nginx`, `MariaDB`,
+`PostgreSQL` etc.
+
+NixOS will be a consumer of this separated service layer.
+
+## Generator functions for other kinds of process managers
+
+With a separated services tree and command-line tool to manage instances of
+services, it should also become more convenient to target different kinds of
+process managers.
+
+The purpose of this RFC is not to have a debate about systemd's usefulness.
+systemd has IMO many good features, and I believe it should remain the default
+option in NixOS.
+
+However, there are also good reasons to sometimes consider a different process
+manager:
+
+* systemd is a Linux-specific tool. If we want to deploy a service on another
+  operating system, we need to pick a different solution that is compatible with
+  that operating system.
+* systemd is glibc-specific. Sometimes, it may also be desired to manage
+  services on a Linux system that uses a different libc, such as musl. Then
+  you need a process manager that is compatible with that different flavour of
+  libc.
+* We may want to deploy services on older Linux distros or distros with LTS
+  support. These distros may still work with sysvinit scripts (also known as
+  [LSB Init compliant scripts](https://wiki.debian.org/LSBInitScripts)).
+* Sometimes we may want to experiment with services as an unprivileged user
+  that does not have the ability to control systemd. Then it can be very
+  convenient to use a process manager that can be used as an unprivileged user,
+  such as `supervisord`.
+* Although it is possible to run systemd in a Docker container to manage
+  multiple processes, it is typically more preferable to use a more Docker
+  friendly alternative
+
+With a separated service layer in Nixpkgs, it should also become more convenient
+to use a different generator module for a different kind of process manager.
+
+Similar to the systemd generator module, we can provide NixOS generator modules
+for other process managers as well, that provide a one-on-one translation
+between properties in the Nix expression language and generated config file
+properties.
+
+For example, we can also develop an abstraction that generates `sysvinit` scripts
+(also known as LSB Init compliant scripts). The nix-processmgmt framework allows
+you to generate a `sysvinit` script as follows:
+
+```nix
+{createSystemVInitScript, nginx}:
+
+let
+  configFile = ./nginx.conf;
+  stateDir = "/var";
+in
+createSystemVInitScript {
+  name = "nginx";
+  description = "Nginx";
+  activities = {
+    start = ''
+      mkdir -p ${stateDir}/logs
+      log_info_msg "Starting Nginx..."
+      loadproc ${nginx}/bin/nginx -c ${configFile} -p ${stateDir}
+      evaluate_retval
+    '';
+    stop = ''
+      log_info_msg "Stopping Nginx..."
+      killproc ${nginx}/bin/nginx
+      evaluate_retval
+    '';
+    reload = ''
+      log_info_msg "Reloading Nginx..."
+      killproc ${nginx}/bin/nginx -HUP
+      evaluate_retval
+    '';
+    restart = ''
+      $0 stop
+      sleep 1
+      $0 start
+    '';
+    status = "statusproc ${nginx}/bin/nginx";
+  };
+  runlevels = [ 3 4 5 ];
+}
+```
+
+(As a sidenote: the nix-processmgmt framework uses different conventions than
+NixOS, which I will explain later)
+
+The above code example invokes the `createSystemVInitScript` function to generate
+a `sysvinit` script to manage Nginx:
+
+* The `name` of the script is: `nginx`
+* The `description` in the comment is: Nginx
+* The `activities` parameter specifies the implementation of all the activities
+  (process lifecycle operations) that the init script exposes (as bash shell
+  instructions), such as starting, stopping, restarting and reloading the
+  service.
+
+Since the kind of activities that you need to implement are so common, it is
+also possible to add a little bit of abstraction on top of the generator
+function:
+
+```nix
+{createSystemVInitScript, nginx}:
+
+let
+  configFile = ./nginx.conf;
+  stateDir = "/var";
+in
+createSystemVInitScript {
+  name = "nginx";
+  description = "Nginx";
+  initialize = ''
+    mkdir -p ${stateDir}/logs
+  '';
+  process = "${nginx}/bin/nginx";
+  args = [ "-c" configFile "-p" stateDir ];
+  runlevels = [ 3 4 5 ];
+}
+```
+
+In the above example, we have simplified the previous example -- we no longer
+specify the implementation of any of the activities, but we specify which
+process we want to be managed and how it can be initialized. From this
+high-level specification the activities are automatically derived.
+
+Currently, the Nix process management framework provides generator functions for
+the following backends:
+
+* `sysvinit` scripts (also known as LSB Init compliant scripts). Because it is
+  standardize by the LSB, widely used and the most primitive, I have decided to
+  experiment with this first.
+* `systemd` services
+* `supervisord` services
+* `launchd` services
+* `bsdrc` scripts
+* `s6-rc` services
+* Windows services (deployed by calling: `cygrunsrv`)
+
+In addition to the above process managers, it also possible to directly generate
+Docker containers for each process instance. Although the framework was not
+originally designed for this purpose, I have noticed that the abstractions are
+generic enough to make such deployments possible.
+
+In NixOS, for each backend, we can add a generator module. Currently, a NixOS
+module for systemd already exists, but we can also add generator modules for
+different kinds of process managers.
+
+## A high-level, target-agnostic specification format for running processes
+
+During the development of the Nix process management framework I have written
+configurations for various kinds of services for various kinds of process
+managers.
+
+If you look closely at their configurations, then I have noticed that they have
+much in common. For example, to generate a `sysvinit` script for managing Nginx,
+I can use the following Nix expression:
+
+```nix
+{createSystemVInitScript, nginx,  stateDir}:
+{configFile, dependencies ? [], instanceSuffix ? ""}:
+
+let
+  instanceName = "nginx${instanceSuffix}";
+  nginxLogDir = "${stateDir}/${instanceName}/logs";
+in
+createSystemVInitScript {
+  name = instanceName;
+  description = "Nginx";
+  initialize = ''
+    mkdir -p ${nginxLogDir}
+  '';
+  process = "${nginx}/bin/nginx";
+  args = [ "-c" configFile "-p" stateDir ];
+  runlevels = [ 3 4 5 ];
+
+  inherit dependencies instanceName;
+}
+```
+
+For `supervisord` I could write:
+
+```nix
+{createSupervisordProgram, nginx, stateDir}:
+{configFile, instanceSuffix ? ""}:
+
+let
+  instanceName = "nginx${instanceSuffix}";
+  nginxLogDir = "${stateDir}/${instanceName}/logs";
+in
+createSupervisordProgram {
+  name = instanceName;
+  command = "mkdir -p ${nginxLogDir}; "+
+    "${nginx}/bin/nginx -c ${configFile} -p ${stateDir}";
+}
+```
+
+And for `systemd` I could write:
+
+```nix
+{createSystemdService, nginx, stateDir}:
+{configFile, instanceSuffix ? ""}:
+
+let
+  instanceName = "nginx${instanceSuffix}";
+  nginxLogDir = "${stateDir}/${instanceName}/logs";
+in
+createSystemdService {
+  name = instanceName;
+  Unit = {
+    Description = "Nginx";
+  };
+  Service = {
+    ExecStartPre = "+mkdir -p ${nginxLogDir}";
+    ExecStart = "${nginx}/bin/nginx -c ${configFile} -p ${stateDir}";
+    Type = "simple";
+  };
+}
+```
+
+As you may probably have already seen, in all the examples I specify the same
+kinds of configuration properties:
+* The initialization steps
+* The executable to run with the appropriate command-line arguments
+* Meta information such as the name and description
+
+We could also create a high-level abstraction that captures these properties:
+
+```nix
+{createManagedProcess, nginx, stateDir}:
+{configFile}:
+
+let
+  nginxLogDir = "${stateDir}/${instanceName}/logs";
+in
+createManagedProcess {
+  name = "nginx";
+  description = "Nginx";
+  initialize = ''
+    mkdir -p ${nginxLogDir}
+  '';
+  process = "${nginx}/bin/nginx";
+  args = [ "-c" configFile" -p" "${stateDir}/${instanceName}" ];
+}
+```
+
+The above code example uses a high-level function: `createManagedProcess` that
+specifies an `initialize` script and a process that we need to launch (`process`)
+with a number of command-line parameters (`args`). This high-level specification
+can be translated to a configuration file for various kinds of process managers.
+
+Currently, the `createManagedProcess` function in the Nix process management
+framework supports the following properties:
+
+* `name`. Name of the service
+* `description`. Description of the service
+* `initialize`. Initialization script that initializes state on first startup.
+* `foregroundProcess`. Path to an executable that launches a process in
+  foreground mode
+* `foregroundProcessArgs`. Command-line parameters propagated to the foreground
+  process
+* `daemon`. Path to an executable that launches a process in daemon mode
+* `daemonArgs`. Command-line parameters propagated to the daemon
+* `process`. When this property is specified, it translates both to:
+  `foregroundProcess` and `daemon`.
+* `args`. When this property is specified, it translates both to:
+  `foregroundProcessArgs`, `daemonArgs`.
+* `foregroundProcessExtraArgs`. Extra command-line parameters appended to the
+  list of command-line parameters when the process runs in foreground mode
+* `daemonExtraArgs`. Extra command-line parameters appended to the
+  list of command-line parameters when the process runs in daemon mode
+* `environment`. An attribute set with environment variables
+* `user`. Name of the user we should change to
+* `directory`. The current working directory we should change to before
+  executing the process
+* `nice`. Nice level
+* `umask`. Umask
+* `dependencies`. Specifies the process dependencies: dependencies on other
+  running processes that should be deployed first.
+
+### About foreground processes and daemons
+
+The function abstraction makes a distinction between *foreground processes* and
+*daemons*. Most modern process managers
+(e.g. `systemd`, `launchd`, `supervisord`) have a preference for the former,
+because they can be more reliably managed (i.e. no PID files). However,
+`sysvinit` (which is an LSB standard) and `bsdrc` scripts require processes to
+daemonize by themselves, otherwise the scripts will block.
+
+Most services have the option to support both deployment scenarios. For an
+optimal user experience, it is also best to support both. However, this is not a
+strict requirement -- if a service is only offered as a foreground process then 
+it can still be daemonized by calling an external tool, such as:
+[libslack's daemon](https://libslack.org/daemon).
+
+The inverse scenario is also possible -- if a service always daemonizes, you can
+add a substitute proxy that runs in foreground mode that watches the lifecycle
+of the daemon.
+
+In the Nix process management framework, the `createManagedProcess` abstraction
+will automatically use a simulation strategy when a deployment scenario is
+unspecified. For example, if you only specify a `foregroundProcess` then it will
+transparently call the `daemon` tool if a process manager requires a process to
+daemonize by itself.
+
+### Process dependencies
+
+Another interesting concept is the notion process dependencies (`dependencies`
+parameter). Sometimes, a service requires the presence of another service to
+work reliably. For example, if you have a PHP application deployed to Apache
+that uses a MariaDB database for data storage, then MariaDB needs to be
+activated before Apache.
+
+The `dependencies` parameter can be used to automatically generate a
+configuration in which the order is preserved. For example, when generating
+a `systemd` configuration, this property can be translated into `Wants=` and
+`Requires=` directives. For `sysvinit` we can derived sequence numbers.
+
+In NixOS, there is no high-level concept of process dependencies. It is still
+possible to directly control the activation order by modifying systemd's `Wants=`
+and `Requires=` directives through overrides, but this is not really
+user-friendly.
+
+### Process manager-specific overrides
+
+Obviously, it is impossible to standardize all properties of all process
+managers. Sometimes, it may still be desired to specify properties that are
+unique to certain process managers.
+
+For example, `sysvinit` has the notion of run levels that are not used by other
+process managers. Specifying runlevels for `sysvinit` scripts can be done by
+defining a process-manager specific override:
+
+```nix
+{createManagedProcess, nginx, stateDir}:
+{configFile}:
+
+let
+  nginxLogDir = "${stateDir}/${instanceName}/logs";
+in
+createManagedProcess {
+  name = "nginx";
+  description = "Nginx";
+  initialize = ''
+    mkdir -p ${nginxLogDir}
+  '';
+  process = "${nginx}/bin/nginx";
+  args = [ "-c" configFile" -p" "${stateDir}/${instanceName}" ];
+
+  overrides = {
+    sysvinit = {
+      runlevels = [ 3 4 5 ];
+    };
+  };
+}
+```
+
+In the above example, the `overrides.sysvinit` property specifies
+`sysvinit`-only options. We use the override to define in which runlevels the
+script should be started.
+
+We can also use overrides for other process managers. For example, if we want
+to use the readiness-check property of systemd (which is not standardized) or
+Linux namespaces and cgroups, we can define a `systemd`-specific override.
+
+## Writing services as units of instantiation rather than singleton objects
+
+Compared to NixOS and other solutions, such as nix-darwin and system-manager,
+another feature of the Nix process management framework is that services are
+units of *instantiation* rather than *singleton* objects -- in NixOS, services
+are bound to modules. As a result, services are singleton objects that only have
+a single implementation and a single configuration.
+
+For many kinds of services (e.g. PostgreSQL, Apache HTTP server etc), it
+typically suffices to only have a single instance running on a machine. For
+microservices, it may actually be quite useful to have the ability to easily
+create multiple instances. For testing/experimentation purposes it may also be
+desired to run multiple instances of a system service, such as a web server.
+
+It is also not easily possible in NixOS to switch to alternative implementations
+of a service. For many services, alternative implementations (e.g. `cron`,
+`ssh`) are not required, but some services accept complex configurations, such
+as the Apache HTTP server and Apache Tomcat.
+
+Finding a universal abstraction in NixOS to facilitate all possible use cases
+may be impossible. Moreover, it may also be desired for developers to write
+their own abstractions on top of existing functionality.
+
+To support abstractions and multiple instances, the Nix process management
+framework does not use the NixOS module system, but a convention that is based
+on the organisation of "ordinary" Nix packages in the Nixpkgs repository.
+
+In Nixpkgs, most package follow the convention that a separate Nix file defines
+a function in which the function header refers to the build-time dependencies to
+construct a package from source. Packages are composed in the top-level Nix
+expression (`all-packages.nix`) that invokes these functions with the required
+build-time parameters. For most packages, only a single variant is composed, but
+it is also possible to build different kinds of package variants by changing the
+function parameters.
+
+The Nix process management framework follows the same principles and extends the
+Nixpkgs convention with a number of additional concepts:
+
+* Every service is instantiated from a *constructor* which is a nested function:
+  the outer function refers to the build-time dependencies that are required to
+  build the service from source code.
+* The inner function refers to the service *instance parameters* -- calling the
+  constructor with different parameters makes it possible to compose a service
+  with an alternative configuration and in such a way that they do not conflict
+  with other instances.
+* The build-time dependencies of constructor functions are composed in a Nix
+  expression that is typically called: `constructors.nix` in which every
+  constructor function is called with its required build-time dependency
+  parameters.
+* The process instances of a system are composed in a Nix expression that is
+  typically called: `processes.nix` in which every constructor function is
+  called with its required instance parameters.
+
+```nix
+{createManagedProcess, lib, postgresql, su, stateDir, runtimeDir}:
+
+{ port ? 5432
+, instanceSuffix ? ""
+, instanceName ? "postgresql${instanceSuffix}"
+, configFile ? null
+, postInstall ? ""
+}:
+
+let
+  postgresqlStateDir = "${stateDir}/db/${instanceName}";
+  dataDir = "${postgresqlStateDir}/data";
+  socketDir = "${runtimeDir}/${instanceName}";
+
+  user = instanceName;
+  group = instanceName;
+in
+createManagedProcess rec {
+  inherit instanceName user postInstall;
+
+  path = [ postgresql su ];
+  initialize = ''
+    mkdir -m0755 -p ${socketDir}
+    mkdir -m0700 -p ${dataDir}
+
+    chown ${user}:${group} ${socketDir}
+    chown ${user}:${group} ${dataDir}
+
+    if [ ! -e "${dataDir}/PG_VERSION" ]
+    then
+        su ${user} -c '${postgresql}/bin/initdb -D ${dataDir} --no-locale'
+    fi
+
+    ${lib.optionalString (configFile != null) ''
+      ln -sfn ${configFile} ${dataDir}/postgresql.conf
+    ''}
+  '';
+
+  foregroundProcess = "${postgresql}/bin/postgres";
+  args = [ "-D" dataDir "-p" port "-k" socketDir ];
+  environment = {
+    PGDATA = dataDir;
+  };
+
+  overrides = {
+    sysvinit = {
+      runlevels = [ 3 4 5 ];
+    };
+  };
+}
+```
+
+The above Nix expression shows a constructor function for PostgreSQL and has the
+following properties:
+
+* The outer function-header (first line) refer to the build inputs of the
+  service. For example, to construct a configuration from a high-level
+  specification for any supported process manager, we propagate the
+  `createManagedProcess` as a parameter.
+* The inner-function header refers to the instance parameters of a PostgreSQL
+  instance. When it is desired to have two instances deployed to the same
+  machine we must make sure that there is no conflict between the state of
+  the two instances. Avoiding a conflict is possible by assigning a unique TCP
+  port number and making sure that the state directories do not conflict. The
+  `instanceName` parameter is appended to the name of the state directories to
+  make this possible.
+* In the body of the inner function, we invoke `createManagedProcess` to
+  agnostically generate a configuration for a supported process manager.
+  In the configuration, we provide an initialization script and we call the
+  `postgres` service with its required command-line parameters and environment
+  variables.
+
+As explained earlier, a constructor function is a nested function that we need to compose
+twice. The build-time dependencies are composed in a Nix expression typically called
+`constructors.nix`:
+
+```nix
+{ nix-processmgmt ? ../../nix-processmgmt
+, pkgs
+, stateDir
+, logDir
+, runtimeDir
+, cacheDir
+, spoolDir
+, libDir
+, tmpDir
+, callingUser ? null
+, callingGroup ? null
+, processManager
+, ids ? {}
+}:
+
+let
+  createManagedProcess = import "${nix-processmgmt}/nixproc/create-managed-process/universal/create-managed-process-universal.nix" {
+    inherit pkgs runtimeDir stateDir logDir tmpDir processManager ids;
+  };
+in
+{
+  postgresql = import ./postgresql {
+    inherit createManagedProcess stateDir runtimeDir;
+    inherit (pkgs) lib postgresql su;
+  };
+}
+```
+
+In the above Nix expression, we invoke the `postgresql` constructor function and
+we propagated its required build-time dependencies, such as the `postgresql`
+package and the `createManagedProcess` abstraction function.
+
+The instance parameters are composed in a Nix expression that is typically called
+`processes.nix`:
+
+```nix
+{ pkgs ? import <nixpkgs> { inherit system; }
+, system ? builtins.currentSystem
+, stateDir ? "/var"
+, runtimeDir ? "${stateDir}/run"
+, logDir ? "${stateDir}/log"
+, spoolDir ? "${stateDir}/spool"
+, cacheDir ? "${stateDir}/cache"
+, libDir ? "${stateDir}/lib"
+, tmpDir ? (if stateDir == "/var" then "/tmp" else "${stateDir}/tmp")
+, processManager
+, nix-processmgmt ? ../../../nix-processmgmt
+}:
+
+let
+  constructors = import ../../services-agnostic/constructors.nix {
+    inherit pkgs stateDir runtimeDir logDir tmpDir cacheDir spoolDir libDir processManager nix-processmgmt;
+  };
+in
+rec {
+  postgresql = rec {
+    port = 5432;
+
+    pkg = constructors.simplePostgresql {
+      inherit port;
+    };
+  };
+
+  postgresql-secondary = rec {
+    port = 5433;
+
+    pkg = constructors.simplePostgresql {
+      inherit port;
+      instanceSuffix = "-secondary";
+    };
+  };
+}
+```
+
+In the above Nix expression, we compose two PostgreSQL instances that run on the
+same machine simultaneously. This is possible due to the fact that they use
+instance parameters that avoid their states to conflict.
+
+The above expression imports the `constructors.nix` expression so that each
+process instance can be composed from the PostgreSQL constructor function.
+
+In NixOS, we use the NixOS module system that makes it possible to define
+multiple aspects of a system in a module (such as packages, config files,
+systemd units and environment variables) and combine the aspects of each module
+into a single configuration from which a complete system is deployed, which is
+very powerful.
+
+Combining the conventions of the nix-processmgmt framework with the NixOS module
+system is something that still needs to be discussed. There are variety of
+options possible, such as:
+
+* Combining the convention with process instances as described earlier by
+  referring to a process instance from a NixOS module, similar to how NixOS
+  modules can also refer to ordinary Nix packages
+* Using sub modules, in which every sub module can define its own variant of a
+  service.
+
+### Unprivileged user deployments
+
+Another interesting feature of Nix is to allow unprivileged users to deploy
+packages. With services, this is also possible, but there are some things we
+need to take into account, such as:
+
+* The state directories must refer to a directory that has writable permissions
+  for the user
+* We cannot bind to a TCP/UDP port below 1024
+
+We can also introduce a configuration flag and configure a service in such a way
+that the above criteria are met:
+
+```nix
+{createManagedProcess, lib, postgresql, su, stateDir, runtimeDir, forceDisableUserChange}:
+
+{ port ? 5432
+, instanceSuffix ? ""
+, instanceName ? "postgresql${instanceSuffix}"
+, configFile ? null
+, postInstall ? ""
+}:
+
+let
+  postgresqlStateDir = "${stateDir}/db/${instanceName}";
+  dataDir = "${postgresqlStateDir}/data";
+  socketDir = "${runtimeDir}/${instanceName}";
+
+  user = instanceName;
+  group = instanceName;
+in
+createManagedProcess rec {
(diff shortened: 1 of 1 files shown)


rfcs/0160-portable-service-layer.md
```diff
@@ -0,0 +1,893 @@
...
+to use the readiness-check property of systemd (which is not standardized) or
+Linux namespaces and cgroups, we can define a `systemd`-specific override.
+
+## Writing services as units of instantiation rather than singleton objects
+
+Compared to NixOS and other solutions, such as nix-darwin and system-manager,
+another feature of the Nix process management framework is that services are
+units of *instantiation* rather than *singleton* objects -- in NixOS, services
+are bound to modules. As a result, services are singleton objects that only have
+a single implementation and a single configuration.
```
infinisil, line 486: I talked to you at Nixcon about exactly this, where I even created [a gist](https://gist.github.com/infinisil/e29ee58a8184eeb7a2e139c2b2d3e4ff) for you to demonstrate how the module system can be used to support instances of services. The module system is slowly becoming the dominant way to write composable and modular Nix code, and I really think it should be used for this too. This will make integration into NixOS easier too.

rfcs/0160-portable-service-layer.md
```diff
@@ -0,0 +1,893 @@
...
+making sure that state directories and port numbers do not conflict through
+function parameters.
+
+It is also possible to isolate services by using concepts such as Linux
+namespaces. Then it is not required to avoid certain kinds of conflicts.
+Unfortunately, using these features is restricted to Linux only -- this RFC aims
+to be operating system agnostic.
+
+# Prior art
+[prior-art]: #prior-art
```
infinisil, line 849: I'm surprised for you to not mention your own [Disnix](https://github.com/svanderburg/disnix)/[Dysnomia](https://github.com/svanderburg/dysnomia) projects. It's clear that that's where this RFC comes from.

rfcs/0160-portable-service-layer.md
```diff
@@ -0,0 +1,893 @@
...
+Aside from the first option, the biggest drawback is that these solutions are
+not generic and there is only little reuse possible between service
+configurations.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+The integration with NixOS module system still needs to be worked out in detail.
+Currently, I can think of two scenarios. For this a discussion in the Nix
+community is required.
```
infinisil, line 885: I am in full support of this general idea. However this RFC feels just like a blog post as is, it needs to be much more concrete about how this idea should be implemented.

infinisil, line 885: This proposal might combine very well with the new `pkgs/by-name` directory introduced with [RFC 140](https://github.com/NixOS/rfcs/pull/140), because it could be extended to support a standard `service.nix` file to specify a standardised service definition for a package, without having to be tied to NixOS.

infinisil (changes requested):

infinisil:
PSA: RFC numbers should match the PR number, so this is RFC 163. I updated the title, but make sure to also update the file path.

svanderburg:
> PSA: RFC numbers should match the PR number, so this is RFC 163. I updated the title, but make sure to also update the file path.

Ah, I wasn't aware of this. I just updated the file

2xsaiko:
The title of this PR should probably be "[RFC 163] Portable Service Layer" or something along those lines, not "Initial commit" :^)

rfcs/0160-portable-service-layer.md
aanderse: @infinisil can you please elaborate on your comment a bit? I'd like to make sure I understand you clearly. I'm not sure if you mean that this `service.nix` file would be an interface (`options`) without implementation (`config`), or something else... and to what extent the separation would go 

I can imagine a `service.nix` which declares a common set of options for a module and then several other files (`systemd.nix`, `supervisord.nix`, `launchd.nix`, etc...) which define different implementations of these options (maybe even declaring some implementation specific options of their own). I love how this could tie into RFC 140!

Or did you have something else in mind?

rfcs/0163-portable-service-layer.md
```diff
@@ -0,0 +1,893 @@
...
+[system-manager](https://github.com/numtide/system-manager)
+that makes it possible to only generate systemd units that can be used outside
+NixOS, but it can only be used on Ubuntu and is still under heavy development).
+
+Another project that makes services management available in combination with
+Nix is [nix-darwin](https://github.com/LnL7/nix-darwin). This project uses
+launchd (on macOS) in combination with Nix to support process management.
+
+To make service deployment accessible to a broader audience, it would be nice to
+create a separation in the Nixpkgs tree for portable systemd services (e.g. a
```
arianvp, line 107: FYI: We already have support for building portable service images https://nixos.org/manual/nixpkgs/stable/#sec-pkgs-portableService though I'd love it to be generalized to using the host's nix store instead of building a completely new image

rfcs/0163-portable-service-layer.md
```diff
@@ -0,0 +1,893 @@
...
+systemd has IMO many good features, and I believe it should remain the default
+option in NixOS.
+
+However, there are also good reasons to sometimes consider a different process
+manager:
+
+* systemd is a Linux-specific tool. If we want to deploy a service on another
+  operating system, we need to pick a different solution that is compatible with
+  that operating system.
+* systemd is glibc-specific. Sometimes, it may also be desired to manage
```
arianvp, line 135: Note that we have a community-maintained set of patches that make systemd run on musl

svanderburg:
> The title of this PR should probably be "[RFC 163] Portable Service Layer" or something along those lines, not "Initial commit" :^)

Indeed. I just updated the commit and PR title

rfcs/0160-portable-service-layer.md
alper: Looking for something like this, I wast pointed to https://devenv.sh/services/ as the most likely current candidate. I think it would make sense to add that to prior art as well.

rfcs/0160-portable-service-layer.md
svanderburg: I'm going to add examples for the other two scenarios as well. Indeed, the gist that you gave me is a way to use the modular-style for multi-instance purposes.

I'd also like to explore the third option that I gave -- the hybrid approach between the functional style and the NixOS module system.

I'd like to show all three scenarios, because each of them have pros and cons. The benefit of the functional/pkgs style is that it is less memory consuming and components are configured on a need-to-know basis. The downside is obviously that you can only define one system aspect and not combine multiple aspects into one (e.g. such as adding programs to the PATH etc). Also, the NixOS module system has a "type system" which the functional style doesn't use.

The NixOS modular style is beneficial because it does fit well in the current conventions, it can be used to combine multiple system aspects into one and use of the type system to prevent certain kinds of errors. One of its major drawbacks is memory consumption/evaluation time.

The hybrid approach is something I still need to explore. Maybe it can be good a trade-off between the convenience of combining multiple aspects into one and the need-to-know properties of the functional style.

I'd like to see that we consciously pick the most suitable option out of these three.

Exploring this may take some time, maybe I'll ask you for some help @infinisil

rfcs/0160-portable-service-layer.md
svanderburg: Maybe that's true. On the other hand, the challenge is that the Nix process management is addressing multiple concerns at the same time that can be implemented in multiple ways. Choosing the implementation strategy is also something I'd like discuss. A notable example is the style in which process instances are specified.

For that reason, I think it's good to also provide a bit of context rather than only an implementation strategy, because this still needs to be decided.

Atemu:
(The rendered link is still wrong.)

rfcs/0163-portable-service-layer.md
```diff
@@ -0,0 +1,893 @@
...
+system's package configuration.
+
+For Nix, it is *not* even required to have *super-user privileges* to deploy
+packages -- thanks to its isolation properties this can be done without
+interfering with the packages of another user or the system-wide set of packages.
+
+Moreover, the Nix package manager is also a *portable* solution. In addition
+to Linux, Nix also has first-class support for macOS/Darwin. With some effort,
+it can also be used on other UNIX-like operating systems, such as FreeBSD and
+Cygwin.
```
AndersonTorres, line 39: This effort should be better quantified.
Nix repo does have CI jobs only for MacOS and (Ubuntu) Linux only. We don't have a precise measure about how hard is to support FreeBSD or Cygwin.

Further, I am not aware of "portability policies" adopted by Nix source code repo.
Certainly we have the ongoing (and very advanced, I dare to say) https://github.com/NixOS/rfcs/pull/132, but before this I have seen no such attempt of portability _as a policy_.

Even further, as far as I can remember, the default install of Nix requires some, ehr,  post-install service management. And even among Linux distros it is not very portable.

Atemu (changes requested):
I am absolutely in favour of achieving this RFCs goal but there are two non-negotiable aspects about this topic that the RFC must address IMO:

### It must stay compatible with NixOS module-style configuration.

If NixOS has some service defined, the user must be able to control its behaviour in every detail via their own module(s) using standard module functions such as `mkMerge`, `mkForce` and the like. In other words: It must stay composable.

### It cannot come at the cost of unreasonable complexity or loss of functionality to NixOS maintainers.

If I, as a maintainer for some NixOS module, do not care for other init systems but systemd, I must be able to write the module in such a way that it is only compatible with systemd. Additionally, if another person who does care for other inits comes along and makes it compatible, I (and, by extension, the module's users) must still be able to configure every aspect of the systemd service without limitations.  
Abstractions that make certain common things such as the service's exectuable portable are welcome and (IMO) do not constitute unreasonable complexity but I must still be able to use e.g. systemd's ExecStart with its exact semantics without having to think about other inits' equivalents should there not be an abstract interface for this.

Requiring NixOS maintainers to muck with bash scripts or whatever in order to implement mechanisms for which systemd has a standardised interface would be an absolute no-go for me.

Related to this: If we want alternative inits to have a chance, I think it'd also be a good idea to be able to explicitly declare a service as incompatible. If my module makes use of an init-specific feature and the service's execution is dependant on it, I need to be able to tell the user that my module won't work for them at eval time.  
This also goes the other way: If at some point service modules are implemented by non-systemd users making use of non-systemd init features, they must be able to declare their modules incompatible with systemd. (This would also be useful for modules that only make sense on a specific platform.)

If you can achieve that, I'm all for it.

AndersonTorres (commented):
Since I am very interested in getting rid of systemd dependency, I am very excited about this RFC!

svanderburg:
> (The rendered link is still wrong.)

Ah yes. I've just fixed it!

rfcs/0160-portable-service-layer.md
roberth: > I talked to you at Nixcon about exactly this [...]

... too, but I did not write a gist. My suggestion was to use `imports`.

After all, anything you can do by calling an arbitrary function, you can do by _importing into a submodule_.

Using it might look something like

```nix
# NixOS configuration.nix
{ config, lib, serviceModules, ... }:
{
  services.app-db = {
    imports = [ serviceModules.postgresql ];
    listenPort = 15432;
  };
}
```



> One of its major drawbacks is memory consumption/evaluation time.

That's mostly a scaling problem with the current (anti?)pattern of using a massive module list that's always loaded.
Using imports as above, that's solved.

A constant factor overhead does remain, but it's a very small price worth paying for a few service definitions, as opposed to, say, large package sets.

rfcs/0160-portable-service-layer.md
roberth: Or kubenix, or arion (which piggybacks on NixOS instead of rewriting the service library again). Maybe more?
You might want to draw the line. They're just custom service managers for which this RFC proposes to create a common interface. If someone has a repo for services on S6, would that change the design? Probably not.

rfcs/0163-portable-service-layer.md
roberth: > portability _as a policy_.

It's a policy-free package manager ;)

This kind of thing is more a matter of resources, prioritization (usually by those who have the resources), and a hint of perseverence.

Recently we did merge https://github.com/NixOS/nix/pull/8887 but that's really just a starting point, an empty space for others to contribute and nowhere near a commitment or policy. If someone can seriously help out, please do.

CMCDragonkai:
Does this RFC cover dynamic service orchestration or is that our of scope?

rfcs/0160-portable-service-layer.md
svanderburg: I've expanded this section a bit. The reason why I didn't add my own projects is because they are only loosely coupled to process management -- in its current state, whenever Disnix needs to manage processes it invokes the Nix process management framework for that purpose.

I have also added `devenv`.

rfcs/0163-portable-service-layer.md
svanderburg: Nix does not have an official policy to support other platforms than Linux and macOS/Darwin, but the reason is just simply because of a lack of resources: we don't have manpower and resources (e.g. Hydra build machines) to effectively maintain other ports. On the other hand, Nix does not restrict other platforms and accepts contributions for other platforms.

I myself, for example, have contributed fixes for exotic platforms in several occasions.

If the user base for other platforms grows, then there is little reason to ignore them.

From a historic perspective -- in the early days e.g. 2009-2012 we had all kinds of exotic build machines in our Hydra cluster at Delft University of Technology, such as `FreeBSD`, `OpenBSD` and `OpenSolaris`. Nix could be used on all of these operating systems. Moreover, in the `stdenv/` directory of Nixpkgs you can still find traces of code that makes it possible to build stuff on such platforms.

AndersonTorres:
> If I, as a maintainer for some NixOS module, do not care for other init systems but systemd, I must be able to write the module in such a way that it is only compatible with systemd. 

A mandatory non-portable module? Why?

> I must still be able to use e.g. systemd's ExecStart with its exact semantics

You want access to systemd the same way the keyword `asm` provides access to the assembler?

Also, how portable is that systemd interface?

svanderburg:
> Does this RFC cover dynamic service orchestration or is that our of scope?

What do you exactly mean with "dynamic"? Is it about deploying processes that are not part of a NixOS system profile?

rfcs/0160-portable-service-layer.md
Ericson2314: I think I prefer something closer to @roberth's, because explicitly collecting all the instances in one place is also not terribly modular. (Feels to SQL like me, with out enough type vs table distinction.)

CMCDragonkai:
> > Does this RFC cover dynamic service orchestration or is that our of scope?
> 
> What do you exactly mean with "dynamic"? Is it about deploying processes that are not part of a NixOS system profile?

I like to think of first-class orchestrators. One can imagine this system as an orchestrator. If this system deploys an orchestrator as a service, then that results in some level of dynamic orchestration.

DrRuhe:
To add a bit of other "prior art", I'd like to mention[ a project I've been working on recently](https://github.com/DrRuhe/nix2dream), which extends the [drv-parts](https://github.com/DavHau/drv-parts)(Now part of [dream2nix](https://github.com/nix-community/dream2nix)) to define a generic service layer with the module system which can be rendered into configurations for multiple process managers. The project is still in development, but the core module api is pretty much finnished. It allows defining multiple services together in a straightforward way. See [this example]([source](https://github.com/DrRuhe/nix2dream/blob/main/examples/deployments/hello.nix#L9C5-L32C7)):

```nix
deployments.hello.services = {
      hello = {
        imports = [
          ../packages/hello.nix #use the hello package as this service
        ];

        service = {
          args = [
            "-g"
            ''"Hello there, $GENERAL!"''
          ];
          env.GENERAL = "General Kenobi";
        };
      };
      other-service = {
        imports = [
          ../packages/hello.nix
        ];
        service.args = [
          "-g"
          ''"Hello from the other service!"''
        ];
      };
    };
```

Note that this setup could also easily be used without drv-parts/dream2nix, by requiring services to set a "mainProgram" option instead.

rfcs/0163-portable-service-layer.md
```diff
@@ -872,9 +872,19 @@ solution:
...
+  with a variety container-friendly/non-PID-1 based process managers, such as
+  process-compose, overmind, and honcho.
+* [Dysnomia](https://github.com/svanderburg/dysnomia) is a system to manage the
+  life-cycle of services deployed by
+  [Disnix](https://github.com/svanderburg/disnix) in a generic way. It has its
+  own `process` module that deploys services as daemons running the background.
+  It can also use the Nix process management framework to deploy services that
+  are managed by any kind process manager supported by the framework.
+
+Aside from the first and last option, the biggest drawback is that these
```
AndersonTorres, line 886: Why not group the "best options"  and say "Aside from the last two options..."?

rfcs/0163-portable-service-layer.md
```diff
@@ -0,0 +1,903 @@
...
+  (of different services) in a Nix-based development environment. It works a
+  with a variety container-friendly/non-PID-1 based process managers, such as
+  process-compose, overmind, and honcho.
+* [Dysnomia](https://github.com/svanderburg/dysnomia) is a system to manage the
+  life-cycle of services deployed by
+  [Disnix](https://github.com/svanderburg/disnix) in a generic way. It has its
+  own `process` module that deploys services as daemons running the background.
+  It can also use the Nix process management framework to deploy services that
+  are managed by any kind process manager supported by the framework.
+
```
Aleksanaa, line 885:
```markdown
* [NixNG](https://github.com/nix-community/NixNG/tree/master) is considered a late sibling to NixOS and shares many designs of it. It also implements the function of generating services for several kinds of init systems from the same nix expressions.
```

rfcs/0160-portable-service-layer.md
8573: > Feels to SQL like me, with out enough type vs table distinction.

Do you have a link to a page that expands on this issue? I at least am not familiar enough with SQL (or its intersection with type theory) to understand this remark.

rfcs/0163-portable-service-layer.md
AndersonTorres: Try to use the suggestion feature.

rfcs/0160-portable-service-layer.md
infinisil: Pretty much something like that yes. One of the potential future ideas with `pkgs/by-name` was to use it not only for package builds, but also the package-specific modules. So in addition to the current `package.nix` file in each directory, there could be a `module.nix` file or so too. This would then also be automatically discovered in some way and added to NixOS (or as here, more generic process management).

The reason I'm suggesting it here is because I imagine we need some kind of migration from the current NixOS modules to more generic modules, and combining that with a transition to `pkgs/by-name` could work nicely. I think of `pkgs/by-name` as a sort of strictly type-checked part of Nixpkgs, where CI prevents you from e.g. committing `package.nix` files that aren't packages. Similarly it could check that `module.nix` is a module with specific options or so.

rfcs/0163-portable-service-layer.md
AndersonTorres: > It's a policy-free package manager ;)

Technically, there is RFC #46 for Nixpkgs. But I understood the point.

rfcs/0160-portable-service-layer.md
aanderse: I love it! Thanks for confirming?

AndersonTorres: > a standardised service definition

In other words, an API?

roberth: > This would then also be automatically discovered in some way and added to NixOS

We don't even have a module based implementation yet, so it's way too soon to start automating and imposing restrictions. This is a very different problem.
I'm all for grouping files by functionality instead of technical type, but easy on the rules please.



> an API?

An interface.

"Application programming" just muddies the waters. It's 3 syllables either way.

infinisil: Then it sounds like this RFC should be a draft instead

ehmry:
Excellent. Immediate benefits aside this kind of abstraction would be immensely helpful for OS research.

I'm wondering how far the scope of "services" extends above the system-level. Could graphical desktop applications be included? That's a big scope-creep but I imagine there is more incentive to write fresh desktop services than to migrate NixOS services that already work fine where they are.

I run [Synit](https://synit.org/) as a second-level init system and I can't be bothered to rewrite the systemd services from NixOS, but I do write services from-scratch for my desktop environment.

rfcs/0160-portable-service-layer.md
Ericson2314: I do not, but perhaps i just explain it. Suppose in some language I declare a type (`struct`, whatever) called `Foo`. I can't express "give me all the `Foo`s", I have to find them myself. But in SQL I can.

roberth: We could call this "enumerability" of instances. In the end, for the configuration to be buildable, all services have to be enumerated anyway, so I guess this is a bit about how it "feels" more than what is fundamentally possible. 
Although I think we could make this idea work by talking about the enumerability of service "subtypes" where a subtype is for example a postgresql instance (ie not talking about the other dimension, which is different supervisors).

Whether making enumeration "hard" is a good thing, I'm not sure about. For instance, it will be somewhat harder to configure a database management thing for all local databases.
Though arguably that's not a question of configuring services, but rather a constellation of services that interact. How to do this would tend to involve the specifics of the "medium" in which these services live, whether that's a NixOS configuration, or a NixOps network, or something more dynamic, etc.
Solving these as part of the (individual) service abstraction may well be a mistake. Instead, we'd have to abstract over larger systems, which seems like a different proposition altogether.

I think it'd be fine to solve reuse of fairly uninvolved service definitions before talking more about modeling larger systems of services.

AndersonTorres:
> Could graphical desktop applications be included?

It looks very natural, since Home-Manager has support for X11 window managers.

AndersonTorres:
I have a plan to assimilate [Home Manager](https://github.com/nix-community/home-manager) to Nixpkgs repo.
Can your RFC help this transition, since it would provide a portable service layer to HM too?

rfcs/0163-portable-service-layer.md
```diff
@@ -0,0 +1,925 @@
...
+The `dependencies` parameter can be used to automatically generate a
+configuration in which the order is preserved. For example, when generating
+a `systemd` configuration, this property can be translated into `Wants=` and
+`Requires=` directives. For `sysvinit` we can derived sequence numbers.
+
+In NixOS, there is no high-level concept of process dependencies. It is still
+possible to directly control the activation order by modifying systemd's `Wants=`
+and `Requires=` directives through overrides, but this is not really
+user-friendly.
```
Animeshz, line 436: Home Manager has ability to [create a DAG](https://github.com/nix-community/home-manager/blob/master/modules/lib/dag.nix) (also the [Manual](https://nix-community.github.io/home-manager/#sec-option-types)), such concept maybe useful to some extent to specify dependency and reverse-dependency of the services at a high-level.

Animeshz: Nixpkgs also have [some similar module for strings](https://github.com/NixOS/nixpkgs/blob/master/lib/strings-with-deps.nix) which is not as generic as home-manager has, but it is used for ordering the system activation script.

7c6f434c:
>> If I, as a maintainer for some NixOS module, do not care for other init systems but systemd, I must be able to write the module in such a way that it is only compatible with systemd.
>
>A mandatory non-portable module? Why?

I assume the point is «I want some weird kind of dependency that I barely managed to figure out with systemd and don't know to express in anything more generic»

rfcs/0160-portable-service-layer.md
7c6f434c: > I think it'd be fine to solve reuse of fairly uninvolved service definitions before talking more about modeling larger systems of services.

I'd say that packages with overrides in Nixpkgs are simpler to reason about than modules in NixOS, so it would be nice to have the lower implementation level to be functions, and the module-based whole function composition sit on top.

(Would also allow slightly less horribly slow to evaluate implementations of whole-system composition to share more code with NixOS)

aanderse: > Using it might look something like

I understand at a high level what you're saying but I'm not sure it's immediately clear to everyone the _details_ of what this would actually look like. I don't understand, at least.

Maybe you can lead me in the right direction. For simplicity and so that we're all clear let's pretend we're implementing this system and we happen to be implementing a `systemd` based backend - again, for simplicity and clarity using conventions we're all used to.

Back to your example. You defined an option `listenPort` in your submodule and I understand what this would look like. What I don't understand is what the implementation of this would look like - what configuration would you merge into `config` based on any `options`, like `listenPort`? In NixOS merging configuration into `config` works so well because of a number of high level abstractions defined in other modules, like `users.users`, `users.groups`, `environment.etc`, `environment.systemPackages`, and `systemd` to have a few of the most common and valuable.

In your example I don't understand how we build anything useful without those abstractions, or if we have those abstractions where they come from... unless you say each submodule `import`s everything it needs, like `systemd.nix`, `users.nix`, `environment.nix`, etc... Even if you say each submodule needs to `import` what it needs I could still use some clarification, though. Doesn't this mean a bit of redundant evaluation because multiple modules are `import`ing the same dependencies (like `systemd.nix`) over and over? If so, I guess these `import`ed files would have to be carefully designed to be used in submodules.

Any clarification you can provide would be greatly appreciated.

rfcs/0163-portable-service-layer.md
infinisil: I also have [a PR](https://github.com/NixOS/nixpkgs/pull/97392) to add the DAG stuff to Nixpkgs

Atemu:
> A mandatory non-portable module? Why?

As I elaborated in my initial comment, this is in order to preserve the ability to use every feature of systemd with its exact semantics as we currently can.

I'm not saying this should be the default or even desirable but it should be an option; a valid one at that.

> You want access to systemd the same way the keyword `asm` provides access to the assembler?

Yes. In most cases you shouldn't need it and I too would prefer to use more abstract interfaces but we must also preserve the ability to use the "bare metal" access.

Niche process manager support cannot come at significant burden for everyone else.

> Also, how portable is that systemd interface?

What do you mean by "portable"? Systemd only supports Linux currently.

As a NixOS user and maintainer, I should not be forced to care about non-systemd+linux platforms. I should have the option to but not the obligation.

> I want some weird kind of dependency that I barely managed to figure out with systemd and don't know to express in anything more generic

The problem isn't just knowledge, it's having to care; if I cared, I'd know.  
I don't want to have to care about the specifics of 6 different other service managers that were all replaced by a (IMO) superior option that is in use today (status quo).

Also, forcing the use of generic interfaces must come at the cost of a loss in configurability as the generic interface cannot be more than the intersection of all process managers' features. If that generic interface was the only option, there'd be no reason to use systemd anymore because none of its desirable features are supported by the legacy process managers it replaced.

Therefore, as a NixOS user and maintainer, I must be able to use non-generic/specific interfaces if I need to.  
(That should ideally work the other way aswell; if some process manager provides an interface that systemd does not, users of said process manager should also be able to use it.)

AndersonTorres:
> Niche process manager support cannot come at significant burden for everyone else.
> 

You are precisely defending the opposite. Systemd is the niche process manager here (even if the hugest one).

**Generalized** process manager support cannot come at significant burden for everyone else that wants to use their niche process managers.

> > Also, how portable is that systemd interface?
> 
> What do you mean by "portable"? Systemd only supports Linux currently.
> 

Besides not being useful outside Linux, I was talking about the systemd as an abstraction.

-------------------------------

In this sense I believe it is a good compromise. A general opt-in interface plus the possibility of writing raw files for direct use. AFAICK Nix provides this already.

rfcs/0163-portable-service-layer.md
```diff
@@ -0,0 +1,925 @@
...
+  running processes that should be deployed first.
+
+### About foreground processes and daemons
+
+The function abstraction makes a distinction between *foreground processes* and
+*daemons*. Most modern process managers
+(e.g. `systemd`, `launchd`, `supervisord`) have a preference for the former,
+because they can be more reliably managed (i.e. no PID files). However,
+`sysvinit` (which is an LSB standard) and `bsdrc` scripts require processes to
+daemonize by themselves, otherwise the scripts will block.
```
SuperSandro2000, line 402: This speaks big for modern process managers. With systemd processes cannot that easily become rogue and linger around in your system. For a small embedded system like my router that is not a problem but on one of my servers PIDs are already above the 4M mark and modern software just solved this problem. I do not think it is time efficient to ignore those conveniences.

SuperSandro2000 (commented):
I appreciate and like the idea but adding extra layers of abstractions even for NixOS is not a good looking sign. Also that all alternatives to systemd lack major features of it is really worrying me.
Also you completely ignored some of systemd's features like service hardening, sandboxing, tmpfiles.d, journald, networkd.
Reimplementing that all in shell script is probably possible but at what cost?

PS: I would appreciate a more accessible writing style like one sentence per line.

Animeshz:
> Also you completely ignored some of systemd's features like service hardening, sandboxing, tmpfiles.d, journald, networkd.
Reimplementing that all in shell script is probably possible but at what cost?

Why to reimplement anything? Where are these really used that are really important? Point of other service manager is to reduce what's not important.

Also, I believe having an alternating implementation added allows for future flexibility as well, maybe about another thing, before it deprecates, who knows?

SuperSandro2000:
> Why to reimplement anything? Where are these really used that are really important? Point of other service manager is to reduce what's not important.

Because those features are extensible used in NixOS? NixOS is for example one of the frontrunners for systemd's LoadCredential, EnvironmentFile and family. Also services are using hardening in various ways like using BindPaths to make service variations easier https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/security/acme/default.nix#L359-L363 . It is also used to give various CAPs to services without giving them full root access.

Also configuring those things with plain flat config files is a perfect match for NixOS.
Also having it for networking a lot better than doing it with scripts. It is way easier to transition between two states cleanly and without complicated conditions on the configures side.

And journald makes logging easier. You no longer need to convince every application where and how to write it's log files and when to rotate them. It automatically rotates files when the disk gets full and compresses them. Of course you can use logrotate but if the file is filled within a short timeframe, your system might just crash because you ran out of disk space.

> Also, I believe having an alternating implementation added allows for future flexibility as well, maybe about another thing, before it deprecates, who knows?

but do we have actual viable alternative implementation that support the features we are using right now? If the alternatives only support the bare minimum of an init system and we need to take care of all the additional features used somehow ourselves then I am not sure if it actually is one.

That wouldn't be such a big deal if we wouldn't use those features but we are using them because they are convenient, save us time and allow us to focus on the actual service instead of worrying about shellscripts that start it.

AndersonTorres:
> Besides not being useful outside Linux, I was talking about the systemd as an abstraction.

Being a bit more precise: 
1. what does Systemd provide that no other similar software stack (like, say, the native NetBSD or FreeBSD or even GNU Shepherd from Guix) can do?
2. how much can we live without those extras?
3. for the particular cases the extras are needed, what does it affect the general case?

rfcs/0160-portable-service-layer.md
aanderse: @roberth maybe i'm a bit dense and this doesn't make automatic sense to me... but i was really hoping you could explain your ideas to me - sounds really interesting.

sorry for the hassle!

aanderse:
> I appreciate and like the idea but adding extra layers of abstractions even for NixOS is not a good looking sign.

personally i agree with that statement and i think we shouldn't add a layer of abstraction on top of service managers - but i definitely think we should add a "portable service layer"... it just depends how you define that

as @infinisil and i discussed [above](https://github.com/NixOS/rfcs/pull/163#discussion_r1348215130):

> I can imagine a `service.nix` which declares a common set of options for a module and then several other files (`systemd.nix`, `supervisord.nix`, `launchd.nix`, etc...) which define different implementations of these options (maybe even declaring some implementation specific options of their own). I love how this could tie into RFC 140!

i think this is the only way to add more without creating a burden on people who only care about `systemd` implementations

rfcs/0163-portable-service-layer.md
```diff
@@ -0,0 +1,925 @@
...
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+This proposal introduces a generic, portable, high-level process management
+framework that makes it possible to manage a variety of commonly used
+application services on preferably all operating systems/environments that the
+Nix package manager supports.
```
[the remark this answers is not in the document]
AndersonTorres: > I don't want this to take presence in NixOS modules to have 5 workarounds for all kinds of distros in every module. 

This is more or less similar to a compiler keeping different backends and frontends. The workarounds should be treated as distro-dependent code.

Aleksanaa:
> Also you completely ignored some of systemd's features like service hardening, sandboxing, tmpfiles.d, journald, networkd.

It's worth mentioning that some systemd-* commands work well as standalone binaries, while some need a small patchset. Artix and Gentoo have been doing this for a long time.

Animeshz:
Ah yes, that reminds me of, eudev and elogind, that has been forked out from systemd as standalone package by gentoo and its even used by void and other distributions.

AndersonTorres:
About this extirpation from systemd:

https://guix.gnu.org/sk/blog/2016/gnome-in-guixsd/

ehmry:
> all alternatives to systemd lack major features of it is really worrying me.

I don't want to use systemd but it really does have the best features of all the init systems available. Something better than systemd is going to come along so we should be ready to pivot rather than have a messy transition with a lot of bad feelings, which happened with systemd and some of the other distros.

Doing the abstraction now avoids fighting and forking later.

AndersonTorres:
> Something better than systemd is going to come along

Some time ago we talked about initware, a portable implementation of systemd.

https://discourse.nixos.org/t/should-we-support-initware-is-a-fork-of-systemd-ported-to-different-bsd-systems/14400

roberth:
Ah, the perpetual question about abstraction.
What's nice about the module system is that abstractions can coexist with their implementation and lower level counterparts within the same configuration system.
Whether a module uses high level or low level options to achieve its goals should be up to the module's maintainers; not a decision that is preempted by others.

In this case, assuming a module-based implementation of the portable service "layer", the portable part could be considered the high level definitions, which can be supplemented with "low level" systemd-specific configuration.

So we can have all three kinds of service modules.
- Service modules that are simply portable.
- Service modules that have systemd-specific optimizations. They configure both portable and systemd-specific options.
- Serivce modules that are non-portable but use some portable options for convenience.
- Non-portable modules that only define systemd-specific options. Some people prefer this, and that's ok. Could be personal preference, or the nature of the service. For instance, a portable `journald` service seems over-engineered.

edolstra:

rfcs/0160-portable-service-layer.md
amjoseph-nixpkgs: > This proposal might combine very well with the new `pkgs/by-name` directory introduced with [RFC 140](https://github.com/NixOS/rfcs/pull/140), because it could be extended to support a standard `service.nix` file to specify a standardised service definition for a package, without having to be tied to NixOS.

There is already at least one project doing exactly this.

rfcs/0160-portable-service-layer.md
amjoseph-nixpkgs: > I even created [a gist](https://gist.github.com/infinisil/e29ee58a8184eeb7a2e139c2b2d3e4ff) for you to demonstrate how the module system can be used to support instances of services.

The fact that there are no examples of this technique actually being used is probably meaningful here.

@svanderburg's point remains: with Nix functions and packages you don't have to do anything special in order to "multi-instance enable" them.  It's automatic.  The gist above requires extra work in order to make a service "multi-instance aware", and comes at a cost to people who only want one instance.

We should strive for a service abstraction that makes multiple instances effortless, the way `/nix/store` makes it effortless to have multiple versions of the same package installed.

rfcs/0163-portable-service-layer.md
amjoseph-nixpkgs: > Technically, there is RFC #46 for Nixpkgs. But I understood the point.

That RFC is about documentation.  Unfortunately the title it chose does not mention that.

KFearsoff (changes requested):
I like the RFC, but I fear it is already ambitious, and will only have more feature creep on top.

I don't think discussing whether support for additional service managers is valuable. I think the discussion we should be having is **what** service managers we want to support, and how.

For me, an easy one that comes to mind is Tier 1 support for systemd, userland systemd and containers. We could have Tier 2 or 3 support for launchd. Everything else sounds fun, but we really can't make any guarantees, I think. If someone wants to write glue to port services to other init systems - they are welcome, but I don't think we have the capacity to provide full-fledged support for those.

I imagine you would specify the service manager in the top-level attributes (something like `service-manager.type = "systemd";`, and then everything would inherit from that, but you can override it case by case basis (something like `services.postgres.runs-on = "container";`.

Then, we'd be able to build NixOS services, NixOS containers, OCI-containers-ran-on-NixOS, Home Manager and Nix Darwin on the same abstract base, getting rid of a lot of inconsistencies and glue throughout them. I feel like that's extremely relevant to the success of this RFC, and what we should spend more time discussing.

Stakeholders in all of the mentioned subsystems welcome to share their input.

Animeshz:
> If someone wants to write glue to port services to other init systems - they are welcome, but I don't think we have the capacity to provide full-fledged support for those.

Its exactly about making the current service layer co-existable with other service manager. It is by no means expecting all the packages to support every service manager.

Quoting back from [roberth's comment](https://github.com/NixOS/rfcs/pull/163#issuecomment-1775011732),
> Whether a module uses high level or low level options to achieve its goals should be up to the module's maintainers; not a decision that is preempted by others.

SuperSandro2000:
> I don't want to use systemd but it really does have the best features of all the init systems available. Something better than systemd is going to come along so we should be ready to pivot rather than have a messy transition with a lot of bad feelings, which happened with systemd and some of the other distros.
> 
> Doing the abstraction now avoids fighting and forking later.

The problem with that is, that we don't know how that future will look like and we could focus abstraction on an area which is later identically and we missed abstraction in an area which could be totally different.

Also if the feature sets we abstract around would be at least somewhat similar then we would have a good basis to start things off but right now it looks more like that the alternatives implement only a tiny fraction of what we are using and lots of things would need workarounds to continue to function.

> Whether a module uses high level or low level options to achieve its goals should be up to the module's maintainers; not a decision that is preempted by others.

Up to some extend. Obviously we shouldn't replicate higher level options with lower level ones that already exist in existing modules no matter that taste of the maintainer.

> Service modules that have systemd-specific optimizations. They configure both portable and systemd-specific options.

Wouldn't this make modules rather complex? eg you configure most things for systemd in the systemd options but not how to start things that needs to be done in that nixos specific abstraction. I'd imagine that a major hurdle which wouldn't make contributing up to standard things a bit harder. We'd probably would need some tooling which could automatically convert those things or automatically populate those fields.

Also I am worried that the module system could slow down. It isn't the fastest right now and some of my machines already take multiple minutes to evaluate and tracking down performance problems in my own config isn't usually that easy to do. Maybe bringing some laziness back into the system and improving overhaul performance by maybe using native types alongside this could improve things to not notice the extra abstraction overhead.

roberth:
> Also I am worried that the module system could slow down.

Unfortunately NixOS is not a good example of how to use the module system in a scalable way.
This service layer is an opportunity to effectively implement #22 minimal module list, because it gets rid of those global and always-loaded `services.foo.enable` options and all the overhead parsing their files brings along.
@infinisil's [gist](https://github.com/NixOS/rfcs/pull/163#discussion_r1347854623) doesn't show how to solve the performance problem, but [mine](https://github.com/NixOS/rfcs/pull/163#discussion_r1348841973) does, because of the explicit `imports`.
Explicit imports could be added to his proposal too, but that may feel like unnecessary duplication.

> Maybe bringing some laziness back into the system

The module system itself is already rather lazy, but it depends on how you use it. Besides the non-scalable pattern of NixOS currently importing each and every service always, there's very little to worry about.

rfcs/0160-portable-service-layer.md
roberth: We might end up using both something like the gist and my example because they solve different problems.

 - @infinisil's [gist](https://gist.github.com/infinisil/e29ee58a8184eeb7a2e139c2b2d3e4ff) operates at the NixOS level, allowing a "constellation" of services to be configured in one go. For example you could have a module that sets up an entire web application with local reverse proxy, database, etc, which would be counterproductive to do in a single systemd service.
 - [my example](https://github.com/NixOS/rfcs/pull/163#discussion_r1348841973) operates within a single service. This solves one of the gripes people have, that a NixOS module can affect *any* aspect of the system - no principle of least authority. In my example, the service module can only affect the system in ways that are explicitly facilitated by the system. For instance it can not add a file to `/etc` or create a user unless a proper NixOS module (ie the service abstraction layer implementation) inspects the service submodules and acts on them as it sees fit.

I imagine a crossover is also possible, where such "constellations" are defined in isolation just like individual services, but are only allotted a certain prefix of unit names for example.

Btw surely someone has come up with a better name than constellation for this.

I'll try to demonstrate the principle with a working example. (no constellations probably. that's mostly just the same pattern applied twice. interesting though. we'll see)

aanderse: > In my example, the service module can only affect the system in ways that are explicitly facilitated by the system. For instance it can not add a file to /etc or create a user unless a proper NixOS module (ie the service abstraction layer implementation) inspects the service submodules and acts on them as it sees fit.

thanks for the explanation! looking forward to seeing your example.

roberth: Ok, did the thing. A bit rushed maybe, but a lot more complete than anything I wrote before this.

- https://github.com/NixOS/nixpkgs/pull/267111

KFearsoff:
roberth put out a PoC. Great job, thank you for your effort!

One thing immediately sticks out to me, though: we need a lot of glue code in Nix in order to bridge the module definition to a specific implementation (such as systemd). I fear this might become an issue: NixOS modules already have messy errors sometimes, and adding a lot of glue on top will exacerbate the problem when writing your own modules.

By the way, can't we translate the current `services.systemd` definitions into abstract definitions? I think it has enough information, and if we are able to pull it off, then the migration to the new system is basically free.

ehmry:
> By the way, can't we translate the current `services.systemd` definitions into abstract definitions? I think it has enough information, and if we are able to pull it off, then the migration to the new system is basically free.

A few people have tried this and it's not been practical. In my experience one can sometime reuse `config.systemd.services.«…».serviceConfig.ExecStart` but not without some manipulation. Something to consider is that systemd security features are opt-in, and an init system designed for security is going to be opt-out for hardening. This means that there is little logical overlap for options and services need to be ported by hand.

Aleksanaa:
> can't we translate the current services.systemd definitions into abstract definitions? 

https://git.sr.ht/~guido/nixos-init-freedom/tree/master/item/systemd-to-s6.nix

However, although it exists as a workaround, it doesn't make sense to follow the step of systemd forever.

KFearsoff:
Makes sense. Let's add those bits of info to the RFC text, cos it's pretty important.

roberth:
I'd like to nominate myself as shepherd.

AndersonTorres:
I have few to no experience on this field.
However I am interested on this because of my next RFC draft.

Recruit Torres, presenting myself as shepherd.

ehmry:
I have experience with unconventional init systems and I have some understanding of nix-processmgmt (https://github.com/svanderburg/nix-processmgmt/pull/5). I want to realize a portable service abstraction but not without a performance regressions.

I nominate myself as a shepherd.

roberth:
> As for my experience with nix, even though I'm <1 yr experienced with nix, I've read the original thesis and following up nix development with NixCon and NixHour.

While I can't assess experience from a distance, I would say that relative inexperience could be an asset to the shepherd team, for more of a fresh perspective on the significant "human interface" aspect of the proposal. :+1:

rfcs/0163-portable-service-layer.md
```diff
@@ -0,0 +1,925 @@
+---
+feature: portable_service_layer
+start-date: 2023-09-10
+author: Sander van der Burg
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
```
lheckemann, line 7:
```suggestion
shepherd-team:  roberth, AndersonTorres, ehmry, Animeshz, IreneKnapp
shepherd-leader: AndersonTorres
```
Thanks for all your nominations!

IreneKnapp:
is there a chat room or anything for the shepherd team? should we set one up?

ehmry:
I would also like to get moving on this. I don't use matrix though so I would ask that a chat be bridged to IRC.

roberth:
I believe it's important for services to be able to include other services as part of their implementation. I'd like this to be possible by means of an `attrsOf submodule`, which automatically takes care of adding a prefix to the final (e.g. systemd) service names. This seems feasible, and somewhat suggests that there's a redundancy between my poc's ability for a portable service to declare e.g. multiple service units. However, I'm not yet entirely sure of that, because of non-service units. Needs more thought.

Motivating use case: typical web apps require a database, and agreeing on the version is somewhat [hard](https://github.com/NixOS/nixpkgs/pull/223064#issuecomment-1887148975) and even if we solve that, we may still want to run more than one version anyway. "Having to" agree on a version is also somewhat counter to the spirit of Nix, where store paths can always coexist rather well.

roberth:
@ehmry did you make "hackint IRC Bridge status" contact me on matrix some time ago, or is that a coincidence? Did you set up a room on IRC?

7c6f434c:
> "Having to" agree on a version is also somewhat counter to the spirit of Nix, where store paths can always coexist rather well.

I'd argue passing an overridable version of a library as an input is pretty similar. If the goal is to have useful layering, maybe this could be the view of the lower level portable-service, with a higher level passing the stuff one way or another.

roberth:
@7c6f434c I agree that's a valid perspective.

The module system provides overriding. For instance, the web app service module could declare an option that contains the postgres service *module* (ie `types.deferredModule`) to use for implementing its postgres dependency. If that option has a default, that means we're either hardcoding implementations or doing some sort of dependency injection.

Also worth considering is the use case where admin wants to use an external database; just a connection configuration. Indeed this seems to lead to a distinction between a low level module (scoped to approximately just the `.service`) and a high level service that is self-sufficient and sets up a database. The high level module could import the low level module and add/"implement" operational dependencies such as a postgres instance.

DrRuhe:
Hi again,

While I did not have the time to make substantial further progress, I had some time to tinker on [nix2dream](https://github.com/DrRuhe/nix2dream) (Which, although sounds similar to - and also currently depends on - [dream2nix](https://github.com/nix-community/dream2nix), is not officially affiliated to dream2nix)).
With that experience, I'd like to nominate myself as a shepherd as well.

IogaMaster:
I would love to see this become a reality, if this becomes merged I will be a user of the runit (or openrc) backend (helping fix services, testing etc)

IreneKnapp:
I went ahead and asked about how to make a channel. It turns out it's easy. On Matrix, join #nix-rfc-163:hackint.org. on Hackint IRC, join #nix-rfc-163. this is for the shepherd team to coordinate, not for public feedback, just to set that expectation, but I'm not gatekeeping who joins or anything.

IreneKnapp:
okay, I sent Matrix invites to everyone who listed a Matrix account in the maintainer list. @svanderburg @ehmry @Animeshz you're the ones not listed there, please join at your convenience and let's start building some momentum on this :D

ehmry:
After playing a bit with nix-processmgmt the missing piece that concerns me most is how to generate documentation from code. In terms of feature parity I don't care about timers and socket activation (everything to implement those is present), I want documentation parity. For me `man configuration.nix` is a killer feature of NixOS.

roberth:
We could do that with the module system. Translating function calls to modules is fairly straightforward, so we could approach it that way. I didn't base my proof of concept on `nix-processmgmt`, but by doing such a translation, we could inherit its knowledge and experience.

> don't care about timers and socket activation (everything to implement those is present)

I don't think you have to care. It will be modular like that. :+1:

jonringer:
Hey shepherds, we would appreciate a status update or some activity. If there are any blockers, please let us know.

If you all are not able to progress on this at the moment, feel free to mark as draft until a better time in the future.

nyabinary:
> Hey shepherds, we would appreciate a status update or some activity. If there are any blockers, please let us know.
> 
> If you all are not able to progress on this at the moment, feel free to mark as draft until a better time in the future.

Is there a Matrix room for discussing the RFC?

Aleksanaa:
> Is there a Matrix room for discussing the RFC?

https://matrix.to/#/#nix-rfc-163:hackint.org

lheckemann:
@AndersonTorres any chance of a shepherd team meeting to get this moving onwards?

deviantsemicolon:
I have some concerns about this. 

Out of everyone who uses Linux, there's a mere fraction of them who don't use systemd. Out of everyone who uses Linux and doesn't use systemd, there's probably an even smaller fraction of them who'd be willing to learn and use NixOS. It seems like a large burden to put on nixpkgs maintainers to satisfy such a small group of users. 

On that note, I'm not a hard no. I'd be curious to hear more about what this would entail development-wise. But I'm concerned about what it would mean for people who use systemd exclusively in terms of functionality. Would we lose functionality by making NixOS more init-agnostic? If not, how would we develop support for that many init systems plus the entirety of systemd with the already small userbase we have?

Animeshz:
@deviantsemicolon I think its been answered a few times and written in the detailed design section as well. Services configuration remains similar for most init systems, its just the templating language feels different. The added abstraction will sufficiently provide generic closure that will transpile into any of the service manager and provide overrides for specific service manager if its really important for the services to depend on such constructs (e.g. runlevels).

na-sa-do:
I would also point out, @deviantsemicolon, that systemd is the only supported init for more or less every major Linux distribution. The number of people using it is unlikely to reflect the number of people who would _choose_ it, given the possibility of switching without moving to a niche and worse-supported distro.

SuperSandro2000:
I cannot imagine an outcome of this RFC where it will not become a burden for most people that are using the default init system. We can already observe this with Darwin or musl support. For Darwin people not owning an expensive Apple device or investing many hours into a hackintosh, have no chance to debug errors and even then errors are often byzantin and require deep apple knowledge (missing headers, linking errors) and even very knowledgeable Linux people often just mark things broken for the platform.
People caring about musl often have to chase after updates and fix musl support afterwards, making it an time expensive endeavor using it. When not doing that for a few months, the very few people maintaining it, sometimes need to fix things very deep down again because software doesn't comply.

I cannot imagine how supporting other init systems will not suffer from the same problems.
It will also probably place code in maintainers modules which they don't test on a daily basis or on updates or maybe don't even want it there and it will also place reviewers into hard spots when updates break things but the person that contributed the other init system is not available to fix the breakage and the author does not have the time to deeply dig into it.

We are also using more and more systemd specific features (hardening, dynamic user) which can break things in subtle ways and where porting is not easy.

And we are moving towards having networkd and systemd in the initrd by default, not because people love systemd so much or want to force people onto it, but because it is so much easier and more compatible and fixes many bugs in more complicated setups which before required unreasonable high complex shell scripts.

Also I don't think the question got answered, when a module maintainer does not want to support other init systems and sees it as a to high burden for him/her/it/... self especially if the module is important.

therealpxc:
Re: maintenance burdens, one thing that is exciting to me about this RFC is that it could be a step towards sharing code between module systems like NixOS and Nix-Darwin, where currently a lot of Nix-Darwin modules are born as copy/pastes of NixOS modules but then maintained separately. We already *have* whole separate module systems for init systems other than systemd (like Nix-Darwin or the new [NixBSD](https://github.com/nixos-bsd/nixbsd), the process management features on devenv.sh, the process-compose flake-parts module, etc.) in the wider community.

I understand you looking out for NixOS/Nixpkgs first, @SuperSandro2000, but I think the opportunities for code sharing here could be really good for the wider Nix world. It could reduce some currently duplicated efforts.

Atemu:
> Is there any reason this would work any differently to how e.g. musl does today?

Making a package work for musl usually involves a patch, special configure flag or disabling some optional feature. Though they may get outdated when maintainers cannot/do not want to test for musl when the packages gets updated, these usually do not get in the way of a maintainer. They can also be removed again quite easily if they break and the maintainer does not want to bother with the support as it's just a bits of code integrated with the regular derivation.  
There is a slight maintenance burden here as the maintainer has to at least think about the bits of code concerning use-cases they don't care about but I'd consider this degree to be reasonable.   

As I understand the current proposal, making a modern NixOS module that makes proper use of systemd features support other service managers would be akin to rewriting the entire derivation to use an entirely different interface where you'd have to manually execute many commands that were previously abstracted away and might not even be able to support the same features of the derivation, just to support musl.  
That would not be reasonable to expect current or future maintainers to support, even if the transition was contributed by someone else and I apply the same line of reasoning to a potential portable service layer.

Atemu:
@therealpxc I don't think anyone is arguing against the benefits of code sharing and abstractions.

roberth:
Meeting notes

Attendees: @Animeshz, @ehmry, @roberth 

- We agree that module system would be good interface
  - Sander, who wasn't in the meeting, would also agree. He and Robert talked at NixCon and FOSSDEM and he's opened the RFC exactly so that the community can improve it and give it a NixOS flavor (paraphrasing)

- RFC would have been more successful if it had an accompanying NixOS PR. It couldn't yet, because the design / interface design is still open.

- Preserving effort already put into the Nixos modules is important.

- roberth *might* be able to reserve some time for work on a non-toy implementation, translating Sander's logic into modules

- emery is using nix-processmgmt with nixos already and is willing to help integrate some form of it in nixpkgs 

- Animesh is also using runit and openrc implementations personally for his own systems i.e. nix on top of other distros, and willing to contribute / integrate some part of it, although does not hold expertise in nix as a language.

- No deadlines or timeline defined.

- Looking for projects that wish to reuse Nixos modules externally and can contribute

- Author of [Synit](https://synit.org/) is interested in basing on Nixos, but not without an abstraction to hook into.

ehmry:
I'll add that it was mentioned that substantial parts of this RFC may be rewritten after we have a serious implementation, and discussion around that will probably happen on nixpkgs/nixos PRs that will reference back to this RFC.

ibizaman:
> Looking for projects that wish to reuse Nixos modules externally and can contribute

Can you elaborate on what exactly does this mean?

I do have a project relying heavily on NixOS modules and actually have been using disnix/dysnomia for that very same project a while ago. I switched since to NixOS modules because I wanted to reuse services from nixpkgs instead.

I’d be happy to help this RFC move along. I personally am seduced by the appeal of easily expressing having multiple instances of a service live side by side.

rfcs/0163-portable-service-layer.md
Ericson2314: Now that we have https://github.com/nixos-bsd, we know a lot more about this!!

Ericson2314:
I would definitely talk to @rhelmot and @artemist about NixBSD, and also @netbsduser the Initware person.

I think approaching this stuff from both ends (generalization of NixOS, dedup of Home Manager, https://github.com/nixos-bsd/nixbsd, etc.) is how we get this done. We can iteratively arrive at a design with all that.

ehmry:
> > Looking for projects that wish to reuse Nixos modules externally and can contribute
>
> Can you elaborate on what exactly does externally mean?

This means projects that are NixOS alternatives that aren't using s******. If you would like to work on making this possible from within the nixpkgs repo there will friendly reviewers to stand behind you.

deviantsemicolon:
To clarify, would this allow people to use init systems other than systemd?

Animeshz:
@deviantsemicolon NixOS is currently highly dependent on systemd, so there's neither an ETA nor plans to make it replaceable, but surely this will fulfil the usages of nix as a portable package manager and build system -- allowing it to integrate nicely over macos, other linux distribution, ci/cd environments and more. Additionally this will serve as first step in modularising the core that it'd benefit if in case it is ever needed to shift from systemd in the future (highly unlikely though, for now).

AndersonTorres:
> To clarify, would this allow people to use init systems other than systemd?

I believe we can be a bit more ambitious and write a distro in this portable service framework.

KFearsoff:
Tangentially related, but we could also specify "storage tiers": ephemeral/persistent/essential/cache. That would make `impermanence` project also a first-class citizen. See also: https://github.com/nix-community/impermanence/issues/10#issuecomment-2198392482

aanderse:
> I believe we can be a bit more ambitious and write a distro in this portable service framework.

@AndersonTorres sounds super fun to prototype - any plans personally to start something like that at some point?

asymmetric:
RFCSC: There hasn't been much activity here in the last few months, has there been activity through other channels? If so, please provide an update here. 

Otherwise, are there any actions you can foresee to take this RFC towards conclusion? If not, feel free to draft this PR.

/cc'ing shepherd team: @AndersonTorres (lead) @roberth @ehmry @Animeshz @IreneKnapp

AndersonTorres:
> > I believe we can be a bit more ambitious and write a distro in this portable service framework.
> 
> @AndersonTorres sounds super fun to prototype - any plans personally to start something like that at some point?

I have only an Emacs instance and a dream :)

I feel that I need to study more things about Nixpkgs itself. But yes I want to try something, especially testing other init systems.

aanderse:
sounds cool - please ping me if you start hacking on something

i am still trying to understand how some of these concepts practically work, though 🤔

let's take the example @roberth provided in https://github.com/NixOS/nixpkgs/pull/267111 - a pragmatic implementation of some ideas Eelco wrote about [a long time ago](https://gist.github.com/edolstra/29ce9d8ea399b703a7023073b0dbc00d#nixos-services) - at first glance the ability to write something like `services.foo = <module here>` looks great and saves us the security nightmare of a flakes world where we "don't want the PostgreSQL module to have the ability to change your X11 configuration"... but after i play that scenario out in my mind i realize nixos is so awesome because postgres can change my x11 configuration... well not specifically that example, but see any module of a web application which uses a database, redis, web server, etc... if we fully commit to essentially containers this works fine, but i am not sure we want that.

anyone have any thoughts on that?

SuperSandro2000:
> Tangentially related, but we could also specify "storage tiers": ephemeral/persistent/essential/cache. That would make `impermanence` project also a first-class citizen. See also: [nix-community/impermanence#10 (comment)](https://github.com/nix-community/impermanence/issues/10#issuecomment-2198392482)

IMO for a first implementation this would be kind of feature creep and even more discouraging people to adapt this.

> * emery is using nix-processmgmt with nixos already and is willing to help integrate some form of it in nixpkgs

Some notes from me after looking a bit on a random file:

- we could probably template this from an higher level option or straight from tmpfiles.d https://github.com/svanderburg/nix-processmgmt/blob/0e9ec77cc633907aa93bf3d758013f6f699f8842/examples/services-agnostic/constructors/nginx/default.nix#L14-L21
- No idea about "runlevels", I am to young for them, https://github.com/svanderburg/nix-processmgmt/blob/0e9ec77cc633907aa93bf3d758013f6f699f8842/examples/services-agnostic/constructors/nginx/default.nix#L42 but in all examples they seem to be 3 4 5, so maybe we can just hardcode them?
- having to handle pid files is very depending on the program. Most new ones don't have that concept anymore and also many do it wrong. We need to abstract that away https://github.com/svanderburg/nix-processmgmt/blob/0e9ec77cc633907aa93bf3d758013f6f699f8842/examples/services-agnostic/constructors/docker/default.nix#L16
- the modules seem smaller because they lack many options and seem to prefer a function argument approach instead of using the module system. Also writters are not used and RFC42 also not.

----

I think we would loose features like easy socket activation.
Also the service managers seem to lack functionality to implement advanced dependency systems like the acme module has, requiring people to probably fall back to polling directories via shell which is exactly where we do not want to go back.

ehmry:
FWIW I'm waiting on the delivery of some hardware before I do an init overhaul down to pid1.

I don't like how long this issue has been open without progress but this is an issue we can't ignore completely because it will continue to be brought up. I think init portability will be a bit like maintaining powerpc support in nixpkgs, we don't do it because many people ask for it, we do it because it demonstrates that our basic portability is sound.

ylh:
> I don't like how long this issue has been open without progress but this is an issue we can't ignore completely because it will continue to be brought up.

yes, what i'm noticing here is much of this thread seems to be less about drafting the rfc per se, and more about the broad issue of abstracting service management, which, hey, fair enough, it's something people evidently care about.

at the risk of perpetuating the trend,

https://github.com/NixOS/rfcs/blob/3798ca1a0946f3fd745b6765c133838b65ae1960/rfcs/0163-portable-service-layer.md?plain=1#L433-L436

it's worth noting that *systemd* has no high-level concept of dependencies. this bit of docs:

    Note that transactions are generated independently of a unit's
    state at runtime, hence, for example, if a start job is requested on an
    already started unit, it will still generate a transaction and wake up any
    inactive dependencies (and cause propagation of other jobs as per the
    defined relationships). This is because the enqueued job is at the time of
    execution compared to the target unit's state and is marked successful and
    complete when both satisfy. However, this job also pulls in other
    dependencies due to the defined relationships and thus leads to, in our
    our example, start jobs for any of those inactive units getting queued as
    well.

serves to illustrate that transactions are related to units by suggestion, nearly by accident. by similar accident, and the collective input of everyone who converged on systemd as a *platform* (or had it converge upon them), polishing the happy path 99% of users tread, this is usually unnoticeable.

despite nixos's occasional moments butting heads with systemd's concepts (https://github.com/NixOS/nixpkgs/issues/135557, https://github.com/NixOS/nixpkgs/issues/81138, etc.), we largely get by fine even when straining systemd's expectations, thanks to the hermetic nature (closure bloat aside) of nix packaging, the ability to sprinkle tests anywhere in evaluation, generation rollbacks as deep down as the grub menu, etc.

https://github.com/NixOS/rfcs/blob/3798ca1a0946f3fd745b6765c133838b65ae1960/rfcs/0163-portable-service-layer.md?plain=1#L428-L431

https://github.com/NixOS/rfcs/blob/3798ca1a0946f3fd745b6765c133838b65ae1960/rfcs/0163-portable-service-layer.md?plain=1#L244-L246

on the complete opposite end, sysvinit offers no more runtime guarantees than any other shell scripts in a trenchcoat, but yes, a boot sequence for it is a fairly trivial function of a partial ordering expressed by dependencies.

so for any sufficiently nice, general definition of a "service" that expresses dependencies and so on: ...accordingly nice, well-designed target platforms to compile down to—that can actually take advantage of all that generality—are thin on the ground. 

at that rate, [services.kubernetes], [virtualisation.oci-containers], and friends join the list of "prior art", which raises a point: many objections in this thread are about portability efforts adding burden to maintaining plain old systemd services. this initiative has the potential to be more qol than bother for nixos proper, though:

1. right now, if you want to move your `systemd.services` a level above bare metal without rewriting them, your (in-tree) options are the native systemd-nspawn container module, or building a vm image. some nixos users would be happy to see that list grow.
2. singleton services, as mentioned, are a pain. putting in my vote for the humble module-level library to aid turning current singleton services instanced, as some have drafted here. such a library would necessarily implement a number of ideas in this rfc. done right it could go over just as uncontroversially as the existence of `lib.mkEnableOption`.
3. further down the line, systemd tmpfiles, for instance, could be abstracted out in places, as i've seen those abused to generally ensure the presence of certain directories, where [impermanence] as mentioned, and, more generally, reifying storage lifetimes, seems more prudent.

what i'm getting at is that after sufficiently many tangentially-related baby steps that make nixos nicer in general, huge swaths of a portable service layer, if not the whole thing, have a chance to fall out for free.

so, if the strongest intersection of opinion is just that having a portable service layer would be Nice™, i could see this rfc getting merged with most of the code samples omitted, as a largely conceptual/advisory piece, to take note of how conducive future nixos changes are to the overall goal, and once a specific implementation feels close at hand, a subsequent rfc detailing it supersedes this one.

_ps._

https://github.com/NixOS/rfcs/blob/3798ca1a0946f3fd745b6765c133838b65ae1960/rfcs/0163-portable-service-layer.md?plain=1#L862-L865

fwiw, home-manager gets by, for the time being, independently specifying launchd and systemd units for a service definition on a can-be-bothered basis, ex. https://github.com/nix-community/home-manager/blob/c2cd2a52e02f1dfa1c88f95abeb89298d46023be/modules/services/nix-gc.nix#L108-L149

thing is, home-manager services tend to be comparatively low-stakes, to the point that taking on an additional channel or flake input for the sake of a user-level service doesn't feel worth it—take it from someone who's running out of memorable 4-digit ports on localhost courtesy solely of nix-darwin.

there is a big difference between i3bar/godoc and nginx/postgres.

[services.kubernetes]: https://nixos.org/manual/nixos/stable/options#opt-services.kubernetes.roles
[virtualisation.oci-containers]: https://nixos.org/manual/nixos/stable/options#opt-virtualisation.oci-containers.containers
[impermanence]: https://github.com/nix-community/impermanence

AndersonTorres:
> FWIW I'm waiting on the delivery of some hardware before I do an init overhaul down to pid1.

Any news on this,  @ehmry ?

ehmry:
@AndersonTorres I've applied for funding and I'm blocked waiting for a response.

asymmetric:
@roberth, @AndersonTorres, @ehmry, @Animeshz, @IreneKnapp: RFCSC: Could you please provide an update on what is blocking this RFC from moving forward?

@svanderburg please update the shepherd team in the RFC's body, as suggested [here](https://github.com/NixOS/rfcs/pull/163/files#r1426607982).

MMesch:
RFCSC:

There seem to have been no significant updates to this RFC for a while. Please provide an update, otherwise we will close it. Note that it can always be reopened if there is time and interest to work on it again.

Ericson2314:
I still hope that NixBSD can revive this as living, breathing, example of what we want on  multiple axes --- but I leave actually doing that that to @artemist and @rhelmot.

roberth:
I'd like to start an alternate implementation by simplifying the proof of concept I wrote earlier and growing it collaboratively from there. That should avoid our collective analysis paralysis, while providing the necessary input for making a decision later, without turning Portable Service Layer into policy.

Besides that, I believe that this RFC text approaches the problem from the wrong angle, and the original author has disengaged through their virtually complete lack of communication.
With those issues in mind, I don't see the value in keeping this particular RFC open, but I'd be glad to be corrected about that.
(This thread could still be used for posting updates to those who are interested)

roberth:
- https://github.com/NixOS/nixpkgs/pull/372170

jzbor:
I have not been following the discussion here very closely, but I think this project would be of relevance to the discussion: https://codeberg.org/amjoseph/infuse.nix

infinisil:
RFCSC: Without activity by the original author and with the general consensus expressed by @roberth's comment, we're now closing this RFC. Feel free to open a new RFC if necessary.

Ericson2314:
Note that Modular Services (as proposed by @roberth [above](https://github.com/NixOS/rfcs/pull/163#issuecomment-2573549371)) has finally landed!

- Tracking issue: https://github.com/NixOS/nixpkgs/issues/428084
- Docs as of `nixos-unstable`: https://nixos.org/manual/nixos/unstable/#modular-services

PR #168, opened 2024-01-14
Simplify language around functions/function arguments.

bzm3r:
According to @hsjobeki it is a common misunderstanding that RFC 145 does not formalize a standard for how function arguments ought to be documented.

There might be two reasons for this:

1. https://github.com/NixOS/rfcs/blob/c65c8321782b40844167beb48818471f70900d9d/rfcs/0145-doc-strings.md?plain=1#L387

2. The use of "lambda formal" or "formal" to refer to "function argument". 

From cursory research 2. might be [Lisp-culture specific](https://people.csail.mit.edu/jaffer/r5rs/Definitions.html). Church referred to these as [arguments](https://compcalc.github.io/public/church/church_calculi_1941.pdf) in his own work. 

The document also already sometimes says argument (especially in reference to `nixdoc`). (Note: it is also not clear what is meant precisely by the lines referred to in point 1, but that could be addressed in a separate PR once clarification has been provided.)

### DIFF
--- a/rfcs/0145-doc-strings.md
+++ b/rfcs/0145-doc-strings.md
@@ -140,7 +140,7 @@ A **documentable node** can be:
 
 - Expression 
 - Binding 
-- Lambda Formal
+- Function Argument
 
 The following rules apply in descending order of precedence:
 
@@ -156,7 +156,7 @@ The following rules apply in descending order of precedence:
 Only whitespaces between the `documentable node` and the `doc-comment`
 
 ```nix
-/**Doc for anonymous lambda function*/
+/**Doc for anonymous function*/
 ↓
 x: x;
 ```
@@ -172,7 +172,7 @@ listToAttrs [
 It is allowed to write the documentation before the attribute instead of placing it right before the body.
 
 ```nix
-/**Doc for lambda function bound to a variable*/
+/**Doc for a function bound to a variable*/
            ↓
 assigned = x: x;
 ```
@@ -230,12 +230,12 @@ in
 # Documentation can still be retrieved.
 ```
 
-#### Lambda formals
+#### Function arguments
 
 ```nix
-/**Doc for the whole lambda function*/
+/**Doc for the whole function*/
 {
- /**Doc for formal 'a'*/
+ /**Doc for argument 'a'*/
  a
 }:
  a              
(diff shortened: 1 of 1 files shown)


rfcs/0145-doc-strings.md
```diff
@@ -140,7 +140,7 @@ A **documentable node** can be:
 
 - Expression 
 - Binding 
-- Lambda Formal
+- Function Argument
```
jtojnar, line 143: This change makes it even more confusing IMO – I think function argument without further qualification usually refers to “(actual) arguments” – i.e. the expression passed to functions. The names/bindings in the function definition are often called “(formal) parameters” or “formal arguments”. This variant is not specific to Lisp as it is used e.g. in C++ standard. See also https://en.wikipedia.org/wiki/Parameter_(computer_programming)

kevincox: I am happy with "Function parameters". This seems better and is on the correct side of your definition.

hsjobeki:
During the RFC most of us agreed on the term `lambda formal`.

Nix knows two kind of functions. Remember every function can only ever take one parameter. The term `parameter` might again be an alias for `argument`.

Now naming the following an `argument`. Which CANNOT have a doc-comment.

```nix
# argument a
    ↓
f = a: 1; 
```

If we want to write documentation for `a` we can still do that.

```nix
/** 
Documentation for the function `f`

# Arguments

- [a]: Any ignored argument
 
*/
↓
f = a: 1; 
```

The the thing that we called `lambda formal`. Which CAN have an assigned doc-comment

```nix
# formal a
    ↓
f = {a}: 1; 
```
with a doc-comment
```nix
f = {
/**Document formal a*/
a
}: 1;
```

Can we together answer the question; What `a`  inside `{a}:` represents? 

Problem in mind: 'nix' doesnt have a language specification. We should refer to terms that `A` might be used in c++ nix. But `B` are understood by the majority of the community.


`A`: A term that is used in c++ nix is `Formal`which refers to the following struct
https://github.com/NixOS/nix/blob/cbd5553d57ebf5d0532047165a2d81825424bd76/src/libexpr/nixexpr.hh#L292

`B`: Which term is the best alternative?

PR #169, opened 2024-01-21
Feature parameter names

KAction:
RFC to establish rules on how `enable/with/support/use` flags should be named. Essentially, an attempt to bring order and uniformity of Gentoo USE flags into Nixpkgs.

[Pre-RFC discussion on the discourse](https://discourse.nixos.org/t/pre-rfc-standardization-of-feature-parameters/38104)
[Rendered RFC ](https://github.com/KAction/rfcs/blob/feature-flags/rfcs/0168-feature-parameter-names.md)


rfcs/0168-feature-parameter-names.md
```diff
@@ -0,0 +1,559 @@
...
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Nixpkgs usually exports compile-time options of the upstream build system
```
7c6f434c, line 14: Maybe something more specific than «usually» should be said, as the majority of packages exposes no options (which happens sometimes because of upstream not having feature flags and sometimes because the maintainers do not see the benefits as sufficient to justify the effort).

rfcs/0168-feature-parameter-names.md
```diff
@@ -0,0 +1,559 @@
...
+* In existing packages, maintainers MAY start to expose feature parameters and MAY cease exposing them.
+* Further contributions MAY change their package to be compliant with the RFC
+* Further contributions MUST NOT rename non-compliant feature parameters to another non-compliant name.
+* Further contributions MUST NOT rename compliant feature parameters to non-compliant parameter.
+* Further contributions MAY rename compliant feature parameters to another compliant name
+  for the purposes of consistency with other packages.
+* Further contributions that improve compliance with this RFC MAY involve
+  maintainers but SHOULD NOT be required to do so because the rules SHOULD be
+  clear.
+* New packages MUST be compliant with the RFC (either no feature parameters or compliant with this RFC.)
```
7c6f434c, line 192: Does the count starts from RFC passing, from Nixpkgs documentation updates being finalised, from the linter (with a working inhibition mechanism) being available, from the linter being deployed in CI?

rfcs/0168-feature-parameter-names.md
```diff
@@ -0,0 +1,559 @@
...
+with implicit caveat "provided that package maintainer decided to export
+underlying build system configuration option".
+
+Feature parameters MUST only be used according to the rules outlined in this
+RFC. They MUST NOT be used for any other purpose.
+
+Boolean feature parameter MUST be either `with`-parameter or `enable`-parameter.
+Non-boolean (e.g. string or integer) feature parameter MUST be an `conf`-parameter.
+
+Feature paramter that incurs additional runtime dependency MUST be and `with`-parameter.
```
7c6f434c, line 54: This line needs rewriting.

rfcs/0168-feature-parameter-names.md
```diff
@@ -0,0 +1,559 @@
...
+libgif  => with_gif
+curl    => with_curl
+alsa    => with_alsa
+
+# When people say "Xorg" support, they usually mean linking "X11"
+# library which is one of the ways to talk X protocol to X server. "xcb"
+# being another one.
+#
+# I am not aware of anybody implementing X protocol themself, and not
+# using one of these libraries.
```
7c6f434c, line 286: `sbcl.pkgs.clx` does, but I am not aware of any Common Lisp package with an optional dependency on CLX. Rust has e.g. `breadx`.

rfcs/0168-feature-parameter-names.md
```diff
@@ -0,0 +1,559 @@
...
+common term for enabling programs to produce output in non-English languages.
+
+</details>
+
+<details><summary>Rationale</summary>
+
+This distinction between `with` and `enable` feature parameters is based on Autoconf naming conventions.
+
+[Autoconf ENABLE](https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Package-Options.html)
+[Autoconf WITH](https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Package-Options.html)
```
7c6f434c, line 94: Is it intended to be the same link two times with two different labels?

rfcs/0168-feature-parameter-names.md
```diff
@@ -0,0 +1,559 @@
...
+{ lib
+, stdenv
+, enable_foo ? true
+}:
+
+# uses enable_foo
+stdenv.mkDerivation { ... }
+```
+
+## Feature parameter naming guidelines
```
7c6f434c, line 261: There should be some example with a multi-word package being an optional dependency.

7c6f434c:
This is actually 0169 (based on the PR number)

rfcs/0168-feature-parameter-names.md
KAction: From passing of the RFC is implied. But in reality, people will keep violating the rules until there is automation that stops them from doing so. Promise checked, promise kept.

rfcs/0168-feature-parameter-names.md
KAction: Nope, it is bug. Fixed.

rfcs/0168-feature-parameter-names.md
KAction: What exactly is wrong with it?

rfcs/0168-feature-parameter-names.md
7c6f434c: The problem is that without «CI or go away», it's not like everyone will find out at once.

So there will be more useless nitpicking, _and_ less uniformity (because gradual conversion away from established norms). And no time bound on this mess (as it starts with RFC passing and ends when CI is good enough to deploy).

rfcs/0168-feature-parameter-names.md
7c6f434c: It looks like «and» is a remainder of some previous sentence structure? Of course «parmter» is a typo, but I hope it was «parameter».

7c6f434c:
I guess PR title also needs an update.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,562 @@
...
+
+When upstream build system has options to enable or disable a optional features
+or configure some compile-time parameter, like path to default mailbox or
+initial size of hash map, derivation function (for short, package) MAY expose
+compile-time options via feature parameters that match
+`^(with|conf|enable)_[a-z_0-9]$` regular expression. For short, we will refer
+to them as `with`-parameter, `conf`-parameter and `enable`-parameter correspondingly,
+collectively as `feature parameters`.
+
+Note that this RFC does not establish any requires whether package should or
```
7c6f434c, line 42:
```suggestion
Note that this RFC does not establish any requirements whether package should or
```

rfcs/0168-feature-parameter-names.md
KAction: Added one.

rfcs/0168-feature-parameter-names.md
KAction: Well, If tomorrow I batch rename everything related to ALSA to `with_alsa`, new contributor day after will have no reason to invent `alsaSupport` if the only example he can find is `with_alsa`. But I agree, adding linter rule that would ban `alsaSupport`, `withAlsa`, `useAlsa` and other stuff I just renamed from together with the renaming would make it cleaner.

Not sure about performance, but on proof-of-concept level, it should be relatively simple treesitter query

rfcs/0168-feature-parameter-names.md
KAction: `s/and/a/`

rfcs/0169-feature-parameter-names.md
KAction: Thanks. Fixed.

rfcs/0168-feature-parameter-names.md
7c6f434c: For various definitions of new. New contributor probably has already tried some kind of overriding before. Also, what happens to the release branch?

For CI, performance would be a consideration…

KAction:
As proof-of-concept, implemented renaming logic on the `makeOverrideable` level. That means that individual packages
do not need to carry deprecation notices (something that @Atemu called "ffmeg" test).

https://github.com/NixOS/nixpkgs/pull/284107

rfcs/0168-feature-parameter-names.md
KAction: It ended up better than I expected.

I wrote [my tool](https://github.com/KAction/nix-lament) inspired by nixpkgs-lint, and it takes ~30 seconds to process whole nixpkgs, around 15ms for single file.

KAction:
@infinisil Can I have your input? I think at 2024-01-09 nixpkgs architecture meeting you expressed interest in shepherding this one.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,529 @@
...
+has neither, and this RFC strives to rectify it.
+
+# Detailed design
+[design]: #detailed-design
+
+When upstream build system has options to enable or disable a optional features
+or configure some compile-time parameter, like path to default mailbox or
+initial size of hash map, derivation function (for short, package) MAY expose
+compile-time options via feature parameters that match
+`^(with|conf|enable)_[a-z_0-9]$` regular expression. For short, we will refer
```
h7x4, line 38: I'm open to using underscores in these flags, but I think it would require changes to https://github.com/NixOS/nixpkgs/blob/master/CONTRIBUTING.md#syntax, which says

> Use `lowerCamelCase` for variable names, not `UpperCamelCase`. Note, this rule does not apply to package attribute names, which instead follow the rules in [package naming](https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#package-naming).

KAction: Thank you for bringing it up. I'll add into the migration section step of updating this documentation.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,537 @@
...
+[motivation]: #motivation
+
+Developer interface for building environment with specific features enabled and
+disabled is more complicated and requires more knowledge of implementation
+details compared to already existing systems.
+
+For example, Gentoo Linux has exhaustive list of [all known configuration
+flags](https://www.gentoo.org/support/use-flags) and has means to
+programmatically query what flags are available for particular package. Nixpkgs
+has neither, and this RFC strives to rectify it.
```
infinisil, line 29: You seem to describe the problem of `.override` attributes not being easily discoverable, but the RFC doesn't actually address that at all from what I understand.

If that's the actual problem you're trying to solve, the easiest solution might be to make those attributes more easily discoverable, e.g. by rendering attributes with their documentation on search.nixos.org. Or on the deep end, something like https://discourse.nixos.org/t/working-group-member-search-module-system-for-packages/26574

However, the actual problem you seem to want to solve with this RFC is described better in your comment here: https://github.com/NixOS/nixpkgs/pull/234463#issuecomment-1574892207. If that's indeed the case, the motivation section should be rewritten to talk about that.

KAction: It actually does address, since now feature parameters are clearly namespaced from all other. To make it clearer,
I rewrote the motivation part as list of good things that we will get by implementing this RFC. Thoughts?

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,566 @@
...
+```
+
+4. There is no clear separation between feature parameters and other arguments
+   to the package. With naming mandated by this RFC, distinction is clear to
+   cater to automatic queries.
+
+5. As side effect of the proposad migration plan is compiling exahustive list
+   of known feature parameters. It is not huge stretch to have CI check to
+   enforce that this list is current, bringing nixpkgs on par with Gentoo
+   [use-flags](https://www.gentoo.org/support/use-flags).
```
7c6f434c, line 58: Doesn't Gentoo separately invest effort in consistently providing the popular USE flags for almost all the  packages where they make any amount of sense?

KAction: At some point did some manual comparison between Gentoo and Nixpkgs and learned that Nixpkgs exports no less of feature parameters than Gentoo, just in less uniform way. In other words, I expected Gentoo configuration to be more granular than Nixpkgs, but that wasn't the case.

I'll move the last sentence into the future work to not inadvertently imply that RFC demands Nixpkgs to have as much feature parameters as Gentoo has use flags.

7c6f434c: My expectation was that Nixpkgs has a comparable number of feature parameters, but the corresponding feature parameters are provided by fewer packages, isn't it so?

KAction: Probably. Many packages do something like

```
buildInputs = lib.optionals stdenv.isLinux [ foo ];
```

without bothering to introduce `with_foo` flag.

KAction:
@jtojnar
@ghost
@mohe2015 

Inviting reviewers of #234463 into the RFC discussion. 

Feedback is welcome. Finding more people to get RFC

> received ample discussion and enough of the tradeoffs have been discussed.

is also welcome. I think on my side, I did proof-of-concept both on how to lint new packages and how to do renaming in existing. Anything I miss?

7c6f434c:
My current opinion is: motivation mismatches the measures proposed; switching the naming word-handling convention is costly and the changes elsewhere have deprived it of the main benefit in terms of linting; the linting being mandated but not the trigger condition for the more onerous requirements timing is bad. 

Unification of option naming where only the names already used in practice are eligible — via should-merge stipulation for zero-rebuild PRs in that direction — would not bring practical benefits either but at least would be mostly harmless.

I won't actively support even the latter reasonable version of unification, so I guess I am too biased to be a shepherd. I hope though, that I manage to keep the local feedback useful either to reject the substance and not silly technicalities, or to implement it with less unnecessary cost, useful or not. (Costs inherent to the claimed benefits are a trade-off question, we will never get agreement there, only a truce)

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+feature.
+
+# Motivation
+[motivation]: #motivation
+
+Developer interface for building environment with specific features enabled and
+disabled is more complicated and and involves more "you just have to know it"
+compared to already existing systems.
+
+1. Knowing that package has a feature parameter (out of knowning the upstream)
```
AndersonTorres, line 26:
```suggestion
1. Knowing that package has a feature parameter (out of knowing the upstream)
```

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+withX11
+x11Support
+enableX
+```
+
+4. There is no clear separation between feature parameters and other arguments
+   to the package. With naming mandated by this RFC, distinction is clear to
+   cater to automatic queries.
+
+5. Side effect of the proposdd migration plan is compiling of almost
```
AndersonTorres, line 55:
```suggestion
5. Side effect of the proposed migration plan is compiling of almost
```

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+When upstream build system has options to enable or disable a optional features
+or configure some compile-time parameter, like path to default mailbox or
+initial size of hash map, derivation function (for short, package) MAY expose
+compile-time options via feature parameters that match
+`^(with|conf|enable)_[a-z_0-9]$` regular expression. For short, we will refer
+to them as `with`-parameter, `conf`-parameter and `enable`-parameter correspondingly,
+collectively as `feature parameters`.
+
+Note that this RFC does not establish any requirements whether package should or
+should not have any feature parameters, this decision is left at descretion of
```
AndersonTorres, line 71:
```suggestion
should not have any feature parameters, this decision is left at discretion of
```

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+common term for enabling programs to produce output in non-English languages.
+
+</details>
+
+<details><summary>Rationale</summary>
+
+This distinction between `with` and `enable` feature parameters is based on Autoconf naming conventions.
+
+[Autoconf ENABLE](https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Package-Options.html)
+[Autoconf WITH](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/External-Software.html)
```
AndersonTorres, line 123: You are using distinct versions of autotools here.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+
+If upstream uses `--enable-translations` to denote support for NLS,
+corresponding feature parameter SHOULD be `enable_nls`, since it is much more
+common term for enabling programs to produce output in non-English languages.
+
+</details>
+
+<details><summary>Rationale</summary>
+
+This distinction between `with` and `enable` feature parameters is based on Autoconf naming conventions.
```
AndersonTorres, line 120:
```suggestion
This distinction between `with` and `enable` feature parameters is based on Autotools naming conventions.
```

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+}
+```
+
+</details>
+
+The rules are to be enforced by static code analyse linter to be written. Since
+no static code analyzis is perfect, it shall have support for inhibiting
+warnings in unsual cases.
+
+All feature parameters SHOULD be exported in `features` passthrough set. See example above.
```
AndersonTorres, line 195:
```suggestion
All feature parameters SHOULD be exported in `passthru.features` set. See example above.
```

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+2. Overrides become much more verbose. Simple and ubiquitous
+
+   ```
+   bar.override { with_foo = true; }
+   ```
+
+   becomes unwieldy
+
+   ```
+   bar.override (old: old // { features = old.features // { with_foo = true }; })
```
AndersonTorres, line 432: you mean

```suggestion
   bar.overrideAttrs (old: old // { features = old.features // { with_foo = true }; })
```

?

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+As mentioned in motivation part, the best in class of feature flag
+configuration system is Gentoo Linux:
+
+- [https://www.gentoo.org/support/use-flags]
+
+There is not much work I can find about this problem in Nixpkgs other than my
+previous attempts to solve it on package-by-package basis:
```
AndersonTorres, line 475:
```suggestion
previous attempts to solve it on a  per package basis:
```

rfcs/0169-feature-parameter-names.md
KAction: No, I mean `override`. If we do something like this:

```
{ lib, stdenv, features ? { with_foo = true; enable_bar = stdenv.isLinux; } }:

stdenv.mkDerivation { ... }
```

then we will have nice separation of feature flags and other stuff, but overrides become mess.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+underlying library. For example, optional dependecy on `pyqt5` Python bindings
+to `Qt5` library should have `with_qt5` feature parameter.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+This is how one can query list of feature parameters that incur extra runtime dependencies:
+
+```
+$ nix eval --json --impure --expr 'with import ./. {}; builtins.attrNames gnuplot.override.__functionArgs' | jq '.[]| select(startswith("with"))'
```
AndersonTorres, line 341: I would like if we can do it without jq or similar shell script tricks.

But I believe it would require a too huge modification to the language. Nix does not look like a query language.

7c6f434c: It's not that people have not implemented typed (and checked) arguments in Nix, they have, multiple times. So the massive change would be on Nixpkgs side… and the speed impact will be unpleasant

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+# Motivation
+[motivation]: #motivation
+
+Developer interface for building environment with specific features enabled and
+disabled is more complicated and and involves more "you just have to know it"
+compared to already existing systems.
+
+1. Knowing that package has a feature parameter (out of knowing the upstream)
+   does not mechanically translates to the name of feature parameter, it can be
+   named in multiple ways.
```
infinisil, line 28: Doesn't `.overrideAttrs (old: { cmakeFlags = old.cmakeFlags or [] ++ [ "..." ] })` already do this in a fairly standard way? I don't think there's even a need to expose cmake flags further.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+
+2. Writing generic code is complicated. Even if list of packages all have
+   optional dependency on package `foo`, duck-typing like
+
+```
+map (p: p.override { with_foo = false; }) packages
+```
+
+won't work because of lack of uniformity, leading ugly code downstream.
```
infinisil, line 37: Shouldn't this also work with `.overrideAttrs (old: { cmakeFlags = old.cmakeFlags or [] ++ [ "..." ] })`? Of course this only works when all packages actually have this option.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,572 @@
...
+   dependency as much as possible, he needs to set multiple feature parameters,
+   and he still can miss couple. E.g, it is not obvious whether the following
+   list is enough to cover everything X11-related:
+
+```
+withX
+withX11
+x11Support
+enableX
+```
```
infinisil, line 49: It sounds like this would be fixed with a [`config` option](https://github.com/NixOS/nixpkgs/blob/20d0954713ffa76a48507e90db6ad558f1238cf6/pkgs/top-level/config.nix), which is for flags that are generic/useful enough to apply to many packages.

infinisil (changes requested):

rfcs/0169-feature-parameter-names.md
KAction: Not everything uses cmake. There are multiple build systems around, and projects even switch them sometimes. And each project names corresponding flag slightly differently.

The whole point of distribution is to abstract away idiosyncrasies of the upstream. If we say that anybody who wants non-default configuration has to do their own research, a lot of effort will be duplicated.

rfcs/0169-feature-parameter-names.md
KAction: And use the same build system. And name their flags in the same way.

I wish upstreams were to do the work that this RFC tries to do, but it didn't happened in 1991, and it won't happen now.

rfcs/0169-feature-parameter-names.md
KAction: Overrides on individual packages become problematic (discussed in alternatives). Any way, we need some guidelines what exactly name we should to into `config`.

rfcs/0169-feature-parameter-names.md
infinisil: Then I'm not sure if the motivational point of being able to re-use upstream flags really works out.

If you abstract over something, you need to deal with changes when they happen underneath. This works out if upstream never changes any build flags, but we cannot guarantee that. And different build systems might not even support the same naming conventions.

So if upstream ever renames a flag, say from `with_foo` to `with_foo2`, to make the abstraction work, the Nix code has to implement support for both flags and deprecate the older one. Otherwise you risk breaking users `.override`'s.

While this is doable, I don't think package authors should be bothered with this. If the upstream flag changes, the package implementation should just transparently switch to the new name without requiring any changes at `.override` call sites.

So this is why I think this motivational point should be dropped: We shouldn't be required to re-use upstream flags.

rfcs/0169-feature-parameter-names.md
infinisil: Makes sense.

I think a convention to reuse the same `config` option name would be good, e.g. like this:

```nix
{
  config,
  withX11 ? config.withX11,
}:
...
```

Then we have consistency between per-package flags and global flags.

rfcs/0169-feature-parameter-names.md
KAction: I see your point. I'll reword.

My idea is that if I happen to know that emacs has optional dependency on alsa, and I happend to know
that it is exported by nixpkgs, than I can write this:

```nix
emacs.override { with_alsa = false }
```

No need to remember whether it is `supportAlsa` or `withAlsaLib`.

infinisil: How about establishing `.override { alsa = null; }` as a pattern for that instead? This way we save a flag.

In the implementation this would look like

```nix
{
  stdenv,
  alsa,
}:
stdenv.mkDerivation {
  
  configureFlags = [
    (lib.optional (alsa != null) "--with-alsa")

    # Or with a more convenient function
    (lib.withPackageFeature alsa "alsa")
  ];

}
```

Atemu: This is what we used to do in Nixpkgs but I'm not a fan of this. A couple of reasons off the top of my head:

* it only applies to `with` parameters. `enable` or `conf` parameters cannot be covered by this as they aren't usually connected to any one package
* You cannot sensibly disable support for something by default; you'd have to override the thing with null at callPackage site. Apply the "ffmpeg test" if you want to see what I mean.
* Conflates existance of the dep with enablement of dep
* Overriding a dep with null could cause exceptions in the drv because it assumes the dep is an attrset (e.g. `versionAtLeast dep.version "1.0"` would throw if dep was null)

It'd be great if we could have these reasons (and likely more) to the RFC text as an alternative and for historical context as this was the status quo at some point.

KAction: I'll include your nice list into description of `alsa = null` alternative, probably this weekend. Thank you.

rfcs/0169-feature-parameter-names.md
KAction: I don't think we will need `config` for that. If you want global, just `overlays = [self: super: { with_x11 = false; }`

rfcs/0169-feature-parameter-names.md
KAction: Updated RFC to mention `foo ? null` pattern.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,595 @@
...
+
+4. Side effect of the proposdd migration plan is compiling of almost
+   exahustive list of known feature parameters.
+
+
+# Detailed design
+[design]: #detailed-design
+
+When upstream build system has options to enable or disable a optional features
+or configure some compile-time parameter, like path to default mailbox or
```
reckenrode, line 59: Would this rule apply to feature parameters that do not correspond to upstream build system features?

For exampe, [libiconv](https://github.com/NixOS/nixpkgs/blob/fadc1d7ef25cba8580cc0380f07e686b1ad2fe49/pkgs/development/libraries/libiconv/default.nix#L4) has a feature flag `enableDarwinABICompat` for building libiconv in a way that is ABI-compatible with the system-provided libiconv. The plan is to drop darwin.libiconv in favor of the upstream one, but the upstream one needs this compatibility to allow references to the system one to be replace with it (using `install_name_tool`). Otherwise, doing so will result in crashes due to missing symbols.

Additionally, libiconv (and other packages) feature `enableShared` and `enableStatic` flags. Would these be migrated to the proposed feature flags section or left outside of it?

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,595 @@
...
+```
+
+</details>
+
+As special provision, if optional vendored dependency is exposed by feature
+parameter, it MUST be `with`-parameter.
+
+If feature parameter does not incur extra runtime dependencies, corresponding
+feature parameter MUST be `enable`-parameter and its name SHOULD correspond to
+the upstream naming, unless is causes major inconsistency with other packages.
```
reckenrode, line 101: Some projects use build systems that don’t support feature flags but do support customization via environment variables or preprocessor definitions (e.g., makefiles or xcodebuild). Am I correct assuming there be some leeway when translating their flags to appropriate feature flags in nixpkgs?

For example, MoltenVK has an [option](https://github.com/KhronosGroup/MoltenVK?tab=readme-ov-file#metal_private_api) to determine whether to use private APIs. That option is named `MVK_USE_METAL_PRIVATE_API`. A direct translation would be `enable_mvk_use_metal_private_api`, which is verbose and funky. The current package uses [`enablePrivateAPIUsage`](https://github.com/NixOS/nixpkgs/blob/b06025f1533a1e07b6db3e75151caa155d1c7eb3/pkgs/os-specific/darwin/moltenvk/default.nix#L23), which I assume would translate to something like `enable_private_api_usage` or `enable_metal_private_api_usage`.

rfcs/0169-feature-parameter-names.md
KAction: I'd say to migrate these too, in line with suggested `enable_ssl`.

reckenrode: Thanks. My other question was about options that are for nixpkgs use. libiconv is the primary one that comes to mind. It builds libivonv in a specific way for ABI compatibility, so it doesn’t map to a specific build system option.

If I am understanding correctly, that type of option would still be okay under this RFC even though it’s not an upstream-provided option. (In this case, we’re manually recreating what Apple does in their fork’s Xcode project.)

rfcs/0169-feature-parameter-names.md
KAction: Yes, consistency across nixpkgs is more important than matching upstream naming exactly (I think I have this line somewhere).

I think we can call this option `enable_private_api`, and use this name for other, similar situations.
I'll add your example into guidelines.

AndersonTorres:
How many are needed right now?

infinisil:
@AndersonTorres There are no shepherds right now but three are required. (While I would be a good candidate, I'm a bit overbooked right now, so I'd prefer to only give occasional feedback.)

AndersonTorres:
I am interested in seeing this implemented.
So I wanna shepherd this.
However I am on vacation now, I will return after 15 days.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,595 @@
...
+Feature flags that require single external dependency SHOULD be named after
+that dependency. Prefix `lib` SHOULD be removed. For example,
+
+<details><summary>Example</summary>
+
+```
+systemd => with_systemd
+libgif  => with_gif
+curl    => with_curl
+alsa    => with_alsa
```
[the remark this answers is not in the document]
KAction: No. There is rule about removing `lib` prefix.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,595 @@
...
+
+Rules in the first section of this RFC describe different scenarios when
+`with`, `enable` and `conf` feature parameters must be used, but coming with
+exact name that is clear, concise, describes it effect correctly and is
+consistent across the nixpkgs is hard to codify with absolute rule. Instead,
+this RFC provides set of guidelines and examples based on scenarios brought up
+during the discussion.
+
+Feature flags that require single external dependency SHOULD be named after
+that dependency. Prefix `lib` SHOULD be removed. For example,
```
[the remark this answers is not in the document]
KAction: I don't like this change. Why?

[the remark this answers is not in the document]
KAction: Otherwise we will have `with_alsalib` and `with_libgif`.

KAction:
@wolfgangwalther Thank  your for thorough revision of the text.

I find your reasoning about consolidating knowledge of package availability quite compelling. Having `with_` flags naming strictly derived from package name sounds good, but there might be some corner cases where `MUST` will paint us into the corner.

@7c6f434c @Atemu Thoughts?

7c6f434c:
> The justification to use snake case for those feature parameter names is the ability to have a 1:1 mapping for attribute names and with_ parameter names.

However, the currently proposed rules for what is the _real_ dependency parameter to mention are complicated enough that no automation is plausible, so we can just write should-accept for unifying renamings and keep camel case — given that for most cases it is clear what is the proper camel case.

lheckemann:
@wolfgangwalther shepherding is probably the best way you can help this RFC get forwards, so we'd like to nominate you as a shepherd -- let us know if you're alright with that :)

Atemu (commented):
First of all, I'd like to thank @KAction for putting all of this together. From the Pre-RFC to this, it's clear that a lot of thought and consideration was put into this.

There are some minor details I have some opinions on but the large picture gets an ACK from me.

Atemu:
Overall, I do not agree with @wolfgangwalther's proposal to add global parameter mechanism to this RFC or tie parameter naming to a global goal.

The possibility of making such Gentoo-like USE-flags a reality is one fo the goals of this RFC as I understand it. Actually doing that should be explicitly excluded from this RFC though IMO.

As I understand it, limiting it in this regard was intentional and I'd agree with that. A USE-flag like mechanism entails a lot of other design decisions and problems to solve than setting a standard on the naming of parameter names. This standardisation can be useful without the global mechanism and it'd be best to keep the RFC for it minimal.

rfcs/0169-feature-parameter-names.md
```diff
@@ -0,0 +1,595 @@
...
+```
+
+It is definitely looks cleaner, but unfortunately hits several limitations of
+the Nix language.
+
+1. Nix language provides way to [introspect function argument names](https://nixos.org/manual/nix/stable/language/builtins.html#builtins-functionArgs),
+   but no way to learn their default values. Learning default values is fundamentally impossible, since default values of one parameter might depend
+   on the value of another parameter, so default values can't be evaluated until function is called.
+
+2. Overrides become much more verbose. Simple and ubiquitous
```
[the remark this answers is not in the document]
Atemu: I've had a little breakthrough regarding this: https://github.com/NixOS/nixpkgs/pull/312432

infinisil:
RFCSC:
      
@wolfgangwalther: Asking again, would you be willing to shepherd this RFC?
Furthermore, we're also nominating @Atemu as a shepherd, since you recently reviewed it, sound good?

GetPsyched:
RFCSC:

Pandapip1:
> The justification to use snake case for those feature parameter names is the ability to have a 1:1 mapping for attribute names and with_ parameter names. This is not possible with camel case, because attributes don't start with a capital letter.

I don't find this particularly convincing. Having a 1:1 mapping isn't obviously beneficial. The only reason I can think of that would make it somewhat beneficial is if you're going to be doing automated parsing on attribute names. But in that case, you can easily make the first letter after the `use` lowercase. If there's a different rationale for having a 1:1 mapping, please tell me, but I'm strongly against hybrid snake and camel case.

---

What about features that require multiple dependencies to be added at once, and where it doesn't make sense to have only one of the two optional dependencies? Currently, this RFC says there would need to be two parameters, `with_depA` and `with_debB`, but `enable_optionalFeature` would obviously be the better thing to use instead.

---

I'm concerned that preferring `with_dependency` over `enable_feature` for parameters that add a single dependency makes it harder to figure out how to enable optional features. Why don't we get rid of `with_dependency` altogether, and only have `enable_feature`-style parameters?

Atemu:
> What about features that require multiple dependencies to be added at once, and where it doesn't make sense to have only one of the two optional dependencies? Currently, this RFC says there would need to be two parameters, `with_depA` and `with_debB`, but `enable_optionalFeature` would obviously be the better thing to use instead.

Why "instead"? You can have both; `with*` flags and `enable*`. The `with` flags would simply default to the value of the enable.

Atemu:
Ah, that's a good point. Thinking about it again, I actually came to the conclusion that the defaults should work the other way around a while back (you can find it up above somewhere): The enable flag should default to the values of the individual with flags.

Though that of course limits the effect of the enable flag to enabling/disabling the feature, not disabling the dependency; leaving it up to the build system to do the right thing. That should be expected though IMHO as only the `with` flags would directly control dependencies.

Pandapip1:
Again though, why would you want to *add* a dependency if not to enable a feature? It would only add dead code paths.

7c6f434c:
> Again though, why would you want to add a dependency if not to enable a feature? It would only add dead code paths.

Maybe to replace the vendored version…

Pandapip1:
I thought it was standard to to always replace the vendored version of things anyway? Regardless, I still find this an uncompelling use for `with*` arguments:

1. If it's to enable a feature, a change should be upstreamed to nixpkgs that adds an `enable*` argument.
2. If it's to fix a bug, then there's a bug with the derivation and a change to nixpkgs that either makes the derivation use the nixpkgs-packaged library or makes the necessary patches to the nixpkgs-packaged library should be made.

7c6f434c:
Of course replacement of dependencies adds integration bug risk, non-replacement increases the chances that some bug fix that is not in the vendored version is actually relevant to the specific use case, and no maintainer realistically has resources to check either part _completely_.

Or both types of bugs are known to be there, known to require large changes to fix, and happen to be rare enough that for most use cases no more than one of these bugs is actually observable.

PR #171, opened 2024-03-20
[RFC 0171] Default name of fetchFromGithub FOD to include revision

jonringer:
Make fetchFromGithub FOD name more meaningful. This avoids stale artifacts and gives more content fidelity when looking at nix store paths.

Rendered: https://github.com/jonringer/rfcs/blob/jringer/fetch-from-github/rfcs/0171-fetch-from-github.md

Items for further refinement:
- Other version control fetchers
- Prefix vs suffix of `src` label


rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,114 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- All derivations which don't pass a "name" parameter will need to be re-realized
+    - This will be a download-intensive one-time cost to realize the new FOD derivations.
+    - NAR hash should not need to be recommputed assuming it was deterministic and not stale.
+    - Cache should be minimally impacted as NARs are content addressable, thus deterministic sources should not contribute to cache bloat.
+    - Potential for sources which are no longer available to be broken.
+        - These can have their name manually set to "source" to perserve previous behavior.
+        - Ideally source availability would remedied with more appropriate methods. E.g. being made available.
```
dotlambda, line 68:
```suggestion
        - Ideally source availability would be remedied with more appropriate methods. E.g. being made available.
```

rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,114 @@
...
+- "Interchangability" with other fetchers is diminished as the derivation name is diferent
+    - In practice, fetchFromGitHub is never used in this way. It is generally the only fetcher, so there is never another FOD to dedupilicate.
+- Out-of-tree repositories may get hash mismatch errors
+    - If the cause of the mismatch is staleness, this is good and working as intended
+    - If the cause is non-determinism, this is frustrating.
+- Some derivations assume "source" to be the name of sourceRoot
+    - This has been mitigated over two years within Nixpkgs
+    - Out-of-tree code may need break if they assume "source" is the name
```
dotlambda, line 76:
```suggestion
    - Out-of-tree code may break if they assume "source" is the name
```

rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,114 @@
...
+        - Can be mitigated with realease notes describing the issue
```
dotlambda, line 77:
```suggestion
        - Can be mitigated with release notes describing the issue
```

rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Summary
+[summary]: #summary
+
+Updating the hash on Fixed-Output Derivations (FODs) is a very error prone process. It's not intuitive to invalidate the existing hash, attempt to realize the FOD, then replace the hash value with the value that Nix just calculated. This RFC advocates for influencing the default derivation name of the fetchFromGithub helper with the "repo" and "rev" values to ensure that changed URLs force the FOD to be re-built.
+
+# Motivation
+[motivation]: #motivation
+
+This will hopefully provide immediate feedback that an FOD contains a stale hash. One must either build the FOD without the previous FOD in their Nix store, or run the FOD build with `--check` to verify that the FOD is not stale. Although fetchFromGithub is one of many fetchers; it is very common, and generally has a user specify the source information in granular which makes intent clear.
```
jonringer, line 19:
```suggestion
This will hopefully provide immediate feedback that an FOD contains a stale hash. One must either build the FOD without the previous FOD in their Nix store, or run the FOD build with `--check` to verify that the FOD is not stale. Although fetchFromGithub is one of many fetchers; it is very common, and generally has a user specify granular source information which makes differentiating between sources easy.
```

rfcs/0171-fetch-from-github.md
jonringer: Thanks, I really need to install some spell checking into my vim setup.

rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,114 @@
...
+- Similar treatment to similar fetchFromGitX helpers?
+
+# Future work
+[future]: #future-work
+
+- Apply name change to fetchFromGitHub in PR to staging:
+    - Resolve stale FODs, most of these can be PRs made against master
+    - Resolve assumed usage of "source" as `src.name`, most of these changes can be made against master
+    - Revert name to "source" for removed source urls.
+- Release notes around potential breakages with usage of "source"
```
jonringer, line 113:
```suggestion
- Add release notes for potential breakages when assuming the name of the unpacked directory is "source"
```

jonringer: clarify intent here

Aleksanaa:
Just one silly question: why only `fetchFromGitHub` but not including `fetchFromGitLab` and others, besides the reason that it is very common?

> Although fetchFromGithub is one of many fetchers; it is very common, and generally has a user specify granular source information which makes differentiating between sources easy.

jonringer:
> Just one silly question: why only `fetchFromGitHub` but not including `fetchFromGitLab` and others, besides the reason that it is very common?

I did reference this https://github.com/NixOS/rfcs/blob/1ac472d4f0b82ec9f6781cd8c5a4f88fabef631e/rfcs/0171-fetch-from-github.md?plain=1#L104. 

I mention fetchFromGitHub because it's about 20x as common:
```
$ rg -i fetchFromGitHub | wc -l
33178
$ rg -i fetchFromGitlab | wc -l
1478
```

I'm not opposed to doing a similar treatment to the other fetchFromX helpers. I think this RFC could be expanded to encompass them all eventually.

eclairevoyant:
There's already a speculative impl as of 2 weeks ago: https://github.com/NixOS/nixpkgs/pull/294068

Though the format is slightly different (`"source-${owner}-${repo}-${rev}"`, essentially)

ShamrockLee:
> Though the format is slightly different (`"source-${owner}-${repo}-${rev}"`, essentially)

I choose the format to make it easier to inspect which source FOD is going wrong when several packages got rebuild at the same time.

rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,114 @@
...
+        - Ideally source availability would be remedied with more appropriate methods. E.g. being made available.
+- "Interchangeability" with other fetchers is diminished as the derivation name is different
+    - In practice, fetchFromGitHub is never used in this way. It is generally the only fetcher, so there is never another FOD to dedupilicate.
+- Out-of-tree repositories may get hash mismatch errors
+    - If the cause of the mismatch is staleness, this is good and working as intended
+    - If the cause is non-determinism, this is frustrating.
+- Some derivations assume "source" to be the name of sourceRoot
+    - This has been mitigated over two years within Nixpkgs
+    - Out-of-tree code may break if they assume "source" is the name
+        - Can be mitigated with release notes describing the issue
```
ShamrockLee, line 77: All these "drawbacks" are caused by the assumption of `src.name == "source"` being broken.

However, we have already deprecated such assumption as early as Nixpkgs 21.11. Nixpkgs Manual now requires `sourceRoot` to start with `"${src.name}"` instead of `"source"` when `src` is constructed with by `fetchgit`-based fetchers, and tree-wide conversions (#245388, #247977, #248528, #248683, #294334) has been merged.

`fetchFromGitHub` and other output-as-a-directory fetchers can still be used interchangeably if we stick to `${src.name}` instead of `"source"`.

In my opinion, we do not provide [bug-level compatibility](https://github.com/pkgconf/pkgconf#compatibility-with-pkg-config) to packages failing to follow already-stablized specifications, in-tree or out-of-tree.

ShamrockLee:
> maybe this should be relabeled as "Default name of fetchFromX FOD helpers to include revision"

IMO, "fetchers for version-controlled repositories" would be a suitable  target. This includes fetchers based on specific version control systems (e.g. `fetchcvs`, `fetchsvn`, `fetchgit`, `fetchhg`, `fetchbzr`, etc.) and fetchers based on specific service providers (e.g. `fetchFrom*`).

Ericson2314:
Hehe per https://github.com/NixOS/rfcs/pull/133#issuecomment-2016614005, my dream of single hash git fetchers (avoiding the need to put things in the name like this) is now a good bit closer!

ShamrockLee:
> Hehe per https://github.com/NixOS/rfcs/pull/133#issuecomment-2016614005, my dream of single hash git fetchers (avoiding the need to put things in the name like this) is now a good bit closer!

Congratulations!

We still need to fix those fetchers at Nixpkgs level, though, since Nixpkgs often takes years to adopt new Nix features.

rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,114 @@
...
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Full commit hashes could be truncated. This sacrifices a bit of simplicity for better looking derivation names:
+```
+let
+  version = builtins.replaceStrings [ "refs/tags/" ] [ "" ] rev;
+  # Git commit hashes are 40 characters long, assume that very long versions are version-control labels
+  ref = if (builtins.stringLength rev) > 15 then builtins.substring 0 8 version else version;
+in lib.sanitizeDerivationName "${repo}-${ref}-src";
```
SuperSandro2000, line 101: I wouldn't recommend this. Short hashes are not guaranteed to be stable for long term storage.

jonringer: The current "source" isn't that stable either. 8 characters is still a fair amount of entropy, and likely to be different enough for most repositories.

jonringer: Also remember, that it doesn't have to "unique across all time". It just needs to be different than what was there before, so that the combination of name + hash are different.

ShamrockLee:
I'd like to nominate myself to shepherd this RFC. I previously authored a related implementation, NixOS/nixpkgs#294068, which could inform the discussion around this RFC's design.

I'm excited about this RFC and look forward to working with the community to shepherd it through the process.

rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,114 @@
...
+
+- All derivations which don't pass a "name" parameter will need to be re-realized
+    - This will be a download-intensive one-time cost to realize the new FOD derivations.
+    - NAR hash should not need to be recomputed assuming it was deterministic and not stale.
+    - Cache should be minimally impacted as NARs are content addressable, thus deterministic sources should not contribute to cache bloat.
+    - Potential for sources which are no longer available to be broken.
+        - These can have their name manually set to "source" to perserve previous behavior.
+        - Ideally source availability would be remedied with more appropriate methods. E.g. being made available.
+- "Interchangeability" with other fetchers is diminished as the derivation name is different
+    - In practice, fetchFromGitHub is never used in this way. It is generally the only fetcher, so there is never another FOD to dedupilicate.
```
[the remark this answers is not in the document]
jonringer: Agreed my wording was not the best.

jonringer:
> IMO, "fetchers for version-controlled repositories" would be a suitable target. This includes fetchers based on specific version control systems (e.g. fetchcvs, fetchsvn, fetchgit, fetchhg, fetchbzr, etc.) and fetchers based on specific service providers (e.g. fetchFrom*).

If we have a shepards meeting, we can refine the scope.

rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,115 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- All derivations which don't pass a "name" parameter will need to be re-realized
+    - This will be a download-intensive one-time cost to realize the new FOD derivations.
+    - NAR hash should not need to be recomputed assuming it was deterministic and not stale.
+    - Cache should be minimally impacted as NARs are content addressable, thus deterministic sources should not contribute to cache bloat.
+    - Potential for sources which are no longer available to be broken.
+        - These can have their name manually set to "source" to perserve previous behavior.
```
gerpiet, line 67: Typo:
```suggestion
        - These can have their name manually set to "source" to preserve previous behavior.
```

ShamrockLee:
To code the fetcher name into the `name` attribute, the default name could also be `${repo}-${rev}-github-source`.

rfcs/0171-fetch-from-github.md
```diff
@@ -0,0 +1,115 @@
...
+[motivation]: #motivation
+
+This will hopefully provide immediate feedback that an FOD contains a stale hash. One must either build the FOD without the previous FOD in their Nix store, or run the FOD build with `--check` to verify that the FOD is not stale. Although fetchFromGithub is one of many fetchers; it is very common, and generally has a user specify granular source information which makes differentiating between sources easy.
+
+As a secondary effect, this will also give a more meaningful name to the FODs than "source". E.g. `/nix/store/...-source -> /nix/store/...-protobuf-v24.1-src`
+
+# Detailed design
+[design]: #detailed-design
+
+Change the default name of fetchFromGithub fetcher from `"source"` to `lib.sanitizeDerivationName "${repo}-${rev}-src"`.
```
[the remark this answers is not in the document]
jonringer: > How about using pname instead of repo as a sort of middle ground between stability and informative names. Of course then changing the repo url will not automaticaly verify if the hash actually matches, so a tradeoff.

`pname` is a stdenv.mkDerivation convention, this would require people to pass it separately. The goal here also is to avoid hash mismatch, so I don't think is this is enough to solve the problem this RFC is trying to remedy.

> Also how about starting with src- instead of at the end? When listing the nix store, all fetched inputs can then be sorted together.

I wouldn't be opposed to it, but I'm used to `<pname>-<version>-<variant>` naming, source being a variant in this case.

rfcs/0171-fetch-from-github.md
jonringer: thanks

rfcs/0171-fetch-from-github.md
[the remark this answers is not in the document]
ShamrockLee: > The second suggestion is actually a part of a wish I have to prepend src- to all build inputs (tars, patch files, builders, etc), pkg- to all packages, and nixos- to the generated nixos configs, so that it is easier to tell at a glance why a particular store path is being installed, and also would be useful for tools like nvd when listing diffs. Yeah this would a much broader proposal, but one can dream someday it will happen. For now, we can start with just the FOD fetchers.

It sounds like another Nix RFC.

ShamrockLee:
People who use the result FOD produced by `fetchFromGitHub` don't even need to know about how `name` is specified. They'll just use `src.name` to specify `sourceRoot` and use `sourceRoot` in their build commands.

As an RFC, maybe we don't even need to specify what `name` should be, but require it to contain either version or source revision.

rfcs/0171-fetch-from-github.md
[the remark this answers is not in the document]
ShamrockLee: > I'm used to `<pname>-<version>-<variant>` naming, source being a variant in this case.

It's a bit strange to consider package source a "variant" of the package. Nevertheless, such name-suffixing way could already be found in Nixpkgs -- `goModules` used by `buildGoModule` does use the suffixing `name` pattern, and sometimes people manually set the in-package FODs to name the suffixed way.

On the other hand, language-specific packages (e.g. Python packages, Ocaml packages, etc.) often prefix the name with the name and version of the compiler/interpreter.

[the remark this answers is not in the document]
jonringer: I'm not specifically tied to the suffix approach.. I'll add a comment to the initial comment

lheckemann:

philiptaron:
I'll toss my name in for an RFC shepherd, as I believe in the utility of this proposal.

jonringer:
@oxij I would like to nominate you as a shepherd, looks like you've thought about this longer than I've been involved in nixpkgs.

I like your addition of a lib function to make derivation names more consistent in https://github.com/NixOS/nixpkgs/pull/49862

oxij:
Thanks for a nomination, but I don't think this RFC is a proper solution to the stated problem.

Eelco and others objecting to the `<hash>-<name>-<version or revision>-source` naming scheme (and to the original 2018 version of #49862 which wanted to do exactly this, but for all fetchers, not just `fetchFromGitHub`, and, primarily, for a different reason of `/nix/store` discoverability) do have a point: the current most common `<hash>-source` naming scheme is awfully convenient for running a build server (like Hydra) and when you have to frequently switch between fetchers (e.g. when using out-of-Nixpkgs derivations and/or flakes).

IMHO, the problem this RFC tries to address is actually a set of features/bugs/issues of Nix itself:

- Nix reuses existing fixed-output outputs between different derivations without actually checking the derivation actually builds into an output with the given output hash; which is certainly a feature, given how useful it is for switching between different fetchers;

- but, IIRC, last time I tried, explicit `nix-store -r --check <derivation>` over such a derivation will fail to re-build and re-check it; which is either a bug, or there should be a separate option for doing this; for simplicity, let's assume we want a new option for this, and we want to call it `--force-rebuild`;

- `nix-build` and `nix build` lack `--force-rebuild` option (similarly to how they lack `--check` option too), you have to call `nix-store -r` manually (an issue).

(Now that I think about it, I should probably open Nix issues for these. Later, I'm a bit busy ATM.)

If the latter two issues were not such, then you could just do `nix-build --force-rebuild -A hello.src` instead of any of this, and OfBorg could also run that on all changed fixed-output derivations in a PR, ensuring the author did not screw it up regardless of any (non-)changes to derivation names.

Which would be a better solution to the stated problem, since this RFC will fail to solve it for derivations with custom `name`s anyway.

In the meantime, IMHO, since changes like #245388, #247977, #248528, #248683 are generally useful and get merged without objections, I think you should apply #49862 to your Nixpkgs, split yet-unfixed changes to `sourceRoot`s from your #153386 into a separate patchset (I think I covered most of your changes with my PRs, but I probably missed at least some of those changes), rebuild your system with #49862 applied and `nameSourcesPrettily = true` or `nameSourcesPrettily = full` set, and PR any changes you need to make it all work.

Posting a nice review in #49862 so that it would get a chance to actually get merged eventually would be nice too, I suppose.

Then, if nobody ever wants to work on the Nix issues described above, and #49862 gets merged, we could discuss setting `nameSourcesPrettily = "full"` for OfBorg as a workaround.
And, IMHO, it should be `"full"`, not `true` (which is essentially what this RFC proposes) so that changing between `fetchgit`, `fetchFromGitHub`, `fetchFromGitLab`, etc would force a recheck too.

risicle:
I certainly think this conversation needs to be revived because of the security implications of

> Nix reuses existing fixed-output outputs between different derivations without actually checking the derivation actually builds into an output with the given output hash

which leaves us open to a potential cache poisoning attack on `cache.nixos.org`, related to https://github.com/NixOS/ofborg/issues/68#issuecomment-2085327052 - using the same `name` for a large proportion of our FODs significantly exacerbates this.

oxij:
@risicle You are correct, it does have security implications, but even doing the equivalent of `nameSourcesPrettily = "full"` of https://github.com/NixOS/nixpkgs/pull/49862 wouldn't really help in the most general case: @tobiasBora in https://github.com/NixOS/ofborg/issues/68#issuecomment-2083351493 shows how to easily poison OfBorg's cache.

Obviously, it won't work for poisoning Hydra, since the evil PR of his method won't ever get merged.

But now thinking about this, I have an IMHO much simpler and thus scarier attack that _will_ work against Hydra:

- Malice makes an evil PR with a package update that changes the source's revision to a new version but points the hash to a hash of an older version of the same package, a version which has a well-known CVE.
- A very paranoid Nixpkgs maintainer Alice (who does not use any build caches) looks at it, it looks fine, so she builds it, and it works fine, since Alice has the older version of the source in `/nix/store`, and so she merges it.
- All tools that track unfixed CVEs in Nixpkgs only look at package versions, so those will be fooled too.
- Malice can now exploit the CVE.

In fact, if that source derivation also has a custom `name` (which is not uncommon), then _nothing_ will catch this (including me, with all my machines not using the Hydra cache and all using the equivalent of `nameSourcesPrettily = "full"` since 2018).

A scary thought.

So, I guess https://github.com/NixOS/nix/issues/969 is actually a critical bug, then.

risicle:
You're _close_ to the scenario detailed by that author to the security team.

Ultimately we can never completely *stop* a malicious change missing the attention of a reviewer (and this goes for all projects, hashed caching scheme or not), but we *can* make it so the submitter has to jump through some more hoops that will hopefully make the PR look weirder and provoke more attention. Hence small ("small") changes like this.

ShamrockLee:
> Eelco and others objecting to the `<hash>-<name>-<version or revision>-source` naming scheme (and to the original 2018 version of NixOS/nixpkgs#49862 which wanted to do exactly this, but for all fetchers, not just `fetchFromGitHub`, and, primarily, for a different reason of `/nix/store` discoverability) do have a point: the current most common `<hash>-source` naming scheme is awfully convenient for running a build server (like Hydra) and when you have to frequently switch between fetchers (e.g. when using out-of-Nixpkgs derivations and/or flakes).

The objection(https://github.com/NixOS/nixpkgs/pull/49862#issuecomment-436585365 and https://github.com/NixOS/nixpkgs/pull/59858#issuecomment-484896229) is no longer relevant, as `sourceRoot = "source"` is officially [deprecated](https://nixos.org/manual/nixpkgs/stable/#var-stdenv-sourceRoot) since Nixpkgs 23.11 when an FOD returned by a `fetchzip`-based build helper is assigned to `src`. At least five rounds of clean up PRs has been merged to eliminate the "`sourceRoot` is `"source"`" assumption.

oxij:
I still think these are two different points:

- Changing most source derivation names no longer breaks most builds, yes.
  (Some do depend on customly named source derivations still, AFAIK. E.g. Apache Arrow, Datafusion, etc.)

  (Also, as the author of both ccbb065c88080966e14872ea9e0ac577f753d902 and 775f21b9fdc1f2bb46518a575339fac8ff867959 in Nixpkgs I feel it kind of a conflict of interest to push for this for those reasons.)

- But moving away from `*-source` naming scheme will still make using flakes inconvenient.

oxij:
I wrote a long post with pros/cons of different possible solutions in https://github.com/NixOS/nix/issues/969#issuecomment-2096916208

So, to 

> That would be an unfortunate cache miss. Nevertheless, isn't it "the Nixy way" to pursue dependency isolation even at the cost of rebuilds (i.e. changing the version of compiler used to compile Bash triggers a world rebuild)?

my answer is that, personally, I would like my build machines to always check all new fixed-output things coming from Nixpkgs, but never check those made by me (unless I ask for it explicitly).

The linked comment describes one way to implement that.

oxij:
@kevincox

Alternatively, someone could just merge a noop-by-default https://github.com/NixOS/nixpkgs/pull/316668 (which I split off from https://github.com/NixOS/nixpkgs/pull/49862) which will also optionally implement this RFC for anyone with `nameSourcesPrettily = true` in their `config.nix` of Nixpkgs or `nixpkgs.config` of `configuration.nix`.

Then, someone could merge the latest incarnation of https://github.com/NixOS/nixpkgs/pull/49862 (which still needs to go to `staging`, given it is a mass-rebuild) to make all `fetch*` derivation names consistent across the whole Nixpkgs.

Which will then reduce this RFC to an argument about changing the default value of `nameSourcesPrettily`.

This seems a much better plan to me, since with those PRs merged both sides of the argument can get what they want at any time by changing one `config` option, and you can actually try out what both sides want first and check if the trade-offs this discusses matter to you.

kevincox:
I don't see much point in merging an option if we don't know that we want to use it. Otherwise we need to consider taking it back out if it ends up that this isn't desirable. I would only recommend merging the implementation first if there are unknowns that would have a major impact on the RFC itself and that can not be properly tested without shipping to most users behind a config option.

I would recommend going forward with the RFC, then if it is accepted the mentioned PRs should be fairly easy to get in. (Code quality concerns would still have to be addressed of course, but the intention will be known to be desirable.)

jonringer:
> I don't see much point in merging an option if we don't know that we want to use it. Otherwise we need to consider taking it back out if it ends up that this isn't desirable. 

Agreed, the main reason why I drafted the RFC is for correctness. I would like for it to be opt-out rather than opt-in.

Using the opt-in might be nice for PR stabilization, but I wouldn't want that for an end state.

philiptaron:
I'm going to resign as shepherd, since I don't know how to do the job.

I agree with this RFC broadly that making, using, and updating fixed-output derivations is too hard in Nix and Nixpkgs. There are a lot of good suggestions here about how to make it better. Without taking a strong position on any particular suggestion, it's a shame that doing these sorts of upgrades on the status quo are so difficult, and the levers of action seem both far away and not able to be pulled by roughly anybody.

infinisil:
> I'm going to resign as shepherd, since I don't know how to do the job.

Ah that's unfortunate, thanks for being explicit about stepping down. Shepherding has an (imo) decent explanation [here](https://github.com/NixOS/rfcs?tab=readme-ov-file#shepherd-team):

> The responsibility of the team is to guide the discussion as long as it is constructive, new points are brought up and the RFC is iterated on and from time to time summarise the current state of discussion. If this is the case no longer, then the Shepherd Team shall step in with a motion for FCP.

asymmetric:

PR #175, opened 2024-04-25
[RFC 175] Memorandum on Equitable Moderation

nrdxp:
[Rendered](https://github.com/nrdxp/rfcs/blob/rfc-175/rfcs/0175-appeals-council.md)

Join the discussion on matrix. Please note, we only keep it open during certain hours:
https://matrix.to/#/#rfc-175-all-together:matrix.org

PR #98, opened 2021-08-03
Community Team RFC

IreneKnapp:
This is a proposal for the direction of the NixOS community, by myself and ashkitten. We've been working on this for a while and are very pleased to be publicly presenting it. We welcome public discussion of the proposal's merits and will try our best to respond to comments - have at it!

PR #176, opened 2024-04-29
Don't moderate discussions about moderation

clefru:
Implement "nemo iudex in sua causa" to ensure that moderation powers cannot be used to censor discussions about moderation.

I am also removing the twitter link, which is dead.

### DIFF
--- a/rfcs/0102-moderation-team.md
+++ b/rfcs/0102-moderation-team.md
@@ -49,7 +49,7 @@ for, and where all are given opportunity to participate meaningfully. The
 Foundation will work with the community in service of this goal.
 ```
 
-ref: [twitter](https://twitter.com/grhmc/status/1390775249424338944)
+Discussions about moderation must not be moderated.
 
 # Examples and Interactions
 [examples-and-interactions]: #examples-and-interactions


clefru:
Motivation linked here: https://www.reddit.com/r/NixOS/comments/1cfv8vo/moderation_nogo_zones/

FireyFly:
This doesn't seem to follow the RFC process as it patches an already accepted RFC. The removal of a citation seems questionable--surely it'd be better to replace the link with one pointing to wayback or archive.is?

I think the proposed new clause is also untenable, as it would allow any conduct in a discussion that happens to be about moderation, including any amount of ad-hominem attacks etc (since the discussion "must not be moderated"). I don't think anyone wants that?

piegamesde:
While I can get behind "nemo iudex in sua causa", this doesn't mean that there should be no moderation *at all*.

Take https://discourse.nixos.org/t/why-was-jon-ringer-banned-from-github/44114/18 for example, which is literally making shit up. (My first thought when reading this was "huh Jon has a website?") The post follows up with insinuation. This is the level of discourse quality you'd get without moderation.

poscat0x04:
>  Take https://discourse.nixos.org/t/why-was-jon-ringer-banned-from-github/44114/18 for example, which is literally making shit up. (My first thought when reading this was "huh Jon has a website?") The post follows up with insinuation. This is the level of discourse quality you'd get without moderation.

I find this example unconvincing, there was no need to use moderation power there, you could've just pointed it out in a reply

piegamesde:
So the thing is, replying instead of hiding has the following issues:

- You have to accept the framing of the original post in order to reply to it
- You show that lying is a stance worthy of discussion
- You are now discussing that side tangent about something which is blatantly false, which is both derailing the actual discussion and taking up space.

for a healthy discussion culture, it is crucial to make clear that some things are plainly *undiscussable*. This a different kind of disagreement to the usual "I think you're wrong", which indeed can generally be discussed.

So no, hiding it was not an abuse of power, it was the most sensible thing to do for the wellbeing of the overall discussion.

poscat0x04:
A live example of why we shouldn't let mods judge themselves :)

piegamesde:
@poscat0x04 and a live example how a conversation just got derailed by a concern troll because you replied to it :)

poscat0x04:
To say that other people's as *undiscussable* is just... full of arrogance.

piegamesde:
Yes, if people just make things up about me I'll call them a troll. I'm not sure what more evidence you'd need in addition to "says blatantly false things".

Also yes, I am arrogant and condescending right now because after six months of doing this shit I am tired of armchair moderators making blatantly unworkable\* proposals. If you want to be taken seriously, then make a serious proposal. A one-line diff spit onto an existing RFC without even following the RFC process is *not* something which I can take seriously. I am interested in having an honest discussion about moderation, and I know the current issues very well and they need discussing, but it needs to be done in good faith. Which I currently fail to see.

----

\* And I recognize that many of these are unworkable [by](https://github.com/NixOS/rfcs/pull/175#issuecomment-2078721888) [design](https://github.com/NixOS/rfcs/pull/175#discussion_r1580575376), done by people who'd prefer having no moderation at all. (And yes, I'm aware that Hanlon's razor still applies here)

AshleyYakeley:
piegamesde are you even a moderator anymore? I thought you quit.

Anyway while I definitely understand where this is coming from, in practice this RFC is not robust against bad actors and I don't think it's workable.

piegamesde:
@AshleyYakeley correct, I announced to step down a couple of days ago and followed through today.

schmittlauch:
I see massive abuse potential if it suffices to sprinkle in enough "this is actually about moderation" in your post to give you a pass for saying things that are obviously acceptable otherwise. A blatant example is putting out racist statements in a "just asking (moderation) questions" style, and thus getting a pass on hateful incitement.
Or a more obvious but simplified example:
»But given that @.clefru is eating newborns for breakfast, his opinions on moderation are clearly biased.«

I assume this is the idea of some things being plainly undiscussible that @piegamesde mentioned.
It's also still worth to differentiate between the liberty of voicing your own opinion and stating your own (supposed, but verifiable untrue or unfounded) *facts*.

Another issue I see is that in the age of LLMs, discussions might easily be drowned in made-up posts that must not be marked or hidden as long as they resemble a *discussion about moderation* closely enough. [No, feeding the whole discussion into an LLM for summary is not an appropriate response.]

tomberek:
@clefru I wrote the original RFC102 as a compromise and as a way to do something rather than nothing during RFC98 development. It was meant to be something that was as uncontroversial as I could make it and bring in all parties to agree. Perhaps it was wrong, or not enough, but it is what was agreed upon at the time. I don't think a modification to it is helpful right now. Even the twitter link, though dead, helps point to how the contents were put together.

I ask that you close this PR.

I understand and sympathize with the point I think you are trying to make, but this is not the way to go about it. The "nemo iudex in sua causa" concept is about recusing oneself as an individual, not about removing jurisdiction of the court altogether. Yes, there should be an understanding that moderators should refrain from also participating in discussions/debates because it can create the impression of a conflict of interest. Yes, this is a part of why there has been degraded trust between people. What is needed is for there to be greater trust, and I don't see this modification doing that. I hope that the idea behind this can be included in ongoing moderation - but I don't see this specific PR helping, perhaps elsewhere or at a more opportune time.

Note: I made a similar plea during [RFC98  - comment](https://github.com/NixOS/rfcs/pull/98#issuecomment-962058845), for nearly the the same reason - that I did not see that situation proceeding well.

kevincox:
From the RFC Steering Committee:

This RFC does not fit the [RFC Process](https://github.com/NixOS/rfcs/blob/master/README.md). Most notably RFCs should create new RFC documents and existing RFC documents shouldn't have significant changes made to them.

Please adjust this proposal to fit the RFC Process or it will be closed. Note that this doesn't prevent this this or other RFCs on this topic being opened/reopened in the future as log as the proposal follows the defined process.

clefru:
Thanks for the feedback. I agree that this doesn't fit the RFC process and should not go through it, but not for the lack of fitting the RFC process, but for the RFC process being unsuitable for this discussion. It is unfortunate that the 102-moderation-team.md ended up in this repository. It should be moved eventually, as it is undesirable that such a document is immutable.

Please leave the PR open until the new governance structure is in place and can review the new policies. I converted it to a Draft to remove confusion. 

I suspect there will be a large moderation overhaul given how heated up things are at the moment. If no proper accountability of moderation is implemented in due time, I will reserve the right to petition the new governance body for this change, with this PR being the suggested implementation.

kevincox:
Sounds good. As long as the PR is in Draft (or Closed) it is generally expected to be left alone and the RFCSC will just ignore it. So feel to leave it in Draft for as little or as long as you like.

PR #177, opened 2024-04-30
[RFC 177](Pre RFC) Overhaul of CoC and Moderation Practices

poscat0x04:
[rendered](https://github.com/poscat0x04/nix-rfcs/blob/coc-moderation/rfcs/0177-coc-moderation.md)

### DIFF
--- a/rfcs/0177-coc-moderation.md
+++ b/rfcs/0177-coc-moderation.md
@@ -1,0 +1,68 @@
+---
+feature: Coc & Moderation Overhaul
+start-date: 2024-04-30
+author: poscat
+co-authors: (find a buddy later to help out with the RFC)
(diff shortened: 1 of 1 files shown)


tomberek:
Please mark this as a draft. It is not currently in a re-viewable state.

nyabinary:
Shouldn't of this been posted to the Discourse forums first as a pre-rfc before being posted here?

AndersonTorres:
> Shouldn't of this been posted to the Discourse forums first as a pre-rfc before being posted here?

Not necessarily. Usually RFCs are drafted here when the idea is to call the most diverse possible people.
Not everyone has a login on Discourse or other platforms, while everyone that uses Nixpkgs has a GitHub account.

PR #178, opened 2024-05-28
Add `RFCSC:` prefix to canned messages.

kevincox:
The RFC Steering Committee has started using an `RFCSC:` prefix to their messages posted on behalf of the committee to provide indication between personal opinion and communication from the committee. This updates the canned messages to include that prefix for consistency.


MMesch (approved):
LGTM

PR #148, opened 2023-05-23
[RFC 0148] Pipe operator

piegamesde:
[**Rendered**](https://github.com/piegamesde/rfcs-1/blob/pipe/rfcs/0148-pipe-operator.md)

**Discussion notice:** please try to attach all discussions to a thread by using the code review feature. If your comment doesn't refer a specific line to attach to, use the header line instead.

PR #181, opened 2024-07-14
[RFC 0181] List index syntax

rhendric:
[Rendered](https://github.com/rhendric/nix-rfcs/blob/rhendric/list-index-syntax/rfcs/0181-list-index-syntax.md)

### DIFF
--- a/rfcs/0181-list-index-syntax.md
+++ b/rfcs/0181-list-index-syntax.md
@@ -1,0 +1,290 @@
+---
+feature: list-index-syntax
+start-date: 2024-07-14
+author: rhendric
+co-authors:
+shepherd-team:
+shepherd-leader:
+related-issues: https://github.com/NixOS/nix/issues/10949, https://github.com/NixOS/rfcs/pull/137
+---
+
+# Summary
+[summary]: #summary
+
+This proposal extends the attrpath syntax to include `'[' INT ']'` elements that refer to values in lists.
+This would enable expressions such as the following:
+
+```nix
+x.[0]       # = builtins.elemAt x 0
+x.[1] or y  # = if builtins.isList x && builtins.length x > 1 then builtins.elemAt x 1 else y
+x ? [3].y   # = builtins.isList x && builtins.length x > 3 && builtins.elemAt x 3 ? y
+```
+
+# Motivation
+[motivation]: #motivation
+
+I'm in a REPL.
+I'm exploring parts of Nixpkgs.
+I type a partial expression:
+
+```
+nix-repl> someExpr.foo
+{ bar = { ... }; ignoreThis = { ... }; moreStuff = { ... }; }
+```
+
+I hit up-arrow and keep typing to drill deeper:
+
+```
+nix-repl> someExpr.foo.bar
+{ baz = true; qux = [ ... ]; quux = { ... }; }
+
+nix-repl> someExpr.foo.bar.qux
+[ { ... } ]
+
+nix-repl> someExpr.foo.bar.qux.0
+error: attempt to call something which is not a function but a list
+
+       at «string»:1:1:
+
+            1| someExpr.foo.bar.qux.0
+             | ^
+```
+
+Of course that doesn't work.
+I don't actually type that.
+What I actually do is hit up-arrow, then hit Home, then type `builtins.elemAt`, then hit End, then type ` 0`.
+
+```
+nix-repl> builtins.elemAt someExpr.foo.bar.qux 0
+{ greatMoreStuff = { ... }; }
+```
+
+Now what do I get to do?
+That's right, hit up-arrow, then hit Home, then type `(`, then hit End, then type `).greatMoreStuff`.
+
+---
+
+When writing Nix code, it is relatively uncommon to want to index into a list, and `builtins.elemAt` suffices.
+When exploring data in a REPL, however, indexing into lists is more common, and the above example illustrates how `elemAt` is incompatible with the attrpath selector syntax that is the primary means of data exploration.
+Many other programming languages allow syntaxes such as
+`foo.bar.qux[0].moreStuff` (C, many others)
+or `foo.bar.qux(0).moreStuff` (Octave, Scala)
+or `foo.bar.qux.[0].moreStuff` (F#, OCaml)
+or `foo.bar.qux.0.moreStuff` (Coco, LiveScript, the `--attr` option of `nix-instantiate` and other `nix` commands).
+Of these, only `.[]` does not conflict with existing syntax (see [Alternatives] for more details).
+
+# Detailed design
+[design]: #detailed-design
+
+The `attrpath` grammar nonterminal is currently defined as
+
+```
+attrpath
+  : attrpath '.' attr
+  | attrpath '.' string_attr
+  | attr
+  | string_attr
+  ;
+```
+
+where `attr` is a simple identifier and `string_attr` is either a string literal or a bare `${}` interpolation.
+This nonterminal is used in three contexts in the grammar:
+* In a selector (e.g. `expr.attrpath` or `expr.attrpath or default`)
+* As the right-hand side of the `?` operator (e.g. `expr ? attrpath`)
+* As the left-hand side of a binding (e.g. `let attrpath = expr; in body` or `{ attrpath = expr; }`)
+
+This proposal adds two productions to `attrpath`:
+
+```
+  | attrpath '.' '[' INT ']'
+  | '[' INT ']'
+```
+
+(Supporting non-literal expressions is scoped out of this proposal; see [Future work][future].)
+
+In a selector or `?` operator, these new productions have the following semantics:
+* `expr.prefix.[n].suffix` is the result of evaluating `(builtins.elemAt (expr.prefix) n).suffix`
+* `expr ? prefix.[n].suffix` is true if and only if all of the following are true:
+  * `expr ? prefix` (or `prefix` is empty)
+  * `expr.prefix` evaluates to a list with length at least `n + 1`
+  * `expr.prefix.[n] ? suffix` (or `suffix` is empty)
+* `expr.prefix.[n].suffix or expr2` is the result of evaluating `if expr ? prefix.[n].suffix then expr.prefix.[n].suffix else expr2`
+
+It is a syntax error if either new production is used in the left-hand side of a binding.
+
+An implementation of this design is available as patches for Nix at <https://gitlab.com/rhendric/nix-list-index-syntax/>; see instructions there for use.
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+```
+nix-repl> [ 1 4 9 ].[2]
+9
+
+nix-repl> pkgs.maptool.meta.maintainers.[0].github
+"rhendric"
+
+nix-repl> [ 1 4 9 ].[3]
+error: list index 3 is out of bounds
+
+       at «string»:1:1:
+
+            1| [ 1 4 9 ].[3]
+             | ^
+
+nix-repl> [ 1 4 9 ] ? [2]
+true
+
+nix-repl> [ 1 4 9 ] ? [3]
+false
+
+nix-repl> [ 1 4 9 ] ? [0].[0]
+false
+
+nix-repl> [ [ 1 4 9 ] ] ? [0].[0]
+true
+```
+
+The following are all syntax errors:
+
+```
+nix-repl> let n = 0; in [ 4 ].[n]
+error: syntax error, unexpected ID, expecting INT
+
+       at «string»:1:22:
+
+            1| let n = 0; in [ 4 ].[n
+             |                      ^
+
+nix-repl> { a.[0] = true; }
+error: syntax error, index '0' not allowed here
+
+       at «string»:1:3:
+
+            1| { a.[0] = true;
+             |   ^
+
+nix-repl> let a.[0] = true; in a                   
+error: syntax error, index '0' not allowed here
+
+       at «string»:1:5:
+
+            1| let a.[0] = true;
+             |     ^
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The prototype [patch] implementing this feature adds a net 69 lines of code to Nix, excluding tests.
+
+[patch]: https://gitlab.com/rhendric/nix-list-index-syntax/-/blob/main-dist/patches/bracketed/from-90e630a5.patch
+
+As noted in passing in the motivation section, the proposed syntax differs from the syntax already used by Nix tools on the command line for `--attr`.
+`nix-instantiate` has no difficulty with:
+
+```
+$ nix-instantiate --eval '<nixpkgs>' -A maptool.meta.maintainers.0.github
+"rhendric"
+```
+
+This divergence might lead to confusion.
+
+Evolving the syntax of Nix always imposes a cost on third-party tools that process Nix syntax, including syntax highlighters, linters, formatters, static analyzers, and language servers.
+This syntax is not a dramatic extension of the language but would require support from all of the above for them to maintain full functionality.
+
+Finally, there is an opportunity cost to claiming new syntax.
+One could imagine speculative features that might want to use this syntax, such as a list or string slicing syntax, or a ‘list swizzle’ operator that desugars `expr.[ 2 0 1 ]` to `[ (elemAt expr 2) (elemAt expr 0) (elemAt expr 1) ]`.
+It is, in my opinion, unlikely that list and string manipulation (assuming that any feature in competition for this syntax would involve lists or strings somehow) would be so common in Nix to make this a compelling objection.
+
+# Alternatives
+[alternatives]: #alternatives
+
+#### `expr.${0}`
+
+The simplest alternative, aside from doing nothing, would be to reuse the `.${expr}` syntax that Nix already supports in attrpaths.
+In this alternative, when evaluating `expr.${idx}`, the interpreter would determine whether `idx` evaluates to an integer or a string.
+`expr` would be required to be a list if the former and an attribute set if the latter.
+
+While this is parsimonious with respect to syntax, it creates more complexity for static analyzers.
+It is currently statically known that the `expr` in `expr.${idx}` must be a set and that `idx` must be a string.
+Representing a new constraint that _either_ `expr` is a set and `idx` is a string _or_ `expr` is a list and `idx` is an integer might be difficult for such tools and could result in a degraded user experience.
+
+#### `expr[1]`
+
+The most common syntax, by far, for indexing into a list or array in other programming languages is `expr[idx]`.
+Nix is currently whitespace-insensitive with respect to attrpaths, so such an expression is indistinguishable from `expr [idx]`, an application of the function `expr` to the argument `[idx]`.
+It would be possible to adapt the Nix lexer to interpret the `[` character in `expr[idx]` as the opening of an indexer or a list depending on whether there is whitespace before it.
+Among the disadvantages of this approach would be that it would make specifying the grammar of Nix more complicated, and that it may prevent long attrpaths from being broken naturally across lines.
+Perhaps most fatally, despite the odds of someone writing `expr[idx]` and intending a function call being virtually nil, the conservative principles of the Nix team forbid altering the meaning of even such an unlikely bit of syntax without some sort of larger language versioning or deprecation story, which has eluded us for some time.
+
+All of the above objections apply to `expr(1)` as well, with the additional drawback that in a wide array of C-like languages, this syntax represents exactly what it currently (if coincidentally) represents in Nix.
+
+#### `expr.$[2]`
+
+The syntax `expr.$[idx]` was offered as a possibility in [NixOS/nix#10949](https://github.com/NixOS/nix/issues/10949).
+It resembles the `${}` syntax already used in attrpaths, with the change in delimiters suggesting a shift from attribute sets to lists.
+However, it requires an additional character to type and its technical qualities are identical to those of the proposed syntax without the `$` character.
+There is at least some prior art for `.[]` in OCaml and F#; there is none that I know of for `.$[]`.
+
+#### `expr.3`
+
+It is no accident that a simple dotted index `expr.3` was the syntax chosen for attribute paths in `nix-instantiate` and friends.
+It is the most straightforward expression of intent imaginable, if it is known for certain that an attribute path is what is intended.
+Implementing the same syntax in the Nix language would be harmonious.
+The drawback is that, as with the `expr[1]` case, whitespace insensitivity means that `expr.3` is indistinguishable from `expr .3`, the application of a function to a float literal.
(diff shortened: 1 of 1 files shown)


rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
+---
+feature: list-index-syntax
+start-date: 2024-07-14
+author: rhendric
+co-authors:
+shepherd-team:
```
infinisil, line 6: I've been wanting this for so long, very nice proposal, nominating myself as a shepherd!

```suggestion
shepherd-team: @infinisil
```

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+* `expr.prefix.[n].suffix` is the result of evaluating `(builtins.elemAt (expr.prefix) n).suffix`
+* `expr ? prefix.[n].suffix` is true if and only if all of the following are true:
+  * `expr ? prefix` (or `prefix` is empty)
+  * `expr.prefix` evaluates to a list with length at least `n + 1`
+  * `expr.prefix.[n] ? suffix` (or `suffix` is empty)
+* `expr.prefix.[n].suffix or expr2` is the result of evaluating `if expr ? prefix.[n].suffix then expr.prefix.[n].suffix else expr2`
+
+It is a syntax error if either new production is used in the left-hand side of a binding.
+
+An implementation of this design is available as patches for Nix at <https://gitlab.com/rhendric/nix-list-index-syntax/>; see instructions there for use.
```
inclyc, line 115: I don't expect much regression, but have you tested current nixpkgs trunk?

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
+---
+feature: list-index-syntax
+start-date: 2024-07-14
+author: rhendric
+co-authors:
+shepherd-team: infinisil
```
inclyc, line 6:
```suggestion
shepherd-team: infinisil inclyc
```

May I add myself? (Currently maintaining the LSP, so I'd like to subscribe syntax changes!)

sg-qwt:
>  When writing Nix code, it is relatively uncommon to want to index into a list, and `builtins.elemAt` suffices.

My 2cents is that I'd still prefer the simplicity of `builtins.elemAt` than introducing new syntax. The usability of nix repl is poor compared to other lisp languages. However, maybe that's something can be worked on the tooling level so that writing and interacting Nix code in repl can be the same experience as writing Nix code other places.

Thus, if exploring data in a REPL is the only motivation for introducing this new syntax,  I feel like it doesn't hold a very strong argument this case.

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+This would enable expressions such as the following:
+
+```nix
+x.[0]       # = builtins.elemAt x 0
+x.[1] or y  # = if builtins.isList x && builtins.length x > 1 then builtins.elemAt x 1 else y
+x ? [3].y   # = builtins.isList x && builtins.length x > 3 && builtins.elemAt x 3 ? y
+```
+
+# Motivation
+[motivation]: #motivation
```
infinisil, line 24: @sg-qwt Let's try to keep the discussions in threads, so they can be easily marked as resolved later

---

@sg-qwt [said](https://github.com/NixOS/rfcs/pull/181#issuecomment-2228077835):

> >  When writing Nix code, it is relatively uncommon to want to index into a list, and `builtins.elemAt` suffices.
>
> My 2cents is that I'd still prefer the simplicity of `builtins.elemAt` than introducing new syntax. The usability of nix repl is poor compared to other lisp languages. However, maybe that's something can be worked on the tooling level so that writing and interacting Nix code in repl can be the same experience as writing Nix code other places.
>
> Thus, if exploring data in a REPL is the only motivation for introducing this new syntax,  I feel like it doesn't hold a very strong argument this case.

---

A primary other use case is `builtins.{match,split}`, which returns matched groups in indexed lists. A [GitHub search](https://github.com/search?q=elemAt+lang%3ANix&type=code) finds many examples of this. Though for this use case it might be better to introduce matching group labels, so that `(builtins.match "(?<foo>.*)" "bar").foo == "bar"` would work.

rfcs/0181-list-index-syntax.md
rhendric: I tested one build on the `nixos-unstable` channel as of yesterday, but not literally trunk.

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+This divergence might lead to confusion.
+
+Evolving the syntax of Nix always imposes a cost on third-party tools that process Nix syntax, including syntax highlighters, linters, formatters, static analyzers, and language servers.
+This syntax is not a dramatic extension of the language but would require support from all of the above for them to maintain full functionality.
+
+Finally, there is an opportunity cost to claiming new syntax.
+One could imagine speculative features that might want to use this syntax, such as a list or string slicing syntax, or a ‘list swizzle’ operator that desugars `expr.[ 2 0 1 ]` to `[ (elemAt expr 2) (elemAt expr 0) (elemAt expr 1) ]`.
+It is, in my opinion, unlikely that list and string manipulation (assuming that any feature in competition for this syntax would involve lists or strings somehow) would be so common in Nix to make this a compelling objection.
```
piegamesde, line 198: To me, this is a major blocker for the proposed syntax. While I agree that list slices are fairly uninteresting, this syntax could instead be used for *set* slicing, as drafted out here: To me, this is a major blocker for the proposed syntax. While I agree that list slices are fairly uninteresting, this syntax could be used with tremendous benefits for set slicing, as drafted out here: https://md.darmstadt.ccc.de/nix2?view=#Set-slicing-confidence-mid (ignore the comma separated lists, which are a separate language improvement proposal)

Therefore, I propose going with the `list.NUMERAL` approach, at the cost of having to remove the unnecessary `.3` floating literals first.

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+
+# Alternatives
```
piegamesde, line 200: Another alternative which might be worth mentioning would be `foo |> getElem 2` using the pipe operator (RFC #148). I think this may work sufficiently okay in a REPL, however in Nix code this may interfere with existing `|>` pipelines and thus require parentheses

rhendric: Even in a REPL, the motivating case of `(foo.bar.qux |> getElem 0).moreStuff` requires parentheses.

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
+---
+feature: list-index-syntax
```
AndersonTorres, line 2: Since we are here, what about extending this to sets?

```nix
{ x=1; y=2; }.["x"]
```

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+# Summary
+[summary]: #summary
+
+This proposal extends the attrpath syntax to include `'[' INT ']'` components that refer to elements in lists.
+This would enable expressions such as the following:
+
+```nix
+x.[0]       # = builtins.elemAt x 0
+x.[1] or y  # = if builtins.isList x && builtins.length x > 1 then builtins.elemAt x 1 else y
+x ? [3].y   # = builtins.isList x && builtins.length x > 3 && builtins.elemAt x 3 ? y
```
AndersonTorres, line 20: Ugly.

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+
+# Summary
+[summary]: #summary
+
+This proposal extends the attrpath syntax to include `'[' INT ']'` components that refer to elements in lists.
+This would enable expressions such as the following:
+
+```nix
+x.[0]       # = builtins.elemAt x 0
+x.[1] or y  # = if builtins.isList x && builtins.length x > 1 then builtins.elemAt x 1 else y
```
AndersonTorres, line 19: I believe it should abort evaluation unless it is an indexable.

rfcs/0181-list-index-syntax.md
rhendric: Why? You can already write `{ x = 1; y = 2; }."x"`.

rfcs/0181-list-index-syntax.md
rhendric: In the current language, `or` doesn't cause an abort if the selection target isn't a set:

```nix
nix-repl> null.x or false
false
```

Why should this proposal follow different rules?

rfcs/0181-list-index-syntax.md
piegamesde: I think this is a good argument why the `.0` syntax is more consistent and should be preferred.

rfcs/0181-list-index-syntax.md
AndersonTorres: I was hit by a precedent, then forget this.

rfcs/0181-list-index-syntax.md
numinit: I'm in agreement. Unifying all the composite type elements under one operator is probably going to be nicer long-term.

The `.${numeral}` syntax reminds me of JS a bit and is pretty intuitive. If Nix is just high level JSON and we already have `.foo`, we could have `.42` too. (In other words, I get why `[numeral]` was chosen because JS works that way for both objects and arrays, but Nix doesn't need to have two syntaxes for composite type indexing, I don't think.)

rfcs/0181-list-index-syntax.md
inclyc: > Another alternative which might be worth mentioning would be `foo |> getElem 2` using the pipe operator

I agree that "pipe operator" is good but not a real alternative of this case.

rfcs/0181-list-index-syntax.md
MattSturgeon: > On the other hand, in guise of speculation, with the bracketed syntax we can have nice things like
>
> `.[ "x" "y" ]`

I think this should be discussed in its own RFC.

IMO `.int` (e.g. `.2`) is consistent with the attr indexing syntax `."name"`.

Since attr names can't be int literals, this fits in nicely.

`.${expression}` indexing will cover more complex scenarios too.

This would leave `.[/*other syntax*/]` available for any future use, to be tackled by another RFC.

kvtb:
If you ask a LLM to write llistindex-heavy Nix code (e.g. base64 encode/decode), it will likely write it using `!` as list index operator, as though it already exists.

This suggests that there might be a collective inclination towards using `!` in this context.

rfcs/0181-list-index-syntax.md
MattSturgeon: Agreed, `[ "hello" "world" ].1 => "world"` is the way forward IMO.

`builtins.head` would be equivalent to `list: list.0`.

`builtins.elemAt` would be equivalent to `list: idx: list.${idx}`.

piegamesde: > builtins.elemAt would be equivalent to list: idx: list.${idx}.

I'm not so sure about that part. Basically `list.${idx}` is syntactically identical to the existing `set.${attr}`, which means that an implementation would need to make a case distinction at runtime on the type of the selector.

MattSturgeon: I think it only follows naturally. In `{ }.foo` (and `{ }."bar"`) vs `[ ].42` we're already differentiating based on the type of the _literal_, why not also differentiate based on the type of the evaluated `${123}` expression?

piegamesde: One of them can be done at parse time, the other one has to be delayed until the expression is forced. Btw I don't know the actual performance impact this might have, but I'd be very wary of it

rhendric: As I describe [in the Alternatives section](https://github.com/rhendric/nix-rfcs/blob/rhendric/list-index-syntax/rfcs/0181-list-index-syntax.md#expr0), the main concern with that is less about the evaluator's performance and more about external static analysis tools.

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
...
+All of the above objections apply to `expr(1)` as well, with the additional drawback that in a wide array of C-like languages, this syntax represents exactly what it currently (if coincidentally) represents in Nix.
+
+#### `expr.$[2]`
+
+The syntax `expr.$[idx]` was offered as a possibility in [NixOS/nix#10949](https://github.com/NixOS/nix/issues/10949).
+It resembles the `${}` syntax already used in attrpaths, with the change in delimiters suggesting a shift from attribute sets to lists.
+However, it requires an additional character to type and its technical qualities are identical to those of the proposed syntax without the `$` character.
+There is at least some prior art for `.[]` in OCaml and F#; there is none that I know of for `.$[]`.
+
+#### `expr.3`
```
rhendric, line 230: Everyone arguing for this alternative: you don't have to convince me of its technical superiority. I am already convinced, as I describe below. This is the syntax I'm using in my personal Nix build.

Rather, you have to convince me that the Nix maintainers can be persuaded to make a breaking change to float literal syntax on a timeline that is less than, say, five years, when they have expressed reluctance to make ‘breaking changes’ to the language even with respect to [obvious](https://github.com/NixOS/nix/pull/10025) buggy [behavior](https://github.com/NixOS/nix/pull/11121).

Gating this proposal behind something that will never happen is, in practice, equivalent to rejecting the proposal. I would rather see *some* syntax implemented in a couple of years than wait indefinitely for the *ideal* syntax.

And if sane language evolution manages to actually become policy at some point, in that case we'll have the tools we need to deprecate the less-appealing syntax and migrate to the better one, should we choose to do so.

kvtb: With `expr.3` how would be non-literal index look like?

Obviously, `expr.n` is already mean `expr.${"n"}`.

If you introduce `expr.(n)` or `expr.[n]` for non-literal index, should they work with literals, should `expr.(3)` or `expr.[3]` be valid too?

If yes, why duplicating and confusing `expr.3` ?

rhendric: This proposal is only for when the desired index is known; replacing `elemAt` for variable indices is left as [future work](https://github.com/rhendric/nix-rfcs/blob/rhendric/list-index-syntax/rfcs/0181-list-index-syntax.md#future-work).

kvtb: I believe that adopting `expr.3` without addressing the implications for non-literal indices could lead to significant challenges down the road. If we postpone even the discussion on non-literal indices, we risk creating a legacy issue that will be difficult and cumbersome to address in the future.

While we might agree on `expr.3` today, this decision could result in a syntax that feels outdated and problematic later on, making it harder to evolve the language cleanly.

kevincox: I think I agree here. If we agree that `elemAt` is the future then that is fine. But I think if we agree that we want to improve the dynamic syntax in the future it is best to have that discussion now to ensure that we don't shoot ourselves in the foot.

rhendric: There are several options:
* `elemAt` is good enough
* Make `.${n}` work with both numbers on lists as well as strings on sets
* Introduce `.$[n]` as a parallel construction to `.${str}`
* If this proposal is adopted as-is, simply extend `.[n]` to also allow expressions

I'm confident that regardless of whether this proposal is adopted as-is or if the `expr.3` alternative is chosen instead, there are non-problematic ways forward if we *do* agree that `elemAt` doesn't suffice (and I don't think we do; at least, I'm much less concerned about that personally).

kvtb: Also, besides the space issue with float, `expr.3` opens a gate to hard to read space issues like `expr.3a` (which is a function application `expr.3 a`) or `expr.1e+3`

rhendric: It doesn't really open that gate; `3a` and `1e+3` are already ill-advised things to write in Nix.

kvtb: And for floats, even if we forbid the `.3` syntax for floats, how this would be parsed?
`let a0=1; in [3a0.3a0]`

`[3 a 0.3 a0]` or `[3 a0.$[3] a0]`

kvtb: > * Make `.${n}` work with both numbers on lists as well as strings on sets
this is already criticized above for runtime overhead and issues with static analyzers.

> * Introduce `.$[n]` as a parallel construction to `.${str}`
> * If this proposal is adopted as-is, simply extend `.[n]` to also allow expressions

These are not truly parallel, `[]` is a bit different in Nix than `{}` and `[]`: while `${a b}` and `(a b)` are function applications, `[a b]` is not.

How to read `.$[a b]` or `.[a b]` then?

kvtb: I would add `expr.(3)` or `expr.(n)` variant to the proposal, it seems to be the less confusing.

kvtb: > `3a` and `1e+3` are already ill-advised things to write in Nix.

I mean, currently, such code likely hit "attempt to call something which is not a function but an integer" error
but with `expr.3` one could mistakenly write `expr.0x29A` and she may get no error even at runtime.
Just wrong result at the end.

kvtb: I would add `expr.(3)` or `expr.(n)` variant to the proposal, it seems to be the less confusing, for

1. no issues with floats
2. no issues with some letter or symbol glued on the right side
3. no issues with sets: `.(a)` and `.("a")` are already forbidden to use with sets; really forbidden, not just considered ill-formed, there is guarantied no such existing Nix code
4. no issues with spaces: `expr.(a)` is `expr .(a)` is `expr. (a)` is `expr . ( a )`
5. unlike `[]`, priorities inside `()` are surprise-free: `expr.(a b)` means application like in `expr.${a b}`

kvtb:
I would add `expr.(3)` or `expr.(n)` variant to the proposal, it seems to be the less confusing, for

1. no issues with floats
2. no issues with some letter or symbol glued on the right side
3. no issues with sets: `.(a)` and `.("a")` are already forbidden to use with sets; really forbidden, not just considered ill-formed, there is guarantied no such existing Nix code
4. no issues with spaces: `expr.(a)` is `expr .(a)` is `expr. (a)` is `expr . ( a )`
5. unlike `[]`, priorities inside `()` are surprise-free: `expr.(a b)` means application like in `expr.${a b}`

piegamesde:
I personally find it fairly confusing to use parentheses in syntax that is not a grouping operator. We already have a precedent for it in the language, `inherit (from)` and I really dislike it as well.

rfcs/0181-list-index-syntax.md
piegamesde: > Rather, you have to convince me that the Nix maintainers can be persuaded to make a breaking change to float literal syntax on a timeline that is less than, say, five years

At least for Lix, I am currently setting up [infrastructure](https://gerrit.lix.systems/c/lix/+/1735?tab=comments) to [remove URL literals](https://gerrit.lix.systems/c/lix/+/1736), and the `.3` floating literals are guaranteed to be next.

However, I can't guarantee that adding the new syntax will be allowed without proper language versioning tools (which are on the roadmap but it will take years still).

> elemAt is good enough

I'll join that camp. Especially since the main motivation for this RFC is the REPL, where the user knows which index they want to use.

rhendric: (What other projects do is their business; this is a Nix RFC.)

For everyone using original Nix, a reminder that you can have either the proposed syntax or the `expr.3` alternative in your REPL [right now](https://gitlab.com/rhendric/nix-list-index-syntax), on any stable or unstable version of Nix, without waiting for anyone else to make any decisions about what is allowed.

AndersonTorres:
> We already have a precedent for it in the language, `inherit (from)` and I really dislike it as well.

Off-topic: I think we lost a great chance for suggesting `from lib inherit cmakeBool mesonBool;`...

AndersonTorres:
I am inclined to shepherd this for the sake of advancing the RFC, however I am not so sure if I can deal with three RFCs at the same time.

kevincox:
RFCSC: @AndersonTorres if you are able to spare the time it would be much appreciated as that would be enough to complete the shepherd team and move the RFC forward.

AndersonTorres:
OK then, I recruit myself as shepherd.

rfcs/0181-list-index-syntax.md
```diff
@@ -0,0 +1,290 @@
+---
+feature: list-index-syntax
+start-date: 2024-07-14
+author: rhendric
+co-authors:
+shepherd-team: infinisil inclyc
+shepherd-leader:
```
GetPsyched, line 7:
```suggestion
shepherd-team: infinisil inclyc AndersonTorres
shepherd-leader: infinisil
```

GetPsyched (commented):
RFCSC:

This RFC has acquired enough shepherds; moving to in discussion.

rfcs/0181-list-index-syntax.md
infinisil, line 1: @rhendric @inclyc @AndersonTorres (but anybody is welcome to join of course): I created [#nix-rfc181](https://matrix.to/#/#nix-rfc181:matrix.org) to have more synchronous discussions. I'd like to propose having a regular 2-weekly meeting to make continuous progress on this. If that sounds fine, please enter your availability [here](https://crab.fit/rfc-181-625632) so we can find a time that works :slightly_smiling_face:

infinisil: It looks like our availability has close to zero overlap, so I don't think that's going to work. Instead let's discuss it semi-synchrounously in the Matrix channel at least. @rhendric and @inclyc can you also join [#nix-rfc-181:matrix.org](https://matrix.to/#/#nix-rfc181:matrix.org)?

infinisil:
Brief update: All shepherds (and the author) now joined the [Matrix channel](https://matrix.to/#/#nix-rfc181:matrix.org) and we're having some initial discussions.

infinisil:
Brief update: I initiated some discussions in the Matrix room, waiting for the shepherds/author to respond (ping @rhendric, @AndersonTorres, @inclyc)

kevincox:
RFCSC: @infinisil are there any updates on this PR? What needs to be done to move it forward?

rfcs/0181-list-index-syntax.md
roberth: > elemAt is good enough

I agree within the context of the language, but it is not good enough for use in CLI arguments, as those use a custom attribute path syntax that doesn't and shouldn't have function calls.

GetPsyched:
RFCSC:

Since there has been little activity for quite a while, we'll draft this in the next RFCSC meeting (2 weeks from now), unless the shepherds indicate that they'll pick this back up soon.

PR #184, opened 2024-11-25
[RFC 184] versioned flakes references

domenkozar:
[rendered](https://github.com/domenkozar/rfcs/blob/rfc-183/rfcs/0183-versioned-flakes.md)

### DIFF
--- a/rfcs/0183-versioned-flakes.md
+++ b/rfcs/0183-versioned-flakes.md
@@ -1,0 +1,55 @@
+---
+feature: versioned-flakes
+start-date: 2024-11-25
+author: domenkozar
+co-authors: None
+shepherd-team: None
+shepherd-leader: None
+related-issues: None
+---
+## Summary
+
+Introduce a standardized versioning schema for Nix flakes using [SemVer](https://semver.org/),
+following lessons learned from [RFC 144](https://github.com/NixOS/rfcs/pull/144/files).
+
+Adds to flakes:
+
+- `version` attribute to `inputs` in the top-level of `flake.nix` to specify version of the input to be resolved.
+- An attribute `version` in the top-level of `flake.nix`.
+
+## Motivation
+
+Currently, Flakes lack a formal versioning mechanism, leading to potential duplication of dependencies and inefficient evaluations.
+
+By adopting a versioning schema similar to Rust's Cargo system, we can utilize a SAT solver to manage dependencies more effectively, minimizing redundancy and improving performance.
+
+## Explanation
+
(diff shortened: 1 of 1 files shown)


rfcs/0183-versioned-flakes.md
```diff
@@ -0,0 +1,55 @@
...
+start-date: 2024-11-25
+author: domenkozar
+co-authors: None
+shepherd-team: None
+shepherd-leader: None
+related-issues: None
+---
+## Summary
+
+Introduce a standardized versioning schema for Nix flakes using [SemVer](https://semver.org/),
```
inclyc, line 12: There is a problem in semver: 

> how can we *prove* two packages are *semantically* same, even if their versions are different?.

Nix (as a package manager) think it is more reliable if: any package update will trigger   "full-rebuilds" for those dependent on it.

I just feel a little bit strange about introducing "SemVer" here.

calops: I'm confused, isn't this RFC about versioning *flakes*, not packages?

inclyc: > I'm confused, isn't this RFC about versioning flakes, not packages?

That's the interesting paradox: if we don't use semantic versioning (SemVer) for packages, why should we use SemVer for flakes?

If there are valid reasons for using SemVer with 'flake' dependencies, why wouldn't those same reasons apply to packages as well?

domenkozar: How will you apply semver to nixpkgs packages?

2xsaiko: > That's the interesting paradox: if we don't use semantic versioning (SemVer) for packages, why should we use SemVer for flakes?
>
> If there are valid reasons for using SemVer with 'flake' dependencies, why wouldn't those same reasons apply to packages as well?

- Flake version is about the flake's API surface, while package version is about the package's API surface or whatever upstream wants to do, they are orthogonal to each other
- You can't and shouldn't force developers to use a certain version format for their software, so package versions should be as flexible as possible (nixpkgs of course has no restrictions at all really)
- Flake versions could be automatically resolved, it's well known where to search for versions and how to load them, while neither really applies to packages

I don't have an opinion on this RFC either way but the two situations are certainly very different.

MMesch:
RFCSC:

This RFC needs a shepherd team to move forward. Please leave a comment nominating yourself or others.

GetPsyched:
RFCSC:

asymmetric:

PR #189, opened 2025-08-10
[RFC 0189] Contracts

ibizaman:
[Rendered](https://github.com/ibizaman/rfcs/blob/contracts/rfcs/0189-contracts.md) -  ~[Draft PR 1](https://github.com/NixOS/nixpkgs/pull/432529)~ - ~[Draft PR 2](https://github.com/NixOS/nixpkgs/pull/485453)~ - [Draft PR 4](https://github.com/NixOS/nixpkgs/pull/522054)

This PR introduces the contracts RFC. This follows the discussion on the pre-RFC [discourse post](https://discourse.nixos.org/t/pre-rfc-decouple-services-using-structured-typing/58257).

### DIFF
--- a/rfcs/0189-contracts.md
+++ b/rfcs/0189-contracts.md
@@ -9,14 +9,12 @@ related-issues: https://github.com/NixOS/nixpkgs/pull/432529
 ---
 
 # Summary
-[summary]: #summary
 
 In nixpkgs, modules include a lot of duplicate code to set up their dependencies.
 We introduce a pattern for moving this custom code out of the modules and making it shareable
 in an incremental, backwards-compatible, extensible, and testable way.
 
 # Motivation
-[motivation]: #motivation
 
 As a motivating example, let's take a module
 that sets up a service that needs a database
@@ -46,11 +44,11 @@ This proposal resolves all those issues, as well as allowing a few things that a
 - Using stubs in NixOS tests.
 
 # Detailed design
-[design]: #detailed-design
 
 The core idea is to decouple the use of a feature from its implementation.
 
 Let's first introduce some nomenclature:
+
 - _consumer_: The module using or needing a feature.
   Example: Nextcloud, Vaultwarden and others require a database.
 - _provider_: The module implementing a feature.
@@ -76,6 +74,7 @@ which will translate options from the contract
 into options already defined by the existing module.
 
 Some examples of possible contracts:
+
 - File backup
 - Streaming backup (for databases)
 - Secrets (out of store values) provisioning
@@ -95,22 +94,17 @@ nor to identify an exhaustive list of existing contracts, but to define a patter
 These contracts will live under a new option path `contracts`
 like `contracts.fileBackup` and `contracts.streamingBackup`.
 
-See [prior-art][] for some useful comparisons that can help you get a better picture.
+See [prior art] for some useful comparisons that can help you get a better picture.
 
 # Implementation
-[implementation]: #implementation
 
-The implementation was worked out initially in the [SelfHostBlocks][] repo and perfected in the [module interfaces][] repo.
-There are some slight variations proposed in this RFC relative to the module interfaces repo to get it out sooner rather than later. See the [corresponding unresolved section][unresolved-duallink].
+The implementation was worked out initially in the [SelfHostBlocks] repo and perfected in the [module interfaces] repo.
+There are some slight variations proposed in this RFC relative to the module interfaces repo to get it out sooner rather than later. See the [corresponding unresolved section](#dual-link).
 
 It is important to keep in mind that the proposed implementation comes from
 seeing this pattern emerge naturally "in the wild" from trying to increase code reuse, providing solid evidence on the utility of this approach.
 
-[SelfHostBlocks]: https://github.com/ibizaman/selfhostblocks/tree/main/modules/contracts
-[module interfaces]: https://github.com/fricklerhandwerk/module-interfaces
-
 ## Actors
-[actors]: #actors
 
 Before looking at the code, it is useful to get a mental model of the actors involved.
 There are up to 4 different individuals or teams involved for one contract:
@@ -127,9 +121,9 @@ flowchart TD
 ```
 
 1. `Contract Team`: The team maintaining a contract.
-2. `Provider Team`: The team maintaining one module provider of that contract. Each provider of a same contract can have its own team.
-3. `Consumer Team`: The team maintaining one module consumer of that contract. Each consumer of a same contract can have its own team.
-4. `End User`: The end user linking one consumer of their choice with one provider of their choice for that contract.
+1. `Provider Team`: The team maintaining one module provider of that contract. Each provider of a same contract can have its own team.
+1. `Consumer Team`: The team maintaining one module consumer of that contract. Each consumer of a same contract can have its own team.
+1. `End User`: The end user linking one consumer of their choice with one provider of their choice for that contract.
 
 Note that the `Contract` is the central component here.
 The provider and the consumer teams do not need to know what the other team is doing,
@@ -140,7 +134,6 @@ One nice property here is the `End User` can add a new provider or consumer them
 A module can consume or provide multiple instances of the same or different contracts, for example a single HTTP server module might provide `Web Server` and `Reverse Proxy` contracts.
 
 ## Data Flow
-[dataflow]: #dataflow
 
 Another consideration before looking at the code is how data flows through a contract.
 
@@ -167,21 +160,18 @@ sequenceDiagram
 ```
 
 1. A `Consumer` sets the `input` option of the contract.
-2. The `Provider` reads from that `input` option.
-3. The `Provider` optionally accepts provider-specific options set by the `End User`.
-4. The `Provider` does some side effect (otherwise, there's no point).
-5. The `Provider` optionally writes to the `output` of the contract.
-6. The `Consumer` optionally reads from the `output` of the contract.
+1. The `Provider` reads from that `input` option.
+1. The `Provider` optionally accepts provider-specific options set by the `End User`.
+1. The `Provider` does some side effect (otherwise, there's no point).
+1. The `Provider` optionally writes to the `output` of the contract.
+1. The `Consumer` optionally reads from the `output` of the contract.
 
 If you squint, this looks just like a functional application, only applied at the module level.
 
 ## Contract Interface
-[Contract Interface]: #contract-interface
 
-_The following snippets are taken from the [draft PR][draftPR]._
+_The following snippets are taken from the [draft PR][draftpr]._
 _The intended reading order is first this document, then the PR._
-
-[draftPR]: https://github.com/NixOS/nixpkgs/pull/432529
 
 Links to relevant commits:
 
@@ -286,6 +276,7 @@ Now that we have the ability to declare the `input` and `output` options of a co
 we can declare matching `consumer` and `provider` options using dependent types.
 
 - `consumer`: Submodule option with 3 nested options:
+
   - `provider`: The linked `provider` for this consumer.
     This has to be set by the `end user` as they choose which consumer and provider to link.
   - `input`: An option whose type comes from the top-level `input` `deferredModule`.
@@ -295,6 +286,7 @@ we can declare matching `consumer` and `provider` options using dependent types.
     Its default value comes from the linked `provider`'s `output`.
 
 - `provider`: Submodule option with 3 nested options:
+
   - `consumer`: The linked `consumer` for this provider.
     This has to be set by the `end user` as they choose which consumer and provider to link.
     This option is made nullable because the end user is not necessarily required to use a contract.
@@ -327,16 +319,14 @@ config = {
 ```
 
 Notice the `end user` must link the consumer and provider in both directions.
-This is discussed in [the unresolved section][unresolved].
+This is discussed in [the unresolved section](#unresolved-questions).
 
 # Examples and Interactions
-[examples-and-interactions]: #examples-and-interactions
 
 In this section we will explain, for each contract implemented in the PR,
 why they are useful, and their interesting properties. See the PR for actual code.
 
 ## File Backup Contract
-[fileBackupContract]: #file-backup-contract
 
 Links to relevant commits:
 
@@ -429,7 +419,6 @@ They also allow creating automated backups on deploys,
 and restoring from backups on rollbacks too.
 
 ## Streaming Backup Contract
-[streamingBackupContract]: #streaming-backup-contract
 
 Links to relevant commits:
 
@@ -451,7 +440,6 @@ Here though, instead of engineering a stub for a stream, we use
 the `streamingBackup consumer` added to `services.postgresql` directly.
 
 ## Secrets Contract
-[secretsContract]: #secrets-contract
 
 Links to relevant commits:
 
@@ -503,17 +491,15 @@ This new provider has been tested using the [contract's behavior test](https://g
 and has been used in [`services.stash`'s module](https://github.com/NixOS/nixpkgs/pull/432529/commits/19419ad95913fbed4636d0b24d95c80517c18340) as an example.
 
 # Drawbacks
-[drawbacks]: #drawbacks
 
 We are not aware of any because this solution is fully backwards compatible,
 incremental, and has many advantages. It also arose from a real practical need.
 
 Care should be taken to not abuse this pattern though. It should be reserved
 for contracts where abstracting away a `consumer` and `provider` makes sense.
-We didn't find a general rule for that but a good indicator of an unnecessary contract is where we only find one instance of a `consumer` and `provider` pair in the whole of nixpkgs.
+We didn't find a general rule for that, but a good indicator of an unnecessary contract is where we only find one instance of a `consumer` and `provider` pair in the whole of nixpkgs.
 
 # Alternatives
-[alternatives]: #alternatives
 
 This design arose from trying to maximize code reuse.
 We started by fiddling with nix code and the implementation emerged naturally.
@@ -523,7 +509,6 @@ mostly because our attempts to tweak the code often led us often to infinite rec
 so we couldn't stray too far from the way it already works.
 
 # Prior art
-[prior-art]: #prior-art
 
 We did not find any discussion about any of this by the nix community.
 It is a bit self-centered, but the two talks I (`ibizaman`) gave on this subject in nixpkgs can be considered prior art.
@@ -542,10 +527,8 @@ A few useful comparisons beyond nixpkgs:
 - Contracts are reminiscent of the [reverse dependency principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) which is used in many places.
 
 # Unresolved questions
-[unresolved]: #unresolved-questions
 
 ## Dual Link
-[unresolved-duallink]: #unresolved-questions-duallink
 
 The current implementation requires the `end user` to link the consumer and provider
 in both directions:
@@ -567,19 +550,18 @@ In the snippet above, this would remove the need for the `provider to consumer` 
 The issue comes from the `consumer` and `provider` option in the top-level `contracts` definition to be of type `optionType`.
 They don't have access to the actual `input` and `output` values of an instantiated contract.
 
-There are some experiments on this in the [module interfaces][] repo.
+There are some experiments on this in the [module interfaces] repo.
 There, we set the `provider` option as a function which takes an argument
 which is the instantiated `consumer`, so it is not of type `optionType` but of type `submodule`, and has access to the real input and output values.
 Unfortunately, this has two downsides:
 
 1. It requires one more line in each provider definition. This would be okay except for the following downside:
-2. There's no way to write side effects. This means the `provider` can only write to its own `output`, which misses the whole point of having contracts in the first place.
+1. There's no way to write side effects. This means the `provider` can only write to its own `output`, which misses the whole point of having contracts in the first place.
 
 There may be a way to solve this, but we have not yet figured it out. Help would be appreciated!
 Beware though; you will be crossing the edge of the module system and entering the land of infinite recursion.
 
 ## Documentation
-[unresolved-documentation]: #unresolved-questions-documentation
 
 It is not currently possible to build the manual; doing so results in an error:
 
@@ -596,13 +578,16 @@ $ (cd nixos/; nix-build release.nix -A manual.x86_64-linux)
           436|         description = "Path to file containing a secret used to sign JWT tokens.";
 ```
 
-Comments in the [draft PR][draftPR] have been added to indicate what has been tried.
+Comments in the [draft PR][draftpr] have been added to indicate what has been tried.
 We would appreciate help in solving this.
 
 # Future work
-[future]: #future-work
 
-- Solve the [documentation][unresolved-documentation] issue.
+- Solve the [documentation](#documentation) issue.
 - Identify useful contracts and their inputs, outputs, and behavior tests.
 - Identify services that would benefit from being consumers and providers of contracts and add the necessary options.
-- Optionally solve the [dual-link][unresolved-duallink] issue.
+- Optionally solve the [dual-link](#dual-link) issue.
+
+[draftpr]: https://github.com/NixOS/nixpkgs/pull/432529
+[module interfaces]: https://github.com/fricklerhandwerk/module-interfaces
+[selfhostblocks]: https://github.com/ibizaman/selfhostblocks/tree/main/modules/contracts
--- a/rfcs/0189-contracts.md
+++ b/rfcs/0189-contracts.md
@@ -11,79 +11,71 @@ related-issues: https://github.com/NixOS/nixpkgs/pull/432529
 # Summary
 [summary]: #summary
 
-In Nixpkgs, modules duplicate a lot of code to setup their dependencies.
-We introduce a pattern that allows to move
-this custom code out of the modules and make it shareable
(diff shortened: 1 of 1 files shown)


KiaraGrouwstra:
Thanks for filing. I would like to join here to shepherd this.

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+4. The `provider` does some side effect (otherwise, there's no point).
+5. The `provider` optionally writes to the `output` of the contract.
+6. The `consumer` optionally reads from the `output` of the contract.
+
+If you squint, this looks just like function application, only applied at the module level.
+
+## Contract Interface
+[Contract Interface]: #contract-interface
+
+_The draft PR from which the following snippets are taken can be found [here][draft]._
```
KiaraGrouwstra: the link here doesn't render this way

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+as described by the contract `consumer`.
+
+This new provider has been tested using the contract' behavior test
+and has been used in `services.stash`'s as an example.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+We are not aware of any because this solution is fully backwards compatible,
+incremental and has a lot advantages. It also arise from a real practical need.
```
KiaraGrouwstra: arises/arose

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+
+Care should be taken to not abuse this pattern though. It should be reserved
+for contracts where abstracting away a `consumer` and `provider` makes sense.
+We didn't find a general rule for that but a good indication that the pattern gets abused
+is if we only find one `consumer` and `provider` pair in the whole of nixpkgs.
+
+# Alternatives
+[alternatives]: #alternatives
+
+This design arise from trying to minimize code reuse.
```
KiaraGrouwstra: arises/arose, maximize code reuse

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+We started by fiddling with nix code and the implementation came up naturally.
+
+We are not aware of any alternatives to do this,
+mostly because our attempts to tweak the code led us often to infinite recursion or other module issues
```
KiaraGrouwstra: often led us

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+so we couldn't stray too far from the way it is written now.
+
+# Prior art
+[prior-art]: #prior-art
+
+We did not find any discussion about any of this by the nix community.
+It is a bit self-centered but the two talks I (ibizaman) gave on this subject in nixpkgs can be considered prior arts.
```
KiaraGrouwstra: (prior art  :no_mouth:)

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+Note the syntax presented is a bit outdated now but the message and the passion is still relevant:
```
KiaraGrouwstra: (it wasn't clear to me what you meant by 'passion' here)

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+
+- 04/2024: Scale21x in Pasadena: [Easier NixOS self-hosting with module contracts](https://www.youtube.com/watch?v=lw7PgphB9qM)
+- 11/2024 at NixCon2024 in Berlin: [Enabling incremental adoption of NixOS with module contracts](https://www.youtube.com/watch?v=CP0hR6w1csc)
+
+A pre-RFC has been opened [on discourse][prerfc].
+
+[prerf]: https://discourse.nixos.org/t/pre-rfc-decouple-services-using-structured-typing/58257
```
KiaraGrouwstra: prerfc

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+[prior-art]: #prior-art
+
+We did not find any discussion about any of this by the nix community.
+It is a bit self-centered but the two talks I (ibizaman) gave on this subject in nixpkgs can be considered prior arts.
+Note the syntax presented is a bit outdated now but the message and the passion is still relevant:
+
+- 04/2024: Scale21x in Pasadena: [Easier NixOS self-hosting with module contracts](https://www.youtube.com/watch?v=lw7PgphB9qM)
+- 11/2024 at NixCon2024 in Berlin: [Enabling incremental adoption of NixOS with module contracts](https://www.youtube.com/watch?v=CP0hR6w1csc)
+
+A pre-RFC has been opened [on discourse][prerfc].
```
KiaraGrouwstra: link failed to render

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+
+## Dual Link
+[unresolved-duallink]: #unresolved-questions-duallink
+
+The current implementation requires the `end user` to link the consumer and provider
+in both directions:
+
+```nix
+config = {
+  // consumer to provider
```
KiaraGrouwstra: `//` -> `#` (down as well)

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,630 @@
...
+
+       error: attribute 'contracts' missing
+       at /home/timi/Projects/nixpkgs/nixos/modules/services/web-apps/stash.nix:435:16:
+          434|       jwtSecretKeyFile = mkOption {
+          435|         type = config.contracts.secret.consumer;
+             |                ^
+          436|         description = "Path to file containing a secret used to sign JWT tokens.";
+```
+
+Comments in the draft PR have been added to indicate what has been tried.
```
KiaraGrouwstra: link draft PR here

KiaraGrouwstra: https://github.com/NixOS/nixpkgs/pull/432529

rfcs/0189-contracts.md
ibizaman: I mostly meant that passion I had behind this endeavor. :P But I agree it's superfluous.

rfcs/0189-contracts.md
ibizaman: :man_facepalming:

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,658 @@
...
+
+It would be so much nicer if we could somehow require only to give the `consumer` to the `provider`
+and it managed to make the link backwards automatically.
+In the snippet above, this means removing the need for the `provider to consumer` line.
+
+The issue comes from the `consumer` and `provider` option in the top-level `contracts` definition to be of type `optionType`.
+They don't have access to the actual `input` and `output` values of an instantiated contract.
+
+Experimenting on this has been done in the [module interfaces][] repo.
+There, we set the `provider` option as a function which takes an argument
```
KiaraGrouwstra, line 554: maybe for context we could link 'set the `provider` option' to the [lines in question](https://github.com/fricklerhandwerk/module-interfaces/blob/caeb171fde3a6ce2dc61f4e59834df9e7bf584a0/interface.nix#L47-L58)

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,658 @@
...
+and has been used in [`services.stash`'s module][stash: use secret contract for passwords] as an example.
+
+[hardcodedSecret: define behavior test for secret contract]: https://github.com/NixOS/nixpkgs/pull/432529/commits/448410a520225bc71e1616611cef7ad086c64cd1
+[stash: use secret contract for passwords]: https://github.com/NixOS/nixpkgs/pull/432529/commits/19419ad95913fbed4636d0b24d95c80517c18340
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+We are not aware of any because this solution is fully backwards compatible,
+incremental and has a lot advantages. It also arose from a real practical need.
```
7c6f434c: Standard question for any module-system-based increase in layers: what is the estimated evaluation-time resource impact?

ibizaman: That’s a great question. Is there a standard set of benchmarking tooling? After some quick online search, I see some external repos and wiki guides but I was wondering how from scratch I should start from.

7c6f434c: Good question; I don't actually know things beyond `export NIX_SHOW_STATS=1`. There are some scripts in `ci/eval` but I am not sure whether reusing parts of them is worth it (but maybe looking at what is done there could be useful)

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,658 @@
...
+
+Care should be taken to not abuse this pattern though. It should be reserved
+for contracts where abstracting away a `consumer` and `provider` makes sense.
+We didn't find a general rule for that but a good indication that the pattern gets abused
+is if we only find one `consumer` and `provider` pair in the whole of nixpkgs.
```
axelkar: This is confusing. Is this what you meant?
```suggestion
is if we only find one `consumer` and one `provider` usage of a contract in Nixpkgs.
```

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,658 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+This design arose from trying to maximize code reuse.
+We started by fiddling with nix code and the implementation came up naturally.
+
+We are not aware of any alternatives to do this,
+mostly because our attempts to tweak the code often led us often to infinite recursion or other module issues
```
axelkar, line 508: The dual-linking approach strikes dangerously close to infinite recursion in my mind. What do you think?

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,658 @@
...
+- Leads to difference in features:
+  some implementations are more featureful than others,
+  very few modules allow you to choose from multiple dependencies (e.g. PostgreSQL, MySQL or other).
+- Not extendable without changing the source code:
+  a user cannot easily choose to use a dependency the maintainer didn't add code for.
+
+What we propose answers to all those issues
+as well as allows a few things that's not possible currently:
+
+- interfacing with dependencies and services outside of NixOS,
```
axelkar: Is there an example for this? The contract system will also probably need integration with service managers for scheduling and bringing up dependencies.

rfcs/0189-contracts.md
ibizaman: Yes indeed thats what I meant :)

rfcs/0189-contracts.md
ibizaman: I can’t believe I forgot to add an example for this IMO important point. I’ll add that.

The intent here is to provide an escape hatch to the outside (of NixOS) world. Which means you’re free to do whatever you want. The use case I had in mind does imply “manual” scheduling. Let’s say you want to integrate a NixOS service with a database that exists already on another server which does not run NixOS. You would create a database provider which hardcodes the url and credentials to the other server. Then the consumer would be able to access it. Of course you must set up the other server beforehand correctly. And of course the other server could be a NixOS machine too.

I can see a separate contract for interacting with service management. This should be merged with the module interfaces initiative.

rfcs/0189-contracts.md
KiaraGrouwstra: in my understanding the reason the current implementation works is lazy evaluation

ibizaman: Indeed, it's pretty close. It's working because we're linking two different options, depending on the direction.

I'll argue I'm not the first to have though of this pattern. Let's take the secrets contracts as an example. Remember, it looks like so:

```nix
services.myservice.secret.provider = config.sops.secrets."myservice/secret".provider;

sops.secrets."myservice/secret".consumer = services.myservice.secret.consumer;
```

If we compare to the sops-nix module, if we squint we can see the dual linking is nearly there:

```nix
services.myservice.secret = config.sops.secrets."myservice/secret".path;

sops.secrets."myservice/secret" = {};
```

Btw, you would get the infinite recursion if you removed the `.consumer` or `.provider` from the snippet above.

KiaraGrouwstra: i mean, it works until you try and serialize the whole structure i guess :sweat_smile:

rfcs/0189-contracts.md
KiaraGrouwstra: maybe one way to compare would be getting such stats after/before the PR for say that nextcloud test, like: `NIX_SHOW_STATS=1 nix-build --impure nixos/tests/nextcloud -A basic31`.
  
## PR

```json
{
"cpuTime": 0.1535560041666031,
"envs": {
  "bytes": 2400992,
  "elements": 167207,
  "number": 132917
},
"gc": {
  "heapSize": 402915328,
  "totalBytes": 68045680
},
"list": {
  "bytes": 288200,
  "concats": 1487,
  "elements": 36025
},
"nrAvoided": 133723,
"nrFunctionCalls": 122934,
"nrLookups": 66643,
"nrOpUpdateValuesCopied": 1889724,
"nrOpUpdates": 6747,
"nrPrimOpCalls": 60323,
"nrThunks": 171874,
"sets": {
  "bytes": 35799712,
  "elements": 2221018,
  "number": 16464
},
"symbols": {
  "bytes": 335942,
  "number": 34889
},
"values": {
  "bytes": 18234936,
  "number": 759789
}
}
```
  
## parent revision

```json
{
"cpuTime": 0.1342640072107315,
"envs": {
  "bytes": 2304512,
  "elements": 160244,
  "number": 127820
},
"gc": {
  "heapSize": 402915328,
  "totalBytes": 67791424
},
"list": {
  "bytes": 277016,
  "concats": 1378,
  "elements": 34627
},
"nrAvoided": 128713,
"nrFunctionCalls": 118695,
"nrLookups": 63610,
"nrOpUpdateValuesCopied": 1915086,
"nrOpUpdates": 6556,
"nrPrimOpCalls": 58429,
"nrThunks": 161589,
"sets": {
  "bytes": 36067568,
  "elements": 2239118,
  "number": 15105
},
"symbols": {
  "bytes": 299281,
  "number": 30518
},
"values": {
  "bytes": 17928864,
  "number": 747036
}
}
```

i'm not sure which metrics are interesting (note `cpuTime` is unstable across invocations) or if we should compare for a different build attribute.
building the nextcloud module (`nixos/modules/services/web-apps/nextcloud.nix`) instead, for what it's worth, i didn't really get it to show differences beside on a `symbols` metric.

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,658 @@
...
+Comments in the [draft PR][draftPR] have been added to indicate what has been tried.
+Any help is appreciated to solve this.
+
+# Future work
+[future]: #future-work
+
+- Solve the [documentation][unresolved-documentation] issue.
+- Identify contracts and their inputs, outputs and behavior tests.
+- Identify services that would benefit from being consumers and providers of contracts and add the necessary options.
+- Optionally solve the [dual-link][unresolved-duallink] issue.
```
KiaraGrouwstra: clicking these links seems not to actually navigate the user to these on the page - i think it wants the ids instead?

ibizaman: Just checked it is fixed now :)

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,593 @@
...
+for contracts where abstracting away a `consumer` and `provider` makes sense.
+We didn't find a general rule for that, but a good indicator of an unnecessary contract is where we only find one instance of a `consumer` and `provider` pair in the whole of nixpkgs.
+
+# Alternatives
+
+This design arose from trying to maximize code reuse.
+We started by fiddling with nix code and the implementation emerged naturally.
+
+We are not aware of any alternative ways to do this,
+mostly because our attempts to tweak the code often led us often to infinite recursion or other module issues
```
gytis-ivaskevicius, line 508: What about doing something along the lines of this:
```nix
config = {
  
  # each service coming from submodule generator units.xyz is where the magic happens, its fully isolated. I would expect it to run its own modules evaluator in the background, should even help with eval time
  services.nextcloud = units.nextcloud {
     # ... module config ...
     fileBackup.provider = config.services.restic-for-nextcloud.fileBackup;
  };
   
  # Service name does not matter
  services.restic-for-nextcloud = units.restic {
    // Provider-specific options.
    repository = "/var/lib/backups/nextcloud";
    passwordFile = toString (pkgs.writeText "password" "password");
    initialize = true;
  };
};
```

This way we make modules more portable, contract itself is with root module system and units.xyz which we could implement to support other projects such as home-manager or devshells. and integration between modules is dependent on what other modules explicitly export and ofcourse where user explicitly  passes values

Tho there might be cases of infinite recursion sometimes, tho it should be very rare


I think of this as Terraform modules, they have certain inputs and certain outputs - everything else is a black box

KiaraGrouwstra: this also reminds me a bit of [portable services](https://github.com/NixOS/nixpkgs/pull/267111), and reusing those for the existing modules to prevent duplication.
this example does sound easily extendible in terms of adding new providers - @ibizaman would probably know more about how we were doing there so far.
avoiding dual linkage maybe seems an open question, but yeah, let's try stuff.

rfcs/0189-contracts.md
ibizaman: A good example of this would be for a database contract. The options could be:

Filled out by the provider:
- `hostname`
- `database name`
- `username`
- `password`

I actually can't think of options filled out by the consumer. Maybe extensions needed in the database? Charset?

Usually, you would have a provider like the postgresql module that you would create a database in with the (unfortunately but understandably controversial) `ensureUser` and `ensureDatabase` options. Then, you would plug in the new user and database to the consumer that requests a database.

But this would work too outside of the NixOS ecosystem. You could just create a user and database in any system then, inside the NixOS system that has the consumer, just hardcode the 4 options with strings that would point to the other system.

Does that make sense?

rfcs/0189-contracts.md
ibizaman: Same, makes me think of portable services.

I fully agree we should tend to this and it seems indeed compatible with contracts. I would argue though it is orthogonal and a large effort that should be separate from this.

I'm definitely not going to stop anyone from attempting both at the same time but I can't see myself doing it currently.

rfcs/0189-contracts.md
axelkar: I believe those fields are enough for most configurations, though the password should be specified using `passwordFile` (or a secret management contract!).

ibizaman: That’s spot on @axelkar. I should’ve been more precise. In my project I indeed use a contract for the password itself so it’s not the plain password.

ibizaman:
FYI I created a new draft PR with an alternative implementation which is IMO better than the first. It has less weird quirks at least. I added a comparison in the description of the PR. https://github.com/NixOS/nixpkgs/pull/485453

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,593 @@
...
+like `contracts.fileBackup` and `contracts.streamingBackup`.
+
+See [prior art] for some useful comparisons that can help you get a better picture.
+
+# Implementation
+
+The implementation was worked out initially in the [SelfHostBlocks] repo and perfected in the [module interfaces] repo.
+There are some slight variations proposed in this RFC relative to the module interfaces repo to get it out sooner rather than later. See the [corresponding unresolved section](#dual-link).
+
+It is important to keep in mind that the proposed implementation comes from
```
Pamplemousse, line 104: To me, this proposal is screaming "it is a convention to do dependency injection consistently for modules".
"Dependency injection" is a mean to achieve dependency inversion that could (should?) be very familiar to developpers.

It is mentionned in the [Prior Art](https://github.com/ibizaman/rfcs/blob/contracts/rfcs/0189-contracts.md#prior-art) section, but I feel that it could be mentionned here, and even appear earlier (as early as in [Motivation](https://github.com/ibizaman/rfcs/blob/contracts/rfcs/0189-contracts.md#motivation)).

ibizaman: I didn’t see it at first but it’s evident to my too now

rfcs/0189-contracts.md
```diff
@@ -0,0 +1,593 @@
...
+
+# Unresolved questions
+
+## Dual Link
+
+The current implementation requires the `end user` to link the consumer and provider
+in both directions:
+
+```nix
+config = {
```
mrVanDalo, line 537: In my understanding consumer and provider are switched here.

- nextcloud don't need a fileBackup, it provides a fileBackup (I assume these are files that should be backed up). => nextcloud is a provider
- but restic (and borgBackup) need to know which files to backup (fileBackup) so they would consume these files. => restic is a consumer

I would also argue:  a provider provides only one thing but a consumer can consume multiple things.

Maybe this is what is meant here, and I mix up the syntax (or the object naming could be updated), but I couldn't see descriptions on your option definition up there.

mrVanDalo: First of all sorry, if this was already discussed. I read the last hour the discusions specifically on this topic, but couldn't find anything. But I might have missed it.

Anyway I'm a huge dependency injection fan. But I don't see to much benefit on your example here.
Usually dependency injection comes with a injector which handles the wiring for you, which is one of the benefits.
But we have to make the wireing here our self.
The nixos module system already comes with options which can be rewired on demand where ever you like
(outside the module definitions),
which is not possible in programming languages like java or go.
For example your example currently in nixos could be done like this

```nix
services.restic.bacups.nextcloud.paths = [config.services.nextcloud.datadir]
```

This approach don't lack any aspect to using the contracts (At least from my perspective).

The one benefit I see from the contracts concept is that we create a transactional type system, which standardizes service plumbing. Which is not nothing.

If you would tell me nixos module is capable of dependency injection, I would imaging something like

```nix
options.services.nextcloud.datadir = mkProvider fileBackup {
# same like mkOption but type is dictated by fileBackup (e.g.: 2 fileBackup options can't have different types)
};
...
config = .. # nextcloud configuration using datadir like now
```

than you could set.

```nix
serivces.restic.backup.dependencyInjection.paths = allProvidersOf fileBackup;
```


In this example `allProvidersOf <key/type>` will be handled by the dependency injector and
creates a list of config-references.
```nix
services.restic.backup.dependencyInjection.paths = [config.services.nextcloud.datadir];
```
This is because I expect automatic wireing when I hear dependency injection = inversion of control.

I sketched some scenarios (e.g.: reverse proxy -> webapp -> database connetion <- credentials) and I come always to the same conclusion. But I could see a benefit in the credentials part with this approach, but this is not part of you RFC if I understood correctly.

I hope this is helpful.

p.s.: I would be happy for a proper solution here, I faced a similar problem like in your example and solved it like this : https://gist.github.com/mrVanDalo/a1f7ec5cce80d69ac60b6d22feb6d768 (a home manager interface which handles files I need to backup)

mrVanDalo (commented):
I hope this is a constructive comment

rfcs/0189-contracts.md
KiaraGrouwstra: thanks for your response!

i think the dual link here was maybe less intended design, more like a limitation of what we'd managed with the implementation at the time. #506343 addresses that.

> `mkProvider fileBackup { ... }`

the limitation of functions is that they are pure, whereas going thru the module system ensures that we can use the passed information to set `config` as well.

> In this example `allProvidersOf <key/type>` will be handled by the dependency injector and creates a list of config-references.

this type of aggregation is basically what #506343 does, resulting in the `defaultProvider(Name)` to let one wire things up.

on naming, i def had trouble picking things that made sense as well. in the case of secrets, for one, i sort of made sense out of this by speaking of an application _requesting_ a secret, that another module might then _provide_.
in an earlier iteration, i played with terms like input and output, but then ran into similar issues about perspective as you noted here.

KiaraGrouwstra: @mrVanDalo:

> `mkProvider fileBackup { ... }`

at NixOS/nixpkgs#506343 i've now managed to cut down on boilerplate further again using a `mkProviderType` helper now, perhaps not unlike what you'd aimed for with your mentioned `mkProvider`.

> I sketched some scenarios (e.g.: reverse proxy -> webapp -> database connetion <- credentials)

fwiw, i added a test (`nixos/tests/contracts/chaining.nix`) to my commit `nixos/modules/contracts: add NixOS contracts integration` that demonstrates chaining of contracts now.

> `allProvidersOf`

i put up a [branch](https://github.com/KiaraGrouwstra/nixpkgs/compare/contracts-automated...contracts-all-results) exploring this, tho i'm a bit on the fence on when one might want to use this still

> I faced a similar problem like in your example and solved it like this : https://gist.github.com/mrVanDalo/a1f7ec5cce80d69ac60b6d22feb6d768 (a home manager interface which handles files I need to backup)

i think this does sound a bit similar to how i aggregate things in my PR. now, back-ups i think don't make full use of the concept of contracts in the sense i think they don't really need to get a `result` value back in nix, which the contracts facilitate. this is also how we got the idea of splitting the back-ups out of an initial contracts PR.

mrVanDalo: `nixos/tests/contracts/chaining.nix` is quite helpful to understand your setup. 
I added some questions, which might help me  (and others) to understand better what's happening.

ibizaman:
After considering the comments in this RFC and the draft PRs, I came to the conclusion that I started from the wrong end of contracts. I introduced some complex mechanism that is not justified.

The latest PR https://github.com/NixOS/nixpkgs/pull/522054 does not suffer from this issue and introduces a contract without any complex implementation, because it is not required. I believe this PR is the best shot yet at focusing on the essence of one contract so I’m closing the others in order to focus on that one. 

Although discussions about any implementation is very welcomed, I do believe without something tangible, the design space is too big and it’s hard to give concrete arguments for why one implementation should be chosen instead of another.

PR #109, opened 2021-10-12
[RFC 0109] Nixpkgs Generated Code Policy

Ericson2314:
Nixpkgs contains non-trivial amounts of generated, rather than hand-written code. We want to start systematizing to make it easier to maintain. There is plenty of future work building upon this we could do, but we stop here for now to avoid needing to change any tools (Nix, Hydra, etc.).

[Rendered](https://github.com/Ericson2314/nix-rfcs/blob/nixpkgs-ifd/rfcs/0109-nixpkgs-generated-code-policy.md)

PR #191, opened 2025-10-10
[RFC 0191] Lockfile Generation

pyrox0:
[Rendered](https://github.com/pyrox0/nix-rfcs/blob/lockfile-generation/rfcs/0191-lockfile-generation.md)

### DIFF
--- a/rfcs/0191-lockfile-generation.md
+++ b/rfcs/0191-lockfile-generation.md
@@ -1,0 +1,203 @@
+---
+feature: lockfile-generation
+start-date: 2025-10-09
+author: dish (@pyrox0)
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+
+[summary]: #summary
+
+When adding packages to nixpkgs, or migrating them within nixpkgs, some packages do not contain lockfiles
+that would allow us to lock their dependencies correctly. This is an issue with packages based on the
+NPM, Yarn, PNPM, and Cargo package managers, but is not limited to them. Therefore, we should support
+some mechanism for creating these sorts of lockfiles, **without** bloating the nixpkgs tarball with
+megabytes of lockfiles that are only useful for a handful of packages.
+
+# Motivation
+
+[motivation]: #motivation
+
+I have been migrating many packages from the deprecated `nodePackages` package set, into the main nixpkgs
+package set(under `pkgs/by-name`). Many packages, such as [alex](https://github.com/get-alex/alex),
+most(or all) packages by [`sindresorhus`](https://github.com/sindresorhus), and even projects using
+Cargo and Rust such as [`proxmox-backup-client`](https://git.proxmox.com/?p=proxmox-backup.git;a=tree) do not have lockfiles.
+This becomes an issue. Currently, the solution is to generate these lockfiles, and then vendor them within nixpkgs,
+then using the `postPatch` phase(mostly) to copy this lockfile to the source tree, allowing for package
+reproducibility.
+
+The issue, of course, is that this massively increases the size of the nixpkgs tarball. nixpkgs issues such as
+[#327064](https://github.com/NixOS/nixpkgs/issues/327064) and [#327063](https://github.com/NixOS/nixpkgs/issues/327063) have found
+the effects of this are staggering, with nearly 6MiB of size(at the time) added to the nixpkgs tarball,
+and this also contributed a lot of eval time of nixpkgs(See [this comment](https://github.com/NixOS/nixpkgs/issues/320528#issuecomment-2227185095)).
+
+However, I believe there are solutions to this problem. The
+[`rustPlatform.fetchCargoVendor`](https://nixos.org/manual/nixpkgs/stable/#vendoring-of-dependencies)
+function allows for using upstream Cargo.lock files without parsing them within
+nixpkgs, thus solving the previous problem of parsing these lockfiles within
+nixpkgs itself, removing much of the eval time penalty. This does not solve the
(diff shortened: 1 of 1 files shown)


rfcs/0191-lockfile-generation.md
pbsds, line 1: I love the incentive, and this is indeed a problem we would like to solve.

This RFC describes a happy path, but does however not really go into many edge cases.
There are also a couple load-bearing "should"s, that IMO needs fleshing out.

My unresolved questions as so plentiful that I did not bother splitting them into separated threads. I think a second revision is needed before more a granular discussion becomes productive.

* lockfile generation
    * how do we handle edge cases, like the package.json not being at the root directory?
    * can we in any way patch the package.json file before the lockfile is generated?
    * can we in any way patch the lockfile? (Like changing an unsafe dependency.)
    * can package authors/maintainers in any way test the lockfile generation flow locally?
    * can we have multiple lockfiles, in the event of nixpkgs shipping multiple versions of a package? (These may be created through `.override` and not have a unique attribute path.)
    * can we have multiple lockfiles, in the event of a single package having multiple subpackages? (See [pagefind](https://github.com/NixOS/nixpkgs/blob/c9b6fb798541223bbb396d287d16f43520250518/pkgs/applications/misc/pagefind/default.nix#L69-L101).)
    * can we regenerate a lockfile for an older package version on `release-xx.yy`? (in the event of a dep that is insecure or that fails to build.)
* maintaining the lockfile monorepo
    * can we update a lockfile, e.g. in the event of a dep being unsafe, even if the package version is unchanged?
    * what should happen in the event of a package being renamed?
    * what should happen in the event of a false-detection of a package removal? (Eval failures will happen, those can be at the scope level.)
    * what should happen in the event of a package removal not being detected?
    * what should happen in the event of a package not needing a lockfile anymore? Can we automatically detect that?
    * should we even delete old lockfiles? what about reproducability?
* scaling and fetching the lockfile repo/store
    * should all lockfiles be put into a massive monorepo? 
        * How should the folder layout look like?
        * Should a single revision of the whole lockfile monorepo be used throughout all expressions in nixpkgs? How do we prevent treewide rebuilds when bumping that rev?
        * Should multiple copies of the whole lockfile monorepo be used in nixpkgs? How do we prevent this from bloating the nix store of the user?
        * Should individual lockfiles from the lockfile monorepo be fetched per package expression in nixpkgs? How do we maintain all these `fetchurl` expressions? What should happen if we ever need to change the monorepo file layout?
    * should each lockfile be put into their own lockfile repo branch? 
        * What should the branch name schema look like?
        * Will the github UI break if we have hundreds or thousands of branches?
    * should each lockfile be put into their own gist? 
        * Do we have a limit on those?
        * How can a user browse and search in those? 
    * should each lockfile be stored as GHA build artifacts?
        * Do we have a limit on those?
        * Do they last forever?
        * How can a user browse and search in those? 
    * should we outsource the archival of lockfiles to some other actor than GitHub, like archive.org or the Software Heritage?
        * see https://archive.fosdem.org/2025/schedule/event/fosdem-2025-5897-guix-software-heritage-source-code-archiving-to-the-rescue-of-reproducible-deployment/
    * should each lockfile be stored and served by a website we self-host?
        * Should this website rely on being cached on cache.nixos.org, or need it take measures against ddos attacks?
        * Should we ever prune lockfiles from it? Do we consider it responible for reproducability or do we have cache.nixos.org for that?
* r-ryantm, nix-update, and incorrect maintanance
    * Can these automated load-bearing tools somehow correctly invoke the lockfile repo bot, retrieve the resulting lockfile url, and update the package expression correctly?
    * Can we prevent the case where `fetchurl`'ed lockfiles fail to get their hash updated in a package bump PR and go undetected?
    * How can a PR reviewer verify that a `fetchurl`'ed lockfile expression corresponds to the correct package?

I'm sorry if this wall of questions seems daunting. With a renewed view of various edge cases and considerations, I believe that focusing on how to we store the lockfiles will let the rest fall into place. It may also be benefitial to define scope early, on what cases we support and not, and what the fallback should be.

pyrox0: > This RFC describes a happy path, but does however not really go into many edge cases.
> There are also a couple load-bearing "should"s, that IMO needs fleshing out.

100%! I agree that there's a lot to be fleshed out here. This was my original design idea for it, and to be totally transparent, I created this over a few hours yesterday, which was kind of solidifying a general design I've had floating around in my head for a while. I wanted to put this out earlier in a more unfinished state, specifically so I could get input like this. It might take a while for me to answer all of those questions, but I do appreciate them!

Ericson2314:
BTW, the idea of having a separate repo for verbose/generated data was also discussed as part of the https://github.com/NixOS/rfcs/pull/109 shepherd discussion.

mushrowan:
just ran into this issue myself today - really agree that there should be a solution, though also agree that the "how" also needs to be addressed. thanks for making :)

rfcs/0191-lockfile-generation.md
```diff
@@ -0,0 +1,203 @@
...
+
+[alternatives]: #alternatives
+
+* Not packaging packages with missing lockfiles
+  * A very valid idea! However, there are enough of these sorts of packages, and lockfiles are already such an issue
+  for nixpkgs, that we should solve this so that we can move more lockfiles out of tree.
+* Storing lockfiles in-tree
+  * See above issues r.e massively increased tarball size. I aim to prevent that with this, so doing that
+  would be antithetical to this.
+
```
roberth, line 179:
```suggestion
* Git LFS
  * Not supported by `nix-channel` and `github:` flakerefs
  * Supported by `git+https` flakerefs but not yet lazy in `NixOS/nix` as of December 2025
  * Could perhaps be worked around with LFS pointer fetcher?

```

pyrox0:
Note: I am closing this, as I do not have the bandwidth to resolve its issues at the moment(as is hopefully evident by the lack of movement on this RFC)

Thank you to everyone who gave time to review it, if I ever come back to this in future I will be sure to take your questions into consideration.

PR #192, opened 2025-11-23
[RFC 0192] version pins for the `pkgs/by-name` structure

quantenzitrone:
[Rendered](https://github.com/quantenzitrone/rfcs/blob/by-name-version-pins/rfcs/0192-by-name-version-pins.md)

Sorry if the RFC text is short, i don't know what else to write. The feature is quite simple.

### DIFF
--- a/rfcs/0192-by-name-version-pins.md
+++ b/rfcs/0192-by-name-version-pins.md
@@ -65,7 +65,7 @@ In addition to `pkgs/by-name/{shard}/{pname}/package.nix` there can also be
 I think you know where this is going.
 
 Packages with a `pin.nix` will have the versions of packages pinned
-accordingly. This can be easily archived with:
+accordingly. This can be easily achieved with:
 
 ```nix
 if lib.pathExists (packageDirectory + "/pins.nix") then
--- a/rfcs/0192-by-name-version-pins.md
+++ b/rfcs/0192-by-name-version-pins.md
@@ -93,7 +93,36 @@ Probably longer eval time, this has to be tested however.
 
 - Keep the current situation: version pins are either in `all-packages.nix` or
   directly in the `package.nix`.
+
 - Some more generalized `overrides.nix` without the strict requirements.
+  - This would be less useful for automation of dependency pin removal.
+  
+- Inline dependency pins in `package.nix` as default values in the function arguments.
+  ```nix
+  {
+    package-a_version,
+    package-a ? package-a_version,
+  # ...
+  }:
+  # ...
+  ```
+  This would require a custom `callPackage` for `pkgs/by-name` that prefers default values over
+  values from the `pkgs`.
+  - Confusing for contributors if something works differently through the normal `callPackage`.
+
+- Inline dependency pins in `package.nix` with `let in` syntax.
+  ```nix
+  {
+    package-a_version,
+    # ...
+  }:
+  let
+    package-a = package-a_version;
+  in
+  # ...
+  ```
+  - Possibly harder to automate dependency pin removal.
+
 - Your suggestion here.
 
 # Prior art
--- a/rfcs/0192-by-name-version-pins.md
+++ b/rfcs/0192-by-name-version-pins.md
@@ -1,0 +1,109 @@
+---
+feature: by-name-version-pins
+start-date: 2025-11-25
+author: Quantenzitrone
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: https://github.com/NixOS/nixpkgs/pull/464396
+---
+
+# Summary
+[summary]: #summary
+
+Add an additional file `pins.nix` to the `pkgs/by-name` structure that allows overriding/pinning
+package versions.
+
+# Motivation
+[motivation]: #motivation
+
+> Why are we doing this?
+
+- The `pkgs/by-name` structure doesn't have support for overriding packages yet
+  leading to a lot of remaining entries in `pkgs/top-level/all-packages.nix`.
+- Packages with version pins often get neglected and stay on the old version
+  for a long time, even though it already supports the new version. By having
+  version pins in a defined location with a defined structure, package update
+  bots or scripts could check if the package works without the pin and remove it
+  automatically on update.
+
+> What use cases does it support?
+
+- Pinning package versions of packages.
+
+> What is the expected outcome?
+
+- Less confusion if package pins should go into `all-packages.nix` or be inlined
+  in the `package.nix`.
+- Less maintenance burden due to possible automation of pin removal.
+
+# Detailed design
+[design]: #detailed-design
+
+In addition to `pkgs/by-name/{shard}/{pname}/package.nix` there can also be
+`pkgs/by-name/{shard}/{pname}/pins.nix`. This file will have the following structure:
+
+```nix
+{
+  package-a_version,
+  package-b_version,
+  # [...]
+}:
+{
+  package-a = package-a_version;
+  package-b = package-b_version;
+  # [...]
+}
+```
+
+- Every attrName in the resulting attribute set has to be a valid package
+  attribute name.
+- Every attrValue should be the pinned version of the respectice attrName
+  package, this is however hard to check I think.
+- Every attrName has to be a functionArg of the `package.nix`.
+
+I think you know where this is going.
+
+Packages with a `pin.nix` will have the versions of packages pinned
+accordingly. This can be easily archived with:
+
+```nix
+if lib.pathExists (packageDirectory + "/pins.nix") then
+  callPackage (packageDirectory + "/package.nix") (
+    lib.removeAttrs (callPackage (packageDirectory + "/pins.nix") { }) [
+      "override"
+      "overrideDerivation"
+    ]
+  )
+else
+  callPackage (packageDirectory + "/package.nix")
+```
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Probably longer eval time, this has to be tested however.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep the current situation: version pins are either in `all-packages.nix` or
+  directly in the `package.nix`.
+- Some more generalized `overrides.nix` without the strict requirements.
+- Your suggestion here.
+
+# Prior art
+[prior-art]: #prior-art
+
+- [RFC 140](https://github.com/NixOS/rfcs/blob/master/rfcs/0140-simple-package-paths.md)
+  lists the problem this RFC is trying to solve under "Future work".
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+# Future work
+[future]: #future-work
+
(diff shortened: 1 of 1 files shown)


rfcs/0192-by-name-version-pins.md
```diff
@@ -0,0 +1,109 @@
...
+
+# Examples and Interactions
+[examples-and-interactions]: #examples-and-interactions
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Probably longer eval time, this has to be tested however.
+
+# Alternatives
```
kevincox, line 91: I wonder if it would be cleaner to use default function parameters. So something like:

```nix
{
    package-a_version,
    package-a ? package-a_version,
}: ...
```

Although I don't know if this would require extra magic and possibly language features to avoid `callPackage` from overriding `package-a` with the default version. I suspect it would, so the benefit of the current proposal would be that it doesn't need language changes which is definitely compelling.

rfcs/0192-by-name-version-pins.md
```diff
@@ -0,0 +1,109 @@
...
+- Every attrName in the resulting attribute set has to be a valid package
+  attribute name.
+- Every attrValue should be the pinned version of the respectice attrName
+  package, this is however hard to check I think.
+- Every attrName has to be a functionArg of the `package.nix`.
+
+I think you know where this is going.
+
+Packages with a `pin.nix` will have the versions of packages pinned
+accordingly. This can be easily archived with:
```
mdaniels5757:
```suggestion
accordingly. This can be easily achieved with:
```

rfcs/0192-by-name-version-pins.md
GetPsyched, line 91: IMO versions should be a first-class feature of nixpkgs such that, to an end user, searching for a package should give you just _one_ package with all of its versions listed as a part of it as opposed to each pinned version being its own exposed package (this argument could extend to package variants too but that goes out of scope).

This could go hand-in-hand with the current `pins.nix` approach too where the pins in it are not exposed at the top-level but simply as alternative versions to the existing package. As an end user, that's a far better UX than having multiple top-level packages for the same thing.

I know that this would involve significantly more work than the current proposal, but given that Nix's usability has been a concern for a long time, I'd want to push this along with this change since it might be harder down the line.

7c6f434c: The packages will still need to have default versions, doing otherwise is a usability disaster, and for pin.nix nothing changes if `sbcl_2_0_0` is replaced with `sbcl.versions."2.0.0"`

So the change you propose is not reasonably tied to this one.

vcunat: There aren't just different `.version` numbers.  You also have e.g. the `*Full` variants.

robsliwi:
Related: https://github.com/NixOS/nixpkgs/pull/421201

rfcs/0192-by-name-version-pins.md
7c6f434c: I think this adds complexity to the version unification proposal but not to its interaction with the pins!

GetPsyched: I agree that this is not quite intersecting with the proposed change. I hoped to push this along since it's sort of related by increasing the scope of the RFC in case folks were interested. Sigh. Perhaps another day, in another RFC.

quantenzitrone: While this is possible, the way `callPackage` works for `pkgs/by-name` would have to change and I think this would lead to confusion of default values not getting overridden.

quantenzitrone: I added this suggestion under alternatives.

quantenzitrone: I changed the rfc text so it's more clear that it's about pinning dependency versions and not the packages themselves.
I also added pinning versions of the packages themselves under future work.

rfcs/0192-by-name-version-pins.md
```diff
@@ -0,0 +1,141 @@
...
+  package-b = package-b_version;
+  # ...
+}
+```
+
+- Every attrName in the resulting attribute set has to be a valid package
+  attribute name.
+- Every attrValue should be the pinned version of the respectice attrName
+  package, this is however hard to check I think.
+- Every attrName has to be a functionArg of the `package.nix`.
```
gepbird, line 63: Maybe this is out of scope, but can we also pin dependencies that not just differ in their semantic versions like `nodejs = nodejs_22`, but have other overrides or logic?
Some example `pins.nix`:
```nix
{
  darwin,
  stdenv,
}@args:
{
  stdenv = if args.stdenv.hostPlatform.isDarwin then darwin.bootstrapStdenv else args.stdenv;
}
```

```nix
{
  luajit,
}@args:
{
  luajit = args.luajit.override { enable52Compat = true; };
}
```

This would be useful to keep compatibility with previous overrides of a package when [removing by-name overrides](https://github.com/NixOS/nixpkgs/pull/453948#discussion_r2455738507).

rfcs/0192-by-name-version-pins.md
```diff
@@ -0,0 +1,141 @@
...
+- Your suggestion here.
+
+# Prior art
+[prior-art]: #prior-art
+
+- [RFC 140](https://github.com/NixOS/rfcs/blob/master/rfcs/0140-simple-package-paths.md)
+  lists the problem this RFC is trying to solve under "Future work".
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
```
gepbird, line 135: When overriding a package, will it override the the arguments of `pins.nix` or the arguments of `package.nix`?
Perhaps it could override `package.nix` and we could have an `overridePins` function.

quantenzitrone: `.override {}` overrides the `package.nix`
why would someone need to override the pins?
```nix
package.overridePins {
  dependency_3 = my-dependency_3;
}
```
would be the same as
```nix
package.override {
  dependency = my-dependency_3;
}
```

rfcs/0192-by-name-version-pins.md
quantenzitrone: i think we lose some of the usefulness if we allow that
also like with version pins we can already do that just in the package.nix with `let in`

rfcs/0192-by-name-version-pins.md
7c6f434c: I think there is an ambiguity here. Overrides relative to `callPackage path/to/package.nix {}`  are not provided at all, because this would break more than help (and also `override` works relative to a package that is defined, and the package defined does use pins). You can just override all the pinned stuff one-by-one, though, if for some specific package it makes sense.

quantenzitrone: sorry, i don't understand what you are trying to say

rfcs/0192-by-name-version-pins.md
gepbird: Yes, `let in` handles most of the cases, but there are a few edge cases when it breaks override compatibility.

## package called in all-package.nix

Consider this package and it's possible usage before it being fully self contained in `by-name`:

```nix
# file: pkgs/all-packages.nix
{
  aegisub = callPackage ../by-name/ae/aegisub/package.nix {
    luajit = luajit.override { enable52Compat = false; };
  };

  # ...
}
```

```nix
# file: pkgs/by-name/ae/aegisub/package.nix
{
  stdenv,
  luajit,
  # ...
}:
stdenv.mkDerivation {
  buildInputs = [ luajit ];
  # ...
}
```

```nix
# file: configuration.nix
{ pkgs, ... }:
{
  environment.systemPackages = [
    (pkgs.aegisub.override { luajit = luajit_2_0.override { enable52Compat = true; })
  ]
}
``` 

The above code builds fine, `luajit_2_0.override { enable52Compat = true;` will be in `buildInputs`.

## package is only in by name using `let in`, breaks override compatibility

Now let's move the override from all-packages to the by-name package:

```nix
# file: pkgs/all-packages.nix
{
  # no override for aegisub
}
```

```nix
# file: pkgs/by-name/ae/aegisub/package.nix
{
  stdenv,
  luajit,
  # ...
}:
let
  luajit' = luajit.override { enable52Compat = true; };
in
stdenv.mkDerivation {
  buildInputs = [ luajit' ];
  # ...
}
```

```nix
# file: configuration.nix
{ pkgs, ... }:
{
  environment.systemPackages = [
    (pkgs.aegisub.override { luajit = luajit_2_0.override { enable52Compat = true; })
  ]
}
``` 

The above code fails to build, `(luajit_2_0.override { enable52Compat = true; }).override { enable52Compat = false; }` will be in `buildInputs`. There is no way to override a variable in a let binding.

## package using pins, override compatible and doesn't touch all-packages

To solve this, we may use versions pins:

```nix
# file: pkgs/all-packages.nix
{
  # no override for aegisub
}
```

```nix
# file: pkgs/by-name/ae/aegisub/package.nix
{
  stdenv,
  luajit
  # ...
}:
stdenv.mkDerivation {
  buildInputs = [ luajit ];
  # ...
}
```

```nix
# file: pkgs/by-name/ae/aegisub/pins.nix
{
  luajit
}:
{
  luajit = luajit.override { enable52Compat = false; };
}
```

```nix
# file: configuration.nix
{ pkgs, ... }:
{
  environment.systemPackages = [
    (pkgs.aegisub.override { luajit = luajit_2_0.override { enable52Compat = true; })
  ]
}
```

The above code should build and should produce the same derivation as the first, all-packages example.

(The details in this example is unrealistic: `enable52Compat` is set to `true` in nixpkgs and it's unlikely for someone to override it. The PR I linked in my previous comment talked about overriding platforms, but I didn't want to complicate this example by introducing platforms.)

rfcs/0192-by-name-version-pins.md
gepbird: >  why would someone need to override the pins?

I forgot to mention it's heavily related to my [other comment](https://github.com/NixOS/rfcs/pull/192#discussion_r2683573412), so it's only useful if we also want to pin non-versioned packages.

Following the last example in that thread, we have:

```nix
# file: pkgs/by-name/ae/aegisub/pins.nix
{
  luajit
}:
{
  luajit = luajit.override { enable52Compat = false; };
}
```

```
# file: pkgs/by-name/ae/aegisub/package.nix
{
  stdenv,
  luajit
  # ...
}:
stdenv.mkDerivation {
  buildInputs = [ luajit ];
  # ...
}
```

The `package.nix` override that turns on `enable52Compat` could be written as:
```nix
aegisub.override {
  luajit = luajit.override { enable52Compat = true; };
}
```

And a `pins.nix` override that uses `luajit_2_0` instead of `luajit` could be written as:
```nix
aegisub.overridePins {
  luajit = luajit_2_0;
}
```

The latter will result in `luajit_2_0.override { enable52Compat = false; }` getting into `buildInputs`, because it replace's the pin's input, the pin adds the `override { enable52Compat = false; }` and passes this to the package's input.

rfcs/0192-by-name-version-pins.md
quantenzitrone: I think this falls under
> - Some more generalized `overrides.nix` without the strict requirements.

While I get the problem this is trying to solve, in my opinion it overcomplicates things.
Especially the resulting need to have `overrideOverrides`/`overridePins` is too confusing for people I think.

7c6f434c: Why would you need overridePins? This would be rare enough that «just copy the override and tune it, still better than let-in» should be sufficient.

The current attempts in minimising all-packages.nix are pure regressions for overriding interface quality, this would be an improvement.

PR #193, opened 2025-12-07
[RFC 0193] TOML Flakes

roberth:
As a mere community member I have a proposal for an improvement to flakes.
This can be done concurrently with the efforts to bring already developed flake related features up to scratch as part of stabilization.
Let's get out of analysis paralysis and move the project forward in a way that's positive, not just focused on tech debt and inclusive of community input.
Yes, I believe we need more flakes RFCs, precisely because they didn't fit into the original one.

[Rendered](https://github.com/roberth/rfcs/blob/declarative-flakes/rfcs/0193-toml-flakes.md)

### DIFF
--- a/rfcs/0193-toml-flakes.md
+++ b/rfcs/0193-toml-flakes.md
@@ -27,7 +27,7 @@ This creates several problems:
 
 1. **User uncertainty**: The current structure creates confusion about what language features are available
    and when evaluation restrictions apply,
-   because inputs and outputs are mixed in the same file.
+   because inputs and outputs are mixed in the same file, and inputs are not allowed to be arbitrary Nix expressions.
    See [issue #4945](https://github.com/NixOS/nix/issues/4945).
    Note that while the question was originally asked in a partly rhetorical manner,
    it is still a valid question,
--- a/rfcs/0193-toml-flakes.md
+++ b/rfcs/0193-toml-flakes.md
@@ -16,7 +16,7 @@ Make flakes easier to use, automate and learn.
 Introduce `flake.toml` as the new leading file for flakes,
 separating input specifications from output definitions.
 The `flake.toml` file contains declarative metadata and input declarations,
-while `flake.nix` (read by the built-in entrypoint) or a framework focuses on output definitions.
+while `flake.nix` focuses on output definitions.
 
 Fixes [#4945 What language is flake.nix written in?](https://github.com/NixOS/nix/issues/4945)
 
@@ -40,9 +40,6 @@ This creates several problems:
 2. **Limited automation**: Programmatically editing flake inputs requires Nix AST manipulation,
    which is complex and error-prone compared to editing structured data formats.
 
-3. **Boilerplate**: Common flake patterns require repetitive Nix code
-   that could be handled by frameworks if inputs were separated from implementation.
-
 By moving input specifications to a simpler format,
 we enable better tooling,
 reduce user confusion about evaluation restrictions,
@@ -57,11 +54,7 @@ and create a clearer separation of concerns between dependency declarations and 
 containing input sources and follows relationships.
 It complies with the [Nix JSON guideline](https://nix.dev/manual/nix/latest/development/json-guideline.html) (modulo `null`).
 
-Flake output invocation is changed to go through an "entrypoint", which is the flake that provides a function to help produce the flake output attributes based on the source tree of `flake.toml`.
-If `inputs.entrypoint` exists, it becomes the entrypoint.
-Otherwise if `flake.nix` exists, the built-in entrypoint is used.
-
-`flake.nix` is read by the built-in entrypoint and defines outputs.
+`flake.nix` remains and defines outputs.
 
 ## Choice of TOML
 
@@ -74,18 +67,16 @@ It has a wide ecosystem of libraries for parsing, as well as good number of libr
 ## Relationship to existing files
 
 - `flake.lock` remains unchanged
-- Alternative entrypoints (specified via `inputs.entrypoint` in `flake.toml`) can read different files
-- `flake.nix` remains supported as the file read by the built-in entrypoint, and as a legacy format for the `inputs` and other metadata.
+- `flake.nix` remains and defines outputs, and as a legacy format for the `inputs` and other metadata.
 
 ## Compatibility
 
 The design should support reading legacy `flake.nix` files that contain inline input specifications.
 
 The following negative space is changed and may require a few projects to adapt if they already use these:
-- `inputs.entrypoint` is assigned meaning, changing how flake outputs are formed.
 - A file named `flake.toml` will shadow `flake.nix` in file discovery
 
-Flakes that do not have any of the above elements remain compatible.
+Flakes that do not have a `flake.toml` file remain compatible.
 
 A TOML flake can not be evaluated by older implementations of Nix.
 
@@ -126,7 +117,7 @@ description = "A simple example flake exporting GNU hello for x86_64-linux"
 url = "github:NixOS/nixpkgs/nixos-unstable"
 ```
 
-**flake.nix** (read by built-in entrypoint):
+**flake.nix**:
 ```nix
 {
   outputs = { self, nixpkgs }:
@@ -141,76 +132,36 @@ url = "github:NixOS/nixpkgs/nixos-unstable"
 
 **flake.lock** remains unchanged.
 
-## Proposed structure, framework entrypoint
+## Proposed structure with follows
 
 **flake.toml**:
 ```toml
-description = "A simple example flake exporting GNU hello for x86_64-linux"
+description = "Example with follows relationships"
 
 [inputs.nixpkgs]
 url = "github:NixOS/nixpkgs/nixos-unstable"
 
-[inputs.entrypoint]
+[inputs.flake-parts]
 url = "github:hercules-ci/flake-parts"
 
-[inputs.entrypoint.inputs.nixpkgs-lib]
+[inputs.flake-parts.inputs.nixpkgs-lib]
 follows = "nixpkgs"
 ```
 
-**parts.nix** (read by flake-parts entrypoint):
-```nix
-{ ... }: {
-  perSystem = { pkgs }: {
-    packages.default = pkgs.hello;
-  };
-}
-```
-
-**flake.lock** remains unchanged.
-
-## Entrypoint function
-
-`flake.nix` serves as the base that bootstraps entrypoints.
-This example shows what authoring a framework looks like.
-An entrypoint is invoked as `<flake outputs>.lib.callFlakeEntrypoint`:
-
+**flake.nix**:
 ```nix
 {
-  outputs = { self, nixpkgs }:
-    let
-      system = "x86_64-linux";
-      pkgs = nixpkgs.legacyPackages.${system};
-    in {
-      packages.${system}.default = pkgs.hello;
-
-      # Entrypoints are accessed here
-      lib.callFlakeEntrypoint =
-        { 
-          # The resolved and invoked/dependency-injected flake inputs
-          inputs,
-          # The parsed flake.toml
-          flakeToml,
-          # The invoked/dependency-injected *result*
-          self,
-          # The base directory of flake.toml
-          flakeDir,
-          # The sourceInfo, where `sourceInfo.outPath` may be `flakeDir` or any
-          # parent of it (if applicable; subdirectory flakes)
-          sourceInfo,
-          # ... is mandatory for forward compatibility
-          ...
-        }:
-        # Imagine some useful expression, returning the usual flake outputs
-        {
-          packages = { ... };
-          apps = { ... };
-        };
+  outputs = { self, nixpkgs, flake-parts }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      systems = [ "x86_64-linux" ];
+      perSystem = { pkgs, ... }: {
+        packages.default = pkgs.hello;
+      };
     };
 }
 ```
 
-The built-in entrypoint reads `flake.nix` and expects an `outputs` attribute.
-Alternative entrypoints (specified via `inputs.entrypoint`) can implement different conventions.
+**flake.lock** remains unchanged.
 
 The exact schema shown above is illustrative.
 The final design will follow the Nix JSON guideline for extensibility,
@@ -235,28 +186,6 @@ Maintain the status quo including drawbacks described in the [motivation].
 
 Improve Nix AST manipulation tools instead of introducing a new format.
 Does not solve user uncertainty.
-Does not solve boilerplate.
-
-## Explicit `entrypoint` field in flake.toml
-
-An earlier design included an explicit `entrypoint` field in `flake.toml`:
-
-```toml
-entrypoint = "my-framework"
-
-[inputs.my-framework]
-url = "github:..."
-```
-
-This would take precedence over the `inputs.entrypoint` convention.
-
-It seemed beneficial for the initial migration,
-where flakes that already had an `entrypoint` input for other reasons (unlikely) would not have to rename that input.
-However, that was not a real solution,
-and a special null value would have to be added in order to select the built-in (`flake.nix`) entrypoint,
-since TOML does not have a distinct `null` type.
-This would not have lead to a great solution,
-and wasn't worth the permanent complexity of an unnecessary indirection.
 
 ## Use JSONC
 
@@ -281,28 +210,21 @@ where binary caches aren't shared between Nix versions and old expressions produ
 It incorporated a "poor man's module system" into Nix where it would be of very limited functionality,
 while it ossifies due the consequences in Nix of [Hyrum's law](https://www.hyrumslaw.com/) and the second order effects of reproducibility.
 
+## Flake entrypoints RFC
+
+A companion RFC proposes a flake entrypoint mechanism that allows frameworks to handle output generation.
+Combined with TOML flakes,
+this provides the bigger picture of removing custom Nix code from flake metadata altogether,
+allowing frameworks to handle both input processing and output generation declaratively.
+
+The entrypoint mechanism can be implemented independently of TOML,
+but the two features work well together.
+
 ## devenv
 
 The devenv tool uses a separate configuration format for specifying dependencies
 and generating Nix configurations,
 demonstrating that declarative input specifications can work well in practice.
-
-They have also integrated input declarations and input usage in their application of the module system.
-This RFC could be extended or followed up with a metadata feature that generalizes this idea and allows an entrypoint to achieve a similar effect.
-
-```toml
-[inputs.entrypoint]
-url = "github:cachix/git-hooks.nix"
-# forwarded as e.g. `inputUsages.<input>` to entrypoint, so it can import/enable/etc as intended.
-usage = [ "flake-parts" ]
-```
-
-## flake-parts, flake-utils-plus and blueprint
-
-Framework systems like flake-parts, flake-utils-plus and blueprint already provide abstractions over flake outputs.
-This proposal would enable these frameworks to be specified declaratively,
-in a standard way,
-without boilerplate.
 
 ## Other language ecosystems
 
--- a/rfcs/0193-toml-flakes.md
+++ b/rfcs/0193-toml-flakes.md
@@ -16,7 +16,7 @@ Make flakes easier to use, automate and learn.
 Introduce `flake.toml` as the new leading file for flakes,
 separating input specifications from output definitions.
 The `flake.toml` file contains declarative metadata and input declarations,
-while `flake.nix` (read by the default entrypoint) or a framework focuses on output definitions.
+while `flake.nix` (read by the built-in entrypoint) or a framework focuses on output definitions.
 
 Fixes [#4945 What language is flake.nix written in?](https://github.com/NixOS/nix/issues/4945)
 
@@ -54,15 +54,14 @@ and create a clearer separation of concerns between dependency declarations and 
 ## File structure
 
 `flake.toml` becomes the leading file,
-containing input sources, follows relationships, and entrypoint selection.
+containing input sources and follows relationships.
 It complies with the [Nix JSON guideline](https://nix.dev/manual/nix/latest/development/json-guideline.html) (modulo `null`).
 
 Flake output invocation is changed to go through an "entrypoint", which is the flake that provides a function to help produce the flake output attributes based on the source tree of `flake.toml`.
-If `flake.toml` has an `entrypoint` field, it must name one of the `inputs` which will serve as the entrypoint.
-Otherwise, if an input named `"entrypoint"` exists, it becomes the entrypoint.
-Finally if `flake.nix` exists, the entrypoint is the default entrypoint.
+If `inputs.entrypoint` exists, it becomes the entrypoint.
+Otherwise if `flake.nix` exists, the built-in entrypoint is used.
 
-`flake.nix` is read by the default entrypoint and defines outputs.
+`flake.nix` is read by the built-in entrypoint and defines outputs.
 
 ## Choice of TOML
 
@@ -75,15 +74,15 @@ It has a wide ecosystem of libraries for parsing, as well as good number of libr
 ## Relationship to existing files
 
 - `flake.lock` remains unchanged
-- Alternative entrypoints (referenced in `flake.toml`) can read different files
-- `flake.nix` remains supported as the default entrypoint, and as a legacy format for the `inputs` and other metadata.
+- Alternative entrypoints (specified via `inputs.entrypoint` in `flake.toml`) can read different files
+- `flake.nix` remains supported as the file read by the built-in entrypoint, and as a legacy format for the `inputs` and other metadata.
 
 ## Compatibility
 
 The design should support reading legacy `flake.nix` files that contain inline input specifications.
 
 The following negative space is changed and may require a few projects to adapt if they already use these:
-- A flake input named `entrypoint` is assigned meaning, changing how flake outputs are formed.
+- `inputs.entrypoint` is assigned meaning, changing how flake outputs are formed.
 - A file named `flake.toml` will shadow `flake.nix` in file discovery
 
 Flakes that do not have any of the above elements remain compatible.
@@ -127,7 +126,7 @@ description = "A simple example flake exporting GNU hello for x86_64-linux"
 url = "github:NixOS/nixpkgs/nixos-unstable"
 ```
 
-**flake.nix** (read by default entrypoint):
+**flake.nix** (read by built-in entrypoint):
 ```nix
 {
   outputs = { self, nixpkgs }:
@@ -158,7 +157,7 @@ url = "github:hercules-ci/flake-parts"
 follows = "nixpkgs"
 ```
 
-**parts.nix** (read by default entrypoint):
+**parts.nix** (read by flake-parts entrypoint):
 ```nix
 { ... }: {
   perSystem = { pkgs }: {
@@ -210,8 +209,8 @@ An entrypoint is invoked as `<flake outputs>.lib.callFlakeEntrypoint`:
 }
 ```
 
-The default entrypoint reads `flake.nix` and expects an `outputs` attribute.
-Alternative entrypoints (specified in `flake.toml`) can implement different conventions.
+The built-in entrypoint reads `flake.nix` and expects an `outputs` attribute.
+Alternative entrypoints (specified via `inputs.entrypoint`) can implement different conventions.
 
 The exact schema shown above is illustrative.
 The final design will follow the Nix JSON guideline for extensibility,
@@ -237,6 +236,27 @@ Maintain the status quo including drawbacks described in the [motivation].
 Improve Nix AST manipulation tools instead of introducing a new format.
 Does not solve user uncertainty.
 Does not solve boilerplate.
+
+## Explicit `entrypoint` field in flake.toml
+
+An earlier design included an explicit `entrypoint` field in `flake.toml`:
+
+```toml
+entrypoint = "my-framework"
+
+[inputs.my-framework]
+url = "github:..."
+```
+
+This would take precedence over the `inputs.entrypoint` convention.
+
+It seemed beneficial for the initial migration,
+where flakes that already had an `entrypoint` input for other reasons (unlikely) would not have to rename that input.
+However, that was not a real solution,
+and a special null value would have to be added in order to select the built-in (`flake.nix`) entrypoint,
+since TOML does not have a distinct `null` type.
+This would not have lead to a great solution,
+and wasn't worth the permanent complexity of an unnecessary indirection.
 
 ## Use JSONC
 
--- a/rfcs/0193-toml-flakes.md
+++ b/rfcs/0193-toml-flakes.md
@@ -18,6 +18,8 @@ separating input specifications from output definitions.
 The `flake.toml` file contains declarative metadata and input declarations,
 while `flake.nix` (read by the default entrypoint) or a framework focuses on output definitions.
 
+Fixes [#4945 What language is flake.nix written in?](https://github.com/NixOS/nix/issues/4945)
+
 # Motivation
 [motivation]: #motivation
 
--- a/rfcs/0193-toml-flakes.md
+++ b/rfcs/0193-toml-flakes.md
@@ -1,0 +1,309 @@
+---
+feature: toml-flakes
+start-date: 2025-12-07
+author: Robert Hensing (@roberth)
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Make flakes easier to use, automate and learn.
+
+Introduce `flake.toml` as the new leading file for flakes,
+separating input specifications from output definitions.
+The `flake.toml` file contains declarative metadata and input declarations,
+while `flake.nix` (read by the default entrypoint) or a framework focuses on output definitions.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, flakes combine two distinct concerns in a single `flake.nix` file:
+declaring dependencies (inputs) and implementing functionality (outputs).
+This creates several problems:
+
+1. **User uncertainty**: The current structure creates confusion about what language features are available
+   and when evaluation restrictions apply,
+   because inputs and outputs are mixed in the same file.
+   See [issue #4945](https://github.com/NixOS/nix/issues/4945).
+   Note that while the question was originally asked in a partly rhetorical manner,
+   it is still a valid question,
+   a variation of which pops into new users' minds.
+   Some learning is always required,
+   but this is an unnecessary bump in the curve.
+
+2. **Limited automation**: Programmatically editing flake inputs requires Nix AST manipulation,
+   which is complex and error-prone compared to editing structured data formats.
+
+3. **Boilerplate**: Common flake patterns require repetitive Nix code
+   that could be handled by frameworks if inputs were separated from implementation.
+
+By moving input specifications to a simpler format,
+we enable better tooling,
+reduce user confusion about evaluation restrictions,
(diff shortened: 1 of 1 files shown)


rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,309 @@
...
+# Summary
+[summary]: #summary
+
+Make flakes easier to use, automate and learn.
+
+Introduce `flake.toml` as the new leading file for flakes,
+separating input specifications from output definitions.
+The `flake.toml` file contains declarative metadata and input declarations,
+while `flake.nix` (read by the default entrypoint) or a framework focuses on output definitions.
+
```
roberth, line 20:
```suggestion

Fixes [#4945 What language is flake.nix written in?](https://github.com/NixOS/nix/issues/4945)

```

- https://github.com/NixOS/nix/issues/4945

limwa:
Couldn't this be achieved by doing something like the following?

```nix
{
    inherit (builtins.fromTOML (builtins.readFile ./flake.toml)) description inputs;

    outputs = ...;
}
```

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,309 @@
...
+# Motivation
+[motivation]: #motivation
+
+Currently, flakes combine two distinct concerns in a single `flake.nix` file:
+declaring dependencies (inputs) and implementing functionality (outputs).
+This creates several problems:
+
+1. **User uncertainty**: The current structure creates confusion about what language features are available
+   and when evaluation restrictions apply,
+   because inputs and outputs are mixed in the same file.
```
roberth:
```suggestion
   because inputs and outputs are mixed in the same file, and inputs are not allowed to be arbitrary Nix expressions.
```

roberth:
@limwa Nix restricts flake `inputs` to a subset of the language that does not permit that. I'll add https://github.com/NixOS/rfcs/pull/193#discussion_r2596709067 to address this.

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,311 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Introduces another file format into the Nix ecosystem
+- Migration cost for existing flakes
+- Potential confusion during transition period with two ways to specify inputs
+- Requires tooling updates across the ecosystem
+
+# Alternatives
```
rhendric, line 178: One alternative worth considering: use Dhall instead of TOML.

Many of the complaints about the flake.nix ‘language’ seem to be rooted in the desire to abstract out parts of the inputs for reuse. TOML doesn't actually address this desire; it merely makes it clearer to the user that the desire can't be satisfied. The motivation to restrict how inputs are expressed seems to be to prevent arbitrary code from running, either for security reasons or so that flake input analysis can be guaranteed to terminate. Dhall addresses both concerns, while allowing many of the abstractions flake users are trying to write; it's pretty much made for exactly this use case.

roberth: That direction is more in line with the discussion that followed in https://github.com/NixOS/nix/issues/4945.
Dhall is still somewhat impractical, unless perhaps the Rust implementation could be wrapped such that Nix can use it as a library.
I believe its effect on the learning curve is also negative.
The main use case of programmatic inputs seem to be auto-follows, which could just be a feature, or getting inputs from some shared expression.
I personally wouldn't mind allowing evaluation in `inputs`. If someone wants to make them complicated, non-terminating, whatever, that's on them.

Regarding Dhall as a point in the configuration language design space, I might upset some folks, including my past self who was a big fan of that kind of tech before using it in conjunction with Nix.
While I appreciate the science of it, I find Turing incompleteness to be a bit of a gimmick and an unnecessary constraint for this kind of application. It has a real cost in terms of what kinds of programs can be expressed with it, while gaining very little practical value: either you get an error, or you don't.
The tradeoff is between getting 1 success and 1 mediocre error from a language like Nix vs 1 unnecessary impossibility and 1 nicer error from Dhall.
Or we could just keep it boring and use TOML.

I guess we could add a helper trampoline primop to Dhall to add some Turing completeness for the algorithmically inclined?

I'll boil this rant down into a more reasonable addition later, or I'd be happy to take a GitHub suggestion.

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,311 @@
...
+**parts.nix** (read by default entrypoint):
+```nix
+{ ... }: {
+  perSystem = { pkgs }: {
+    packages.default = pkgs.hello;
+  };
+}
+```
+
+**flake.lock** remains unchanged.
```
7c6f434c: I guess you implicitly add locking the entrypoint flake?

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Keep current structure
+
+Maintain the status quo including drawbacks described in the [motivation].
+
+## Pure Nix with better tooling
+
+Improve Nix AST manipulation tools instead of introducing a new format.
+Does not solve user uncertainty.
+Does not solve boilerplate.
```
crertel: I'll point out that the main complaints about this somewhat go away with taking  "pure nix" tooling and also taking  the admission that both user uncertainty and boilerplate in 2025 are concerns that are rapidly declining.

Bluntly: any of the major LLMs right now are perfectly fine at making decent `flake.nix` files, and where they fall short a different configuration language or helper probably wouldn't be much of an improvement.

rfcs/0193-toml-flakes.md
TLATER: Yaml could also take care of a lot of the desires here, without being fully turing complete, and without being as obscure. It has a variety of variable-like features.

That said, I think simplifying the format as much as possible, and indicating that it *is* deliberately limited is part of the appeal of this proposal. Making inputs templateable just leads to complexity that ends up confusing newcomers to the ecosystem.

I think most use cases for evaluation in inputs that don't just become features outright could be handled in external scripts instead, given the file would be much more programmatically editable.

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,311 @@
...
+
+## File structure
+
+`flake.toml` becomes the leading file,
+containing input sources, follows relationships, and entrypoint selection.
+It complies with the [Nix JSON guideline](https://nix.dev/manual/nix/latest/development/json-guideline.html) (modulo `null`).
+
+Flake output invocation is changed to go through an "entrypoint", which is the flake that provides a function to help produce the flake output attributes based on the source tree of `flake.toml`.
+If `flake.toml` has an `entrypoint` field, it must name one of the `inputs` which will serve as the entrypoint.
+Otherwise, if an input named `"entrypoint"` exists, it becomes the entrypoint.
```
TLATER: I can imagine use cases for declaring multiple entrypoints that define different sets of outputs.

If multiple are defined, nix could basically just recursively `//` the `outputs` of the function returns; collisions should be uncommon if the ecosystem evolves around their possibility, and most use cases for this would probably be to provide two different outputs which would be completely orthogonal anyway.

E.g., having the option of specifying multiple entrypoints would make writing a flake that has both `nixosConfiguration`s *and* `homeManagerConfiguration`s possible, without the boilerplate of maintaining a specialized entrypoint flake for this use case.

rfcs/0193-toml-flakes.md
rhendric: I'm not myself convinced that Dhall would be the best choice, but a lot of these objections strike me as not very strong.

> Dhall is still somewhat impractical, unless perhaps the Rust implementation could be wrapped such that Nix can use it as a library.

Even if there is no convenient binding, Nix could execute `dhall-to-json --file flake.dhall` and use the output; I'd be shocked if the overhead of a child process when interpreting flake input files would have any human-noticeable impact on perf.

> I believe its effect on the learning curve is also negative.

Simple flake.dhall file:

```dhall
{
  nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable",
  entrypoint = {
    url = "github:hercules-ci/flake-parts",
    inputs.nixpkgs-lib.follows = "nixpkgs",
  },
}
```

I don't think this would present more of a learning challenge to people not trying to dive into the intricacies of Dhall than the equivalent TOML would.

In other words: the strength of the argument that people don't need abstractions in their inputs is inversely proportional to the strength of the argument that it would be difficult to read and write inputs written in Dhall. If many people want it, all the more reason to use a non-data-only language. If few people want it, we won't see ‘advanced’ Dhall features (scare quotes because, you know, I'm talking about things like let bindings) very often in the wild (and users can make up their own mind whether having them is worth learning about them).

And it's worth emphasizing that this is not a hypothetical ‘what if people want it’. We know users want it. Here's [this example](https://github.com/NixOS/nix/issues/4945#issuecomment-1869931785) as it might have been written in Dhall:

```dhall
let dep = \(url : Text) -> { url = url, inputs.nixpkgs.follows = "nixpkgs" }
in {
  nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable",
  entrypoint = {
    url = "github:hercules-ci/flake-parts",
    inputs.nixpkgs-lib.follows = "nixpkgs",
  },
  nix-darwin = dep "github:LnL7/nix-darwin",
  home-manager = dep "github:nix-community/home-manager",
  zig = dep "github:Cloudef/nix-zig-stdenv",
  zls = dep "github:zigtools/zls",
  hyprland = dep "github:hyprwm/Hyprland",
  eww = dep "github:elkowar/eww",
}
```

This strikes me as something that said user would have been able to learn how to write, given what they wrote in Nix.

> It has a real cost in terms of what kinds of programs can be expressed with it

When compared to TOML as the alternative, this isn't any sort of disadvantage; the programs TOML is able to express are a strict subset of the programs Dhall is able to express, obviously.

> The tradeoff is between getting 1 success and 1 mediocre error from a language like Nix vs 1 unnecessary impossibility and 1 nicer error from Dhall.
> Or we could just keep it boring and use TOML.

Using (unrestricted) Nix would be keeping it boring. In the same framework, the tradeoff is between TOML's 2 unnecessary impossibilities (not-provably-total programs, and simple deduplication of the sort we actually see people trying to write) and 1 nicer error, Dhall's 1 unnecessary impossibility (NPT programs) and 1 nicer error, or Nix's 0 unnecessary impossibilities and 1 mediocre error. On just these axes, TOML is not on the efficient frontier.

rfcs/0193-toml-flakes.md
roberth: It is a regular input, so it is locked like any other input.
In an earlier iteration of this RFC I had it as a separate lock entry `entrypoint.url` that wasn't part of `inputs`, but that would just complicate the locking mechanism.

7c6f434c: Oh right, I see that in some place you say that `entrypoint` is used if specified but elsewhere it is `inputs.entrypoint`. (Maybe it is better to make all mentions fully qualified?)

rfcs/0193-toml-flakes.md
roberth: This was in reference to the outputs function more so than the inputs.

Boilerplate wasn't much of a problem even before LLMs.
Realistically how does reading 20-40% extra code which is *visually similar* stack up against having to evaluate code before being able to see the outcome?
Explicit is good.

rfcs/0193-toml-flakes.md
roberth: A simple flake could do that too. An entrypoint flake is just a flake with a specific function in it; anyone can call them. So you could have a flake that performs this merging task.

I'd argue that flake-parts already performs this task well, providing a decent framework for interaction between _modules_, as opposed to entrypoints which would typically assume ownership of the outputs, and collide at the various `default` attributes.

rfcs/0193-toml-flakes.md
roberth: > Nix could execute `dhall-to-json --file flake.dhall`

Performance is probably acceptable, but the static Nix binary would have to pull that in from somewhere for basic functionality.

> simplifying the format as much as possible, and indicating that it _is_ deliberately limited is part of the appeal of this proposal.

This. It brings clarity and better scripting.

The comparison in this thread was between Dhall and full-Nix, but the RFC does not propose either.

rfcs/0193-toml-flakes.md
roberth: The possibility of specifying a different entrypoint than `"entrypoint"` in `flake.toml`'s `entrypoint` field would make that inaccurate. Clearly this needs further simplification. Let me strip that indirection out.

rfcs/0193-toml-flakes.md
ruro: > A simple flake could do that too. ..., as opposed to entrypoints which would typically assume ownership of the outputs, ....

Actually, this begs the question if the special `inputs.entrypoint` magic should even be part of this proposal. It seems to me that everything that `inputs.entrypoint` does can already be performed by essentially doing `outputs = inputs: inputs.entrypoint.makeFlake inputs ./. ...;`.

Based on the title of this RFC, it feels like `inputs.entrypoint` might be out of scope? Maybe extra convenience features that extend the current functionality of flakes like `inputs.entrypoint` should be considered in separate RFCs (possibly together with stuff like https://github.com/NixOS/nix/issues/5663, https://github.com/NixOS/nix/issues/3843, https://github.com/NixOS/nix/issues/7422#issuecomment-2112502283, https://github.com/NixOS/nix/issues/5157, https://github.com/NixOS/nix/issues/5567)?

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,331 @@
...
+# Motivation
+[motivation]: #motivation
+
+Currently, flakes combine two distinct concerns in a single `flake.nix` file:
+declaring dependencies (inputs) and implementing functionality (outputs).
+This creates several problems:
+
+1. **User uncertainty**: The current structure creates confusion about what language features are available
+   and when evaluation restrictions apply,
+   because inputs and outputs are mixed in the same file, and inputs are not allowed to be arbitrary Nix expressions.
```
ruro, line 32: I feel like this RFC should not just **assume** that the currently enforced status-quo limitation of "Flake input metadata can't contain arbitrarily complex, possibly non-terminating computations" must be kept as is, without justification.

I'm not necessarily saying that we should remove this restriction, but the RFC should **definitely** explore the pros / cons of this requirement more.

For example, do we know if this restriction is actually useful / needed? I think, I read somewhere that FlakeHub is relying on the fact that flake metadata can be computed statically-ish (I've never used FlakeHub, so I am not sure if this is actually the case). Are there any more examples where this property is actually useful?

If FlakeHub is the only (or one of the few) users of this restriction / guarantee, then maybe it should be implemented as a simple evaluation "depth" limiter on the side of FlakeHub, instead of in Flakes themselves?

Or maybe it's enough to have a simple way to opt-out of this restriction? Something like
```nix
{
    dynamic = true;
    inputs = {
        # Arbitrary nix expressions are now allowed inside inputs,
        # because `dynamic` was set to true. This flake might not have
        # full functionality on services like FlakeHub, but this is fine.
    };
}
```

rfcs/0193-toml-flakes.md
roberth: Scope could be reduced in that way. I like the bigger picture of the combined changes, but for clarity that seems like a good idea actually.

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,253 @@
...
+{
+  outputs = { self, nixpkgs }:
+    let
+      system = "x86_64-linux";
+      pkgs = nixpkgs.legacyPackages.${system};
+    in {
+      packages.${system}.default = pkgs.hello;
+    };
+}
+```
```
ruro, line 131: The choice of "reusing" / keeping `flake.nix` for the `outputs` part of the flake has the potential downside that in older versions of nix and other frameworks, this will look either like a malformed flake (that uses some inputs in its `outputs`, but doesn't declare them as `inputs`) or in this particular case the `nixpkgs` input would fall back to the global flake registry.

Maybe the split should be `flake.nix` -> `flake.toml` + `outputs.nix` instead of `flake.toml` + `flake.nix`?

rfcs/0193-toml-flakes.md
roberth: - #194 :tada:

avnik:
I like idea, that entrypoint could be used from imported entity (like in `parts.nix` example)

GoldsteinE:
If this is accepted, it would be nice to see a section for tool-specific metadata. [unflake](https://codeberg.org/goldstein/unflake/) could use it to store its config and, generally, it would provide an extension point for userspace flake-related utilities.

roberth:
I've moved the entrypoint bits to https://github.com/NixOS/rfcs/pull/194

KiaraGrouwstra:
What would be the expected usage if one were to not declare dependencies? I imagine not placing a TOML file could instead notify the user it could not find one. For reference, flakes' `inputs` attribute at present is optional.

roberth:
@KiaraGrouwstra Simple answer is "of course" you'd put a `description` there, but you raise an interesting question!

I viewed `flake.nix` as a secondary but sufficient file for defining a flake. That's how it used to be done, and I see no overriding reason to break that.
I don't feel strongly about warning such a situation, but leaning towards allowing and not even warning about a "missing" `flake.toml`.
I realize that's at the cost of eventual uniformity, but I don't feel we're at a stage to call such a decision anytime soon.

- [ ] Explain in RFC

KiaraGrouwstra:
(for what it's worth, i believe the description field to be optional as well - maybe just `outputs` wasn't so far)

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,253 @@
...
+
+See also the [alternatives] section.
+
+A nice aspect of TOML is that its non-nested syntax aligns with part of a definition of _declarative systems_, having a set of _independent statements_.
+
+It has a wide ecosystem of libraries for parsing, as well as good number of libraries that support round tripping edits.
+
+## Relationship to existing files
+
+- `flake.lock` remains unchanged
```
GetPsyched, line 69: Since `flake.toml` defines just the inputs and the lockfile is also just metadata about inputs being pinned, could we merge `flake.lock` with `flake.toml`? Is this too radical of a change? I personally would like to not maintain 3 files for flakes;
- one for inputs `flake.toml`,
- one for outputs `flake.nix`/`outputs.nix`, and
- one for pinning the inputs `flake.lock`.

limwa: I also think that using three separate files is a bit excessive for flakes. And since flakes are still experimental, supporting users who prefer non-experimental workflows means you’ll need to provide a `shell.nix` if your development setup relies on flakes, as well as a `default.nix` if you export packages so they remain compatible with `import` clauses.

Regarding the idea of merging `flake.toml` and `flake.lock`, I don’t believe that would be advisable. For example, template repositories (such as those in [NixOS/templates](https://github.com/NixOS/templates/blob/master/.gitignore)) depend on listing `flake.lock` in their `.gitignore` file to ensure it’s not committed to git. If `flake.toml` and `flake.lock` were combined, git wouldn’t offer a way to ignore only the lock changes made to that file. As a result, updating a template would require manual review to make sure no lock information is inadvertently added to `flake.toml`.

This is especially important because Nix automatically generates `flake.lock` (and, in turn, would update `flake.toml` once that file is introduced). That behavior could easily lead to unintended changes being committed without the developer noticing.

username-generic: Correct me if I'm wrong, but aren't `flake.lock` files usually automatically generated and left untouched by the user until an upgrade command is run?

Aside from tooling simplicity and reliability, for the use case of creating templates, wouldn't it make omitting the usually unnecessary version locks much more convenient for template creators?
As of now, you can keep version locks out of your repository with a simple `.gitignore` entry and keep iterating on the template directly with no additional work. 
With the merge of `flake.toml` and `flake.lock` files, the process would not be as simple, requiring things like omitting generated changes from commits, using special command arguments, or just biting the bullet.

I'm not saying that it'd be a _big_ burden with either case, but keeping the files separate just seems like the cleaner and easier to manage solution.

rfcs/0193-toml-flakes.md
crertel: That's a fair point.

rfcs/0193-toml-flakes.md
username-generic: @limwa Oh sorry, I didn't see your comment before starting to write mine! But I guess I'm not alone on this, since we wrote up the exact same example.

roberth: `flake.lock` would no longer be a throw-away file, and the editing experience would get a bit worse.
We'd have to ensure that the most relevant bits stay at the top.
JSON doesn't support comments (or automation support may regress if you do allow it)
GitHub hides lockfile changes, and even if that's changed, you'd have to open very large diffs even if you trust the committer.

I lean towards preserving full `flake.nix` support without warnings, so that might be a good solution for you.

Also, fwiw, if you're using a framework, then combined with #194, you may not need a `flake.nix` or `outputs.nix` anymore.

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,253 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Introduces another file format into the Nix ecosystem
+- Migration cost for existing flakes
+- Potential confusion during transition period with two ways to specify inputs
+- Requires tooling updates across the ecosystem
+
+# Alternatives
+[alternatives]: #alternatives
```
matklad, line 179: One more alternative is to explicitly define "data only" subset of `nix` which guarantees `O(input length)` eval times, and introduce special name&extension&file for it.

C.f `build.zig` and `build.zig.zon` (Zig object notation) in Zig, or, for that matter, `index.js` and `package.json`. 

We could have `flake.nix` and `flake.nix.nox` (Nix Object eXpression). 

This of course presupposes that we do not want to "program" `flake.nox`.

Given that `nix` can be on the bootstrap path, it seems relatively valuable to stick to what we already have, rather than introduce extra dependencies on other formats.

EDIT: we could also think about using nox instead of aterms in derivations?

rfcs/0193-toml-flakes.md
de11n: TOML has been chosen by projects like Rust and Python for the exact reason that it is *necessarily* simple. It doesn't let you get too fancy. YAML was ruled out because it allows *too much* flexibility. YAML is also bad at making things *clear* as complexity increases. Anyone complaining about Nix as an obscure language will not find Dhall an improvement. Anyone who needs the complexity should use Nix. Anyone who doesn't need it can use TOML.

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,253 @@
...
+# Motivation
+[motivation]: #motivation
+
+Currently, flakes combine two distinct concerns in a single `flake.nix` file:
+declaring dependencies (inputs) and implementing functionality (outputs).
+This creates several problems:
+
+1. **User uncertainty**: The current structure creates confusion about what language features are available
+   and when evaluation restrictions apply,
+   because inputs and outputs are mixed in the same file, and inputs are not allowed to be arbitrary Nix expressions.
```
truh: Adding a second language for flakes doesn't make it easier for new users to understand flakes, since they now need to learn how to use flakes in two languages instead of one.

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,253 @@
...
+we enable better tooling,
+reduce user confusion about evaluation restrictions,
+and create a clearer separation of concerns between dependency declarations and output implementations.
+
+# Detailed design
+[design]: #detailed-design
+
+## File structure
+
+`flake.toml` becomes the leading file,
```
truh, line 53: I think it's one of the greatest features of the nix ecosystem that everything is configured in a single language. 

I doubt that adding more configuration languages makes nix more approachable, since people will still have to understand both languages.

rfcs/0193-toml-flakes.md
roberth: Maybe it's because I have a simple mind, but I have great difficulty with nox if we could call it _Nixon_.

Anyway, on a more serious note, but still indulging myself, those would be *highly* similar formats.
I'd say step 1: allow Nix identifiers as object keys, trailing comma, remove decimal numbers => Nixon.
Step 2: rebase the Nix language on top of that, removing floats and `with`, maybe some other things => Jix

Material for some follow-up RFCs ;)

rfcs/0193-toml-flakes.md
```diff
@@ -0,0 +1,253 @@
...
+# Prior art
+[prior-art]: #prior-art
+
+## Experimental `configs` commit by Eelco
+
+[1dc3f53](https://github.com/NixOS/nix/commit/1dc3f5355a3786cab37a4de98ca46a859e015d89), part of stable branch [`configs`](https://github.com/NixOS/nix/compare/configs) implemented a `flake.toml` file.
+It incorporated a "poor man's module system" into Nix where it would be of very limited functionality,
+while it ossifies due the consequences in Nix of [Hyrum's law](https://www.hyrumslaw.com/) and the second order effects of reproducibility.
+
+## Flake entrypoints RFC
```
truh, line 213: I still don't really understand why we can't have a more dynamic flake.nix entrypoint. This would make it trivial to implement something like flake.toml in nix code.

```nix
{
  inputs = builtins.fromTOML (builtins.readFile ./inputs.toml); 
  outputs = import ./outputs.nix;
}
```

edolstra mentioned that the flake format needs to be simpler to limit complexity while querying flake metadata, but couldn't that metadata be written to the flake.lock file? The locking process being complex shouldn't really be a problem.

KiaraGrouwstra: on simplifying `outputs` as mentioned here, a current limitation is it must be explicitly defined as a function, i.e. one cannot currently use beta reduction to delegate its parameters to the `import` function as you did here - that is to say, using `outputs = args: import ./outputs.nix args;` instead does work currently.

on `inputs` using a simplified language, while for reading purposes any of those ways could likely work, do note that at least one tool (flakehub cli) writes to the inputs section to update individual inputs. that usage, for one, may in fact depend on such limited complexity.

truh: Tools like flakehub could instruct their users to write flake inputs in specific ways. This doesn't seem like a reason to limit how people who don't use flakehub write their flakes.

rfcs/0193-toml-flakes.md
iFreilicht: I do love that as well, but there is a big difference between "nix the turing-complete functional language" and "nix the language for flake inputs, flake config and flake metadata that only allows attrsets and literals".

Maybe there could be a separate file format for a subset of nix called "nix object notation" or "nix light", a language that has nix syntax but doesn't allow functions at all?

This is pretty similar in concept to zig's `.zon` format, which reuses zig's syntax for objects to create a json-like format.

It would just need a different file ending (though I'm not sure `.nixon` is a great idea :D). You could `import` files like that just like regular nix files, but for certain things (like the flake metadata in this RFC), only this subset of nix is allowed.

EDIT: Just noticed someone else brought that up already: https://github.com/NixOS/rfcs/pull/193/files#r2602798399

rfcs/0193-toml-flakes.md
iFreilicht: Not sure if that's a follow-up RFC or a prerequisite. A simple format like this could be quite useful for many situations. `flake.lock` could be transitioned to a format like this, the profile manifest as well. It seems like a pretty useful idea to formalize this "data only subset" of nix.

It would also alleviate the concerns raised here: https://github.com/NixOS/rfcs/pull/193/files#r2603388837

iFreilicht:
> What would be the expected usage if one were to not declare dependencies? I imagine not placing a TOML file could instead notify the user it could not find one. For reference, flakes' `inputs` attribute at present is optional.

@KiaraGrouwstra how about we use this opportunity to add a `version` field to flakes? One big criticisms of flakes is that because their format is unversioned, it's nearly impossible to make any backwards-compatible changes to it.

The initial version of `flake.toml` could be such that `version` is the only field in it. If `flake.toml` is missing, `version` is assumed to be `1`. This would be a very simple PR and something that is probably quite easy to reach a consensus on. It could be merged quickly and work to prepare for other versions of flakes could start immediately while the details are still being worked out.

rfcs/0193-toml-flakes.md
eclairevoyant: @de11n toml is _not_ that simple though, especially with the table vs array syntax - e.g. this is quite atrocious: https://toml.io/en/v1.0.0#array-of-tables

Let's just stick to Nix.

truh:
@iFreilicht flake authors that want to offer support for older versions of their flakes could do this with release branches.

rfcs/0193-toml-flakes.md
tomberek: Let me wander a bit........ what happens here is that the inputs section is evaluated (1), then those values are used in another evaluation (2) with the outputs. How (1) works is different than how (2) works. We may be able to loosen the restrictions a bit on (1), but it will still need to be different in some way. Different restrictions, different access, etc. It isn't hard to remove the restrictions in Nix's source code, but while I see how it isbammoying sometimes, I dont see what large differences there would be in end-user outcomes. Perhaps you wnd up with some easier tooling to manage the inputs, but a jq-but-for-nix (eg: nix-editor) can do this already. Some enterprising individuals would embed much more complicated logic into the inputs section, create some abstractions. I've wanted something like this where an input could be just a plain nix value allowing for easier overriding or CLI options to be declared.

Putting "inputs" into toml or another format makes this distinction more clear, but doesn't change its underlying nature.

iFreilicht:
@truh that just puts the burden for support on the authors, though. It would be very helpful if Nix itself was smart enough to understand what version of the flake schema a certain flake is written in, and whether it is able to work with that or not. Right now the format of flakes is impossible to change because there's no control over backwards- or forwards-compatibility.

Of course I understand that Nix supporting multiple versions of the flake schema at the same time is pretty much impossible because flakes are so deeply integrated into the evaluator. But even just getting a nice error message would be a big upgrade. Thankfully nix makes it quite trivial to temporarily install another version of nix.

truh:
@iFreilicht Oh you mean versions of the flake format. I thought you meant semantic versioning for flakes like flakehub. We would have to be be careful to give it a name that makes it explicit that the version number refers to the flake format so that it doesn't confuse anyone.

Honestly I find it hard to imagine benefits that breaking & non-interoperable changes to the flake format could bring that would outweigh the frustration of interoperability problems. Especially if stabilization remains such a difficult process. I have this concern about TOML flakes as well. It fragments the community, makes it more difficult for users to find information for their nix versions/approach and makes it more difficult for community members to offer support to people seeking help.

rfcs/0193-toml-flakes.md
KiaraGrouwstra: this seems a distinct proposal, indeed - perhaps it could be split out

iFreilicht:
> We would have to be be careful to give it a name that makes it explicit that the version number refers to the flake format so that it doesn't confuse anyone.

How about `schema_version`?

> Honestly I find it hard to imagine benefits that breaking & non-interoperable changes to the flake format could bring that would outweigh the frustration of interoperability problems.

We already have such breaking changes, there's just no way to control them. Because all the parts of flakes are so tightly interconnected, changing any part of it is nearly impossible. For example, `builtins.fetchTree` is [still not stabilized](https://github.com/NixOS/nix/milestone/31), but it lies at the core of how flakes work. All `flake.lock` does is record arguments to `fetchTree`, https://github.com/NixOS/nix/issues/5260, https://github.com/helix-editor/helix/issues/9866

But even if all of this was already stabilized, you don't know what the future brings. Maybe https://github.com/NixOS/rfcs/pull/194 is something we want to explore, or we decide to split `flake.nix` into multiple files? I don't know, maybe flakes will be a done feature with no breaking changes for the next 30 year, but being prepared for handling change well (such as supporting two versions of flakes for a transition period) is a much better option than simply pretending it will never happen.

Eveeifyeve:
Personally I think it doesn't make sense to have a TOML format, because pretty much the community has standardized on `.nix` format and some of the features that is proposed could be done in nix. This rfc adds a bit more complexity to the user by knowing nix and toml, but also keeping it as a separate file adds more complexity to people in big repos like more files to manage (eg.`flakes.toml`, `outputs.nix`, `compose.yml` and etc.)

Ideas of versioning could happen like:
```nix
version = 1; # Optional
description = "Desc"; # Optional

inputs = {
	nixpkgs.url = "github:NixOS/nixpkgs";
	# Or
	nixpkgs.url = builtins.fetchFromGitHub {
		...
	};
}
```

I see how nix could be better for this case with inputs because you can declare it both ways via using a fetcher or a direct url.

TLATER:
The point - that is in fact the very first item in the reasoning for this RFC, but I guess it bears repeating - is that `flake.nix` is only barely nix. Your example notably is *not* a valid flake:

```nix
{
  inputs = {
    nixpkgs.url = builtins.fetchFromGitHub {
      ...
    };
}
```

... making you in fact prove the very point of this RFC; Even with notes about this detail *right there* and (presumably) a fair bit of experience given you're commenting here, people still get this wrong.

See [this thread](https://github.com/NixOS/rfcs/pull/193/files#r2597533442) for the discussion about whether *that* should be changed instead.

Assuming this limitation stays in place, however, `flake.toml` would take away a confusing detail, since TOML doesn't lie about supporting dynamic content.

Maybe have another read through instead of just rejecting it at a high level, there are other good points, especially in combination with the sibling RFC.

KiaraGrouwstra:
@roberth i presume if `inputs`' restrictions were lifted, that could also facilitate your TOML idea (`inputs = builtins.fromTOML (builtins.readFile ./inputs.toml);`)?

Malix-Labs:
https://github.com/NixOS/rfcs/pull/193#issuecomment-3796693964 from @KiaraGrouwstra:
> if `inputs`' restrictions were lifted

If all those restrictions were lifted, this RFC would be a non starter

And those restrictions are stemming from preventing underlying problems such as those raised in https://github.com/NixOS/nix/issues/4945#issuecomment-868572873 :

> to prevent arbitrarily complex, possibly non-terminating computations while querying flake metadata

Also using IFD to compute flake inputs, which would break the whole philosophy of nix, and imo (and this is the whole debate around NixOS/nix#4945) sensical to prevent (at least by default)

However, the point raised in https://github.com/NixOS/nix/pull/15132#issuecomment-4207131122 by @roberth :
> Arguably it's more controlled than the alternative, which is that users have to cobble together input modifications using scripts that perform text replacements in flake.nix, or things along those lines.

Is a strong argument;
People really wanting to compute their flakes inputs are just doing it regardless, in an unsafer way.

And other non-flakes solutions (I'm thinking of https://github.com/andir/npins) typically can allows for such behavior.

In my opinion, the biggest pain point that this RFC is solving is the confusion about `flake.nix` not being nix but a subset of nix with a specific schema and thus misleading (what lead to the creation of NixOS/nix#4945) (also see https://github.com/NixOS/nix/issues/4945#issuecomment-4206680838)

If "flakes inputs really shouldn't allow for non-trivial expressions": they should not have been primarily nix to begin with, and this RFC should have been the standard default;
Else: NixOS/nix#15132 makes sense and this RFC shouldn't be accepted.

Malix-Labs:
Is there an experimental implementation for this RFC?

We would be interested for https://github.com/NixOS/nixos-search/blob/main/flakes/manual.toml (also see NixOS/nixos-search#1183)

d-hain:
I don't want to come in here and just hate for no reason, but personally I really don't see a need as to why one would switch to TOML-based Flakes. 
We already have an entire configuration language, even if the flake.nix file is only a subset of it - I am aware.
But if that subset-nature of Flakes would just be properly documented at every step, it wouldn't be a problem at all.

On that note, we also don't want to have both as that is just utterly unnecessary.

Malix-Labs:
> we also don't want to have both as that is just utterly unnecessary

I agree.

https://github.com/NixOS/nix/issues/4945#issuecomment-4774692162 from @Malix-Labs : 
> I am also convinced that if NixOS/nix#4945 was to be rejected, then NixOS/rfcs#193 (or another non-turing configuration language) AND deprecating `flake.nix` would be the only legitimate alternative to the concept of nix flakes;
> 
> Either flakes inputs are defined in a non-turing configuration language, either they are defined in Nix
> 
> The current state of flake being a loosely defined subset of nix is objectively wrong because of this indecision

PR #195, opened 2026-01-15
[RFC 0195] init: Binary Cache Index Protocol

kalbasit:
[Rendered](https://github.com/kalbasit/rfcs/blob/cache-index-rfc/rfcs/0195-binary-cache-index-protocol.md)

Previous discussions and references:
- https://github.com/kalbasit/ncps/issues/343
- https://github.com/NixOS/nix/issues/649
- https://github.com/NixOS/nix/issues/7221
- https://github.com/NixOS/infra/issues/776
- https://discourse.nixos.org/t/garbage-collecting-cache-nixos-org/74249

cc @Mic92 @edef1c @brianmcgee @roberth @zimbatm @Kernald

### DIFF
--- a/rfcs/0195-binary-cache-index-protocol.md
+++ b/rfcs/0195-binary-cache-index-protocol.md
@@ -135,11 +135,7 @@ IndexFormat: hlssi
 IndexCreatedAt: 2026-01-13T12:00:00Z
 IndexItemCount: 1200000000
 IndexShardingDepth: 2
-IndexShardingAlphabet: 0123456789abcdfghijklmnpqrsvwxyz
-IndexEncodingType: golomb-rice
 IndexEncodingParameter: 8
-IndexHashBits: 160
-IndexPrefixBits: 10
 IndexJournalBase: https://cache.example.com/nix-cache-index/journal/
 IndexShardsBase: https://cache.example.com/nix-cache-index/shards/
 IndexDeltasBase: https://cache.example.com/nix-cache-index/deltas/
@@ -161,11 +157,7 @@ Clients that understand these fields use the index. Clients that don't recognize
 - `IndexCreatedAt`: ISO 8601 timestamp of when this index configuration was generated
 - `IndexItemCount`: Total number of store path hashes indexed across all shards. This is approximate and provided for client information and debugging purposes; it may drift slightly between compactions.
 - `IndexShardingDepth`: Number of prefix characters used for partitioning (0–4)
-- `IndexShardingAlphabet`: The base32 alphabet used for shard prefixes. MUST match Nix's alphabet: `0123456789abcdfghijklmnpqrsvwxyz`
-- `IndexEncodingType`: Compression algorithm for shard files
 - `IndexEncodingParameter`: Golomb-Rice divisor exponent (M = 2^parameter)
-- `IndexHashBits`: Total bits in a full store path hash (160 for Nix)
-- `IndexPrefixBits`: Bits consumed by the shard prefix, used to compute suffix size. For depth=2, this is 10 bits (2 characters × 5 bits each).
 - `IndexJournalBase`: Base URL for journal segment files
 - `IndexShardsBase`: Base URL for shard files
 - `IndexDeltasBase`: Base URL for delta files and checksums
@@ -176,6 +168,10 @@ Clients that understand these fields use the index. Clients that don't recognize
 - `IndexDeltasEnabled`: Whether differential updates are available (see Section 10)
 - `IndexDeltasOldestBase`: Oldest epoch from which deltas can be applied. Clients with a local epoch older than this must perform a full download.
 - `IndexDeltasCompression`: Compression algorithm for delta files (`zstd`)
+
+**Format Invariants**: The `hlssi` format fixes several properties that are therefore not advertised as fields. The index is defined over Nix store-path digests, which are 160-bit hashes encoded in Nix's base32 alphabet (`0123456789abcdfghijklmnpqrsvwxyz`, 5 bits per character). The number of bits consumed by the shard prefix is therefore `IndexShardingDepth × 5`, and the per-shard suffix width is `160 − (IndexShardingDepth × 5)`; neither is advertised separately. Shard suffixes are Golomb-Rice coded (tuned by `IndexEncodingParameter`) and every shard file is zstd-compressed (see Section 5). Any change to these invariants would require a new `IndexFormat` (and, if not backward-compatible, a new `IndexVersion`).
+
+**Backward Compatibility**: These fields are purely additive. Nix's `nix-cache-info` parser reads line-oriented `key: value` pairs, recognizes only `StoreDir`, `WantMassQuery`, and `Priority`, and silently ignores every other key (`src/libstore/binary-cache-store.cc`). This behaviour is unchanged across all Nix 2.x releases (verified from 2.3 through the current development branch): unrecognized keys fall through the parser's `if`/`else if` chain with no error. Existing clients are therefore unaffected by the `Index*` fields: caches that advertise an index remain fully usable by clients that predate this RFC.
 
 **Caching**: Servers SHOULD use the `Cache-Control` HTTP header to specify the caching duration of `nix-cache-info`. Clients SHOULD respect this header to allow the server to control how long the index metadata is cached. Revalidation using `If-Modified-Since` or `ETag` SHOULD also be used.
 
@@ -197,8 +193,8 @@ The journal captures recent mutations.
 
 - Lines beginning with `+` indicate additions
 - Lines beginning with `-` indicate deletions (tombstones)
-- Hash is the 32-character store path hash (no `-name` suffix)
-- Encoding is ASCII (all characters are in the ASCII range: `+`, `-`, the 32 Nix base32 characters, and newline `\n`)
+- Hash is the 160 bit store path digest, nixbase32-encoded (32 characters, no `-name` suffix)
+- The whole file is ASCII (as `+`, `-`, the 32 Nix base32 characters, and newline `\n` are in the ASCII range)
 
 **Segment Lifecycle**:
 1. Writer appends to current segment file
@@ -209,13 +205,18 @@ The journal captures recent mutations.
 **Write Protocol** (for cache operators):
 ```
 On artifact push:
-  1. Upload .narinfo and .nar to storage
+  1. Upload the artifacts to storage: the .narinfo, the .nar (only if its
+     contents are not already present on the cache), and any additional
+     objects the configuration produces (e.g. build logs, debug symbols)
   2. Append "+<hash>\n" to current journal segment
 
 On garbage collection:
-  1. Delete .narinfo and .nar from storage
+  1. Delete the .narinfo from storage; delete the .nar only if it is no
+     longer referenced by any other .narinfo
   2. Append "-<hash>\n" to current journal segment
 ```
+
+**Concurrency**: Because the journal is an append-only log, all appends to a given segment MUST be serialized through a single logical writer per cache. This is a notable change from current practice: today multiple uncoordinated agents may upload `.narinfo`/`.nar` objects directly to raw object storage in parallel. Artifact uploads MAY remain concurrent, but recording those mutations in the journal requires a serialization point (a dedicated writer/coordinator, or a queue that batches appends). Caches that cannot provide a single writer can instead rebuild journal segments periodically by listing storage, at the cost of higher latency before new paths appear in the index.
 
 **Implementation Optimizations**: Servers MAY implement HTTP range requests to allow clients to efficiently catch up on journal segments. Servers with dynamic capabilities MAY implement long polling for near-real-time updates. These optimizations are not required by the protocol but can improve performance.
 
@@ -849,11 +850,7 @@ IndexFormat: hlssi
 IndexCreatedAt: 2026-01-13T12:00:00Z
 IndexItemCount: 487
 IndexShardingDepth: 0
-IndexShardingAlphabet: 0123456789abcdfghijklmnpqrsvwxyz
-IndexEncodingType: golomb-rice
 IndexEncodingParameter: 6
-IndexHashBits: 160
-IndexPrefixBits: 0
 IndexJournalBase: https://homelab.local/nix-cache-index/journal/
 IndexShardsBase: https://homelab.local/nix-cache-index/shards/
 IndexDeltasBase: https://homelab.local/nix-cache-index/deltas/
@@ -912,11 +909,7 @@ IndexFormat: hlssi
 IndexCreatedAt: 2026-01-13T12:00:00Z
 IndexItemCount: 98452103
 IndexShardingDepth: 2
-IndexShardingAlphabet: 0123456789abcdfghijklmnpqrsvwxyz
-IndexEncodingType: golomb-rice
 IndexEncodingParameter: 8
-IndexHashBits: 160
-IndexPrefixBits: 10
 IndexJournalBase: https://cache.example.org/index/journal/
 IndexShardsBase: https://cdn.example.org/index/shards/
 IndexDeltasBase: https://cdn.example.org/index/deltas/
--- a/rfcs/0195-binary-cache-index-protocol.md
+++ b/rfcs/0195-binary-cache-index-protocol.md
@@ -118,90 +118,74 @@ for char in base32_string:
     result = (result << 5) | nix_base32_value(char)
 ```
 
-## 3. Layer 0: Manifest
+**Note on Nix32 Byte Order**: Nix's base32 encoding processes bytes in reverse order compared to RFC4648. This means prefix-based sharding on Nix32 strings corresponds to suffix-based sharding on native hash bytes. This is intentional—it preserves the intuitive correspondence between a hash's visible prefix (`b6gv...`) and its shard location (`b6/`). Implementers should use established Nix base32 libraries (such as `go-nix/pkg/nixbase32`) rather than adapting standard base32 implementations.
 
-The manifest is a JSON file that describes the index topology. Clients discover the manifest URL via the `index-url` field in the cache's `nix-cache-info` file.
+## 3. Layer 0: Manifest (via nix-cache-info)
 
-**Discovery**: Caches advertise their index by adding an `index-url` field to `nix-cache-info`:
+The index manifest is embedded directly in the cache's `nix-cache-info` file using `Index`-prefixed fields. This eliminates the need for a separate manifest file and allows clients to discover index capabilities with no additional HTTP requests.
+
+**Discovery**: Caches advertise their index by adding `Index*` fields to `nix-cache-info`:
+
 ```
 StoreDir: /nix/store
 WantMassQuery: 1
 Priority: 40
-index-url: https://cache.example.com/nix-cache-index/manifest.json
+IndexVersion: 1
+IndexFormat: hlssi
+IndexCreatedAt: 2026-01-13T12:00:00Z
+IndexItemCount: 1200000000
+IndexShardingDepth: 2
+IndexShardingAlphabet: 0123456789abcdfghijklmnpqrsvwxyz
+IndexEncodingType: golomb-rice
+IndexEncodingParameter: 8
+IndexHashBits: 160
+IndexPrefixBits: 10
+IndexJournalBase: https://cache.example.com/nix-cache-index/journal/
+IndexShardsBase: https://cache.example.com/nix-cache-index/shards/
+IndexDeltasBase: https://cache.example.com/nix-cache-index/deltas/
+IndexJournalCurrentSegment: 1705147200
+IndexJournalRetentionCount: 12
+IndexEpochCurrent: 42
+IndexEpochPrevious: 41
+IndexDeltasEnabled: true
+IndexDeltasOldestBase: 35
+IndexDeltasCompression: zstd
 ```
 
-Clients that understand this field fetch the manifest from the specified URL. Clients that don't recognize the field continue with standard HTTP probing. Caches without an index simply omit this field.
-
-**Example Manifest**:
-```json
-{
-  "version": 1,
-  "format": "hlssi",
-  "created_at": "2026-01-13T12:00:00Z",
-  "item_count": 1200000000,
-  "sharding": {
-    "depth": 2,
-    "alphabet": "0123456789abcdfghijklmnpqrsvwxyz"
-  },
-  "encoding": {
-    "type": "golomb-rice",
-    "parameter": 8,
-    "hash_bits": 160,
-    "prefix_bits": 10
-  },
-  "urls": {
-    "journal_base": "https://cache.example.com/nix-cache-index/journal/",
-    "shards_base": "https://cache.example.com/nix-cache-index/shards/",
-    "deltas_base": "https://cache.example.com/nix-cache-index/deltas/"
-  },
-  "journal": {
-    "current_segment": 1705147200,
-    "retention_count": 12
-  },
-  "epoch": {
-    "current": 42,
-    "previous": 41
-  },
-  "deltas": {
-    "enabled": true,
-    "oldest_base": 35,
-    "compression": "zstd"
-  }
-}
-```
+Clients that understand these fields use the index. Clients that don't recognize the fields continue with standard HTTP probing. Caches without an index simply omit these fields.
 
 **Field Definitions**:
 
-- `version`: Protocol version (currently 1). Clients MUST reject manifests with unsupported versions.
-- `format`: Index format identifier (`hlssi` for this RFC)
-- `created_at`: ISO 8601 timestamp of when this manifest was generated
-- `item_count`: Total number of store path hashes indexed across all shards. This is approximate and provided for client information and debugging purposes; it may drift slightly between compactions.
-- `sharding.depth`: Number of prefix characters used for partitioning (0–4)
-- `sharding.alphabet`: The base32 alphabet used for shard prefixes. MUST match Nix's alphabet: `0123456789abcdfghijklmnpqrsvwxyz`
-- `encoding.type`: Compression algorithm for shard files
-- `encoding.parameter`: Golomb-Rice divisor exponent (M = 2^parameter)
-- `encoding.hash_bits`: Total bits in a full store path hash (160 for Nix)
-- `encoding.prefix_bits`: Bits consumed by the shard prefix, used to compute suffix size. For depth=2, this is 10 bits (2 characters × 5 bits each).
-- `urls.journal_base`: Base URL for journal segment files
-- `urls.shards_base`: Base URL for shard files
-- `urls.deltas_base`: Base URL for delta files and checksums
-- `journal.current_segment`: Unix timestamp of the active journal segment
-- `journal.retention_count`: Number of journal segments retained before archival into shards
-- `epoch.current`: Current shard generation number
-- `epoch.previous`: Previous shard generation (for grace period support; see Section 9)
-- `deltas.enabled`: Whether differential updates are available (see Section 10)
-- `deltas.oldest_base`: Oldest epoch from which deltas can be applied. Clients with a local epoch older than this must perform a full download.
-- `deltas.compression`: Compression algorithm for delta files (`zstd`)
+- `IndexVersion`: Protocol version (currently 1). Clients MUST reject indexes with unsupported versions.
+- `IndexFormat`: Index format identifier (`hlssi` for this RFC)
+- `IndexCreatedAt`: ISO 8601 timestamp of when this index configuration was generated
+- `IndexItemCount`: Total number of store path hashes indexed across all shards. This is approximate and provided for client information and debugging purposes; it may drift slightly between compactions.
+- `IndexShardingDepth`: Number of prefix characters used for partitioning (0–4)
+- `IndexShardingAlphabet`: The base32 alphabet used for shard prefixes. MUST match Nix's alphabet: `0123456789abcdfghijklmnpqrsvwxyz`
+- `IndexEncodingType`: Compression algorithm for shard files
+- `IndexEncodingParameter`: Golomb-Rice divisor exponent (M = 2^parameter)
+- `IndexHashBits`: Total bits in a full store path hash (160 for Nix)
+- `IndexPrefixBits`: Bits consumed by the shard prefix, used to compute suffix size. For depth=2, this is 10 bits (2 characters × 5 bits each).
+- `IndexJournalBase`: Base URL for journal segment files
+- `IndexShardsBase`: Base URL for shard files
+- `IndexDeltasBase`: Base URL for delta files and checksums
+- `IndexJournalCurrentSegment`: Opaque monotonically increasing identifier for the active journal segment. Implementations commonly use Unix timestamps but MAY use any scheme that guarantees strictly increasing values.
+- `IndexJournalRetentionCount`: Number of journal segments retained before archival into shards
+- `IndexEpochCurrent`: Current shard generation number
+- `IndexEpochPrevious`: Previous shard generation (for grace period support; see Section 9)
+- `IndexDeltasEnabled`: Whether differential updates are available (see Section 10)
+- `IndexDeltasOldestBase`: Oldest epoch from which deltas can be applied. Clients with a local epoch older than this must perform a full download.
+- `IndexDeltasCompression`: Compression algorithm for delta files (`zstd`)
 
-**Caching**: Servers SHOULD use the `Cache-Control` HTTP header to specify the caching duration of the manifest. Clients SHOULD respect this header to allow the server to control how long the manifest is cached. Revalidation using `If-Modified-Since` or `ETag` SHOULD also be used.
+**Caching**: Servers SHOULD use the `Cache-Control` HTTP header to specify the caching duration of `nix-cache-info`. Clients SHOULD respect this header to allow the server to control how long the index metadata is cached. Revalidation using `If-Modified-Since` or `ETag` SHOULD also be used.
 
-**Integrity Verification**: Clients SHOULD verify manifest integrity using HTTP-level mechanisms (`ETag`, `Content-MD5`). Cryptographic signing of index files is deferred to future work (see Future Work: Index Signing and Trust).
+**Integrity Verification**: Clients SHOULD verify `nix-cache-info` integrity using HTTP-level mechanisms (`ETag`, `Content-MD5`). Cryptographic signing of index files is deferred to future work (see Future Work: Index Signing and Trust).
 
 ## 4. Layer 1: Journal (Hot Layer)
 
 The journal captures recent mutations.
 
-**URL Pattern**: `{urls.journal_base}<timestamp>.log`
+**URL Pattern**: `{IndexJournalBase}<segment_id>.log`
 
 **Format**: Line-delimited ASCII text, one operation per line:
 ```
@@ -219,7 +203,7 @@ The journal captures recent mutations.
 **Segment Lifecycle**:
 1. Writer appends to current segment file
 2. The segment is rotated at the discretion of the cache operator.
-3. Segments older than `retention_count` are archived into shards during compaction
+3. Segments older than `IndexJournalRetentionCount` are archived into shards during compaction
 4. Archived segments are deleted after successful compaction
 
 **Write Protocol** (for cache operators):
@@ -239,18 +223,18 @@ On garbage collection:
 
 Shards contain the bulk of membership data, compressed using Golomb-Rice coding.
 
-**URL Pattern**: `{urls.shards_base}<epoch>/<prefix>.idx.zst`
+**URL Pattern**: `{IndexShardsBase}<epoch>/<prefix>.idx.zst`
 
-For `sharding.depth = 2`:
+For `IndexShardingDepth = 2`:
 ```
-{shards_base}42/b6.idx.zst    # All hashes starting with "b6"
-{shards_base}42/a1.idx.zst    # All hashes starting with "a1"
+{IndexShardsBase}42/b6.idx.zst    # All hashes starting with "b6"
+{IndexShardsBase}42/a1.idx.zst    # All hashes starting with "a1"
 ...
 ```
 
-For `sharding.depth = 0` (small caches):
+For `IndexShardingDepth = 0` (small caches):
 ```
-{shards_base}42/root.idx.zst  # All hashes in single file
+{IndexShardsBase}42/root.idx.zst  # All hashes in single file
 ```
 
 **Compression**: Shard files MUST be compressed with zstd (indicated by `.zst` extension).
@@ -416,12 +400,12 @@ The number of partitions at depth `d` with Nix's 32-character base32 alphabet is
 ```
 FUNCTION query(target_hash: string) -> {DEFINITE_HIT, DEFINITE_MISS, PROBABLE_HIT}:
 
-    // Step 1: Fetch and parse manifest (cached)
-    manifest = fetch_cached(manifest_url)
+    // Step 1: Fetch and parse nix-cache-info (cached)
+    cache_info = fetch_cached(cache_url + "/nix-cache-info")
 
     // Step 2: Check journal for recent mutations
-    FOR segment IN manifest.journal.segments:
-        journal_url = manifest.urls.journal_base + segment + ".log"
+    FOR segment IN get_journal_segments(cache_info):
+        journal_url = cache_info.IndexJournalBase + segment + ".log"
         journal = fetch_cached(journal_url)
         IF "-" + target_hash IN journal:
             RETURN DEFINITE_MISS  // Recently deleted
@@ -429,13 +413,13 @@ FUNCTION query(target_hash: string) -> {DEFINITE_HIT, DEFINITE_MISS, PROBABLE_HI
             RETURN PROBABLE_HIT   // Recently added (see note below)
 
     // Step 3: Determine shard for this hash
-    prefix = target_hash[0:manifest.sharding.depth]
+    prefix = target_hash[0:cache_info.IndexShardingDepth]
     shard_url = format("{}{}/{}.idx.zst",
-                        manifest.urls.shards_base, manifest.epoch.current, prefix)
+                        cache_info.IndexShardsBase, cache_info.IndexEpochCurrent, prefix)
 
     // Step 4: Fetch and search shard (cached by epoch)
     shard = fetch_cached(shard_url)
-    suffix = parse_hash_suffix(target_hash, manifest.sharding.depth)
+    suffix = parse_hash_suffix(target_hash, cache_info.IndexShardingDepth)
 
     // Step 5: Binary search sparse index
     bracket = binary_search(shard.sparse_index, suffix)
@@ -467,14 +451,14 @@ Clients receiving `PROBABLE_HIT` SHOULD proceed to fetch the `.narinfo` file, tr
 Compaction merges journal entries into shards, producing a new epoch:
 
 ```
-FUNCTION compact(manifest, old_shards) -> new_shards:
+FUNCTION compact(cache_info, old_shards) -> new_shards:
 
     // Step 0: Rotate to a new journal segment
     // This ensures no writes occur to segments we're about to process
-    new_segment_timestamp = rotate_journal_segment()
+    new_segment_id = rotate_journal_segment()
 
     // Step 1: Identify segments to compact (all BEFORE the new segment)
-    segments_to_compact = get_segments_older_than(new_segment_timestamp)
+    segments_to_compact = get_segments_older_than(new_segment_id)
 
     // Step 2: Parse all journal mutations from those segments
     additions = {}  // prefix -> set of hashes
@@ -483,16 +467,16 @@ FUNCTION compact(manifest, old_shards) -> new_shards:
     FOR segment IN segments_to_compact:
         FOR line IN segment:
             hash = line[1:]
-            prefix = hash[0:manifest.sharding.depth]
+            prefix = hash[0:cache_info.IndexShardingDepth]
             IF line[0] == '+':
                 additions[prefix].add(hash)
             ELSE:
                 deletions[prefix].add(hash)
 
     // Step 3: Process each shard independently
-    new_epoch = manifest.epoch.current + 1
+    new_epoch = cache_info.IndexEpochCurrent + 1
 
-    FOR prefix IN all_prefixes(manifest.sharding.depth):
+    FOR prefix IN all_prefixes(cache_info.IndexShardingDepth):
         old_hashes = decode_shard(old_shards[prefix])
 
         // Streaming merge: old + additions - deletions
@@ -501,16 +485,16 @@ FUNCTION compact(manifest, old_shards) -> new_shards:
         )
 
         // Encode new shard
-        new_shards[prefix] = encode_shard(new_hashes, manifest.encoding)
+        new_shards[prefix] = encode_shard(new_hashes, cache_info)
 
     // Step 4: Generate deltas (see Section 10)
-    IF manifest.deltas.enabled:
-        generate_deltas(old_shards, new_shards, manifest)
+    IF cache_info.IndexDeltasEnabled:
+        generate_deltas(old_shards, new_shards, cache_info)
 
-    // Step 5: Update manifest (atomic swap)
-    manifest.epoch.previous = manifest.epoch.current
-    manifest.epoch.current = new_epoch
-    manifest.journal.current_segment = new_segment_timestamp
+    // Step 5: Update nix-cache-info (atomic swap)
+    cache_info.IndexEpochPrevious = cache_info.IndexEpochCurrent
+    cache_info.IndexEpochCurrent = new_epoch
+    cache_info.IndexJournalCurrentSegment = new_segment_id
 
     // Step 6: Delete compacted journal segments
     // Safe because nothing is writing to these anymore
@@ -539,56 +523,56 @@ To prevent race conditions during compaction, servers MUST maintain both the cur
 Without a grace period, the following race can occur:
 
 ```
-T+0ms:    Client fetches manifest.json, sees epoch: 41
+T+0ms:    Client fetches nix-cache-info, sees IndexEpochCurrent: 41
 T+100ms:  Server completes compaction to epoch 42
-T+101ms:  Server updates manifest.json to epoch: 42
+T+101ms:  Server updates nix-cache-info to IndexEpochCurrent: 42
 T+102ms:  Server deletes shards/41/ directory
 T+500ms:  Client requests shards/41/b6.idx.zst
 T+501ms:  Client receives 404 Not Found
 ```
 
-The client read a valid manifest but could not fetch the shards it referenced.
+The client read a valid index configuration but could not fetch the shards it referenced.
 
 ### 9.2 Grace Period Requirements
 
 **Server Requirements**:
 
-1. The manifest MUST include both `epoch.current` and `epoch.previous` fields.
-2. Shards for `epoch.previous` MUST remain available for a **minimum grace period** of:
+1. The `nix-cache-info` MUST include both `IndexEpochCurrent` and `IndexEpochPrevious` fields.
+2. Shards for `IndexEpochPrevious` MUST remain available for a **minimum grace period** of:
    ```
-   grace_period >= 2 × max_manifest_cache_duration + max_request_duration
+   grace_period >= 2 × max_cache_info_cache_duration + max_request_duration
    ```
-   Where `max_manifest_cache_duration` is the maximum value used in the `Cache-Control` header for the manifest. For typical values (max cache duration = 120s, max request = 30s):
+   Where `max_cache_info_cache_duration` is the maximum value used in the `Cache-Control` header for `nix-cache-info`. For typical values (max cache duration = 120s, max request = 30s):
    ```
    grace_period >= 2 × 120 + 30 = 270 seconds (4.5 minutes)
    ```
 3. Servers SHOULD retain previous epoch shards for at least **10 minutes** to provide margin.
-4. Shards for epochs older than `epoch.previous` MAY be deleted immediately.
+4. Shards for epochs older than `IndexEpochPrevious` MAY be deleted immediately.
 
 **Client Requirements**:
 
-1. Clients MUST first attempt to fetch shards from `epoch.current`.
-2. If a shard fetch returns 404 AND `epoch.previous` exists in manifest:
-   - Client SHOULD retry using `epoch.previous`
-   - Client SHOULD refresh the manifest before subsequent queries
-3. If both epochs return 404, client SHOULD refresh manifest and retry once.
-4. Persistent 404s after manifest refresh indicate a server-side issue or deleted content.
+1. Clients MUST first attempt to fetch shards from `IndexEpochCurrent`.
+2. If a shard fetch returns 404 AND `IndexEpochPrevious` exists:
+   - Client SHOULD retry using `IndexEpochPrevious`
+   - Client SHOULD refresh `nix-cache-info` before subsequent queries
+3. If both epochs return 404, client SHOULD refresh `nix-cache-info` and retry once.
+4. Persistent 404s after refresh indicate a server-side issue or deleted content.
 
 ### 9.3 Compaction Lifecycle
 
 ```
 Phase 1: Pre-Compaction
   shards/41/  (current)
-  manifest.json: { epoch: { current: 41, previous: 40 } }
+  nix-cache-info: IndexEpochCurrent: 41, IndexEpochPrevious: 40
 
 Phase 2: Write New Epoch
   shards/41/  (still current)
   shards/42/  (being written, not yet referenced)
 
-Phase 3: Atomic Manifest Update
+Phase 3: Atomic nix-cache-info Update
   shards/41/  (now previous)
   shards/42/  (now current)
-  manifest.json: { epoch: { current: 42, previous: 41 } }
+  nix-cache-info: IndexEpochCurrent: 42, IndexEpochPrevious: 41
 
 Phase 4: Cleanup (after grace period)
   shards/40/  (deleted, no longer referenced)
@@ -598,7 +582,7 @@ Phase 4: Cleanup (after grace period)
 
 ### 9.4 Structural Parameter Changes
 
-Structural parameters (`sharding.depth`, `encoding.parameter`) are expected to remain stable for the lifetime of a cache. However, if an operator needs to change these parameters (e.g., increasing depth as the cache grows), special handling is required.
+Structural parameters (`IndexShardingDepth`, `IndexEncodingParameter`) are expected to remain stable for the lifetime of a cache. However, if an operator needs to change these parameters (e.g., increasing depth as the cache grows), special handling is required.
 
 **The Problem**:
 
@@ -608,14 +592,14 @@ If structural parameters change between epochs, the grace period mechanism break
 Epoch 41 (depth=0): shards/41/root.idx.zst
 Epoch 42 (depth=2): shards/42/00.idx.zst, shards/42/01.idx.zst, ...
 
-Client with stale manifest tries: shards/41/b6.idx.zst → 404!
+Client with stale nix-cache-info tries: shards/41/b6.idx.zst → 404!
 ```
 
 **Requirements for Structural Changes**:
 
-1. Structural changes MUST reset `deltas.oldest_base` to the new epoch (deltas cannot span structural boundaries).
+1. Structural changes MUST reset `IndexDeltasOldestBase` to the new epoch (deltas cannot span structural boundaries).
 2. The previous epoch MUST be retained with its original structure for the grace period.
-3. Clients encountering a structure mismatch (e.g., expected shard file returns 404) SHOULD refresh the manifest and download the current epoch in full.
+3. Clients encountering a structure mismatch (e.g., expected shard file returns 404) SHOULD refresh `nix-cache-info` and download the current epoch in full.
 4. Operators SHOULD treat structural changes as major events requiring extended grace periods (RECOMMENDED: 1 hour minimum).
 
 ## 10. Differential Updates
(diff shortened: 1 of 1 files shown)


rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+
+4. **Poor Offline/Intermittent Connectivity Handling**: Clients cannot make any progress without network access to check each path individually.
+
+## Use Cases Supported
+
+1. **CI/CD Pipeline Acceleration**: A typical NixOS system closure involves 1,000–5,000 store paths. Determining cache hits upfront allows optimal parallelization of downloads vs. builds.
+
+2. **Multi-Cache Federation**: Organizations often chain caches (private → community → cache.nixos.org). An index allows intelligent cache selection without sequential probing.
+
+3. **Offline-First Workflows**: Developers can sync an index and determine buildability without continuous network access.
```
roberth:
```suggestion
3. **Offline-First Workflows**: Developers can sync an index and determine substitutability without continuous network access.
```

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+All files are static and served via standard HTTP from any object storage or web server. No server-side computation is required.
+
+## 2. Store Path Hash Specification
+
+Nix store paths follow the format:
+```
+/nix/store/<hash>-<name>
+```
+
+Where `<hash>` is a 32-character string using Nix's custom base32 alphabet (`0123456789abcdfghijklmnpqrsvwxyz` — notably excluding `e`, `o`, `u`, `t`). This encodes 160 bits of a truncated SHA-256 digest.
```
roberth: Can keep this brief.
```suggestion
[Nix store paths](https://nix.dev/manual/nix/latest/store/store-path.html) generally follow the format `/nix/store/<hash>-<name>`, where the `<hash>` part is a sufficient identifier for the whole store object.
```

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+result = 0
+for char in base32_string:
+    result = (result << 5) | nix_base32_value(char)
+```
+
+## 3. Layer 0: Manifest
+
+The manifest is a JSON file at a well-known path that describes the index topology:
+
+**Path**: `/nix-cache-index/manifest.json`
```
roberth: We could fetch this info with no extra requests by making it part of `nix-cache-info`, and consider separately to move to a nicer JSON-only format.

Alternatively, we could consider this **Path** to be an example, and have a field in `nix-cache-info` to indicate its presence. This also solves the forward compatibility problem of supporting a new cache index format in the future, and/or multiple simultaneously for other purposes (if that even makes sense beyond migrations).

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+
+**Schema**:
```
roberth:
```suggestion
**Example**:
```

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+- `deltas.oldest_base`: Oldest epoch from which deltas can be applied. Clients with a local epoch older than this must perform a full download.
+- `deltas.compression`: Compression algorithm for delta files (`none`, `gzip`, `zstd`)
+
+**Caching**: Clients SHOULD cache the manifest with a short TTL (30–120 seconds) and revalidate using `If-Modified-Since` or `ETag`.
+
+**Integrity Verification**: Clients SHOULD verify manifest integrity using HTTP-level mechanisms (`ETag`, `Content-MD5`). Cryptographic signing of index files is deferred to future work (see Future Work: Index Signing and Trust).
+
+## 4. Layer 1: Journal (Hot Layer)
+
+The journal captures recent mutations with minimal latency.
```
roberth: Not sure about minimal.

This seems to be determined by segment_duration_seconds, and it requires many individual requests to catch up with the log.

Perhaps with HTTP range requests the number of requests could be reduced, turning this into a small number of bulk downloads.

Long polling could be an implementation strategy to make this even more realtime, without the added complexity of a push protocol.
When doing range requests instead of relying on split files, you'd still want a time interval parameter, but instead of `journal.segment_duration_seconds` it would be `journal.segment_query_interval`. Set to `0` for long polling.
If it's a dumb bucket, set a high value to reduce unnecessary / inefficient traffic.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+Offset  Size  Field
+------  ----  -----
+0       8     Magic number: "NIXIDX01" (ASCII)
+8       8     Item count (uint64, little-endian)
+16      1     Golomb parameter k (uint8, where M = 2^k)
+17      1     Hash suffix bits (uint8, typically 160 - prefix_bits)
+18      8     Sparse index offset from start of file (uint64, little-endian)
+26      8     Sparse index entry count (uint64, little-endian)
+34      8     XXH64 checksum of encoded data section (uint64, little-endian)
+42      22    Reserved for future use (must be zeros)
```
roberth: Should a client check that it's zeroes or will the magic number be bumped when old clients need to fail?
I assume you want clients to be lenient so that information can be added in a minor update of the protocols without breaking compatibility.
Breaking changes could be managed through the manifest, or avoided by having multiple index manifests in the `nix-cache-info` file.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+
+## 1. Additional Infrastructure Complexity
+
+Cache operators must run a compaction process (cron job or similar) to maintain the index. This adds operational burden compared to the current zero-maintenance model.
+
+**Mitigation**: Provide reference implementations as NixOS modules and container images.
+
+## 2. Index Staleness Window
+
+There is an inherent delay between when an artifact is pushed and when the index reflects it. With 5-minute journal segments, the worst-case staleness is ~5 minutes for clients that don't fetch the current journal.
```
roberth, line 1160: - Range requests might improve this. (If the other cache processes support that; I haven't validated)
- 5 minutes is well within the expected latencies of `cache.nixos.org`, but not good enough for a CI/CD scenario, though this could be mitigated by configuring *deployment targets* to ignore the index and assume everything was pushed before deployment was initiated. Deployment targets don't build, and if they do, typically only a dozen.
  - Example: those would tend to be "unpredictable" NixOS config file derivations. E.g. you build an "archetype" config in CI but have many configs IRL because of unique hostnames

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+**Mitigation**: Clients SHOULD always fetch the current journal segment with a short cache TTL.
+
+## 3. Privacy Concerns
+
+Unlike Bloom filters, the HLSSI format stores actual hashes, allowing enumeration of cache contents. This may be undesirable for private caches.
+
+**Mitigation**:
+- For truly private caches, consider HMAC-transforming hashes with a shared secret
+- Or accept this as a reasonable trade-off given that `.narinfo` URLs are already guessable
+
```
roberth, line 1173:
```suggestion
- The latency of a non-indexed binary cache may be acceptable.
- Critical private caches should require authentication. Authorization may be simple, or rely on HMAC instead.

```

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+## 4. Client Implementation Effort
+
+Existing Nix clients must be modified to support this protocol. Until adoption is widespread, cache operators must maintain backward compatibility.
+
+**Mitigation**: Design the protocol to be purely additive—index files don't interfere with existing cache access patterns.
```
roberth: The `.narinfo` files are still around, right?

This seems to be done, so it's a requirement, not a drawback.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+**Recommendation**: Allow segments to grow unbounded but trigger early rotation at a configurable threshold (e.g., 10,000 entries).
+
+## 3. Backward Compatibility Path
+
+How should cache operators transition from no-index to indexed caches? Options:
+- Manual opt-in via configuration
+- Automatic index generation on first compaction
+- Nix client version negotiation
+
+**Recommendation**: Opt-in via presence of `manifest.json`. Clients that don't find it fall back to current behavior.
```
roberth: Again, the index is just an addition, right?
Old clients can just keep doing their thing.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+**Recommendation**: Treat index as advisory. On any parse error or inconsistency, fall back to standard HTTP probing. Cryptographic signing is deferred to future work.
+
+## 5. Multi-Output Derivations
+
+Store paths include `-<name>` suffixes (e.g., `-dev`, `-lib`). Should the index:
+- Store only hashes (current design)
+- Store hash-name pairs
+- Have separate indices per output type
+
+**Recommendation**: Store only hashes. The `.narinfo` fetched on hit contains the full `StorePath` field. The index is merely a pointer to the `.narinfo`, not a replacement for it.
```
roberth: Multi-output is fine, as mentioned in my first suggestion, the hash part is a sufficient identifier.

```suggestion
## 5. Build traces (CA realisations)

(Note: CA realisations are getting renamed to build traces)
In the current implementation of content addressing derivations in the binary cache, realisations (= build traces) are stored in a separate directory.
Either a separate index could be created for build traces, or they could be mapped into the same index.
A client knows whether it it wants a realisation or narinfo, and since realisation and narinfo hashes do not collide (probabilistic truth of course), we do not expect any problems there.
Remaining question: does a separate realisation index improve or degrade performance?

```

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+
+# Future work
+[future]: #future-work
+
+## 1. Index Signing and Trust
+
+Extend the existing binary cache signing mechanism to cover index files, allowing clients to verify index authenticity before trusting membership results. This would include:
+- Defining a signature file format (e.g., `.sig` files)
+- Specifying which files are signed (manifest, shards, or both)
+- Key distribution and trust model
```
roberth, line 1467: Somewhat minor because to my understanding the index (store path availability) does not affect the integrity bearing parts of the realisation system. It can just make it fail or slow down.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+
+Specify standard paths for persistent client-side index caching:
+```
+~/.cache/nix/indices/<cache-hash>/manifest.json
+~/.cache/nix/indices/<cache-hash>/shards/...
+```
+
+## 4. Integration with Nix Flake Registries
+
+Explore automatic index URL discovery via flake registries, reducing configuration burden for common caches.
```
roberth: Solved. We have discoverable binary caches through `nixConfig`, and those refer to their indices through `nix-cache-info` or the conventional path if we keep that.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+
+## 5. Compression Algorithm Alternatives
+
+Evaluate alternative compression schemes as they mature:
+- ANS (Asymmetric Numeral Systems) for better compression ratios
+- SIMD-accelerated decoders for faster queries
```
roberth: I don't understand.
This is an I/O bound system. We're not doing any new bulk hashing, are we?

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+- GPU-based batch queries for large dependency closures
```
roberth: Skeptical, but tell me more.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+co-authors: (to be determined)
+shepherd-team: (to be nominated and accepted by RFC steering committee)
+shepherd-leader: (to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+This RFC proposes a standardized client-side indexing protocol for Nix Binary Caches that enables efficient store path discovery without requiring per-path HTTP requests. The protocol introduces a **Hybrid Log-Structured Sharded Index (HLSSI)** that uses deterministic compressed hash storage, enabling clients to definitively determine cache membership before issuing network requests. The design supports caches ranging from ~100 items (homelabs) to 1+ billion items (cache.nixos.org) while maintaining sub-second freshness for CI/CD pipelines and supporting garbage collection without requiring expensive storage listing operations.
```
Mic92, line 14: Overall I think for this RFC, we should be leading with an implementation & benchmarks to make sure the proposal works.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+
+## 6. Index-Aware Garbage Collection
+
+Develop GC strategies that use index metadata to make smarter retention decisions:
+- Keep items referenced by recent index epochs
+- Prioritize deletion of items not queried recently (requires query logging)
+
+## 7. P2P Index Distribution
+
+Explore peer-to-peer distribution of index files and deltas, reducing load on central cache servers. This could leverage protocols like BitTorrent or IPFS for large index distribution.
```
roberth, line 1493: For future compatibility, rely more on URLs in the spec instead of relative paths (hardcoded relative paths or otherwise).

roberth (commented):
A bit rough around the edges (I know it's draft!), but seems like a good starting point. Do you plan to validate this?

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+3. **Offline-First Workflows**: Developers can sync an index and determine buildability without continuous network access.
+
+4. **Cache Analytics**: Operators can analyze index files to understand cache composition, hit rates, and optimize retention policies. Since HLSSI stores actual hashes (not lossy Bloom filter bits), operators can: count exact items, analyze hash distribution across shards, compare indices between caches to measure overlap, and identify "miss patterns" to improve cache coverage.
+
+## Expected Outcome
+
+A client implementing this protocol will:
+- Reduce cache lookup latency by 90%+ for cache misses
+- Eliminate unnecessary HTTP requests entirely for definitive misses
+- Support caches from 100 to 1,000,000,000+ items with proportional bandwidth costs
```
Mic92: Nit: I think 100 or 500 items as mentioned here multiple times for homelab use is a bit low considering that a single NixOS closure will be in the magnitude of 1000 store paths.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+- `encoding.hash_bits`: Total bits in a full store path hash (160 for Nix)
+- `encoding.prefix_bits`: Bits consumed by the shard prefix, used to compute suffix size. For depth=2, this is 10 bits (2 characters × 5 bits each).
+- `journal.current_segment`: Unix timestamp of the active journal segment
+- `journal.segment_duration_seconds`: How often segments rotate (e.g., 300 = 5 minutes)
+- `journal.retention_count`: Number of journal segments retained before archival into shards
+- `epoch.current`: Current shard generation number
+- `epoch.previous`: Previous shard generation (for grace period support; see Section 9)
+- `deltas.enabled`: Whether differential updates are available (see Section 10)
+- `deltas.oldest_base`: Oldest epoch from which deltas can be applied. Clients with a local epoch older than this must perform a full download.
+- `deltas.compression`: Compression algorithm for delta files (`none`, `gzip`, `zstd`)
```
Mic92: Do we even want to support `gzip` to begin with? `zstd` seems more efficient on all accounts.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+17      1     Hash suffix bits (uint8, typically 160 - prefix_bits)
+18      8     Sparse index offset from start of file (uint64, little-endian)
+26      8     Sparse index entry count (uint64, little-endian)
+34      8     XXH64 checksum of encoded data section (uint64, little-endian)
+42      22    Reserved for future use (must be zeros)
+------  ----
+Total:  64 bytes
+```
+
+**Implementation Note**: The header is designed to avoid struct padding issues. All multi-byte integers are little-endian. Implementations in C/Rust should use explicit byte-level serialization or `#pragma pack(1)` / `#[repr(packed)]` to ensure correct layout.
```
Mic92: Did you not specify big-endian above?

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+```
+Client: GET /b6gvzjyb2pg0kjfwn6a6llj3k1bq6dwi.narinfo
+Server: 200 OK (cache hit) or 404 Not Found (cache miss)
+```
+
+When a path is missing, the client moves to the next cache in its substituter list. This approach has several critical deficiencies:
+
+1. **Latency Accumulation**: Each cache miss incurs a full HTTP round-trip. For builds with thousands of dependencies querying multiple caches, this compounds to significant delays.
+
+2. **Bandwidth Inefficiency**: HTTP headers, TLS handshakes, and connection overhead dominate when the actual answer is a single bit of information (present/absent).
```
kevincox: A "reasonable" implementation would use HTTP 2 or 3 where headers and handshakes don't need to be sent except for the first request. So the TLS handshake is basically irrelevant and the only header that really needs to be sent in the response is `Date`. So I find this point very unconvincing.

rfcs/0195-binary-cache-index-protocol.md
roberth: Depends whether it's a full cache or an overlay cache that depends on `cache.nixos.org` contents for the bulk of it.
Still low though. You'll at least have a bunch of generations in there, each with a bunch of unique config files.
Thing is, it doesn't really matter how low the number gets. An empty cache should also work, right? Without consuming hundreds of megabytes of effectively unused space?
Easy fix: adjust the number. Maybe rephrase for the idea of efficient low end?

```suggestion
- Support caches from 10,000 to 1,000,000,000+ items with proportional bandwidth costs
```

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+4. Operators SHOULD treat structural changes as major events requiring extended grace periods (RECOMMENDED: 1 hour minimum).
+
+## 10. Differential Updates
+
+For large caches, downloading full shard files on every epoch change is bandwidth-intensive. Differential updates allow clients to download only the changes between epochs.
+
+### 10.1 The Bandwidth Problem
+
+For cache.nixos.org with ~1.06 billion items (as of January 2026):
+- Full index size: ~1.5 GB
```
Mic92, line 615: 1.5G sounds okay for downstream caches, but too large for all clients especially clients that only have a short live time i.e. github actions. But I think reduce the size significantly by limiting how many years we include in this index for cache.nixos.org i.e. 2-3 years instead of everything since 2015.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+- `journal.current_segment`: Unix timestamp of the active journal segment
+- `journal.segment_duration_seconds`: How often segments rotate (e.g., 300 = 5 minutes)
+- `journal.retention_count`: Number of journal segments retained before archival into shards
+- `epoch.current`: Current shard generation number
+- `epoch.previous`: Previous shard generation (for grace period support; see Section 9)
+- `deltas.enabled`: Whether differential updates are available (see Section 10)
+- `deltas.oldest_base`: Oldest epoch from which deltas can be applied. Clients with a local epoch older than this must perform a full download.
+- `deltas.compression`: Compression algorithm for delta files (`none`, `gzip`, `zstd`)
+
+**Caching**: Clients SHOULD cache the manifest with a short TTL (30–120 seconds) and revalidate using `If-Modified-Since` or `ETag`.
```
kevincox: I think we should recommend that clients respect standard HTTP caching headers. This lets the cache operator decide. We can provide 30-120s as a fallback if there is no `Cache-Control` header but IMHO it seems better to just very frequently revalidate in the case as it is easy for the cache operator to add that header once cost becomes more important then freshness.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+### 10.2 Delta File Format
+
+**Path Pattern**: `/nix-cache-index/deltas/<from_epoch>-<to_epoch>/<prefix>.delta`
+
+Delta files use a simple line-oriented format listing the operations needed to transform the source epoch shard into the target epoch shard:
+
+```
+# Example: deltas/41-42/b6.delta
+# Operations are sorted by hash for efficient streaming application
+-b6a1c2d3e4f5g6h7i8j9k0l1m2n3o4p5
```
Mic92, line 630: What is the encoding of this format? Ascii?

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+```
+
+- Lines beginning with `-` indicate hashes present in epoch N but absent in epoch N+1
+- Lines beginning with `+` indicate hashes absent in epoch N but present in epoch N+1
+- Operations MUST be sorted by hash (lexicographically) for efficient streaming merge
+- Empty delta files (no changes to that shard) MAY be omitted entirely
+
+**Compression**: Delta files SHOULD be served with compression. The manifest field `deltas.compression` indicates the algorithm:
+- `none`: No compression
+- `gzip`: Gzip compression (`.delta.gz`)
```
Mic92: Same question as above, do we need gzip?

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
++b6z1a2b3c4d5e6f7g8h9i0j1k2l3m4n5
+```
+
+- Lines beginning with `-` indicate hashes present in epoch N but absent in epoch N+1
+- Lines beginning with `+` indicate hashes absent in epoch N but present in epoch N+1
+- Operations MUST be sorted by hash (lexicographically) for efficient streaming merge
+- Empty delta files (no changes to that shard) MAY be omitted entirely
+
+**Compression**: Delta files SHOULD be served with compression. The manifest field `deltas.compression` indicates the algorithm:
+- `none`: No compression
```
Mic92: Maybe we can also not have an uncompressed version of this for simplicity.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+- `created_at`: ISO 8601 timestamp of when this manifest was generated
+- `item_count`: Total number of store path hashes indexed across all shards. This is approximate and provided for client information and debugging purposes; it may drift slightly between compactions.
+- `sharding.depth`: Number of prefix characters used for partitioning (0–4)
+- `sharding.alphabet`: The base32 alphabet used for shard prefixes. MUST match Nix's alphabet: `0123456789abcdfghijklmnpqrsvwxyz`
+- `encoding.type`: Compression algorithm for shard files
+- `encoding.parameter`: Golomb-Rice divisor exponent (M = 2^parameter)
+- `encoding.hash_bits`: Total bits in a full store path hash (160 for Nix)
+- `encoding.prefix_bits`: Bits consumed by the shard prefix, used to compute suffix size. For depth=2, this is 10 bits (2 characters × 5 bits each).
+- `journal.current_segment`: Unix timestamp of the active journal segment
+- `journal.segment_duration_seconds`: How often segments rotate (e.g., 300 = 5 minutes)
```
kevincox: Using time in a distributed system always makes me uneasy. I think the protocol mostly doesn't actually depend on the time (as the client uses `journal.current_segment`) but I wonder if it is better to just make these abstract integers to make it clear that time isn't relevant.

I think the main complication is to know when to avoid replaying the journal and just start fresh, but IMHO just a 404 on the next journal segment that you would need to read to catch up will effectively do that.

rfcs/0195-binary-cache-index-protocol.md
roberth: Unless there's an overriding reason, pick little endian. Big endian is just not how computers work anymore.

{big-endian above} had an overriding reason, which is the correspondence between lexicographic sorting and the ordering of number sorting. But that's part of the *domain*, whereas this here is just a trivial implementation-level detail.
For comparison, we wouldn't pick big/little endian because e.g. accounting applications have Arabic numerals which are big endian.

rfcs/0195-binary-cache-index-protocol.md
roberth: It does make sense. @Mic92 gives another circumstance that could motivate having multiple. 

> 1.5G sounds okay for downstream caches, but too large for all clients especially clients that only have a short live time i.e. github actions. But I think reduce the size significantly by limiting how many years we include in this index for cache.nixos.org i.e. 2-3 years instead of everything since 2015.

&mdash; https://github.com/NixOS/rfcs/pull/195#discussion_r2694318995

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+    "compression": "zstd"
+  }
+}
+```
+
+**Shard Statistics**:
+- Total partitions: 32² = 1,024
+- Average items per shard: ~96,000
+- Average shard size: ~180 KB
+- Full index size: ~180 MB
```
Mic92, line 931: I think this where we should do some experiment:

- how much time do we spend on fetching the index
- how much time do we safe on fetching narinfos

There are different types of clients:

1. CI workload, might not be able to prefetch and potentially could be slow down by first having to download the index before being able to download any store path.
2. Workstation, can potentially prefetch the index and amortize the cost

The result might be not black and white. I could see also potentially different settings for different types of clients as the index can co-exists with the current narinfo lookup.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1532 @@
...
+1. Fetch private cache index (~10 KB total)
+   → Check all paths against single shard
+   → Build set of DEFINITE_MISS paths
+
+2. For remaining paths, fetch community cache index
+   → Only fetch relevant shards
+   → Build set of paths to fetch from community
+
+3. For still-remaining paths, fetch official cache index
+   → Only fetch relevant shards
```
Mic92, line 1024: One thing to note is that we potentially can save for cache.nixos.org on s3 costs here. Instead of having a lot of narinfo GET requests that hit our s3 backend, we would have fewer index requests that can be served from a CDN.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Agree, I already started a reference implementation in ncps.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: I agree, but I'm not sure that all current cache implementations are leveraging connection pools to benefit from this. Happy to remove this point if the consensus is that it's irrelevant.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Happy to do either here. But yea the idea is to aim to support very large indexes as supporting small ones (under 10,000) is trivial. Maybe my choice of a 100 here does not convey reality as @Mic92 has mentioned. I will accept your suggestion for now @roberth but happy to change it as needed.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: The absence of `e`, `o`, `u` and `t` is not mentioned in the documentation you linked to. I think it's important for the implementers of this RFC to be aware of this to avoid trying to compute hashes in the Golomb-rice that includes these chars.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: The problem with the `nix-cache-info` is that it requires a custom parser in all languages. JSON is well-defined format and most/all languages have standard libraries to parse it. One could argue that we already do this with the current nix-cache-info, case in point [ncps](https://github.com/kalbasit/ncps/blob/main/pkg/nixcacheinfo/parser.go), but I just feel that it makes the whole thing fragile.

Additionally, this RFC assumes that the manifest ETAG is the way clients could tell if the cache they have is up-to-date or not, something that existing caches might not implement for the existing nix-cache-info and that may be an issue for them depending on how they server it (static vs generated).

Finally, I think standardizing on well-known file path is reasonable here, I'm not sure what would the benefit be from having `nix-cache-info` convey where it is, but I am happy to adopt this if you think it opens the door for a future change. I sort of designed for that change with the `version` in the manifest to allow for future changes and updates.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: I agree that this particular field `journal.segment_duration_seconds` is not helpful. I was debating with myself wether to include this or not; The main reason I included it because I thought it would help other caches decide on how long to cache the manifest before making a call to check if the etag has changed. Happy to remove it.

With that said, I still think an epoch is better here than a number, it's not really for the "time" of it, it just makes things easier on the server-side. Instead of read current value, increment, write back it would be just stick in the current epoch. However, if you believe that the time may end up confusing implementers, I'm not opposed to reverting to an integer here since the added complexity of what I described is quite minimal compared to the rest.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Agree, I will remove gzip.

kalbasit:
```suggestion
- `deltas.compression`: Compression algorithm for delta files (`none`, `zstd`)
```

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good point. I will update the RFC to rely on `Cache-Control` instead of arbitrary number of seconds of their choosing. Thank you @kevincox this is great!

rfcs/0195-binary-cache-index-protocol.md
kalbasit:
```suggestion
```

kalbasit:
Thank you @kevincox, @Mic92, @roberth for your quick review and suggestions. I have resolved some of your comments and I will attend to the rest in a few hours after work. Thank you again 🙏🏼

rfcs/0195-binary-cache-index-protocol.md
kalbasit: I removed `journal.segment_duration_seconds`.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: This is done.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good suggestion to keep it brief. I've shortened the general description but kept the alphabet specification since it's not in the linked documentation and is critical for correct implementation—using the standard base32 alphabet would produce incorrect index lookups.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: You're right regarding not hardcoding the path now that I had a moment to think this through...

This raises some design questions I'd like input on. For the `nix-cache-info` field, we have a few options:

1. **Single `index-url`**: Simple, one index per cache. The CI use case (smaller/recent-only index) would be handled by using a different substituter pointing to a separate cache.

2. **Multiple `index-urls`**: Supports multiple indices per cache, but raises semantic questions:
   - Are they mirrors/alternatives (client picks one)?
   - Are they different scopes to be merged (client downloads all)?
   - How does a client know which index to use for its use case?

3. **Single `index-url` now, defer multi-index to future work**: Get real-world experience first, then design multi-index semantics based on actual needs.

The temporal coverage approach (index advertises 'only covers items from last N years') has issues since store path hashes don't encode creation time, and old paths can be re-added.

What are your thoughts on the right level of complexity for v1 of this protocol?

rfcs/0195-binary-cache-index-protocol.md
kalbasit: You're right that 'minimal' oversells it. I'll soften the language. Regarding optimizations: range requests for catch-up and long polling for real-time are good implementation strategies, but I'm inclined to keep them out of the spec itself since they're optimizations that servers/clients can adopt independently without protocol changes. The protocol just needs to not preclude them. Would a note in the implementation considerations section acknowledging these optimization opportunities be sufficient?

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good question. The intent is lenient: clients SHOULD ignore the reserved bytes to allow minor, backward-compatible additions without breaking old clients. Breaking changes would bump the magic number (e.g., NIXIDX02) or the manifest version field. I'll clarify this in the spec - something like: 'Clients MUST ignore non-zero values in reserved bytes to allow backward-compatible extensions. Incompatible format changes will use a new magic number.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: To clarify what might be confusing: the RFC uses both, intentionally:

- **Section 2.1 (hash interpretation)**: Big-endian so that lexicographic string ordering equals numeric ordering - required for prefix-based sharding to work correctly.
- **Section 5.1 (header integers)**: Little-endian for uint64 fields like item count and offsets - just binary serialization matching modern CPUs.

These aren't contradictory; they serve different purposes. I'll add a note to Section 5.1 clarifying why header integers use little-endian while hash interpretation uses big-endian.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Circling back on this. You're right, this is the weakest point in the motivation. I'll remove or significantly downplay the HTTP overhead argument. The stronger motivations are: (1) eliminating requests entirely for known misses, (2) reducing backend costs (S3 per-request pricing), and (3) enabling offline/intelligent multi-cache workflows. These hold regardless of HTTP version.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good point about ephemeral CI. A few thoughts:

1. The sharded design already helps here - a CI runner only needs to fetch shards for prefixes in its closure, not all 1024 shards. For a typical closure of ~2000 paths, this might be 200-400 shards (~300-600MB) rather than the full 1.5GB.
2. CI runners could also skip the index entirely and use HTTP probing - the index is additive, not required.
3. The temporal filtering idea is tricky since store path hashes don't encode creation time.

Should I add guidance in the RFC about partial shard fetching strategies for resource-constrained clients?

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good catch - I should specify this explicitly. The journal format is ASCII (equivalently UTF-8, since all characters are in the ASCII range): only `+`, `-`, the 32 Nix base32 characters, and newline (`\n`) appear. I'll add this to the spec.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Agreed. I'll mandate compression (zstd) for shard files and remove the uncompressed option. Simplicity is worth more than the marginal benefit of avoiding compression overhead on tiny caches.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good point about the CI→deploy latency gap. The 5-minute window is fine for general use but problematic when CI pushes and immediately deploys. I'll add a note about this pattern: deployment targets that receive paths from a known CI pipeline can be configured to skip index lookups and assume availability, since they're fetching paths that were just pushed. This is an operational choice rather than a protocol change.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: You're right - this is poorly framed. The `.narinfo` files remain, so backward compatibility is inherent in the design, not a mitigation for a drawback. I'll reframe this section to clarify that the protocol is purely additive: old clients continue working unchanged, new clients can optionally use the index. The only 'drawback' is that new clients need implementation work to benefit from the index.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Correct - old clients continue unchanged. I'll simplify this section. The only 'transition' question is discovery, which we've addressed by advertising the index URL in `nix-cache-info`. Clients that understand the field use the index; clients that don't (or caches that don't advertise one) continue with HTTP probing. No version negotiation needed.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Exactly right. A corrupted or malicious index can only degrade performance (unnecessary HTTP requests or missed optimization opportunities), not compromise store path integrity - that's still protected by `.narinfo` signatures and NAR content hashes. I'll add this clarification to frame index integrity as a performance concern rather than a security concern, which supports the 'treat as advisory' recommendation.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good point - this is already covered by the existing mechanism. Flakes specify substituters via `nixConfig`, substituters expose `nix-cache-info`, and `nix-cache-info` advertises the index URL. I'll remove this from future work.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Fair point - decoding is unlikely to be the bottleneck since this is I/O bound. SIMD-accelerated decoders are probably unnecessary optimization. I'll remove this item in future work.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: On reflection, I can't justify this one. The index is I/O bound as you noted, and even for in-memory batch queries the CPU→GPU transfer overhead would likely negate any benefit. I'll remove it.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Agreed. I'll update the spec to use URLs rather than assuming paths relative to the cache root. The manifest can specify base URLs for journal and shard locations, giving operators flexibility to host index files on CDNs or different infrastructure. This also future-proofs against path structure changes.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Agreed - this needs empirical validation. The reference implementation in ncps will measure both scenarios. I expect the trade-off to favor workstations more than CI, but the sharded design helps CI too (only fetch needed shards). I'll add benchmark results to the RFC once available. For now, the protocol is designed to allow clients to choose: use the index, fall back to HTTP probing, or do both in parallel.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good point - I'll add this to the motivation section. The index files are static and highly CDN-cacheable, so cache.nixos.org could serve index requests from edge nodes rather than hitting S3 for each `.narinfo` lookup. This reduces both latency and S3 request costs.

kalbasit (commented):
Thanks again for the review. I have addressed all your comments. I wasn't sure about the etiquette regarding resolving threads—should I leave them open for you to resolve if you are satisfied, or should I resolve them? I did go ahead and resolve the obvious code changes since I adopted those directly.

rfcs/0195-binary-cache-index-protocol.md
Mic92: We need probably to measure first.

rfcs/0195-binary-cache-index-protocol.md
roberth: It would fail to parse.
Also note the possibly unexpected byte order.
- https://github.com/NixOS/nix/pull/15004

rfcs/0195-binary-cache-index-protocol.md
roberth: I meant to incorporate those HTTP features in order to simplify the protocol.
By appending to a large journal file and relying on this feature, you may *both* reduce the spec complexity and improve performance, in terms of latency and number of requests.

Unless we have an overriding reason to provide this inefficient multi-file scheme, I think we'd be better off treating an append only log as an append only log at the HTTP level.

rfcs/0195-binary-cache-index-protocol.md
roberth: > 'drawback'

I see absolutely no residual drawback-ness to warrant even a scare-quoted 'drawback' unless we're competing with an alternative that requires no protocol changes at all, which seems infeasible.

rfcs/0195-binary-cache-index-protocol.md
roberth: Interestingly, Nix32 encoding follows the lexicographic sort of the reversed byte string.
So prefix-based sharding for Nix32 paths is - behind the scenes - suffix-based sharding as it relates to native hash bytes and the base-16 encoding.
See docs pr https://github.com/NixOS/nix/pull/15004 referenced earlier.

I feel uneasy to perpetuate this syntactic quirk.
If I understand correctly, it causes the byte sequences in this spec to be reverse of the native hash bytes. That is very very ugly.

rfcs/0195-binary-cache-index-protocol.md
roberth: Resource constrained clients should probably just use a nearby proxy, such as the improved cache.nixos.org CDN, volunteer mirrors, a mirror/"proxy" in the local network, etc.

rfcs/0195-binary-cache-index-protocol.md
kevincox: I also agree that while it would have been better for `nix-cache-info` to be in a more standard format I think it is still better to stick with it rather than adding yet another manifest.

As @roberth said it is possible that support for a version of the entire manifest in another format is still on the table. But it is best to do that separately and then these fields will go along for the ride rather than having skew likely forever.

rfcs/0195-binary-cache-index-protocol.md
kevincox: > With that said, I still think an epoch is better here than a number, it's not really for the "time" of it, it just makes things easier on the server-side. Instead of read current value, increment, write back it would be just stick in the current epoch.

It seems like the server should still be doing something like this to ensure that it doesn't end up reusing a timestamp in the face of time jumps or other issues.

It sounds to me that the protocol can say that this is an opaque monotonically increasing number. If the implementation wants to use a timestamp that would be fine (as long as it makes sure it is increasing).

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1547 @@
...
+- `journal.retention_count`: Number of journal segments retained before archival into shards
+- `epoch.current`: Current shard generation number
+- `epoch.previous`: Previous shard generation (for grace period support; see Section 9)
+- `deltas.enabled`: Whether differential updates are available (see Section 10)
+- `deltas.oldest_base`: Oldest epoch from which deltas can be applied. Clients with a local epoch older than this must perform a full download.
+- `deltas.compression`: Compression algorithm for delta files (`zstd`)
+
+**Caching**: Servers SHOULD use the `Cache-Control` HTTP header to specify the caching duration of the manifest. Clients SHOULD respect this header to allow the server to control how long the manifest is cached. Revalidation using `If-Modified-Since` or `ETag` SHOULD also be used.
+
+**Integrity Verification**: Clients SHOULD verify manifest integrity using HTTP-level mechanisms (`ETag`, `Content-MD5`). Cryptographic signing of index files is deferred to future work (see Future Work: Index Signing and Trust).
```
kevincox: `ETag` provides no verification capabilities as it as an opaque string.

Mentioning `Content-MD5` is fine but seems not worth it to me:
1. It is super rare.
2. TLS provides some form of verification anyways.
But I guess it doesn't hurt.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: You're right that this is an unfortunate consequence of Nix's non-standard base32 encoding. The big-endian interpretation in Section 2.1 is required for the protocol to work correctly—it ensures lexicographic string ordering equals numeric ordering, which is essential for prefix-based sharding and delta encoding.
The alternative would be to have the protocol convert Nix32 → native bytes → use native byte order, but this would:

1. Add a conversion step on every operation
1. Break the correspondence between string prefixes and shard assignment
1. Add complexity without functional benefit

I can add a note acknowledging that the 160-bit integers in the index are byte-reversed relative to the native hash representation, but I don't see a way to avoid this without significantly complicating the protocol. Is there a specific problem you foresee this causing?

I considered reversing the string so that shard prefixes correspond to native hash byte prefixes, but this would break the intuitive correspondence between a hash's visible prefix (`b6gv...`) and its shard location (`b6/`). Operators debugging cache issues would need to mentally reverse hashes to find the right shard.

The current design's 'ugliness' is confined to the internal byte representation, which most implementers won't encounter directly (they'll use libraries like go-nix). The reverse approach would surface the ugliness to every user interaction.

I'm open to other suggestions, but I think preserving `hash_prefix == shard_name` is worth the internal byte-order quirk.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: That's a fair point — I'll inline the manifest fields directly into `nix-cache-info` rather than having a separate `manifest.json`. This eliminates an HTTP request and avoids adding yet another file format.

I'll use an `Index` prefix for all fields to namespace them clearly (e.g., `IndexVersion`, `IndexFormat`, `IndexShardsBase`, etc.). The nested structures in the current manifest (like `sharding.depth`) would flatten to `IndexShardingDepth`.

Does this approach work? Any preferences on the field naming convention?

Example `nix-cache-info`

```
StoreDir: /nix/store
WantMassQuery: 1
Priority: 40
IndexVersion: 1
IndexFormat: hlssi
IndexItemCount: 1200000000
IndexShardingDepth: 2
IndexShardingAlphabet: 0123456789abcdfghijklmnpqrsvwxyz
IndexEncodingType: golomb-rice
IndexEncodingParameter: 8
IndexHashBits: 160
IndexPrefixBits: 10
IndexJournalBase: https://cache.example.com/nix-cache-index/journal/
IndexShardsBase: https://cache.example.com/nix-cache-index/shards/
IndexDeltasBase: https://cache.example.com/nix-cache-index/deltas/
IndexJournalCurrentSegment: 1705147200
IndexJournalRetentionCount: 12
IndexEpochCurrent: 42
IndexEpochPrevious: 41
IndexDeltasEnabled: true
IndexDeltasOldestBase: 35
IndexDeltasCompression: zstd
```

rfcs/0195-binary-cache-index-protocol.md
kalbasit: I've been thinking about how to implement a single-file journal over dumb storage (S3 + CDN), and I don't see a clean path. S3 objects are immutable, so "appending" requires re-uploading the entire file. The multi-segment approach lets writers upload small files without touching existing data.

For smart servers (with actual append support or a proxy layer), a single-file journal with range requests would indeed be simpler and more efficient. But I want the baseline protocol to work with just static file hosting.

I see two options:

**Option A: Define both modes in the spec**
- Add a field to indicate journal mode (`segments` vs `single`)
- Clients implement both code paths
- More flexible, but adds complexity to every client implementation

**Option B: Segments as the only mode**
- Keep segments as the baseline (works with dumb storage)
- Smart servers like ncps could still optimize internally but serve the segment format for compatibility
- Simpler client implementations

Given that cache.nixos.org (the largest cache) runs on dumb S3, I'm leaning toward Option B. But I'm open to Option A if you think the range-request efficiency is worth the added client complexity.

What's your preference?

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good point — I'll update the spec to describe `current_segment` (now `IndexJournalCurrentSegment`) as an opaque monotonically increasing identifier rather than explicitly a Unix timestamp. Implementations can still use timestamps if convenient, but the protocol only requires monotonic ordering.

I'll update the field description accordingly.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Agreed — let's wait for benchmark data before adding more prescriptive guidance. The RFC already notes that resource-constrained clients can fetch only the shards they need, and the index is purely additive (clients can always fall back to HTTP probing).

The reference implementation in ncps is in progress (https://github.com/kalbasit/ncps/pull/552 and subsequent PRs). Once it's ready, I'd appreciate help gathering numbers for various scenarios including CI-like workloads (ephemeral client, partial closure lookups). If you'd like to participate in benchmarking, please reach out!

rfcs/0195-binary-cache-index-protocol.md
kalbasit: You're right — 'clients need to implement the protocol to use it' isn't a drawback, it's just the nature of any new feature. I'll remove this section from Drawbacks entirely.

rfcs/0195-binary-cache-index-protocol.md
kevincox: That looks great to me.

kalbasit: Great! I'll wait for consensus on https://github.com/NixOS/rfcs/pull/195#discussion_r2699769513 and update the RFC with this change as well.

roberth: > The problem with the `nix-cache-info` is that it requires a custom parser in all languages.

This file has irked me for a while, and I fully agree with you that it is not a desirable thing to work with.

My first thought was to gradually make it JSON, perhaps embedding for a while: `JSON: { "foo": ..., ... }` field, but then I realized this should be generalized to solve a set of problems in a unified way, so I've opened #196.

IMO you should largely ignore #196 for the purposes of this RFC. I would like it to be in place for you, especially if you find that it makes your job with the index easier, but ultimately your focus should be on the far more consequential overall design and algorithmic choices.
#195/#196 interactions can be tidied up after both are accepted, no problem. It basically means adding a JSON representation. Trivial work.

Mic92:
> Thanks again for the review. I have addressed all your comments. I wasn't sure about the etiquette regarding resolving threads—should I leave them open for you to resolve if you are satisfied, or should I resolve them? I did go ahead and resolve the obvious code changes since I adopted those directly.

I think resolving them is fine. We can always re-open if we feel like this is missing the point.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: nice! I think #196 is a step forward in this, yes it always irked me as well. This RFC can be easily updated before/after #196.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: @roberth circling back on this, which option should I put in the RFC?

kalbasit: @roberth what do you think about the options here? I think I'll keep it as-is for now (I will update the RFC with other comments from last round). Let me know if you prefer to alter this section.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: This is now addressed in the RFC.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: This is now addressed.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: This should be addressed now.

kalbasit:
Most comments are now address, can you give it another review? Let me know if this is ready for next steps.

kalbasit:
@Mic92 @roberth do you have any more comments on this RFC or can we move forward with next steps?

Mic92:
> @Mic92 @roberth do you have any more comments on this RFC or can we move forward with next steps?

Sorry for the reply. Have you attempted to implement this index. Otherwise I could just try to give this a go.

roberth:
As far as I'm concerned, prototyping implementations of this is the next step.
Main goals:
1. find ways to improve the design
2. find any holes in the spec

For the latter it helps to have multiple implementations that are built from this RFC exclusively, without knowledge of other implementations.

kalbasit:
I'd be happy to implement this RFC in ncps to give it a try. What are the chances we can implement this over `cache.nixos.org`? Even if it's `cache-v2.nixos.org` or something?

Mic92:
> I'd be happy to implement this RFC in ncps to give it a try. What are the chances we can implement this over `cache.nixos.org`? Even if it's `cache-v2.nixos.org` or something?

Sounds realistic. Maybe not in short-term because currently we are stabilizing our queue-runner v2 to get it into production. This would enable mirrors in other countries. Queue-runner-v2 in hydra shares code with Harmonia. So if I had a prototype in harmonia-cache that works we could port it over to hydra.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1507 @@
...
+IndexJournalCurrentSegment: 1705147200
+IndexJournalRetentionCount: 12
+IndexEpochCurrent: 42
+IndexEpochPrevious: 41
+IndexDeltasEnabled: true
+IndexDeltasOldestBase: 35
+IndexDeltasCompression: zstd
+```
+
+Clients that understand these fields use the index. Clients that don't recognize the fields continue with standard HTTP probing. Caches without an index simply omit these fields.
```
flokli, line 151: Did you validate Nix versions out there do not stumble over unknown keys in there, but just skip them over?

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1507 @@
...
+
+**Field Definitions**:
+
+- `IndexVersion`: Protocol version (currently 1). Clients MUST reject indexes with unsupported versions.
```
flokli, line 155: This describes fields that are in *nix-cache-info*. I don't understand why `IndexShardingAlphabet` and `IndexHashBits` are in here, and MUST be set to 0123456789abcdfghijklmnpqrsvwxyz and 160.

I'd rather skip them from the field alltogether, this index is only intended for store path digests, and we know how these look like. If we'd ever need to change this, we'd probably also introduce another version here.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1507 @@
...
+```
++b6gvzjyb2pg0kjfwn6a6llj3k1bq6dwi
++a1b2c3d4e5f6g7h8i9j0k1l2m3n4o5p6
+-x9y8z7w6v5u4t3s2r1q0p9o8n7m6l5k4
++q1w2e3r4t5y6u7i8o9p0a1s2d3f4g5h6
+```
+
+- Lines beginning with `+` indicate additions
+- Lines beginning with `-` indicate deletions (tombstones)
+- Hash is the 32-character store path hash (no `-name` suffix)
```
flokli:
```suggestion
- Hash is the 160 bit store path digest, nixbase32-encoded (32 characters, no `-name` suffix)
```

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1507 @@
...
+- Encoding is ASCII (all characters are in the ASCII range: `+`, `-`, the 32 Nix base32 characters, and newline `\n`)
```
flokli:
```suggestion
- The whole file is ASCII (as `+`, `-`, the 32 Nix base32 characters, and newline `\n` are in the ASCII range)
```

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1507 @@
...
+
+**Segment Lifecycle**:
+1. Writer appends to current segment file
```
flokli, line 200: It should be noted somewhere that this implies all writes have to go through a single writer. This currently is not the case, multiple places can write to a raw object storage directly.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1507 @@
...
+2. The segment is rotated at the discretion of the cache operator.
+3. Segments older than `IndexJournalRetentionCount` are archived into shards during compaction
+4. Archived segments are deleted after successful compaction
+
+**Write Protocol** (for cache operators):
+```
+On artifact push:
+  1. Upload .narinfo and .nar to storage
```
flokli: For store paths pointing to contents that already exist on the binary cache, only the .narinfo file is uploaded. Also, there's other things like logs and debug symbols uploaded, depending on the configuration.

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1507 @@
...
+  2. Append "+<hash>\n" to current journal segment
+
+On garbage collection:
+  1. Delete .narinfo and .nar from storage
```
flokli: The NAR, only if the NAR is not referenced by any other .narinfo (which makes GC so annoying)

rfcs/0195-binary-cache-index-protocol.md
```diff
@@ -0,0 +1,1507 @@
...
+{IndexShardsBase}42/a1.idx.zst    # All hashes starting with "a1"
+...
+```
+
+For `IndexShardingDepth = 0` (small caches):
+```
+{IndexShardsBase}42/root.idx.zst  # All hashes in single file
+```
+
+**Compression**: Shard files MUST be compressed with zstd (indicated by `.zst` extension).
```
flokli, line 243: Then why is there a `IndexEncodingType` in the `nix-cache-info`, if this has to be zstd?

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Validated. Nix's `nix-cache-info` parser (`src/libstore/binary-cache-store.cc`) reads line-oriented `key: value` pairs, recognizes only `StoreDir`, `WantMassQuery`, and `Priority`, and silently ignores every other key — unrecognized keys fall through the `if`/`else if` chain with no error. I checked this is unchanged from Nix 2.3 (2019) through the current development branch. Added a **Backward Compatibility** note to the RFC stating this explicitly.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Agreed — these encode invariants of the format, not per-cache configuration. Dropped `IndexShardingAlphabet` and `IndexHashBits` (and, by the same reasoning, `IndexEncodingType` and `IndexPrefixBits`). Their meaning now lives in a new **Format Invariants** note: the index is over 160-bit Nix store-path digests in Nix base32, suffixes are Golomb-Rice coded, shards are zstd-compressed, and prefix width is `IndexShardingDepth × 5`. Changing any of these would require a new `IndexFormat`/`IndexVersion`.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Applied.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Applied.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Good catch. Added a **Concurrency** note: because the journal is append-only, appends to a segment MUST be serialized through a single logical writer per cache — a change from today's practice, where multiple uncoordinated agents write `.narinfo`/`.nar` directly to object storage in parallel. Artifact uploads MAY stay concurrent, but recording mutations in the journal needs a serialization point (a dedicated writer/queue), or the cache can periodically rebuild segments by listing storage.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Right. Reworded the write protocol: upload the `.narinfo`, the `.nar` only if its contents are not already present on the cache, plus any additional objects the configuration produces (e.g. build logs, debug symbols).

Also made explicit that those auxiliary objects are **not** indexed: build logs are keyed by derivation and debug symbols by build-id, not by store-path digest, so they produce no journal entry. Each push/GC of a store path records exactly one journal line (`+<hash>`/`-<hash>`) regardless of how many auxiliary objects accompany it.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: Fixed. The GC step now deletes the `.narinfo`, and the `.nar` only if it is no longer referenced by any other `.narinfo`.

rfcs/0195-binary-cache-index-protocol.md
kalbasit: This was a naming collision on my part: the suffixes are Golomb-Rice *encoded*, then the whole shard file is zstd-*compressed* — two distinct layers. Rather than disambiguate the field, I dropped `IndexEncodingType` entirely, since golomb-rice is fixed by the `hlssi` format (the tunable Golomb parameter remains as `IndexEncodingParameter`). The new **Format Invariants** note spells out the encode-then-compress layering.

kalbasit (commented):
Thanks for the thorough review, @flokli — these were all on point. I've pushed a commit addressing every comment:

- **Dropped `IndexShardingAlphabet`, `IndexHashBits`, `IndexEncodingType`, and `IndexPrefixBits`** from `nix-cache-info`. They encode invariants of the `hlssi` format rather than per-cache configuration, so they're now captured in a single **Format Invariants** note; changing any of them would require a new `IndexFormat`/`IndexVersion`.
- **Confirmed and documented** that old Nix clients ignore unknown `nix-cache-info` keys — verified the parser is unchanged from 2.3 through the development branch.
- **Documented the single-writer implication** of the append-only journal, which is a change from today's parallel direct-to-storage writes.
- **Corrected the write/GC protocol**: conditional NAR upload/delete, auxiliary objects (build logs, debug symbols), and clarified that those auxiliary objects are not indexed.
- **Applied your wording suggestions** for the journal hash and ASCII lines.

Replies inline on each thread. Thanks again!

PR #196, opened 2026-01-18
[RFC 0196] Self-describing store

roberth:
This RFC proposes that stores provide a `nix-store.json` file declaring their properties, enabling:

- Discovery of store settings (storeDir, encoding, etc.)
- Forward compatibility through explicit feature requirements
- A unified replacement for the ad hoc `nix-cache-info` format

Rendered: https://github.com/roberth/rfcs/blob/self-describing-store/rfcs/0196-self-describing-store.md

Shepherd nominations welcome!

### DIFF
--- a/rfcs/0196-self-describing-store.md
+++ b/rfcs/0196-self-describing-store.md
@@ -196,6 +196,10 @@ losing the actual file, and new Nix versions can recover or regenerate it.
 
 The file resides at the root: e.g., `https://cache.nixos.org/nix-store.json`
 
+### Build sandbox
+
+The build sandbox store is intentionally out of scope; see [Future work](#future-work).
+
 ### Client behavior
 
 1. Query `nix-store.json`.
@@ -396,3 +400,17 @@ Examples:
 A JSON Schema file may be maintained.
 We recommend it to be lenient,
 and for `$schema` to reference a mutable resource that provides the latest schema.
+
+## Build sandbox store
+
+The build sandbox is intentionally kept minimal to control build impurities.
+Making this "build store" self-describing is a worthwhile goal,
+but should be done with great care and is out of scope for this RFC.
+
+Considerations:
+- Should only appear if requested by the derivation
+- Should only contain the bare minimum information that is useful to a build
+- The set of relevant contents may depend on ongoing work on dynamic derivations and content-addressed derivations
+- Mounting the correct `nix-store.json` requires a recombination of mounts instead of mere yes/no filtering of the normal user-visible store;
+  this may unnecessarily complicate some platforms' sandbox implementations and/or reduce portability
+- Possible alternative: provide the JSON location or contents through an environment variable that applies only to the store at `$NIX_STORE`
--- a/rfcs/0196-self-describing-store.md
+++ b/rfcs/0196-self-describing-store.md
@@ -1,0 +1,398 @@
+---
+feature: self-describing-store
+start-date: 2026-01-18
+author: Robert Hensing
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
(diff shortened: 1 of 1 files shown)


rfcs/0196-self-describing-store.md
```diff
@@ -0,0 +1,416 @@
...
+[RFC 195](https://github.com/NixOS/rfcs/pull/195) proposes a significant but
+compatible addition that adds some structured data to the binary cache protocol,
+synced periodically.
+
+If that RFC is accepted, combining its data into `nix-store.json` could improve
+`nix-cache-info`'s cache staleness on clients from up to 7 days to mere minutes,
+as well as slightly reduce the number of requests.
+The details of such an integration would be decided by those involved with RFC 195.
+
+## Concrete example: the case hack
```
flokli, line 86: I'm missing two very simple examples, a directory containing NARInfo files and NAR files (with whatever compression), and the one at `../../var/nix/db/nix-store.json`.

There's arguably a big difference to storing Nars or unpacked files, and I feel like we're missing some format field of some sort. It doesn't map nicely to the more additive nature of features.

7c6f434c: Maybe it is also filesystem object encoding like casehack?

rfcs/0196-self-describing-store.md
```diff
@@ -0,0 +1,416 @@
...
+  "usedFeatures": [],
+  "substituterDefaults": {
+    "wantMassQuery": true,
+    "priority": 40
+  }
+}
+```
+
+## Fields
+
```
roberth, line 139:
```suggestion

The following list of fields is exhaustive until new fields are specified in follow-up RFCs.
Unspecified fields MUST be ignored.
Instead, fields that are essential to the store's integrity should be specified, and settings that are not should be handled through other means, such as the client's configuration system.

```

E.g. [Nix local store](https://nix.dev/manual/nix/2.34/store/types/local-store.html#local-store) has options that are implementation-specific or not even applicable:
- `build-dir`: Nix settings currently combine some build settings into the store. They're correlated but only as an implementation detail.
- `use-roots-daemon`: has little bearing on the store storage and it can continue to live in `nix.conf`
- `real` (implied in the context of `nix-store.json`, as the client knows where it is)

rfcs/0196-self-describing-store.md
```diff
@@ -0,0 +1,416 @@
...
+- [NixOS/nix#9318](https://github.com/NixOS/nix/issues/9318) — `use-case-hack` is a property of a store, not a global setting
+- [NixOS/nix#1446](https://github.com/NixOS/nix/issues/1446) — Darwin does not imply case-insensitive file system
+
+# Detailed design
+[design]: #detailed-design
+
+## Metadata file
+
+Stores provide a `nix-store.json` file containing metadata in JSON format.
+
```
roberth, line 120:
```suggestion

This file should be regarded as persistent state of the underlying store representation rather than a configuration file.

```

rfcs/0196-self-describing-store.md
```diff
@@ -0,0 +1,416 @@
...
+```json
+{
```
roberth, line 122: Perhaps it should even include:

```suggestion
{
  "_comment": "DO NOT EDIT THIS FILE BY HAND. Doing so may violate invariants that cause the store to be corrupted.",
```

rfcs/0196-self-describing-store.md
```diff
@@ -0,0 +1,416 @@
...
+  "storeDir": "/nix/store",
+  "fileSystemObjectEncoding": {
+    "caseHack": true
+  },
+  "requiredFeatures": [],
+  "expectedFeatures": [],
+  "usedFeatures": [],
+  "substituterDefaults": {
+    "wantMassQuery": true,
+    "priority": 40
```
roberth, line 133: Worth noting that these were grandfathered in.

Especially `priority` should have been client-side configuration.
It is weird and of questionable utility.

roberth (commented):
Clarification: store state and metadata, not all settings.

PR #197, opened 2026-01-24
[RFC 0197] package sets by-name

quantenzitrone:
A new location and a unified interface for package sets in nixpkgs.

[rendered](https://github.com/quantenzitrone/rfcs/blob/packagesets/rfcs/0197-package-set-definitions.md)

### DIFF
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -202,6 +202,7 @@ lib.makeScope newScope (
         (import ./overrides.nix)
         (import ./functions.nix)
         (import ./aliases.nix)
+        (import ./manual-packages.nix)
       ]
   )
 )
@@ -247,6 +248,7 @@ pkgs
 │   │   │           └── package.nix
 │   │   ├── aliases.nix
 │   │   ├── functions.nix
+│   │   ├── manual-packages.nix
 │   │   ├── mkFishPlugin.nix (could also be inlined in functions.nix)
 │   │   └── overrides.nix
 │   │
@@ -269,11 +271,12 @@ pkgs
 │   │   │   └── buildPythonPackage.nix
 │   │   ├── aliases.nix
 │   │   ├── functions.nix
+│   │   ├── manual-packages.nix
 │   │   ├── overrides.nix
 │   │   └── packageset.nix ← entrypoint for versioned python3Packages
 │   ...
 └── top-level
-    ├── by-name-overlay.nix ← aliases for deprecated versions of package sets
+    ├── aliases.nix ← aliases for deprecated versions of package sets
     ├── by-name-overlay.nix ← used to autocall sharded packages (no change required)
     ...
     └── package-sets.nix ← calls all package sets in pkgs/sets
@@ -307,6 +310,8 @@ lib.pipe ../sets [
       ))
       # overlay functions
       (lib.extends (import (../sets/${name}/functions.nix)))
+      # overlay manual packages
+      (lib.extends (import (../sets/${name}/manual-packages.nix)))
       # overlay overrides
       (lib.extends (import (../sets/${name}/overrides.nix)))
       # make a new scope and recurse into it
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -21,7 +21,12 @@ Different ideas on how to handle package sets in nixpkgs.
   - auto updates with r-ryantm + merge bot maintainer merging
 - *your goal here*
 
+<details>
+<summary>
+
 # Idea 1: new directory for package sets
+
+</summary>
 
 Proof-Of-Concept implementation in https://github.com/NixOS/nixpkgs/pull/482538 and https://github.com/NixOS/nixpkgs/pull/482543
 
@@ -49,6 +54,8 @@ Proof-Of-Concept implementation in https://github.com/NixOS/nixpkgs/pull/482538 
 ## Unresolved Questions
 
 - *your question here*
+
+</details>
 
 # Idea 2: nested by-name structure
 
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -194,17 +194,13 @@ python:
 lib.makeScope newScope (
   self:
   lib.fix (
-    lib.extend
-      (x: {
-        inherit python;
-      })
-      [
-        (import ../../top-level/by-name-overlay.nix ./by-name)
-        (self: super: lib.optionalAttrs config.allowAliases (import (./aliases.nix) self super))
-        (import ./functions.nix)
-        (import ./manual-packages.nix)
-        (import ./overrides.nix)
-      ]
+    lib.pipe (x: { inherit python; }) [
+      (lib.extends (import ../../top-level/by-name-overlay.nix ./by-name))
+      (lib.extends (self: super: lib.optionalAttrs config.allowAliases (import ./aliases.nix self super)))
+      (lib.extends (import ./functions.nix))
+      (lib.extends (import ./manual-packages.nix))
+      (lib.extends (import ./overrides.nix))
+    ]
   )
 )
 ```
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -27,9 +27,6 @@ This doesn't apply to package sets that are auto-generated like `haskellPackages
   - vetting
 - *your goal here*
 
-<details>
-<summary>
-
 # Detailed design
 [design]: #detailed-design
 
@@ -179,7 +176,7 @@ pkgs
 <details>
 <summary>
 
-## Idea 3: package sets in `pkgs/by-name` (scrapped)
+## Idea 3: package sets in `pkgs/by-name`
 
 </summary>
 
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -151,6 +151,7 @@ pkgs
 - unresolved questions
 - some package sets like `lixPackages` (behind all `lib*` packages) will not be accessible through
   GitHub UI
+- having package sets in `pkgs/by-name` may not fit the spirit of rfc 140
 - *your drawback here*
 
 ## Unresolved Questions
@@ -210,6 +211,7 @@ pkgs
     huge as well.
 - some package sets like `lixPackages` (behind all `lib*` packages) will not be accessible through
   GitHub UI
+- having package sets in `pkgs/by-name` may not fit the spirit of rfc 140
 - directory names as attrpaths is sketchy
 - unresolved questions
 
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -149,6 +149,8 @@ pkgs
 ## Drawbacks
 
 - unresolved questions
+- some package sets like `lixPackages` (behind all `lib*` packages) will not be accessible through
+  GitHub UI
 - *your drawback here*
 
 ## Unresolved Questions
@@ -206,6 +208,8 @@ pkgs
   - currently only few shards like `li` are too large for GitHubs UI, but with this idea a lot of
     shards like `py` (`python3Packages`), `ha` (`haskellPackages`) or `vi` (`vimPlugins`) will be
     huge as well.
+- some package sets like `lixPackages` (behind all `lib*` packages) will not be accessible through
+  GitHub UI
 - directory names as attrpaths is sketchy
 - unresolved questions
 
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -240,6 +240,7 @@ TODO: move other designs here once decided
 # Prior art
 [prior-art]: #prior-art
 
+- `by-name` stucture for `python3Packages` https://github.com/NixOS/nixpkgs/pull/449896 https://github.com/NixOS/nixpkgs-vet/pull/180
 - https://github.com/NixOS/nixpkgs/issues/482537
 - https://github.com/NixOS/nixpkgs/issues/432625
 - `tclPackages` has their own `by-name` structure https://github.com/NixOS/nixpkgs/pull/344716
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -23,7 +23,7 @@ Package sets (definition see [Detailed Design](#detailed-design) e.g.`fishPlugin
     with a sharded `by-name` like structure
     that is checked in the CI with `nixpkgs-vet`
     - packages in this structure will get all the benefits of `pkgs/by-name`
-  - they must have files `overrides.nix`, `aliases.nix`, `functions.nix` and `manual-packages.nix`
+  - they _must_ have files `overrides.nix`, `aliases.nix`, `functions.nix` and `manual-packages.nix`
     containing aliases, overrides, functions and packages outside the `by-name` structure
   - are optionally are autocalled
     - this would be the case for manually defined package sets
@@ -34,9 +34,9 @@ Package sets (definition see [Detailed Design](#detailed-design) e.g.`fishPlugin
 - can be versioned for which they:
   - are called multiple times through `packageset.nix`
     with the different interpreter/compiler/... versions
-  - may have a top-level function `<setname>For` that allows convenient creation
+  - _may_ have a top-level function `<setname>For` that allows convenient creation
     of new versions
-  - may be in `passthru` of the package that they depend on
+  - _may_ be in `passthru` of the package that they depend on
     - this can be archieved by using the `<setname>For` function
 
 # Motivation
@@ -178,9 +178,13 @@ which allows package sets to be flexible.
 Calling these package sets should be done in a file in `pkgs/top-level` e.g.
 - `top-level/package-sets.nix` → have all package sets in one place
 - `top-level/all-packages.nix` → only have autocalled packages in a separate file → file is small
+The `packageset.nix` is especially useful for versioned package set,
+but could also be used for other package sets that need special handling.
+
 
 If a package set does not have a `packageset.nix`,
 it is autocalled by a file in `pkgs/top-level` (e.g. `top-level/package-sets.nix`).
+This would be the case for most simple or auto generated package sets.
 
 Versioned package sets
 like `python*Packages`, `nextcloud*Packages`
@@ -212,14 +216,14 @@ lib.makeScope newScope (
 )
 ```
 
-Versioned package sets may define a top-level function wrapping this e.g.
+Versioned package sets _may_ define a top-level function wrapping this e.g.
 ```nix
 {
   pythonPackagesFor = callPackage ../sets/pythonPackages/packageset.nix {};
 }
 ```
 
-Versioned package sets may be in `passthru` of the package that they depend on e.g.
+Versioned package sets _may_ be in `passthru` of the package that they depend on e.g.
 ```nix
 {
   lib,
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -200,10 +200,7 @@ lib.makeScope newScope (
       })
       [
         (import ../../top-level/by-name-overlay.nix ./by-name)
-        (lib.extends (
-          self: super:
-          lib.optionalAttrs config.allowAliases (import (../sets/${name}/aliases.nix) self super)
-        ))
+        (self: super: lib.optionalAttrs config.allowAliases (import (./aliases.nix) self super))
         (import ./functions.nix)
         (import ./manual-packages.nix)
         (import ./overrides.nix)
@@ -309,8 +306,8 @@ lib.pipe ../sets [
       (lib.extends (import ./by-name-overlay.nix ../sets/${name}/by-name))
       # overlay aliases
       (lib.extends (
-        self: super:
-        lib.optionalAttrs self.config.allowAliases (import (../sets/${name}/aliases.nix) self super)
+        setself: setsuper:
+        lib.optionalAttrs self.config.allowAliases (import (../sets/${name}/aliases.nix) setself setsuper)
       ))
       # overlay functions
       (lib.extends (import (../sets/${name}/functions.nix)))
@@ -328,7 +325,7 @@ lib.pipe ../sets [
 // {
   # python
   # we only recurse into the newest 2 python versions
-  python3PackagesFor = self.callPackage ../sets/python3Packages/packageset.nix {};
+  python3PackagesFor = self.callPackage ../sets/python3Packages/packageset.nix { };
   python311Packages = self.python3PackagesFor self.python311;
   python312Packages = self.python3PackagesFor self.python312;
   python313Packages = lib.recurseIntoAttrs (self.python3PackagesFor self.python313);
@@ -339,7 +336,7 @@ lib.pipe ../sets [
   pypy3Packages = pypy312Packages;
 
   # nextcloud
-  nextcloudPackagesFor = self.callPackage ../sets/nextcloudPackages/packageset.nix {};
+  nextcloudPackagesFor = self.callPackage ../sets/nextcloudPackages/packageset.nix { };
   nextcloud31Packages = self.nextcloudPackagesFor "31"; # idk if it works like this, this is an example
   nextcloud32Packages = self.nextcloudPackagesFor "32";
   nextcloud33Packages = self.nextcloudPackagesFor "33";
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -149,8 +149,7 @@ pkgs
 ## Drawbacks
 
 - unresolved questions
-- some package sets like `lixPackages` (behind all `lib*` packages) will not be accessible through
-  GitHub UI
+- `lixPackages` (behind all `lib*` packages) will not be accessible through GitHubs UI
 - having package sets in `pkgs/by-name` may not fit the spirit of rfc 140
 - *your drawback here*
 
@@ -206,9 +205,13 @@ pkgs
 ## Drawbacks
 
 - huge shards due to package sets
-  - currently only few shards like `li` are too large for GitHubs UI, but with this idea a lot of
-    shards like `py` (`python3Packages`), `ha` (`haskellPackages`) or `vi` (`vimPlugins`) will be
-    huge as well.
+  - currently only few shards like `li` are too large for GitHubs UI, but with this idea more shards
+    will be huge as well
+    - specifically 12 more shards `em`, `gn`, `ha`, `ho`, `oc`, `pe`, `py`, `rp`, `sb`,
+      `te`, `ty`, `vi` and `vi` (for `emacsPackages`, `gnomeExtensions`, `haskellPackages`,
+      `home-assistant-component-tests`, `ocamlPackages`, `pearlPackages`, `python3Packages`,
+      `rPackages`, `sbclPackages`, `texlivePackages`, `typstPackages` and `vimPlugins`) will be
+      inaccessible.
 - some package sets like `lixPackages` (behind all `lib*` packages) will not be accessible through
   GitHub UI
 - having package sets in `pkgs/by-name` may not fit the spirit of rfc 140
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -90,6 +90,7 @@ pkgs
 - reuse of RFC 140 concepts
 - making maintainer merging work for this is probably relatively simple
   - implement tooling to work for a non-sharded by-name structure
+  - implement tooling to work for multiple directories https://github.com/NixOS/nixpkgs-vet/pull/180
   - enable tooling on all subdirectories of `pkgs/sets-by-name`
 
 ## Drawbacks
@@ -117,8 +118,8 @@ Proof-Of-Concept implementation in https://github.com/NixOS/nixpkgs/pull/483432
 
 - Package sets are done in a nested `by-name` structure under `pkgs/by-name`, e.g.
   `fishPlugins.puffer` would be in `by-name/fi/fishPlugins/pu/puffer`.
-- To distinguish packages from packagesets we have to mark packagesets somehow, for example with a
-  file `by-name/fi/fishPlugins/.packageset`, if that file exists, it's called as a packageset,
+- To distinguish packages from package sets we have to mark package sets somehow, for example with a
+  file `by-name/fi/fishPlugins/.packageset`, if that file exists, it's called as a package set,
   if not the `package.nix` must exist and is called as a package.
 
 ```
--- a/rfcs/0197-package-set-definitions.md
+++ b/rfcs/0197-package-set-definitions.md
@@ -1,5 +1,5 @@
 ---
-feature: nixpkgs package sets by-name
+feature: nixpkgs package sets
 start-date: 2026-01-24
 author: quantenzitrone
 co-authors: (find a buddy later to help out with the RFC)
@@ -11,10 +11,33 @@ related-issues: https://github.com/NixOS/nixpkgs/pull/482538
 # Summary
 [summary]: #summary
 
-Package sets like `fishPlugins` or `python3Packages` move to a `pkgs/by-name` like structure in
-`pkgs/sets-by-name/<setname>`.
+Package sets (definition see [Detailed Design](#detailed-design) e.g.`fishPlugins`, `python3Packages`)
 
-This doesn't apply to package sets that are auto-generated like `haskellPackages`.
+- should be top-level attribute sets that contain
+  - packages
+  - functions helping these packages
+  - variables like `version`, `withFoo`, ...
+  - TBD: other package sets?
+- should move to `pkgs/sets/<setname>` where
+  - they must have a directory `pkgs/sets/<setname>/by-name`
+    with a sharded `by-name` like structure
+    that is checked in the CI with `nixpkgs-vet`
+    - packages in this structure will get all the benefits of `pkgs/by-name`
+  - they must have files `overrides.nix`, `aliases.nix`, `functions.nix` and `manual-packages.nix`
+    containing aliases, overrides, functions and packages outside of the `by-name` structure
+  - are optionally are autocalled
+    - this would be the case for manually defined package sets
+      and auto generated package sets
+  - if there exists a file `pkgs/sets/<setname>/packageset.nix`,
+    they have to be manually called
+    - this would be the case for versioned package sets
+- can be versioned for which they:
+  - are are called multiple times through `packageset.nix`
+    with the different interpreter/compiler/... versions
+  - may have a top-level function `<setname>For` that allows convenient creation
+    of new versions
+  - may be in `passthru` of the package that they depend on
+    - this can be archived by using the `<setname>For` function
 
 # Motivation
 [motivation]: #motivation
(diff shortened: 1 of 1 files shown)


rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Drawbacks
+
+- unresolved questions
+- *your drawback here*
+
+## Unresolved Questions
+
+- How do we handle functions like `fishPlugins.buildFishPlugin`?
+- How do we handle aliases?
```
quantenzitrone: This could probably be handled with fixpoint evaluation like `top-level/stage.nix` and having an `overlay.nix` that contains functions, aliases and packages with overrides.
I tried to implement that, but failed.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,151 @@
...
+## Advantages
+
+- flexibility
+- easy transition
+  - for a lot of package sets just move the folder and rename the `default.nix` to `packageset.nix`
+
+## Drawbacks
+
+- all packagesets have to handle their shit themselves, so no maintainer merging, by-name
+  autocalling and stuff without the additional work
```
Sigmanificient: I am not aware of how the current system works, but why does this block maintainer merging? I'd like to know a bit more about the details on the technical limitation what would cause an issue

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,151 @@
...
+- The `packageset.nix` files are autocalled by some `pkgs/top-level/by-name-overlay.nix` like file.
+- Versioned package sets like `python316Packages` are done in `all-packages.nix` by calling the
+  `packageset.nix` again with different arguments.
+
+## Advantages
+
+- flexibility
+- easy transition
+  - for a lot of package sets just move the folder and rename the `default.nix` to `packageset.nix`
+
```
Sigmanificient, line 364: Could locality be also considered as an advantage? every packages sets are grouped in one place and easily accessible through classic navigation

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,151 @@
...
+- *your goal here*
+
+# Idea 1: new directory for package sets
+
+Proof-Of-Concept implementation in https://github.com/NixOS/nixpkgs/pull/482538 and https://github.com/NixOS/nixpkgs/pull/482543
+
+## Detailed design
+
+- Create a new directory under `pkgs/`, e.g. `pkgs/sets` that contains package sets.
+  - For example `pkgs/sets/fishPlugins` `pkgs/sets/python3Packages`.
```
Sigmanificient: I think adding an simplified tree to give a visual would be very nice (on each ideas)

### Idea 1

```
pkgs
└── sets
    ├── fishPlugins
    │   ├── as
    │   │   └── async-prompt
    │   ├── au
    │   ...
    │   └── z_
    │       └── z
    └── python3Packages
    │   ├── a2
    │   │   └── a2wsgi
    │   ├── aa
    │   ...
    │   └── zx
    │       ├── zxcvbn
    │       ├── zxcvbn-rs-py
    │       └── zxing-cpp
    ...
```

### Idea 2

```
pkgs
└── by-name
    ├── 0_
    ...
    ├── fi
    │   ├── fiano
    │   ├── fiche
    │   ├── ...
    │   ├── fishnet
    │   ├── fishPlugins
    │   │   ├── as
    │   │   │   └── async-prompt
    │   │   ├── au
    │   │   ...
    │   │   └── z_
    │   │       └── z
    │   ├── fishy
    │   ...
    ...
```

### Idea 3

```
pkgs
└── by-name
    ├── 0_
    ...
    ├── fi
    │   ├── fiano
    │   ├── fiche
    │   ├── ...
    │   ├── fishnet
    │   ├── fishPlugins.async-prompt
    │   ├── fishPlugins.autopair
    │   ├── fishPlugins.z
    │   ├── fishy
    │   ...
    ...
```

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,151 @@
...
+- Each packageset has a `packageset.nix`, that functions as an entrypoint to the package set.
```
Sigmanificient: Is there a reasoning behind not being `default.nix`?

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,151 @@
...
+## Advantages
+
+- clear package separation
+- reuse of RFC 140 concepts
+- making maintainer merging work for this is probably relatively simple
+
+## Drawbacks
+
+- unresolved questions
+- *your drawback here*
```
Sigmanificient: As stated in my comment on the implementation pr, i think a drawback of this approach is that it may make some packagesets inaccessible from github navigation, like `lixPackageSets` (which would be in `1087th` within `pkgs/by-name/li`).

it also seems like it is fitting 2 different ideas into `pkgs/by-name`, by mixing packages and sets

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,151 @@
...
+
+- clear package separation
+- reuse of RFC 140 concepts
+- making maintainer merging work for this is probably easy
+
+## Drawbacks
+
+- huge shards due to package sets
+  - currently only few shards like `li` are too large for GitHubs UI, but with this idea a lot of
+    shards like `py` (`python3Packages`), `ha` (`haskellPackages`) or `vi` (`vimPlugins`) will be
```
Sigmanificient: Can we add a precise number on affected shards for this?

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Drawbacks
+
+- all packagesets have to handle their shit themselves, so no maintainer merging, by-name
+  autocalling and stuff without the additional work
+- *your drawback here*
+
+## Unresolved Questions
+
+- *your question here*
```
Sigmanificient: it seems that we would end up with a structure like this:

```
pkgs
├── by-name/
│   ├── _0
│   ├── _1
│   ...
│   └── zz
└── sets
    ├── fishPlugins
    │   ├── as
    │   ├── au
    │   ...
    │   └── z_
    │       └── z
    └── python3Packages
    │   ├── a2
    │   │   └── a2wsgi
    │   ├── aa
    │   └── zx
    ...
```

My pattern recognition brain-cell tells me that `by-name` is just like any other entries of `pkgs/sets`. What about exploring the idea of having `pkgs/sets/top-level`?

```
pkgs
└── sets
    ├── fishPlugins
    │   ├── as
    │   ├── au
    │   ...
    │   └── z_
    │       └── z
    ...
    ├── python3Packages
    │   ├── a2
    │   │   └── a2wsgi
    │   ├── aa
    │   └── zx
    ...
    ├── toplevel
    │   ├── _0
    │   ├── _1
    │   ...
    │   └── zz
    ...
```

Sigmanificient (commented):
First time interacting with an RFC, i hope i am not making comment that are too dumb 😅

rfcs/0197-package-set-definitions.md
quantenzitrone: following the pattern of using more descriptive names like `package.nix` instead of `default.nix`

rfcs/0197-package-set-definitions.md
qweered: I like tweaked `Idea 1` with support for plain structure from `Idea 3` for small sets

**Large sets** (sharded like regular by-name):
```
pkgs/sets-by-name/fishPlugins
├── as
│   └── async-prompt
│       └── package.nix
├── au
│   └── autopair
│       └── package.nix
...
```

**Small sets** (plain structure, no sharding):
```
pkgs/sets-by-name/nodePackages
├── typescript
│   └── package.nix
├── eslint
│   └── package.nix
...
```

rfcs/0197-package-set-definitions.md
qweered: Aggree with `packageset.nix` convention

rfcs/0197-package-set-definitions.md
quantenzitrone: i think you misunderstand idea 1 a bit, i don't specify the directory structure inside `pkgs/sets`

<s> however that is a good Idea 4</s>

i changed idea 1 to this instead of creating idea 4

rfcs/0197-package-set-definitions.md
qweered: There is ongoing development for by-name-like structure for pythonPackages in https://github.com/NixOS/nixpkgs/pull/449896 and https://github.com/NixOS/nixpkgs-vet/pull/180. Tooling is almost there so i don't think its a problem

rfcs/0197-package-set-definitions.md
qweered: I am for by-name/sets, cause they would be sorted and in by-name like structure, also clear separation from "old" nixpkgs

rfcs/0197-package-set-definitions.md
Sigmanificient: @quantenzitrone well, i would like to see the proper tree of idea 1 then, that you help me to get it :)

rfcs/0197-package-set-definitions.md
Sigmanificient: yep, i also do, this was more of a curiosity. What do you think having `aliases.nix` directly within `pkgs/sets/xxx` too?

rfcs/0197-package-set-definitions.md
qweered: Updated comment

Sigmanificient:
thanks for adding trees :heart:

rfcs/0197-package-set-definitions.md
qweered: we can merge them directly into `packagesets.nix`, its just `lib.optional config.allowAliases`

Sigmanificient:
i think a drawback from idea 3 is that we end up with attrpath in directory names

rfcs/0197-package-set-definitions.md
Sigmanificient: yeah this seems like a reasonable approach for sets with small sets. What would be the threshold to move to sharding? 100?

rfcs/0197-package-set-definitions.md
quantenzitrone: nah, we can't have a random 4 character folder in the 2 character shard jungle

Sigmanificient: what about a `pkgsets` folder?

rfcs/0197-package-set-definitions.md
quantenzitrone: i have been trying to implement an additional `aliases.nix` for my idea 2 proof of concept, and i don't seem to understand fixpoint functions (i always get infinite recursion errors :/)

rfcs/0197-package-set-definitions.md
```diff
@@ -74,6 +90,28 @@ Proof-Of-Concept implementation in https://github.com/NixOS/nixpkgs/pull/483432
...
+pkgs
+└── by-name
+    ├── 0_
+    ...
+    ├── fi
+    │   ├── fiano
+    │   ├── fiche
+    │   ├── ...
+    │   ├── fishnet
+    │   ├── fishPlugins
```
Sigmanificient, line 429:
```suggestion
    │   ├── fishPlugins
    │   │   ├── .packageset
```

rfcs/0197-package-set-definitions.md
quantenzitrone: what about `pkgs/sets-by-name`

rfcs/0197-package-set-definitions.md
qweered: Yeah 100 is a reasonable threshold, maybe 150. Most sets are either enormous (like python and perl) or small. Here's the full breakdown:

| Count | Package Set |
|------:|-------------|
| 31,372 | rPackages |
| 19,160 | haskellPackages |
| ~10,806 | python3Packages (+ variants) |
| 6,564 | emacsPackages |
| 4,725 | sbclPackages |
| 4,658 | texlive |
| 4,218 | typstPackages |
| 2,080 | vimPlugins |
| 2,030 | perlPackages |
| 1,171 | ocamlPackages |
| 966 | gnomeExtensions |
| 539 | kdePackages |
| 423 | rubyPackages |
| 415 | nodePackages |
| 369 | akkuPackages |
| 229 | xorg |
| 218 | plasma5Packages / libsForQt5 |
| ~204 | luaPackages (+ variants) |
| 191 | linuxPackages |
| 181 | darwin |
| 132 | tree-sitter-grammars |
| 126 | hunspellDicts |
| 123 | gnome |
| 114 | phpExtensions |
| 106 | rocmPackages / qt6Packages |
| 102 | pantheon / kodiPackages |
| 97 | cudaPackages / coqPackages |
| 95 | idrisPackages |
| 94 | aspellDicts |
| 83 | hyphenDicts |
| 78 | xfce |
| 75 | postgresqlPackages / octavePackages |
| 70 | gopkgs / cni-plugins |
| 64 | tmuxPlugins |
| 60 | mdbook-plugins |
| 57 | mate / gnome2 |
| 55 | obs-studio-plugins |
| 51 | llvmPackages |
| 48 | cinnamon |
| 47 | lxqt |
| 42 | kakounePlugins |
| 36 | beamPackages / tclPackages |
| 31 | jetbrains |
| 24 | gst_all_1 |

Sigmanificient: wow rPackages is huge

rfcs/0197-package-set-definitions.md
qweered: > nah, we can't have a random 4 character folder in the 2 character shard jungle

yeah, didn't think about that. Then yeah, somewhere in `pkgs`, i like `sets-by-name`

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,258 @@
...
+- accessibility through github ui navigation (better than idea 2 and 3)
+- clear package separation
+- reuse of RFC 140 concepts
+- making maintainer merging work for this is probably relatively simple
+  - implement tooling to work for a non-sharded by-name structure
+  - enable tooling on all subdirectories of `pkgs/sets-by-name`
+
+## Drawbacks
+
+- autocalling logic has to be duplicated for each package set
```
Sigmanificient: What would like look like concretely? Any way to factor out common parts?

rfcs/0197-package-set-definitions.md
quantenzitrone: with the revised idea 1 this is resolved

rfcs/0197-package-set-definitions.md
quantenzitrone: `em` -> `emacsPackages`
`gn` -> `gnomeExtensions`
`ha` -> `haskellPackages`
`ho` -> `home-assistant-component-tests`
`oc` -> `ocamlPackages`
`pe` -> `pearlPackages`
`py` -> `python3Packages`
`rp` -> `rPackages`
`sb` -> `sbclPackages`
`te` -> `texlivePackages`
`ty` -> `typstPackages`
`vi` -> `vimPlugins`

so 12

well and `li` -> `lixPackages` but that shard is huge for other reasons

all other packageset and shards are small

Sigmanificient: Well, I am sold to the idea 1 since you made the first pkgs/sets pr tbh

rfcs/0197-package-set-definitions.md
Sigmanificient: I'm also okay with it :)

GetPsyched: FWIW, I prefer `pkgs/sets` and the renaming of `by-name` to `pkgs/top-level`. The only issue there is that `pkgs/top-level` already exists, making iterative migration infeasible.

I'm happy with `sets-by-name` as an inferior alternative.

quantenzitrone: Once every package has migrated to `pkgs/by-name` or `pkgs/sets-by-name` (also `pkgs/pkgs-lib` and `pkgs/build-support` is probably staying) we can move stuff from `pkgs/top-level` to `pkgs` and rename some folders to get an optimal clean `pkgs` directory.

rfcs/0197-package-set-definitions.md
Sigmanificient: @qweered curious on how you generated that list

qweered: > @qweered curious on how you generated that list

i just asked claude. here the script
```shell
{
    nix eval --json '.#legacyPackages.x86_64-linux' --apply '
      pkgs: builtins.filter (name: 
        builtins.match ".*(Packages|Pkgs|packages|pkgs|Plugins|plugins|Dicts|Grammars|Extensions)$" name != null
      ) (builtins.attrNames pkgs)
    ' 2>/dev/null | jq -r '.[]'
    echo -e "xorg\ngnome\ngnome2\nlibsForQt5\nqt6\nmate\ncinnamon\npantheon\nlxqt\nxfce\ndarwin\nlinuxKernel\ngst_all_1\ntexlive\ntree-sitter-grammars\njetbrains\nobs-studio-plugins\nkakounePlugins"
  } | sort -u | while read pkg; do
    count=$(nix eval --json ".#legacyPackages.x86_64-linux.${pkg}" --apply 'ps: builtins.length (builtins.attrNames ps)' 2>/dev/null || echo "0")
    echo "$count $pkg"
  done | sort -rn | grep -vE '^(26077|26076) (targetPackages|buildPackages|__splicedPackages|pkgs)$' | head -100
```

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+- Create a new directory under `pkgs/`, e.g. `pkgs/sets-by-name` that contains package sets.
+  - For example `pkgs/sets-by-name/fishPlugins`, `pkgs/sets-by-name/python3Packages`.
+- Each packageset has a `packageset.nix`, that functions as an entrypoint to the package set.
+  - The `packageset.nix` has some code like `top-level/by-name-overlay.nix` that autocalls packages
+    in the folder.
+    - with sharding for large package sets like `python3Packages`
+    - without sharding for small package sets like `fishPlugins`
+    - threshold for sharding is to be discussed
+      - 1000 (including extra files like `README.md` and `packageset.nix`) should the the absolute
+        maximum because of GitHubs UI
```
quantenzitrone: how should sharding threshold be handled?
what happens if a package sets exceeds or inceeds the threshold?
how does CI know if a package set is sharded or not?

quantenzitrone: <details>
<summary>list of package sets:</summary>

```sh
nix eval --impure --expr \
  'let pkgs = import <nixpkgs> {config.allowAliases = false;}; lib = pkgs.lib; in lib.filterAttrs (_: v: (__tryEval v).success && !(lib.isDerivation v) && __isAttrs v && !(lib.isFunction v)) pkgs |> __mapAttrs (_: v: __length (__attrNames v))' --json \
  | jq '. | to_entries[] | "\(.value) \(.key)"' -r | sort -nr \
  | rg -v '2413[0-2]' \
  | rg -v 'Packages[_-]' \
  | rg -v ' _|(gnome|postgresql|zabbix|beamMinimal|beam|gnat|php|python3|python2|pypy|rust_|wine|wineWow|lua|perl)[0-9]' \
  | rg -v ' (pypyPackages|python2Packages|lib|nixosTests|threads|nextcloud31Packages|yarn-berry_3-fetcher)$'
```

number of packages | set name
--- | ---
31372 | rPackages
19078 | haskellPackages
10256 | python3Packages
6558 | emacsPackages
4722 | sbclPackages
4658 | texlive
4651 | texlivePackages
3923 | typstPackages
1930 | perlPackages
1908 | vimPlugins
1155 | ocamlPackages
1026 | home-assistant-component-tests
960 | gnomeExtensions
537 | kdePackages
423 | rubyPackages
385 | vscode-extensions
369 | akkuPackages
223 | xorg
204 | luaPackages
204 | luajitPackages
192 | freebsd
188 | plasma5Packages
188 | libsForQt5
188 | azure-cli-extensions
173 | linuxPackages
172 | terraform-providers
171 | defaultPkgConfigPackages
132 | tree-sitter-grammars
126 | hunspellDicts
124 | nltk-data
119 | darwin
112 | phpExtensions
112 | libretro
112 | defaultGemConfig
109 | openbsd
104 | qt6Packages
100 | kodiPackages
97 | rocmPackages
97 | coqPackages
96 | cudaPackages
94 | aspellDicts
92 | pantheon
92 | idrisPackages
81 | home-assistant-custom-components
80 | pkgsCross
75 | octavePackages
74 | postgresqlPackages
73 | nerd-fonts
72 | dotnetCorePackages
65 | tests
63 | tmuxPlugins
63 | netbsd
61 | defaultCrateOverrides
58 | qt5
57 | mate
57 | lomiri
56 | mpvScripts
55 | indi-3rdparty
54 | obs-studio-plugins
53 | nginxModules
52 | qt6
52 | minimal-bootstrap
51 | llvmPackages
49 | writers
49 | fishPlugins
47 | maple-mono
47 | lxqt
45 | yaziPlugins
45 | optifinePackages
39 | grafanaPlugins
36 | tclPackages
35 | unixtools
35 | beamPackages
35 | beamMinimalPackages
35 | azure-sdk-for-cpp
34 | kakounePlugins
34 | home-assistant-custom-lovelace-modules
33 | vmTools
33 | translatelocally-models
33 | pkgsForCudaArch
30 | skawarePackages
30 | mopidyPackages
29 | platforms
29 | gnatPackages
29 | elmPackages
28 | nodePackages
27 | CuboCore
26 | papermcServers
26 | config
25 | pidginPackages
25 | dockerTools
25 | astal
24 | skkDictionaries
24 | linphonePackages
23 | gst_all_1
23 | graylogPlugins
23 | dotnetPackages
23 | apacheHttpdPackages
22 | minecraftServers
22 | gerbil-support
22 | gawkextlib
22 | agdaPackages
21 | vdrPlugins
21 | jetbrains
21 | ghidra-extensions
21 | formats
20 | ue4demos
20 | ibus-engines
19 | weechatScripts
19 | testers
19 | openlilylib-fonts
19 | nushellPlugins
19 | hyprlandPlugins
19 | dwarf-fortress-packages
19 | dictdDBs
19 | cmdpack
18 | wineWowPackages
18 | winePackages
18 | stdenvAdapters
18 | phpPackages
18 | lohit-fonts
18 | graalvmPackages
18 | gimpPlugins
18 | gimp2Plugins
17 | swiftPackages
17 | rustPlatform
17 | rocqPackages
17 | pulumiPackages
17 | ocaml-ng
17 | gaugePlugins
17 | dprint-plugins
16 | yarn2nix-moretea
16 | wordpressPackages
16 | pythonInterpreters
16 | openmodelica
16 | nixVersions
16 | nix-gitignore
16 | linuxKernel
16 | eclipses
15 | zathuraPkgs
15 | yarn-berry_4-fetcher
15 | nagiosPlugins
15 | icu-versions
15 | enlightenment
15 | dockapps
15 | dhallPackages
14 | windows
14 | vimUtils
14 | texFunctions
14 | rustPackages
14 | releaseTools
14 | pianoteq
14 | emiluaPlugins
14 | buildbotPackages
14 | ankiAddons
13 | reposilitePlugins
13 | mailmanPackages
13 | iosevka-comfy
13 | hareThirdParty
13 | factorPackages
13 | elasticsearchPlugins
13 | bat-extras
13 | arm-trusted-firmware
12 | zncModules
12 | vscode-utils
12 | open-music-kontrollers
12 | nimOverrides
12 | lumina
12 | luaInterpreters
12 | lixPackageSets
12 | kubernetes-helmPlugins
12 | hunspellDictsChromium
12 | haxePackages
12 | gnuradioPackages
12 | chickenPackages
11 | rust
11 | neovimUtils
11 | magnetophonDSP
11 | hyphenDicts
11 | gnome
11 | freshrss-extensions
11 | freshBootstrapTools
11 | cataclysmDDA
11 | bitscope
11 | arrayUtilities
10 | zabbix
10 | wayfirePlugins
10 | sourceHanPackages
10 | rxvt-unicode-plugins
10 | roundcubePlugins
10 | nvtopPackages
10 | nvidiaCtkPackages
10 | nextcloud32Packages
10 | matrix-synapse-plugins
10 | flutterPackages
10 | fedora-backgrounds
10 | cygwin
10 | buildRustCrateHelpers
10 | buildbot-plugins
10 | appimageTools
10 | androidenv
9 | yosys-symbiflow
9 | tex-gyre
9 | passExtensions
9 | octodns-providers
9 | nixDependencies
9 | jsonSchemaCatalogs
9 | coreboot-toolchain
8 | valeStyles
8 | pythonManylinuxPackages
8 | pythonDocs
8 | minimal-bootstrap-sources
8 | idris2Packages
8 | haskell
8 | gradle-packages
8 | driversi686Linux
8 | deadbeefPlugins
8 | cbqn-bootstrap
7 | xfce
7 | unixODBCDrivers
7 | thunderbirdPackages
7 | sourceAndTags
7 | seaborn-data
7 | opentelemetry-collector-releases
7 | nixops_unstablePlugins
7 | multitran
7 | kernelPatches
7 | javaPackages
7 | inkscape-extensions
7 | godotPackages
7 | gns3Packages
7 | gambit-support
7 | emscriptenPackages
7 | dotnetfx40
7 | dotnetfx35
7 | dartHooks
7 | aroccPackages
7 | androidStudioPackages
6 | zigPackages
6 | varnish60Packages
6 | szurubooru
6 | sweethome3d
6 | steampipePackages
6 | scx
6 | poetryPlugins
6 | opensshPackages
6 | octave-kernel
6 | npmHooks
6 | intel-oneapi
6 | icingaweb2Modules
6 | hspellDicts
6 | coolercontrol
6 | beam_minimal
6 | beam
5 | xcodeenv
5 | uefitoolPackages
5 | ubootPythonTools
5 | tflint-plugins
5 | tex-gyre-math
5 | sublime3Packages
5 | singularity-tools
5 | pub2nix
5 | postgresqlVersions
5 | postgresqlJitVersions
5 | mplus-outline-fonts
5 | kicadAddons
5 | fusePackages
5 | electron-source
5 | devShellTools
5 | curaPlugins
5 | cpp2a-kernel
5 | cpp17-kernel
5 | cpp14-kernel
5 | cpp11-kernel
5 | binlore
5 | androidStudioForPlatformPackages
4 | webos
4 | varnishPackages
4 | varnish77Packages
4 | rizinPlugins
4 | plover
4 | openraPackages
4 | openconnectPackages
4 | nixos-artwork
4 | mkspiffs-presets
4 | kakouneUtils
4 | jupyter-kernel
4 | jupyter-console
4 | isabelle-components
4 | gancioPlugins
4 | factorio-utils
4 | error-inject
4 | deviceTree
4 | cutterPlugins
4 | clojupyter
4 | checkpointBuildTools
4 | authentik-outposts
4 | anki-utils
3 | wolfram-for-jupyter-kernel
3 | sway-contrib
3 | sane-drivers
3 | pythonCondaPackages
3 | ociTools
3 | moodle-utils
3 | interception-tools-plugins
3 | expidus
2 | supercolliderPlugins
2 | nixBufferBuilders
1 | texinfoPackages
1 | perlInterpreters

</details>

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+  - the `packageset.nix` can include aliases, functions like `fishPlugins.buildFishPlugin` and
+    overrides.
+- The `packageset.nix` files are autocalled by some `pkgs/top-level/by-name-overlay.nix` like file.
```
quantenzitrone: idea: scrap the autocalling of the `packageset.nix`, we don't have that many package sets anyway (only ~300, ~450 with versions)
this way we can also have package sets without a "default" version
e.g. `nextcloudPackages` does not exist, only `nextcloud{31,32}Packages`

rfcs/0197-package-set-definitions.md
SigmaSquadron: In my opinion, we should shard all package sets. Yes, even the ones that have ~20 odd packages.

Not sharding some of them means we'll have two different directory interfaces, which means future updates to `nixpkgs-vet` (because the whole point of this is to more accurately vet these packages, like in `by-name`) will be needlessly complex because they'd need to support the differences between sharded and non-sharded package sets.

The pros and cons of sharding everything are clear: for the cost of some trees looking awkward with few shards, we can reduce the overall maintenance cost by applying one overkill solution for every set.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+
+# Summary
+[summary]: #summary
+
+Different ideas on how to handle package sets in nixpkgs.
+
+# Motivation
+[motivation]: #motivation
+
+- get rid of the package categories as directories.
```
SigmaSquadron: I think we should have clearer, straightforward goals. "Getting rid of the package categories" is meaningless; why do we want to get rid of them? Hygiene? Style? Why go through all that effort if it's not going to result in something meaningful?

The three wonderful things about `by-name` are scalability, isolation and vetting. It is infinitely more scalable than a category or directory system which was used before by top-level packages and is currently used by package sets. It also isolates each individual package in their own `directory/package.nix` cluster, which allows for the entirety of by-name to be migrated in the future. It can also be vetted, preventing a malicious PR from attaching malicious code to an entire tree of packages. These benefits are currently unavailable for package sets, and I think we should have this goal (bring the benefits of `by-name` to package sets) clearly in mind.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+- make packages in package sets take advantage of the by-name greatness
+  - auto updates with r-ryantm + merge bot maintainer merging
```
SigmaSquadron: I think our implementation plan should be separated into four stages:

1. Create interfaces for package sets that uses the same functions as `by-name` to `callPackage` packages from sharded directories, with additional extensions for package set versioning.
2. Update `nixpkgs-vet` to expand the existing linting checks to the new package sets.
3. Migrate all package sets into the new interface when it makes sense to do so, or move the children of smaller sets into `by-name` as top-level packages.
4. Add new rules to `nixpkgs-merge-bot` to allow unprivileged merging to package sets.

We can't move into the next stage before the previous one is complete, so let's not think about the later stages just now.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+- How do we handle aliases?
+- How do we handle versioned package sets?
+- How do we move large package sets?
+
+</details>
+
+# Detailed design
+[design]: #detailed-design
+
+TODO: decide for one of the above ideas
```
SigmaSquadron: Let's drop ideas 2 and 3 early. They're objectively worse in every way from idea 1, and we don't need the mental clutter. Improving idea 1 to be as simple to implement and as robust as possible should be our main focus.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+      inaccessible.
+- some package sets like `lixPackages` (behind all `lib*` packages) will not be accessible through
+  GitHub UI
+- having package sets in `pkgs/by-name` may not fit the spirit of rfc 140
+- directory names as attrpaths is sketchy
+- unresolved questions
+
+## Unresolved Questions
+
+- How do we handle functions like `fishPlugins.buildFishPlugin`?
```
SigmaSquadron, line 513: (Answering because this applies to idea 1)

I never thought it made sense for functions and derivation builders to be inside package sets. I think our Go infrastructure has the right idea: move them to the top-level, and version them in the attribute name. `fishPlugins.buildFishPlugin` can just be moved to the top-level, while `python313Packages.buildPythonApplication` becomes `buildPython313Application` on the top-level. We can deal with this at [stage 3](https://github.com/NixOS/rfcs/pull/197#discussion_r2724948547), where we migrate everything over.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+- How do we handle aliases?
```
SigmaSquadron, line 514: (Answering because this applies to idea 1)

There has been some discussion that generally boils down to adding aliases directly into `packageset.nix`. I'm personally iffy about the whole idea of `packageset.nix`, and I think we should have a more focused `aliases.nix` file on the root of every single package set directory, which is imported automatically by the governing interface, just like the sharded packages. I think `packageset.nix`, should it need to exist, should simply focus on more meta parameters of the package set, instead of being constantly updated with new aliases and such.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+- How do we handle versioned package sets?
```
SigmaSquadron, line 515: (Answering because this applies to idea 1)

I generally think it's a good idea to think of Nixpkgs as having three types of package infrastructure systems:
1. Top-level packages, which nowadays go to `by-name`;
2. Simple package sets, like `fishPlugins`, which are easier to implement in this new interface as we can simply call them as we do `by-name` packages and shove the resulting set into `fishPlugins` from the top-level.
3. Versioned package sets, like `python3Packages`, which are by far the most complex. I think we should absolutely get the input of the primary maintainers of these sets here before we move forward. There are many concerns regarding updating versioned sets, and default aliases, as well as edge cases like `python3.pkgs` and `python3Packages`. In the end, having something that is no worse to maintain as it was before would be ideal.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+- How do we move large package sets?
```
SigmaSquadron, line 516: (Answering because this applies to idea 1)

Some of them are easy enough. If we resolve the aforementioned issues with the python edge cases, it should be a simple matter of moving and renaming files. OCaml should also be pretty simple.

The generated sets, like Haskell, Node, Lisp and R, will be impossible, though. Those packages don't actually exist as separate files in Nixpkgs, and are instead imported from some other upstream.

This is something that again, we'll have to get the input of package set maintainers on. Moving everything over will end up being a case-by-case basis.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+- accessibility through github ui navigation (better than idea 2 and 3)
+- clear package separation
+- reuse of RFC 140 concepts
+- making maintainer merging work for this is probably relatively simple
+  - implement tooling to work for a non-sharded by-name structure
+  - enable tooling on all subdirectories of `pkgs/sets-by-name`
+
+## Drawbacks
+
+- autocalling logic has to be duplicated for each package set
```
SigmaSquadron: No, it doesn't. It should be pretty simple to autocall from top-level or another "internal maintenance" directory in the tree. I don't really understand why your initial implementation PR had the duplicated code when you could just as easily have mapped it inside another directory list reader for the package set attrnames.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+    in the folder.
+    - with sharding for large package sets like `python3Packages`
+    - without sharding for small package sets like `fishPlugins`
+    - threshold for sharding is to be discussed
+      - 1000 (including extra files like `README.md` and `packageset.nix`) should the the absolute
+        maximum because of GitHubs UI
+  - the `packageset.nix` can include aliases, functions like `fishPlugins.buildFishPlugin` and
+    overrides.
+- The `packageset.nix` files are autocalled by some `pkgs/top-level/by-name-overlay.nix` like file.
+- Versioned package sets like `python316Packages` are done in `all-packages.nix` by calling the
```
SigmaSquadron: That seems good to me. As mentioned in https://github.com/NixOS/rfcs/pull/197#discussion_r2724956507, this is conditional on the approval of the current versioned package set maintainers.

rfcs/0197-package-set-definitions.md
SigmaSquadron: I like this idea. You mentioned how there's duplicated code in `packageset.nix`. If `by-name-overlay.nix` can autocall `packageset.nix`, why not skip the middleman and just call the sharded packages while we're at it?

SigmaSquadron (commented):
I was discussing this privately with @Sigmanificient earlier today, and he helped me organise my thoughts on this.

rfcs/0197-package-set-definitions.md
quantenzitrone: skipping `packageset.nix` is basically idea 2 but package sets are in a separate unsharded directory `pkgs/sets-by-name` instead of `pkgs/by-name`

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+└── sets-by-name
+    ├── fishPlugins <- small package set so no sharding
+    │   ├── async-prompt
+    │   │   └── package.nix
+    │   ├── autopair
+    │   │   └── package.nix
+    │   ...
+    │   ├── z
+    │   │   └── package.nix
+    │   └── packageset.nix <- entrypoint to package set
```
infinisil: Idea 1 is how I also always imagined other package sets to work, so +1 to that!

I have a feeling `packageset.nix` could become problematic, because if it becomes large, you might want to split it into smaller files, at which point you can't nest them into a separate directory without complicating the enumeration of packages.

Though one could also deal with this problem as it occurs, especially because if CI is properly integrated (cc https://github.com/NixOS/nixpkgs-vet/pull/180), it will [ensure](https://github.com/NixOS/nixpkgs-vet?tab=readme-ov-file#nix-parser-checks) that

> Each package directory must not refer to files outside itself using symlinks or Nix path expressions.

which means you're pretty much guaranteed that moving package directories won't break any `by-name` packages (though it can break packages outside `by-name`).

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,267 @@
...
+
+Proof-Of-Concept implementation in https://github.com/NixOS/nixpkgs/pull/483432
+
+## Detailed design
+
+- Package sets are done in a nested `by-name` structure under `pkgs/by-name`, e.g.
+  `fishPlugins.puffer` would be in `by-name/fi/fishPlugins/pu/puffer`.
+- To distinguish packages from packagesets we have to mark packagesets somehow, for example with a
+  file `by-name/fi/fishPlugins/.packageset`, if that file exists, it's called as a packageset,
+  if not the `package.nix` must exist and is called as a package.
```
infinisil: -1 from me for Idea 2 because of this. The fact that package sets can be nested is one of the big problems of Nixpkgs imo, because automated systems can't easily know what it needs to evaluate without evaluating. By replicating this into the file structure, we make it harder to get away from that model.

infinisil (commented):
In addition to the comments, also a meta-comment: While it's really great to create RFC _documents_ like these, we don't need the lengthy RFC _process_ anymore to get them approved, because if Nixpkgs committers can't agree on it, we can escalate to the SC-appointed Nixpkgs core team, which has the authority to make a call.

Probably still beneficial for Nixpkgs to keep discussions for such changes outside the Nixpkgs repo itself due to the noise there, but otherwise I'd suggest to PR this to an `adr/`[^1] directory in Nixpkgs.

[^1]: From https://adr.github.io/, which is very similar

rfcs/0197-package-set-definitions.md
quantenzitrone: i scrapped idea 3, however i want to hear a few more opinions until i scrap idea 2 as well

rfcs/0197-package-set-definitions.md
```diff
@@ -117,8 +118,8 @@ Proof-Of-Concept implementation in https://github.com/NixOS/nixpkgs/pull/483432
 
 - Package sets are done in a nested `by-name` structure under `pkgs/by-name`, e.g.
   `fishPlugins.puffer` would be in `by-name/fi/fishPlugins/pu/puffer`.
-- To distinguish packages from packagesets we have to mark packagesets somehow, for example with a
-  file `by-name/fi/fishPlugins/.packageset`, if that file exists, it's called as a packageset,
+- To distinguish packages from package sets we have to mark package sets somehow, for example with a
+  file `by-name/fi/fishPlugins/.packageset`, if that file exists, it's called as a package set,
```
Sigmanificient: This is a bit hard to read with `package sets` appearing 3 time in the same sentence
```suggestion
- A marker is required in order to distinguish package sets from simple packages, such as using a `.packageset` file (example: `by-name/fi/fishPlugins/.packageset`)
```

rfcs/0197-package-set-definitions.md
7c6f434c: This proposal is awkward with Python where you have CPython and PyPy and then language version-dialects within them, and even worse for Common Lisp where you have same base language but multiple compatible implementations with their completely own versioning.

quantenzitrone: i also think keeping functions in package sets is fine

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,274 @@
...
+co-authors: (find a buddy later to help out with the RFC)
+shepherd-team: (names, to be nominated and accepted by RFC steering committee)
+shepherd-leader: (name to be appointed by RFC steering committee)
+related-issues: (will contain links to implementation PRs)
+---
+
+# Summary
+[summary]: #summary
+
+Different ideas on how to handle package sets in nixpkgs.
```
7c6f434c: What _is_ a package set? There are some that do not have a direct top-level name. I think in general some places of the text is written for a package set that comes with its set of expressions, while there are cases where meaningfully sets share a large fraction of the expressions.

rfcs/0197-package-set-definitions.md
quantenzitrone: this is resolved due to autocalled `by-name` structure aliases, overrides and functions

rfcs/0197-package-set-definitions.md
quantenzitrone: sharding all package sets makes things significantly easier

rfcs/0197-package-set-definitions.md
quantenzitrone: `packageset.nix` was dropped in favor of more autocalling

rfcs/0197-package-set-definitions.md
quantenzitrone: i added a more focused `aliases.nix` that is overlayed over the autocalled packages (behind `optionalAttrs config.allowAliases` of course)

rfcs/0197-package-set-definitions.md
```diff
@@ -37,65 +37,83 @@ Proof-Of-Concept implementation in https://github.com/NixOS/nixpkgs/pull/482538
...
-    │   ├── aa
-    │   ...
-    │   ├── zx
-    │   │   ├── zxcvbn
-    │   │   │   └── package.nix
-    │   │   ├── zxcvbn-rs-py
-    │   │   │   └── package.nix
-    │   │   └── zxing-cpp
-    │   │       └── package.nix
-    │   └── packageset.nix <- entrypoint to package set
```
qweered: if we do sharding everywhere than smaller package sets become really unpleasant to work with, basically 2x -1.5x folders for sets with 30-80 packages

rfcs/0197-package-set-definitions.md
quantenzitrone: i split up `packageset.nix` into `aliases.nix`, `overrides.nix` and `functions.nix`

rfcs/0197-package-set-definitions.md
quantenzitrone: I think it's fine.
Packackes will just be one shard in mostly, which doesn't bother much.

```
$ tree --dirsfirst -L 2 pkgs/sets-by-name/fishPlugins/
pkgs/sets-by-name/fishPlugins/
├── as
│   └── async-prompt
├── au
│   └── autopair
├── aw
│   └── aws
├── ba
│   ├── bang-bang
│   └── bass
├── bo
│   ├── bobthefish
│   └── bobthefisher
├── cl
│   └── clownfish
├── co
│   └── colored-man-pages
├── do
│   └── done
├── ex
│   └── exercism-cli-fish-wrapper
├── fi
│   ├── fifc
│   ├── fishbang
│   ├── fish-bd
│   ├── fishtape
│   ├── fishtape_3
│   └── fish-you-should-use
├── fo
│   ├── foreign-env
│   └── forgit
├── fz
│   ├── fzf
│   └── fzf-fish
├── gi
│   ├── git-abbr
│   └── github-copilot-cli-fish
├── gr
│   ├── grc
│   └── gruvbox
├── hu
│   └── humantime-fish
├── hy
│   └── hydro
├── ma
│   └── macos
├── nv
│   └── nvm
├── pi
│   └── pisces
├── pl
│   ├── plugin-git
│   └── plugin-sudope
├── pu
│   ├── puffer
│   └── pure
├── sd
│   └── sdkman-for-fish
├── sp
│   ├── spark
│   └── sponge
├── ti
│   └── tide
├── tr
│   └── transient-fish
├── wa
│   └── wakatime-fish
├── z_
│   └── z
├── aliases.nix
├── buildFishPlugin.nix
├── functions.nix
└── overrides.nix
```

```
$ ls pkgs/sets-by-name/fishPlugins/
as  aw  bo  co  ex  fo  gi  hu  ma  pi  pu  sp  tr  z_           buildFishPlugin.nix  overrides.nix
au  ba  cl  do  fi  fz  gr  hy  nv  pl  sd  ti  wa  aliases.nix  functions.nix
```

qweered: well yeah, its not THAT big of a deal and it simplifies tooling and mental model, so i think it fine

rfcs/0197-package-set-definitions.md
quantenzitrone: 4. Auto generated package sets like `haskellPackages`.

quantenzitrone: <s>i want handle both 2 and 3 with this RFC</s>
actually i want to handle all package sets in this RFC

rfcs/0197-package-set-definitions.md
quantenzitrone: decided on idea 1, moved other ideas to alternatives

rfcs/0197-package-set-definitions.md
quantenzitrone: implemented in idea 1

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,254 @@
...
+      package.
+  - `aliases.nix`: aliases for aliases in package sets behind `optionalAttrs config.allowAliases`
+    (like `top-level/aliases.nix`).
+- All package sets with their sharded packages, overlayed with their functions, overrides and
+  aliases are automatically called by `top-level/package-sets-by-name.nix`.
+- Versioned package sets like `python316Packages` are done in `all-packages.nix` by overriding the
+  default version package set.
+- Versioned package sets without a default version will have to override the default version with
+  an error.
+  - e.g. `nextcloud*Packages` are in `sets-by-name/nextcloundPackages` and thus autocalled by
```
Sigmanificient: `nextcloundPackages` (i think this is a typo)

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,254 @@
...
+# Motivation
+[motivation]: #motivation
+
+- get rid of the package categories as directories (decided in RFC 140 and 146)
+- bring the benefits of by-name to package sets
+  - merge bot maintainer merging
+  - scaleability
+  - isolation
+  - vetting
+- *your goal here*
```
Eveeifyeve, line 52: Adding potentially an GitHub action bot that runs updates????

SigmaSquadron: Updates across package sets are very different. I don't think this should be a goal of this RFC, but instead something that can be developed separately as an `updateScript` that `nixpkgs-update` can pick up.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,254 @@
...
+│   │   ├── aa
+│   │   ...
+│   │   ├── zx
+│   │   │   ├── zxcvbn
+│   │   │   │   └── package.nix
+│   │   │   ├── zxcvbn-rs-py
+│   │   │   │   └── package.nix
+│   │   │   └── zxing-cpp
+│   │   │       └── package.nix
+│   │   ├── aliases.nix
```
7c6f434c: So, a few files mixed within the shard list?

SigmaSquadron: The alternative would be putting them in `pkgs/top-level`. It seems more sensible to have them at the root of the package set instead.

7c6f434c: Another alternative is to have one more level inside the package set before the shards (`by-name` subdirectory, I guess).

Which would also make more manageable packagesets that benefit from having a few more support files inside the package set but not as a packageset package.

SigmaSquadron: That might require us to filter out that directory from the autocaller function, but it's a good idea as well.

quantenzitrone: i think they meant
```
pkgs/sets
├── fishPlugins
│   ├── by-name
│   │   ├── as
│   │   │   ├── async-prompt
│   │   │   ... └── package.nix
│   │   ├── au
│   │   ...
│   │   └── z_
│   │       └── z
│   │           └── package.nix
│   └── top-level
│       ├── aliases.nix
│       ├── functions.nix
│       └── overrides.nix
```

7c6f434c: Well, it is a matter of taste if `top-level` is needed or the files would go into the root of the package set.

Eveeifyeve: > Well, it is a matter of taste if top-level is needed or the files would go into the root of the package set.
Exactly.

So I was thinking like this could happen:
```
pkgs/
├── fi/fishPlugins
│   ├── by-name
│   │   ├── as
│   │   │   ├── async-prompt
│   │   │   ... └── package.nix
│   │   ├── au
│   │   ...
│   │   └── z_
│   │       └── z
│   │           └── package.nix
└── aliases.nix
└── functions.nix
└── overrides.nix
```

By name is used in the case so that way you don't get confused easily with two by-names. 
Another idea to improve this, is making the `by-name` in package sets to something like `packages` could work? So like:
```
pkgs/
├── fi/fishPlugins
│   ├── packages
│   │   ├── as
│   │   │   ├── async-prompt
│   │   │   ... └── package.nix
│   │   ├── au
│   │   ...
│   │   └── z_
│   │       └── z
│   │           └── package.nix
└── aliases.nix
└── functions.nix
└── overrides.nix
```

quantenzitrone: If you look at the POC implementation, you can also see that buildFishPlugin is in `sets-by-name/fishPlugins/buildFishPlugin.nix`.
For some package sets there might be larger amounts of functions so the package set folder could become crowded.
We could also just inline functions into `functions.nix` and require that no additional files are in the package set tree root.

7c6f434c: I don't understand that. This mixes shards and non-shards under `pkgs` itself? And puts the package sets as normal packages? I am probably misreading the diagram.

7c6f434c:
By the way, lispPackages treated as sbclPackages but also needed for `ccl.pkgs` and `ecl.pkgs` seems even worse semantically than override `luajit = lua5_3` or the other way round that current attempts to minimise `all-packages.nix` at any cost promote. So this should include pins from the beginning.

quantenzitrone:
Guys I think we need to decide on and merge https://github.com/NixOS/rfcs/pull/83 first.
Especially the `foo.pkgs` vs `fooPackages` part.
I've always thought of package sets as top level attributes like `fooPackages`, because that is what is exposed on https://search.nixos.org, but it seems that is not always the case.

SigmaSquadron:
#83 is unfortunately abandoned. I don't think we should depend on it, but rather replace it.

rfcs/0197-package-set-definitions.md
Eveeifyeve: > I don't understand that. This mixes shards and non-shards under pkgs itself? And puts the package sets as normal packages? I am probably misreading the diagram.

Well the ideally in the packages it's another set then there would be default.nix which would hook it up.

> I am probably misreading the diagram.

Probably 👀

7c6f434c: OK, so what are the full paths to the fishPlugins async-prompt package.nix and to aliases.nix at the bottom, so that we can interpret the diagram the same way?

7c6f434c:
I think #192 has chances and also is needed (on its own or incorporated) for this one to be good?

quantenzitrone:
Why do we need versioned python, ghc and ocaml packages but not different package sets for other languages like C? Why are there no `{clang{18,19,20,22},gcc{13,14,15},tinyc,arocc,hcc}Packages`?

edit: The out path depends on the version of the interpreter:
- For Python: `$out/lib/python${version}/site-packages/${pname}`
- For Ocaml: `$out/lib/ocaml/${version}/site-lib/${pname}`

7c6f434c:
> Why do we need versioned python, ghc and ocaml packages but not different package sets for other languages like C? 

* The recent stream of PRs setting `--std=gnu17` is addressing a similar issue for C in a different way
* For SBCL we do sometimes have to use `stumpwm = sbcl_<previous_version>.pkgs.stumpwm` somewhere, but on the balance we don't have a full top-level name for previous-version sbclPackages.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,254 @@
+---
```
quantenzitrone, line 1: If we don't incorporate a system for nested package sets/package sets nested in attribute sets, how do we deal with them?
- these could probably be moved to the top level:
  - `emacsPackages.*Packages`
  - `ocaml-ng.ocamlPackages*`
  - `haskell.packages.{,native-bignum.}*`
  - `chickenPackages.chickenEggs`
- `vscode-extensions` could probably be flattened
- `ocamlPackages*.janeStreet` could maybe be merged into `ocamlPackages*`
- `kdePackages.{sources,gear,framework,plasma}`?
- `vimPlugins.nvim-treesitter-{,legacy-}parsers`?
- `linuxKernels.**`?
- `luaPackages.luaLib`?
- `androidenv.**`?
- `nginxModules.*`?

rfcs/0197-package-set-definitions.md
quantenzitrone: the new rfc version does not have this problem

rfcs/0197-package-set-definitions.md
quantenzitrone: i included a definition of package sets
does it suffice?

rfcs/0197-package-set-definitions.md
quantenzitrone: I decided since I also want to include generated package sets and make the migration of large package sets easier, that the by-name structure should be a subdirectory of `pkgs/sets/<setname>/`.
This way package sets can have lots of other files next to the standardized ones.

qweered:
Somewhat related — I've been working on [`lib.mkPackageVariants`](https://github.com/NixOS/nixpkgs/pull/496617), which solves a complementary problem: managing multiple **variants of a single package** (versions + feature configurations).

Currently packages like openssl, python, perl, etc. each reinvent their own multi-version handling with ad-hoc patterns. `mkPackageVariants` standardizes this into a shared utility — you get a `variants.nix` (version/config metadata), a `generic.nix` (curried builder), and a thin `package.nix` wrapper. The default variant is the package itself, and all variants are accessible via passthru (`openssl.v1_1`, `openssl.v3_6`, `openssl.oqs`, etc.).

1. **by-name compatible**: `mkPackageVariants` returns a single derivation (the default variant) from `package.nix`, so it works natively with by-name auto-discovery. The openssl migration in the PR already lives in `pkgs/by-name/op/openssl/`. Versioned aliases (`openssl_1_1`, `openssl_3`, etc.) go in `all-packages.nix` as backward-compat references.

2. **Complements package sets**: This RFC handles *package ecosystems* (pythonPackages, etc.) where multiple packages share infrastructure. `mkPackageVariants` handles the orthogonal case of *one package with multiple versions/configs*. They compose naturally — a package set member could itself use `mkPackageVariants` for its versioning.

3. **Reduces `all-packages.nix` clutter**: Instead of N top-level entries per versioned package (e.g., `openssl_1_1`, `openssl_3`, `openssl_3_5`, `openssl_3_6`, `openssl_oqs`, `openssl_legacy`), there's one by-name entry with variants in passthru and lightweight aliases for backward compat.

4. **Grouped definitions**: Packages using `mkPackageVariants` naturally share files across versions (patches, generic builder), which aligns with RFC 197's grouped definition support. The `package.nix` + `variants.nix` + `generic.nix` convention could become a standard pattern for this.

rfcs/0197-package-set-definitions.md
```diff
@@ -0,0 +1,537 @@
...
+- `vscode-extensions.*` but that's a whole another mess
+- `kdePackages.{gear,frameworks,plasma}`
+- ...
+
+## What should package sets be?
+
+Package sets should be top-level attribute sets. This means: 
+- `ocaml-ng.*` will have to move to the top-level
+- `vscode-extensions` will be flattened
+- TODO: which package sets will have to move?
```
quantenzitrone, line 88: Alternative:
have versioned package sets be in a package-set-set
e.g.
- `ocamlPackageSets.ocaml_5_4`
- `lixPackageSets.lix_2_93` (this is how it currently is)
- `pythonPackageSets.python_3_13`

This could also be used as a nice way of handling versioned packages
e.g.
- `ocamlVersions.ocaml_5_4`, `ocamlVersions.ocaml_4_14_unsafe_string`
- `nixVersions.nix_2_32` (this is how it currently is)
- `pythonVersions.python_3_13`, `pythonVersions.pypy_3_10`
- `ffmpegVersions.ffmpeg_7`, `ffmpegVersions.ffmpeg_7_headless`

yay (👍) or nay (👎)?

Eveeifyeve: What is the problem that could happen with doing both?

qweered: The problem is that we not only have versions, but variants like cmakeMinimal, I'll try to come up a solution to passthru problem in my pr. Also `ocamlPackageSets.ocaml_5_4` is much longer than `ocaml.v5_4`

Eveeifyeve: I agree on the shorter version, however the main question is how can we make sure users don't get confused, especially beginners? My only thoughts are to just produce a error that tells you what is available if it's evals and it's incorrect. Like `ocaml.test` which doesn't exist.

rfcs/0197-package-set-definitions.md
mdaniels5757: This question belongs under "future work" or "unresolved questions", I think. (i.e. I don't think we need to solve this before getting the benefits of top-level package sets.)

rfcs/0197-package-set-definitions.md
qweered: Derivations inside derivations apparently a no-go so maybe we can generate `openssl_major_minor?_patch?` on the fly from versions.nix. It preserves current behavior and unifies it. I still don't like `opensslVersions.openssl_major_minor_patch` cause it is very long to type. Will implement this in my lib.mkPackageVariants pr soon

GetPsyched: > I'll try to come up a solution to passthru problem in my pr.

Unsure if many folks have the context here, however, this is likely the PR being referenced: https://github.com/NixOS/nixpkgs/pull/496617

I second the idea of not having nested package sets for versioned variants. I love the concept of `lib.mkPackageVariants` and IMO we should assume that to be in place for such edge cases in this RFC.

PR #201, opened 2026-05-13
[RFC 0201] security.artifacts: backend-agnostic secret management

Mutasem-mk4:
This RFC proposes \security.artifacts\, a backend-agnostic abstraction layer for managing secrets in NixOS. It decouples NixOS service modules from specific secret providers (like \sops-nix\, \genix\, or \systemd-creds\), allowing users to uniformly declare and inject secrets without vendor lock-in.

*This RFC accompanies the implementation PR at NixOS/nixpkgs#519619*.


Lord-Valen:
PR and file are named wrong. The number should be the same as the PR number, padded to four digits: 0201. Not sure why that isn't written anywhere tbh.

Mutasem-mk4:
Fixed: Renamed RFC file to \
fcs/0201-security-artifacts.md\ and updated the PR title to match the PR number as requested. Thanks for the catch!

rfcs/0201-security-artifacts.md
```diff
@@ -0,0 +1,98 @@
...
+security.artifacts.secrets."postgres-pw" = {
+  owner = "postgres";
+  group = "postgres";
+  mode = "0400";
+};
+
+services.postgresql = {
+  enable = true;
+  # ...
+  # The module can directly read the provisioned artifact or accept the path.
```
truh: NixOS Modules accessing security artifact secrets directly seems more difficult to explain to users than a passwordFile option.

Adding a string based API (the secret name) to NixOS modules doesn't seem ideal.

rfcs/0201-security-artifacts.md
```diff
@@ -0,0 +1,98 @@
...
+This RFC proposes the `security.artifacts` module tree:
+
+1. **Option Declaration**:
+   `security.artifacts.secrets.<name>` accepts a submodule configuring the owner, group, mode, and target path of the secret.
+
+2. **Providers**:
+   The abstraction supports pluggable backends (`security.artifacts.provider`):
+   - `sops-nix`
+   - `agenix`
+   - `systemd-creds`
```
truh, line 38: Do you have an article/blog posts etc that explains the use of systemd-creds with NixOS you have in mind? I haven't encountered it yet and wasn't able to find anything with a quick web search.

rfcs/0201-security-artifacts.md
truh: How do modules consume secrets managed with security.artifacts? I think that's worth including in the RFC.

ruro:
1) It's not immediately clear from the RFC text, how exactly are `services.*` supposed to "natively consume" `security.artifacts`.

    In the postgres example, you use `"postgres-pw"` as the name of the secret. Am I understanding correctly, that this name is just implicitly picked up by `services.postgresql` with no extra configuration? That seems a bit too "spooky action at a distance" to me.

2) IIUC, in the current design it's not possible to use different secret providers for different secrets. Even right now with only 3 providers, I can imagine situations when one might want to mix different providers (ie use `systemd-creds`  for machine-specific secrets and `sops-nix` / `agenix` for shared secrets). It might be worth it to not hardcode the "single provider" assumption into the system.

3) IIUC, the `dummy` provider creates secret files from plain text data. If we are including "built in" providers, it might make sense to include some kind of `externally-provided-path` provider (or extend the `dummy` provider to support this).

4) Speaking of providers - right now the RFC describes them as "pluggable backends", but it's not clear how would one create and "plug in" their own provider.

    Looking at the reference implementation it looks like all of the current providers are just hard coded in nixpkgs and in fact the `security.artifacts` module has bespoke logic for each provider (for example, the `security.artifacts.provider` option is a fixed enum, the `security.artifacts.secrets."*".dummy` option is globally defined by the `security.artifacts` module, not by the `dummy` provider, etc).

P.S. Is this LLM slop?

Mutasem-mk4:
I've updated the RFC and the reference implementation to address the feedback:

1. **Per-Secret Providers**: Now supports mixed configurations (e.g., combining \systemd-creds\ for hardware-bound secrets and \sops-nix\ for shared ones).
2. **External Provider**: Added a provider for secrets provisioned outside of NixOS (cloud-init, etc.).
3. **Implicit vs Explicit**: Added examples clarifying that modules should explicitly reference \config.security.artifacts.secrets.<name>.path\ to avoid 'spooky action'.
4. **Naming**: Renamed the RFC file to \

rfcs/0201-security-artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+# Detailed design
+[design]: #detailed-design
+
+This RFC proposes the `security.artifacts` module tree:
+
+1. **Option Declaration**:
+   `security.artifacts.secrets.<name>` accepts a submodule configuring the owner, group, mode, target path, and **optional per-secret provider**.
+
+2. **Providers**:
+   The abstraction supports pluggable backends (`security.artifacts.provider`):
```
kevincox, line 35:
```suggestion
   The abstraction supports pluggable backends (`security.artifacts.defaultProvider`):
```

Nit:

rfcs/0201-security-artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+   - `sops-nix`
+   - `agenix`
+   - `systemd-creds`
+   - `external` (Assumes secret is provisioned outside of NixOS, e.g. cloud-init)
+   - `dummy` (for CI/CD environments where secrets are mocked plaintext)
+
+   Users can set a global provider or override it per-secret, allowing for mixed configurations (e.g. TPM2-bound local secrets via `systemd-creds` alongside shared GitOps secrets via `sops-nix`).
+
+3. **Systemd Synchronization**:
+   A global systemd target `nixos-artifacts-secrets.target` is introduced. All secrets must be provisioned before this target is reached. Downstream services that consume secrets simply need `Wants = [ "nixos-artifacts-secrets.target" ]` and `After = [ "nixos-artifacts-secrets.target" ]`.
```
kevincox, line 45: I suspect this code be an issue for some providers like sops that do decryption on-host and can have any number of failure modes such as "New key `x` is not decryptable". In this case it is probably better to have a distinct target per-key so that most services can still boot up. Providers that do just copy secrets in one batch or otherwise can't have partial failures can just have these all be the same but providers that do have this issue can create a target per key.

rfcs/0201-security-artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+security.artifacts.provider = "sops-nix"; # Global default
+
+security.artifacts.secrets = {
+  "shared-app-key" = {
+    source = ./secrets/app.yaml;
+  };
+  
+  "local-tpm-secret" = {
+    provider = "systemd-creds"; # Override global default
+    path = "/run/secrets/tpm-key";
```
kevincox, line 86: I worry that figuring out how to ensure that the provider can write to this directory will be complicated. Is it assumed that this directory exists? Do we assume that all providers run as root and can write to any existing directory?

Maybe it would be better to make the `path` an "out parameter" which is set by the provider and the provider picks what directory to put files in (possibly with some provider-specific configuration).

rfcs/0201-security-artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+
+- **Status Quo**: Continue using `passwordFile` options and leave integration up to the user. This keeps NixOS core simpler but pushes complexity to the end-user.
+- **systemd-creds Only**: Force all secrets to use `systemd-creds`. While powerful, `systemd-creds` does not natively support GitOps workflows (e.g. encrypting with age/PGP and committing to a repo) as easily as `sops` or `agenix` without significant wrapper tooling.
+
+# Prior art
+[prior-art]: #prior-art
+
+- **sops-nix**: Pioneered the `sops.secrets.<name>` pattern, which heavily inspired this RFC's interface.
+- **agenix**: Similar pattern using `age.secrets.<name>`.
+
```
kevincox, line 110:
```suggestion
- **colmena**: Similar pattern using `deployment.keys.<name>` (inherited from nixops).

```

Not critical but relevant IMHO. It also shows a different style where the secrets are just copied from the deployment host without any on-host decryption step.

rfcs/0201-security-artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should `security.artifacts` also handle dynamically generated secrets (e.g. `nixos-generate-config` generating an SSH key) rather than just static deployed secrets?
```
kevincox, line 114: This would be very cool. For example sometimes services need secrets for auth between themselves but the secret doesn't matter as long as both services know it. I see a few use cases:

1. Generated per-deploy.
2. Generated one for the lifetime of the host.
3. Generated deterministically but just to use logic (decrypt, base64, ...)

For example colmena supports 1 and 3 via a [`keyCommand` option](https://colmena.cli.rs/unstable/features/keys.html).

kevincox (commented):
Thanks for taking this on. Being able to create a common interface would make using NixOS modules far more pleasant by removing those pesky required fields.

rfcs/0201-security-artifacts.md
```diff
@@ -0,0 +1,120 @@
...
+- **Maintenance of Translation Layers**: The translation layers for out-of-tree backends (`agenix`, `sops-nix`) will need to be kept up to date with upstream changes.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- **Status Quo**: Continue using `passwordFile` options and leave integration up to the user. This keeps NixOS core simpler but pushes complexity to the end-user.
+- **systemd-creds Only**: Force all secrets to use `systemd-creds`. While powerful, `systemd-creds` does not natively support GitOps workflows (e.g. encrypting with age/PGP and committing to a repo) as easily as `sops` or `agenix` without significant wrapper tooling.
+
+# Prior art
+[prior-art]: #prior-art
```
kevincox, line 106: A lot of the prior art allows controlling the mode, user, groups of the secrets but this proposal doesn't mention any of this. This need is somewhat mitigated by systemd's LoadCredential but IMHO there are still many use cases (that for better or worse) still need to have files lying around owned by a particular user or group. What is the plan for supporting these use cases?

KiaraGrouwstra:
this should probably mention clan's vars effort for comparison.

i also see partial overlap with contracts (#189) and its implementations (e.g. #506343), which isn't concerned with deploying secrets directly, but does help decouple logic more broadly (not just for secrets providers), and may also be reusable across nix-based systems (nixos, home-manager, ...), across modular services and across (eval-time) node boundaries.

